PR #8, opened 2015-06-25
Fix FPGA/VLSI Mem Gen for Python 2 and 3 Environments

seldridge:
Running fpga_mem_gen in a Python 3 environment dies due to one use of `range` and a `print` statement that doesn't use parentheses. This trivially fixes both of these without breaking Python 2 compatibility.

Commit message:

Two quick fixes that enable fpga_mem_gen to work with either Python 2 or
Python 3:
* Change an `xrange` instance to `range`
* Wrap the arguments of a bare `print` in parentheses

### DIFF
--- a/fsim/fpga_mem_gen
+++ b/fsim/fpga_mem_gen
@@ -53,7 +53,7 @@ def parse_line(line):
   mask_gran = 1
   tokens = line.split()
   i = 0
-  for i in xrange(0,len(tokens),2):
+  for i in range(0,len(tokens),2):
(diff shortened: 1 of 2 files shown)


seldridge:
I realized the same problem existed in vlsi_mem_gen. Fixed with b4cd8c5.

sbeamer:
Thanks!

In general python3 isn't a priority for us since it isn't yet the default on our platforms (OSX and Ubuntu), but if the change is this minor there is no reason to not do it.

PR #9, opened 2015-07-13
Param-based makefrag generation

hcook:
Uses Chisel Parameters to generate makefrags for use in testing. The Makefrags contain only the targets compatible with the particular processors or accelerators generated by the Parameter evaluations and no others. For now, best practice is to add tests to the TestGeneration object when some particular Build* Parameter is evaluated.


yunsup:
This is really cool.  A couple of suggestions:

1) How about renaming generated-src/Makefrag to generated-src/Makefrag-tests to be more descriptive?
2) With the new refactoring, I noticed that now in a clean emulator/ or vsim/ directory, you can't run "make -j16 run-asm-tests" to build&test, since generated-src/Makefrag-tests isn't there yet.  How about having a separate make target for generated-src/Makefrag-tests to first generated the set of tests before including it?  Something like how makefiles generated the .d file and then including them?

PR #13, opened 2015-08-13
Use new setName method for SeqMem to fix readmemh reference in Zscale

albert-magyar:
The Zscale test harness broke because it used cross-module references in Verilog for readmemh. Although this could be fixed by using a testbench that sets up memory contents via AHB, the addition of `setName` to `SeqMem` that Jim merged in allows this patch to fix Zscale.


colinschmidt:
If you rebase this change off master it now has a newer chisel version(with setName for SeqMem) so would be mergeable.

PR #15, opened 2015-09-03
If you don't have spike-disasm in your path, your path is dumped

ben-k:
to stdout by this line every time you do anything in the entire repo.


palmer-dabbelt:
@yunsup @aswaterman Who merges reference-chip PRs?  Do you want me to deal with these?

aswaterman:
We didn't notice this because, on our systems, "which" doesn't dump the path when it fails.

aswaterman:
@palmer-dabbelt feel free to help merge PRs.

PR #20, opened 2015-10-08
Add the ability to switch between chisel as a library or a project dependency.

ucbjrl:
Add SBT_FLAGS and explicitly set chiselVersion if it's defined in the environment.
Use ProjectRef for chisel to get plugin's included in build.


ucbjrl:
This is a "nice to have" mostly for the purpose of testing Chisel. Closed pending more interest.

PR #24, opened 2015-10-29
Add a "--master" flag to the regression script

palmer-dabbelt:
I want to be able to test the master of riscv-gnu-toolchain against the current
RTL as part of the buildbot.  This flag takes a list of repositories (by their
submodule path) and updates those to the current master, which facilitates that
check.

### DIFF
--- a/regression.sh
+++ b/regression.sh
@@ -9,11 +9,29 @@ set -ex
 echo "Starting Rocket-chip regression test"
 if [ $# -lt 1 ]
 then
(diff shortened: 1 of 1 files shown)


palmer-dabbelt:
@colinschmidt IIRC this is your thing.  I'm assuming this is OK?  If you don't say anything I'll merge it in a few hours...

colinschmidt:
Looks good to me.

PR #28, opened 2015-11-12
Don't re-generate the .d files on "make clean"

palmer-dabbelt:


yunsup:
thanks a lot for fixing this plamer!

aswaterman:
hooray

PR #31, opened 2015-11-26
implement support for multiple RoCC accelerators

zhemao:
This has the rocket changes and some changes to Config.scala.

The main different is that BuildRoCC is now a Seq instead of an Option. In addition, we have two new parameters: RoccOpcodes, which specifies which instructions go to which accelerator, and RoccAcceleratorMemChannels, which specifies how many dmem channels to give each accelerator. The latter should be edited instead of modifying RoccNMemChannels directly.


colinschmidt:
Looks good to me. I'll see if I can add one of these configs to the travis config.

colinschmidt:
Ok travis should run this config and the default config now, so hopefully RoCC breakage happens less often.

PR #33, opened 2015-12-09
Fix ROCKETCHIP_ADDONS chisel_srcs Makefrag parsing

seldridge:
The ROCKETCHIP_ADDONS environment variable is intended to be comma-separated (see project/build.scala). However, the variable chisel_srcs will treat a comma-separated ROCKETCHIP_ADDONS as one variable when evaluating chisle_srcs unless all commas are changed to spaces. This replaces commas with spaces in the normal kludgy Makefil fashion.


colinschmidt:
Why not move the parsing in project/build.scala to use spaces so that ROCKETCHIP_ADDONS matches other similar variables like default_submodules?

seldridge:
@colinschmidt, that makes way more sense to me. See https://github.com/ucb-bar/rocket-chip/pull/34.

PR #34, opened 2015-12-09
build.scala uses space-delimited ROCKETCHIP_ADDONS

seldridge:
This prevents the build from barfing if you use more than one rocketchip addon.


colinschmidt:
Looks good, thanks!

PR #40, opened 2016-01-28
Support make-3.82 and newer

palmer-dabbelt:
make changed its priorties for resolving implicit rules, which causes different
behavior when running "make run-bmark-tests".  This patch changes the hex file
rules to ensure they match between the two versions of make.

I've tried this with both make-3.81 and make-4.1, and they both work for me.


palmer-dabbelt:
@aswaterman how does this one look?

aswaterman:
cool

PR #41, opened 2016-02-01
Support torture regressions

palmer-dabbelt:
This should be pretty boring: it's almost all just regression changes.  I'm making a PR because there's two things that people might care about:

* riscv-tools bump (for the Spike NaN change)
* riscv-torture is added as a submodule of rocket-chip.

I added torture because the I want the nightlies to run it, and since different rocket-chip branches might need different tortures (hwacha needs esp-torture, for example) I think it's easiest to just track this like every other submodule.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -31,3 +31,6 @@
 [submodule "groundtest"]
 	path = groundtest
 	url = https://github.com/ucb-bar/groundtest.git
+[submodule "torture"]
+	path = torture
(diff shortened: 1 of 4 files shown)


colinschmidt:
This looks good to me. I think its a good idea to track torture now that its upstream.
Shouldn't this also be replacing the scripty parts of .travis.yml

palmer-dabbelt:
I'll bump that when I make sure the buildbot works.

PR #49, opened 2016-02-17
SCR file and NastiMemorySelector changes

palmer-dabbelt:
@zhemao these aren't really related, but need to be ordered.


palmer-dabbelt:
@zhemao CR

PR #54, opened 2016-02-23
The FPGA doesn't have an HTIF clock divider

palmer-dabbelt:
We used to just be writing the SCR anyway, but now that the SCR maps are
automatically defined VCS will detect the missing SCR and bail out when
compiling test harness code.  This patch just doesn't write the HTIF SCR when
there isn't one.


colinschmidt:
Seems like a good change to me.

PR #55, opened 2016-02-23
travis-ci.org improvements

palmer-dabbelt:
We used to have two regression scripts, now there's one.  This also allows caching of the riscv-tools builds, which should speed up the regressions a lot.

### DIFF
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -19,8 +19,12 @@ endif
 # The top-level directory that contains rocket-chip
 TOP ?= ..
 
+# The hash of the tools that we're using
+TOOLS_HASH ?= $(shell git -C $(TOP) ls-tree HEAD -- riscv-tools | xargs echo | cut -d' ' -f3)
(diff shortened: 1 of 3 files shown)


regression/Makefile
```diff
@@ -37,21 +41,24 @@ clean:
 # this order, since there's dependencies for everything it doesn't actually
 # matter.  They're here to make running the various targets from the
 # commandline a bit cleaner.
-submodules: stamps/submodules.stamp
-tools: stamps/tools.stamp
+submodules: stamps/other-submodules.stamp stamps/riscv-tools-submodules.stamp
+tools: $(RISCV)/install.stamp
 
+emulator-verilog: stamps/$(CONFIG)/emulator-verilog.stamp
```
colinschmidt: This seems wrong. Running 
`make verilog`
in emulator fails

palmer-dabbelt:
I'm merging this because I think the reason it's not working is because it keeps getting bounced off and other PRs drop the cache.

PR #58, opened 2016-02-24
More travis fixing

palmer-dabbelt:
I believe with this Travis should actually work.


colinschmidt:
cool, lets be done with this finally

PR #63, opened 2016-02-25
RoCC PTW refactoring

yunsup:
@aswaterman @zhemao cleaning up the RoCC PTW ports.

### DIFF
--- a/src/main/scala/Configs.scala
+++ b/src/main/scala/Configs.scala
@@ -153,6 +153,7 @@ class DefaultConfig extends Config (
       }
       case BuildRoCC => Nil
       case RoccNMemChannels => site(BuildRoCC).map(_.nMemChannels).foldLeft(0)(_ + _)
+      case RoccNPTWPorts => site(BuildRoCC).map(_.nPTWPorts).foldLeft(0)(_ + _)
       case RoccNCSRs => site(BuildRoCC).map(_.csrs.size).foldLeft(0)(_ + _)
       case UseDma => false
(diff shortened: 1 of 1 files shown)


src/main/scala/Configs.scala
```diff
@@ -153,6 +153,7 @@ class DefaultConfig extends Config (
       }
       case BuildRoCC => Nil
       case RoccNMemChannels => site(BuildRoCC).map(_.nMemChannels).foldLeft(0)(_ + _)
+      case RoccNPTWPorts => site(BuildRoCC).map(_.nPTWPorts).foldLeft(0)(_ + _)
```
aswaterman: What's with the inconsistent capitalization between RoCC and Rocc?

yunsup: I was trying to be consistent with other parameters around: RoccNMemChannels, RoccNCSRs.

colinschmidt: At some point in the past I think the parameters were changed to match some Scala style (camel case). I personally like the capital Cs but either way we should pick one

aswaterman: It seems that Java and Scala favor using consistent casing for acronyms, even in camelcase (e.g. package xml, class XMLParser.  So RoCC is fine.

yunsup: There should be another commit which changes all Rocc references to RoCC, so I'm submitting this PR again only with the updated rocket hash tag.

PR #65, opened 2016-02-25
Move N_CORES and MMIO_BASE out of the SCRFile class and into the uncore instance

jwright6323:
See https://github.com/ucb-bar/uncore/pull/15


colinschmidt:
Something is wrong with your branch, it has many different commits but its "really" only 1 commit ahead of master, the same is happening with your uncore branch

jwright6323:
I'll fix it, I think @zhemao did some rebasing to clean up my earlier PRs.

jwright6323:
@colinschmidt Fixed

PR #75, opened 2016-03-15
Fix the SCR file for Chisel 3

palmer-dabbelt:
https://github.com/ucb-bar/uncore/pull/28


zhemao:
Looks like everything's good here. I'll take care of merging.

Issue #74, opened 2016-03-15
Atomic increment using LR/SC


mn416:
Hi,

I have a barrier implementation which works when the atomic increment part is implemented using an `amoadd.d`:

```
amoadd.d a0, %1, (%0)
```

but not when using an LR/SC:

```
1:
lr.d   a0, 0(%0)
add    a0, a0, %1
sc.d   a0, a0, 0(%0)
bnez   a0, 1b
```

I've made a simplified, self-contained implementation that demonstrates the problem:

1. Download the source files at:

  http://www.cl.cam.ac.uk/~mn416/rocket-lrsc-issue/

2. Build by running `make.sh` to produce `main.hex`

3. Pass `main.hex` to a rocket-chip emulator compiled with `NTILES=2':

```
> ./emulator +loadmem=main.hex
```

When using `amoadd.d`, the program prints

```
Started
Finished
```

When using LR/SC, we only see `Started`.  The second half of the barrier, which waits until every core has incremented a shared variable, doesn't seem to terminate.

colinschmidt:
Does the same test work in spike with -p2 ?

mn416:
Yes, with the LR/SC code, `spike -p2 main.elf` gives

```
Started
Finished
```

as expected.

aswaterman:
I am looking into this.

aswaterman:
This bug seems to date back to 8b5a2e964b99030eeb44d34396915d8af88b3727, when TileLink Finish messages were removed from the L1 D$.

@hcook the L1 was waiting to send finish messages until the replay queue had drained, which prevented a race between replays and subsequent probes.  In this case, it manifests as livelock on LR/SC, because the probe occurs before the LR replays, so the SC is never able to complete.  I think the cleanest option is to use finish messages.  It's not obvious to me how else to solve this without blocking all probes to addresses that have outstanding Acquires, which would seem to cause deadlock.  (It doesn't suffice to block probes between when the grant occurs and when the replay queue is drained, because the probe could actually arrive before the grant.)

PR #76, opened 2016-03-18
Trace generator: updates and additions to the scripts directory

mn416:
1. Introduce `tracegen.py`, a script that invokes the emulator (built with `TraceGenConfig`), sending a `SIGTERM` once all cores are finished.

2. Update `toaxe.py` to gather some statistics about the trace.

3. Introduce `tracestats.py`, which displays the stats in a useful way.

4. Introduce `tracegen+check.py`, a top-level script that generates traces, checks them, and emits stats.

If this commit is pulled, it should be done after pulling my latest groundtest commit.

`tracegen+check.py` is designed to be run in the emulator directory with the scripts directory in your `PATH`, e.g.

```
mn416@babbage:~/rocket-chip/emulator$ tracegen+check.sh 
Testing against WMO model:
 
       0: .......... .......... .......... .......... .......... 
      50: .......... .......... .......... .......... ..........

OK, passed 100 tests
LR/SC success rate: 1%
Load-external rate: 76%

```

If you want 1000 tests starting from seed 100, you can do `NUM_TESTS=1000 START_SEED=100 tracegen+check.sh`.  There are various other parameters too (look in the script to see).

The current rocket-chip is passing a lot of tests, but note the `LR/SC success rate` at 1% which is probably related to issue #74.  In some cases, we can also observe a successful LR/SC performing non-atomic updates: specifically when the SC is issued before the LR response occurs.  This may be related to issue #74 too, so I'll not post a new issue.

The `Load-external rate` shows the proportion of loads that read a value written by another core.  It is there to show liveness: otherwise, an implementation in which each core is connected to a different memory, with no sharing, would be classed as passing.

There are other potentially useful stats too, e.g. average time between memory request and response.  If these non-functional properties are of interest I can add some more.

### DIFF
--- a/scripts/toaxe.py
+++ b/scripts/toaxe.py
@@ -27,142 +27,177 @@
 
 import sys
 import re
+import sets
 
-if len(sys.argv) != 2:
-  print "Usage: toaxe.py [FILE]"
-  sys.exit()
+def main():
 
-if sys.argv[1] == "-":
-  f = sys.stdin
-else:
-  f = open(sys.argv[1], 'r')
-  if f == None:
-    print "File not found: ", sys.argv[1]
-    sys.exit()
+  if len(sys.argv) < 2:
+    sys.stderr.write("Usage: toaxe.py TRACE-FILE [STATS-OUT-FILE]\n")
+    sys.exit(-1)
 
-lineCount = 0
-def error(msg):
-  print "Error at line ", lineCount, ": ", msg
-  sys.exit()
-
-# Mapping from address to axe address
-addrMap = {}
-nextAddr = 0
-
-# Mapping from (thread id, tag id) to axe operation id
-tagMap = {}
-
-# Mapping from thread id to operation id
-fenceReq = {}
-
-# Mapping from thread id to operation id
-loadReserve = {}
-
-# Array of axe operations
-ops = []
-
-for line in f:
-  # Exit loop at end of trace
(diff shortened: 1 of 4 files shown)


zhemao:
A few comments. 

1. The top level script assumes that the the executable you want to run is always named ./emulator-Top-TraceGenConfig. This is not always true. We might want to give the configuration a different name or use the debug emulator. You should be able to pass this in as a command line argument.

2. Does the script print out the random seed it is using? This would be important for reproducing failing cases.

3. You forgot to commit the updated submodule pointer for groundtest.

mn416:
Thanks for looking at this.

It is possible to change the name/path of the emulator via the command-line, e.g.

```
mn416@babbage:~/rocket-chip/emulator$ EMU=emulator-debug tracegen+check.sh
```

The other possible parameters are listed at the top of tracegen+check.sh.

The script does print the seed when a failure is found, and the seed is also printed every 50 tests before the colon.  Each dot in the output denotes a passed test, there are 50 dots per line, with a space separating every 10.  The seed is incremented by one for each new test.

If I understand correctly, I should have done "git add groundtest" in my commit for rocket-chip to tie the two together?  Have done so now.

Matt

zhemao:
I've pushed an amended version of your original commit (with submodule pointer bumped) to a staging branch. I'll merge it into master once it passes all the checks.

zhemao:
This has been merged.

Issue #67, opened 2016-02-26
Chisel3 builds are broken


palmer-dabbelt:
These have been broken for a day or two

````

[error] /vm/home/buildbot/slave/chisel3-chip_chisel3-chip_CONFIG_DefaultCPPConfig_TORTURE_CONFIG_default/build/uncore/src/main/scala/scr.scala:52: value name is not a member of Chisel.Data
[error]   def attach(data: Data): Data = attach(data, data.name, false, false)
[error]                                                    ^
[error] /vm/home/buildbot/slave/chisel3-chip_chisel3-chip_CONFIG_DefaultCPPConfig_TORTURE_CONFIG_default/build/uncore/src/main/scala/scr.scala:54: value name is not a member of Chisel.Data
[error]   def attach(data: Data, addReg: Boolean): Data = attach(data, data.name, false, false)
[error]                                                                     ^
[error] /vm/home/buildbot/slave/chisel3-chip_chisel3-chip_CONFIG_DefaultCPPConfig_TORTURE_CONFIG_default/build/uncore/src/main/scala/scr.scala:55: value name is not a member of Chisel.Data
[error]   def attach(data: Data, addReg: Boolean, readOnly: Boolean): Data = attach(data, data.name, readOnly, false)
[error]                                                                                        ^
[error] /vm/home/buildbot/slave/chisel3-chip_chisel3-chip_CONFIG_DefaultCPPConfig_TORTURE_CONFIG_default/build/uncore/src/main/scala/scr.scala:68: inferred type arguments [Chisel.Data] do not conform to method apply's type parameter bounds [T <: Chisel.Bits]
[error]       rdata(addr) := Cat(UInt(0, width=(scrDataBits-data.getWidth)),reg)
[error]                      ^
[error] /vm/home/buildbot/slave/chisel3-chip_chisel3-chip_CONFIG_DefaultCPPConfig_TORTURE_CONFIG_default/build/uncore/src/main/scala/scr.scala:68: type mismatch;
[error]  found   : Chisel.UInt
[error]  required: T
[error]       rdata(addr) := Cat(UInt(0, width=(scrDataBits-data.getWidth)),reg)
[error]                              ^
[error] /vm/home/buildbot/slave/chisel3-chip_chisel3-chip_CONFIG_DefaultCPPConfig_TORTURE_CONFIG_default/build/uncore/src/main/scala/scr.scala:68: type mismatch;
[error]  found   : Chisel.Data
[error]  required: T
[...]

Issue #44, opened 2016-02-11
Chisel is being invoked too many times on error (3x)? 


ccelio:
I'm trying to generate C++ emulators, and if I have a Chisel compile error, I get Chisel being invoked three times (I'm expecting Chisel to be invoked twice, the first to attempt to compile the hardware quickly, and the second to generate the proper stack trace as required if an error is found). 

I'm wondering if the emulator/Makefile may be the culprit:

````
44 $(generated_dir)/$(MODEL).$(CONFIG).d $(model_header) $(params_file): $(chisel_srcs)                   
45    cd $(base_dir) && $(SBT) "project $(PROJECT)" "run $(CHISEL_ARGS) --noIoDebug"                      
46                                                                                                        
````

Is anyone else seeing this behavior?

palmer-dabbelt:
This is because there's no "%" in the list of targets of this recipe.  A non-pattern recipe with multiple targets is actually just multiple recipes, so this is directly telling make to run Chisel three times.  You'll probably also see this fall over and die on parallel builds.

The easiest way to fix it is probably to just have a Chisel stamp, but finding a way to massage this into a pattern-based recipe would also fix the problem.

palmer-dabbelt:
I believe I've fixed this in the Chisel 3 Makefile, so I'm going to close the issue since the old version will eventually go away.

ccelio:
Good enough for me!

PR #77, opened 2016-03-22
Preliminary Chisel 3 Support

palmer-dabbelt:
This adds preliminary support for Chisel 3: all the configurations that run in Travis can now run through Chisel 3 and Stanza FIRRTL to generate Verilog on Travis.  This won't test anything because we don't support Verilator yet, but I wanted to get this up as fast as possible to avoid having people break what Chisel 3 support we have now.  This will also allow everyone to start working on Chisel 3 related infrastructure.

If you want to try the Chisel 3 build, do something like "make -C vsim CHISEL_VERSION=3".  The build defaults to Chisel 2, since Chisel 3 isn't working yet.

This does drop support for the Rocket Chip specific backends, since they won't work in Chisel 3.  We don't have a tapeout now, so I figured dropping them would be a good forcing function to make Chisel 3 work before the next one.

Merging this is blocked on https://github.com/ucb-bar/uncore/pull/30


palmer-dabbelt:
Merging this will fix #67

palmer-dabbelt:
Merging this should also fix #44

palmer-dabbelt:
@aswaterman This isn't blocked any more.

PR #78, opened 2016-03-23
Add checks for divide-by-zero

mwachs5:
When trying this script out on some simple inputs for reporting the statistics, I get divide-by-zero errors. Adding checks for these and assigning "N/A" instead of executing the divide-by-zero.


mn416:
Oops, apologies for missing this and thanks for reporting it.

Instead of printing `Load_Ext_Rate=N/A` and `LRSC_Success_Rate=N/A`, it is definitely preferable to not print anything.  This way, the `tracestats.py` script won't get confused when it sees `N/A` where a number is expected.  (It seems I did remember to check for division-by-zero in tracestats.py.)

mwachs5:
I updated the code to do nothing in divide-by-zero cases (vs printing N/A). But @mn416 if you want to adjust this more appropriately I can close my request.

mn416:
Thanks @mwachs5, your latest correction looks fine to me.  I've tried it and it works.

hcook:
Hey @mwachs5, this looks good to me too but something is hung up with the CI server so I can't merge it in. I think the best thing to try is for you to re-merge master and hopefully that will re-trigger the tests.

mwachs5:
Ok, will give it a shot. Let me know if I should squash everything into one commit instead.

mn416:
Does anyone else think that moving the `tracegen` scripts to `groundtest` would be a good idea?

This feels less intrusive, and now that I have write access to `groundtest` I could maintain them directly.

aswaterman:
In general, less stuff directly in rocket-chip is probably better.

hcook:
I agree with this as well. Closing this PR in expectation of one moving tracegen to groundtest.

PR #81, opened 2016-03-30
Removed jackhammer submodule

ericlove:
This was previously removed from the master branch of ucb-bar/rocket-chip, but not from the hwacha branch.  Removing it now should prevent `git submodule update` from failing due to the unavailability of jackhammer.


colinschmidt:
Jackhammer is still available (just not public for some reason) but since we aren't using it I'm fine removing it.

PR #82, opened 2016-03-31
Switch MMIO network to TileLink

zhemao:
This switches the MMIO network to do all of its routing over Tilelink instead of AXI. It uses a Tilelink interconnect generator based off of the existing AXI generator. A separate uncore pull request contains that addition.


aswaterman:
Thanks for doing this.

hcook:
Merged.

PR #85, opened 2016-04-02
Fix narrow read/write behavior for AXI converters

zhemao:
Until recently, we were assuming that the data channel in AXI was always right-justified. However, for narrow writes, the data must actually be aligned within the byte lanes. This commit changes some of the
converters in order to fix this issue.

Changes to junctions, uncore, and groundtest are on separate axi-fix branches as well.


aswaterman:
changes to sub-repos look ok to me.  should we merge them into master before continuing?

zhemao:
Yeah. Just give me a bit. I want to change the new unit test in groundtest to check the full conversion TL -> AXI -> SMI. Once that's done, I'll update, run through Travis and merge.

PR #88, opened 2016-04-06
Remove tracegen scripts, now in groundtest

mn416:
And bump groundtest.


zhemao:
Hmm. Maybe we should have a discussion about which way is best. @aswaterman do you have any opinions one way or another?

aswaterman:
@zhemao about whether it should be in groundtest or rocket-chip?  I putting it in groundtest is fine

mn416:
There was some discussion at the bottom of PR #78.

The advantage of groundtest is that I have write access, so can maintain the scripts more easily.

aswaterman:
retest this please

aswaterman:
wait, does that work with travis?

mn416:
From what I can tell, `.travis.yml` doesn't use these scripts so the commit shouldn't affect it. (?)

aswaterman:
ok.  somehow we need to get it to re-run the tests if I'm to merge this commit :-)

colinschmidt:
In travis you can look at a PR's build and click the restart build button in the upper right...But I don't even see this PR on the rocket-chip travis page. https://travis-ci.org/ucb-bar/rocket-chip/pull_requests

colinschmidt:
Ah someone turned off build pul requests in travis. (presumably to give us more bandwidth) But this means if the PR isn't a branch of rocket-chip we can't merge them (because the tests don't run).

colinschmidt:
The internet told me closing and re-opening would trigger a rebuild but that seems to not be the case...

zhemao:
Okay, I'm just going to push his changes to a separate branch in our repo. Then travis can actually run on this.

PR #95, opened 2016-04-22
Put memory into the address map and no longer use MMIOBase

zhemao:
@aswaterman This adds the change you requested. Instead of routing using a fixed MMIOBase, we add "mem" back to the address map as the first element and use that for routing in the l1tol2 network. We also add cacheability bits to the address map and use those to decide whether to use the MSHR or IOMSHR in Rocket's L1 data cache.


aswaterman:
thanks for that, @zhemao. merged & re-running the pull request hooks.

PR #96, opened 2016-04-22
add default cases to configs that use CDEMatchError

colinschmidt:
this avoids filling in the stack trace every time
a config doesn't contain the parameter

This speeds up a Hwacha config's compile time by 1 minute (from 2 to 1) in Chisel 2 and 4 minutes (6->2) in chisel3.

See https://github.com/ucb-bar/context-dependent-environments/pull/2

and some discussion https://github.com/ucb-bar/chisel3/issues/143

### DIFF
--- a/src/main/scala/Configs.scala
+++ b/src/main/scala/Configs.scala
@@ -11,7 +11,7 @@ import zscale._
 import groundtest._
 import scala.math.max
 import DefaultTestSuites._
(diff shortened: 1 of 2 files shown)


zhemao:
Looks like this needs to be rebased on top of the changes I made. I'll go ahead and take care of it.

PR #98, opened 2016-05-02
Add AXI to AHB converter and more conformant HASTI RAM

zhemao:
The DRAM controller for the Hurricane2 chip has an AHB port for configuration. Thus, we need a way to talk AHB from rocket. This adds an AXI -> AHB (NASTI -> HASTI) converter, some tests in groundtest, and a AHB RAM in uncore.


hcook:
LGTM

PR #102, opened 2016-05-09
Don't make MIFTagBits a computed parameter

zhemao:
For the hurricane2 tape out and other future work, the width of the AXI id fields will be fixed. Thus, the TL -> AXI converter will have to account from this and map TileLink ids to AXI ids.


aswaterman:
Currently, what happens if TileLink IDs are wider?

zhemao:
You mean what happens currently in master? The bits will just be truncated.

aswaterman:
Can we have a require() then?  It would be nice to not have hard-to-debug silent failures.

zhemao:
There is a require in master already. But with this commit there will no longer be a need for it.

aswaterman:
Oh.  I understand now.

zhemao:
So you've merged the uncore branch. Shall I merge this one then?

aswaterman:
I was about to do it

PR #107, opened 2016-05-17
Remove Zscale

zhemao:
Since Zscale is not currently working and we have no plans to use it in the near future, it might be best to remove it from the master branch so that external users don't get confused.


xfguo:
So I should try riscvz-sodor if i want an RV32 CPU and it is written by chisel?

sbeamer:
Sodor might be a little out of date. Rocket now supports RV32, so you could look at `DefaultRV32Config` as a starting point.

xfguo:
@sbeamer sure I will check the `RV32Config`, and do you think **rocket** is a good choice used in embedded/MCU application when it is configured with `Small/RV32` (e.g. compared with pulpino.)

Also, do you have any plan to support RVC in rocket. Thanks.

sbeamer:
Rocket should be able to downsize itself quite well. In addition to using RV32, you can also disable the FPU and reduce the amount of branch prediction.

I haven't compared Rocket with Pulpino, I would recommend trying both out.

There is interest in adding RVC to Rocket, but I am unaware of current development effort on it so I can't provide a release date.

colinschmidt:
Looks good. Thanks.

PR #110, opened 2016-05-18
return non-zero exit codes when an assertion fires

colinschmidt:
This ensures that assertion failures, which currently print a message to
the console but return a successful exit code, now will cause non-zero
exit code. This is meant to help automated tools like travis and
buildbot do a better job at catching assertions.
This impacts the various run-* targets in the simulation
directories.

### DIFF
--- a/src/main/scala/Testing.scala
+++ b/src/main/scala/Testing.scala
@@ -23,10 +23,10 @@ $$(addprefix $$(output_dir)/, $$($makeTargetName)): $$(output_dir)/%: $dir/%
 \tln -fs $$< $$@
 
 run-$makeTargetName: $$(addprefix $$(output_dir)/, $$(addsuffix .out, $$($makeTargetName)))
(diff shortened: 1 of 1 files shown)


colinschmidt:
I have too similar ways of solving this and I'm not sure which is preferred.

The PR has this:
```
perl -ne 'print "  [$1] $ARGV \t$2\n" if ( /\*{3}(.{8})\*{3}(.*)/ || /ASSERTION (FAILED)/i)' output/rv64ui-p-*.out | perl -ne 'if(/(.*)/){print "$1\n"; exit(1) if eof()}'
```

More compact, trades a pipe for a variable though:
```
perl -ne 'if ( /\*{3}(.{8})\*{3}(.*)/ || /ASSERTION (FAILED)/i) { print "  [$1] $ARGV \t$2\n"; $ret=1 }; exit($ret) if eof()'
```

colinschmidt:
@palmer-dabbelt @aswaterman Thoughts on this PR? It works in my tests where a few assertions fire per target.

palmer-dabbelt:
I like the idea.  I can't read the perl, so if you say it works then I'll just believe you :)

PR #113, opened 2016-05-24
ahb: move submodules forward

terpstra:
Include the two ahb changes from uncore and junctions. This commit should be merged after the other two.


aswaterman:
I'm going to PR the new D$ design tonight, so I'll just include these updated submodules as part of that.

PR #115, opened 2016-05-25
fix building with chisel3 without sbt

terpstra:
Don't fail to build if sbt is not installed


aswaterman:
i'm including this in the PR i'm currently trying to ram through.

PR #119, opened 2016-06-02
Instantiate Debug Module

aswaterman:


colinschmidt:
Is this developed enough to have tests? Seems like they could require a bit more new infrastructure and since this is off of most paths I don't think it should block merging this.

Otherwise, looks good to me.

aswaterman:
It's disabled for now.  The intent is to switch the front-end server to use it instead of HTIF (then delete HTIF), so running the ISA tests will server as the tests.

PR #122, opened 2016-06-03
Refactor AddrMap and its usage

aswaterman:


mwachs5:
Does this move PLIC base address?

aswaterman:
Yeah.  I forgot to add that to the commit message.  It is now at 0x4000_0000.

PR #123, opened 2016-06-04
AHB-top

terpstra:
Support AHB buses directly out of rocket-chip.
Merge request pre-approved by Megan!

PR #125, opened 2016-06-06
AHB-top

aswaterman:
Duplicate of #123 to force Travis to run.

### DIFF
--- a/src/main/scala/RocketChip.scala
+++ b/src/main/scala/RocketChip.scala
@@ -104,8 +104,7 @@ object TopUtils {
     rom.order(java.nio.ByteOrder.LITTLE_ENDIAN)
 
     // for now, have the reset vector jump straight to memory
-    val addrHashMap = p(GlobalAddrHashMap)
-    val resetToMemDist = addrHashMap("mem").start - p(ResetVector)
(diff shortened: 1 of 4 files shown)


src/main/scala/RocketChip.scala
```diff
@@ -11,16 +11,26 @@ import rocket.Util._
...
 /** Number of banks per memory channel */
 case object NBanksPerMemoryChannel extends Field[Int]
 /** Least significant bit of address used for bank partitioning */
 case object BankIdLSB extends Field[Int]
 /** Number of outstanding memory requests */
 case object NOutstandingMemReqsPerChannel extends Field[Int]
 /** Number of exteral MMIO ports */
-case object NExtMMIOChannels extends Field[Int]
+case object NExtMMIOAXIChannels extends Field[Int]
+case object NExtMMIOAHBChannels extends Field[Int]
```
colinschmidt: Why not have a TExtMMIOChannels as with the MemoryChannels?

aswaterman: It's probably a safe assumption in the memory case that the channels are homogeneous, but not for MMIO, where having both AXI and AHB bridges is conceivable.

PR #129, opened 2016-06-08
grammar

scottj97:
English major I'm not, but my sister was and she says 'who' is correct here


aswaterman:
Indeed.  I'll merge this as part of another PR.

PR #130, opened 2016-06-08
replace emulator with verilator for chisel3

donggyukim:
This works good in general, but mkpipe for vpd doesn't work, so make run-test-*-debug dumps nothing even though verilator correctly dumps waveform to the file. It's not clear what's happening to the dump file in VerilatedVcdC.


aswaterman:
thanks, @donggyukim, this looks good.

hope you can figure out how to fix vpd, and eventually avoid the special cases in the test bench generator.

PR #133, opened 2016-06-09
Comparator

terpstra:
Include the comparator tile. By default point it at two DDR locations.


zhemao:
Commits have been rebased and merged in.

PR #137, opened 2016-06-11
fix make error mixing implicit and normal rules

yunsup:
Thanks for adding verilator support @donggyukim.  This PR fixes a trivial Makefile problem.


palmer-dabbelt:
So I think you actually want _more_ %, not less.  Note that multi-target make rules with %s in them don't look at all like those without %s.  See

  http://stackoverflow.com/questions/2973445/gnu-makefile-rule-generating-a-few-targets-from-a-single-source-file

From my understanding of make the result of your rule is that SBT will be run 3 times, which is actually racy (but because of locks rarely causes problems).  I've changed this to instead use %-based targets everywhere in the rule, which by my understanding will let make run without introducing a race condition.

yunsup:
Ah, thanks.  Seems like Makefrag-emulator is also written without %s.  Do you want to update and force push?  Also, looking at Makefrags for emulator and verilator, the listed targets aren't complete.  How far should we go to write these outputs down?

palmer-dabbelt:
Well, it's best to have the whole target list, but I usually just end up with whatever outputs other rules depend on because new targets are constantly being added.

I'll push this when Travis finishes.

PR #139, opened 2016-06-14
Dynamically compute number of L1 client channels

zhemao:
Until now, the number of L1 client channels was set statically in the
configuration. This static configuration also assumed the same number of
cached and uncached channels per tile. As we plan to move towards
heterogenous multicore systems, this restriction should be removed.

This commit changes the generator so that number of channels per tile
can be independently set (using cde.Parameters.alterPartial).
The OuterMemorySystem will dynamically compute the number of cached and
uncached channels by summing the number of each kind of channel per core.

There are changes to all of the submodules to make them compatible with the new way of doing things, please check those as well.

### DIFF
--- a/src/main/scala/Configs.scala
+++ b/src/main/scala/Configs.scala
@@ -529,4 +529,5 @@ class EightChannelConfig extends Config(new WithNMemoryChannels(8) ++ new BaseCo
 class WithSplitL2Metadata extends Config(knobValues = { case "L2_SPLIT_METADATA" => true; case _ => throw new CDEMatchError })
 class SplitL2MetadataTestConfig extends Config(new WithSplitL2Metadata ++ new DefaultL2Config)
 
-class DualCoreConfig extends Config(new WithNCores(2) ++ new BaseConfig)
+class DualCoreConfig extends Config(
+  new WithNCores(2) ++ new WithL2Cache ++ new BaseConfig)
--- a/src/main/scala/Configs.scala
+++ b/src/main/scala/Configs.scala
@@ -178,6 +178,8 @@ class BaseConfig extends Config (
         Module(new L2BroadcastHub()(p.alterPartial({
           case InnerTLId => "L1toL2"
           case OuterTLId => "L2toMC" })))
+      case NCachedTileLinkPorts => 1
+      case NUncachedTileLinkPorts => 1
       //Tile Constants
       case BuildTiles => {
         val (rvi, rvu) =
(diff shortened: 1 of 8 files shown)


hcook:
lgtm

zhemao:
@aswaterman @colinschmidt could you take a look as well?

src/main/scala/TestConfigs.scala
```diff
@@ -53,39 +59,53 @@ class WithComparator extends Config(
...
   (pname, site, here) => pname match {
-    case NGenerators => site(NTiles)
-    case GenerateUncached => true
-    case GenerateCached => true
+    case GroundTestCachedClients => 1
+    case GroundTestUncachedClients => 1
+    case GroundTestNPTW => 0
     case MaxGenerateRequests => 128
-    case GeneratorStartAddress => site(GlobalAddrMap)("mem").start
+    case GeneratorStartAddress => BigInt("80001040", 16)
```
colinschmidt: why did this become a magic number?

src/main/scala/RocketChip.scala
```diff
@@ -172,6 +182,8 @@ class Top(topParams: Parameters) extends Module with HasTopLevelParameters {
   */
 class Uncore(implicit val p: Parameters) extends Module
     with HasTopLevelParameters {
+
```
colinschmidt: Extra white space. I wouldn't want to rerun travis for this though.

src/main/scala/Configs.scala
```diff
@@ -324,29 +325,18 @@ class BaseConfig extends Config (
 )
 class DefaultConfig extends Config(new WithBlockingL1 ++ new BaseConfig)
 
-class With2Cores extends Config(knobValues = { case "NTILES" => 2; case _ => throw new CDEMatchError })
-class With4Cores extends Config(knobValues = { case "NTILES" => 4; case _ => throw new CDEMatchError })
-class With8Cores extends Config(knobValues = { case "NTILES" => 8; case _ => throw new CDEMatchError })
+class WithNCores(n: Int) extends Config(
```
colinschmidt: Nice change!

zhemao:
Let's not merge this yet. Henry has some ongoing uncore changes that conflict with mine.

src/main/scala/TestConfigs.scala
zhemao: I didn't want it to conflict with the TohostAddr. I suppose we could compute by using the address one block away from the TohostAddr.

src/main/scala/RocketChip.scala
zhemao: I think I'll have to rebase anyway, I'll get this along the way.

src/main/scala/TestConfigs.scala
aswaterman: That sounds reasonable to me.  (Actually, TohostAddr could be rewritten as site(GlobalAddrMap)("mem").start + 0x1000 for increased clarity.)

PR #140, opened 2016-06-15
Default to Chisel 3

palmer-dabbelt:
Thanks to Donggyu (for making Verilator work) and Colin (for making Rocket Chip run under Chisel 3 in all configurations), we can now switch over to Chisel 3.  This PR cleans up a few rough edges:

* The newest Verilator is installed and used, since nobody has it and we need it.
* Travis runs Chisel 2 and Chisel 3 on all configurations
* CHISEL_VERSION=3 is now the default

I don't think we're quite ready to deprecate Chisel 2 yet, so I'm leaving all the Chisel 2 stuff in place.  This means Travis will take twice as long to test everything.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,17 +32,17 @@ addons:
 
 env:
   matrix:
-    - CONFIG=DefaultConfig
-    - CONFIG=DefaultL2Config
-    - CONFIG=RoccExampleConfig
-    - CONFIG=DualCoreConfig
-    - CONFIG=MemtestConfig
-    - CONFIG=FancyMemtestConfig
-    - CONFIG=BroadcastRegressionTestConfig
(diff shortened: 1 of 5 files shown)


aswaterman:
Increased Travis runtime is really going to hurt us.  Maybe as a compromise we could only run Chisel2 regressions against DefaultConfig, just to make sure the build isn't totally broken.

palmer-dabbelt:
After discussion during the meeting today, we're just turning off Chisel 2 builds since nobody there cares about it.

jackkoenig:
It would also be helpful if we could add rocket-chip builds to Chisel 3 (they might already be on Jenkins), and FIRRTL

ucbjrl:
Will do.

colinschmidt:
Looks good to me.

PR #141, opened 2016-06-16
Don't build the Verilator man pages

palmer-dabbelt:
These failed for Andrew earlier.  While it might be paranioa, there's
really no reason to build the man pages so we might as well not bother.


palmer-dabbelt:
@aswaterman I don't care about this, it hasn't failed for me :)

PR #148, opened 2016-06-24
Merge different configs into regression suites to reduce travis build times

zhemao:
Exactly what it says on the tin. This builds and runs several different configs in the same Travis job so that we don't have to wait for GCC to compile a dozen times.

Right now, I have the configurations separated in two, with one for the Rocket assembly tests and the other for the ground tests, but we could break this up further if necessary.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,18 +32,8 @@ addons:
 
 env:
   matrix:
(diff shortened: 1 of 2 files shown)


hcook:
lgtm

PR #152, opened 2016-07-05
PRINTF_COND was flipped for Verilator

palmer-dabbelt:


aswaterman:
I fixed this in 4a8e6c773a367204e787d4697a570b5f5220e448

palmer-dabbelt:
OK, thanks -- I guess we're just behind master a bit.

PR #153, opened 2016-07-07
Fix converter and L2

zhemao:
This fixes an issue with the L2WritebackUnit which was causing it to incorrectly get voluntary releases allocated to it.

It also adds buffering and locking to the TL -> AXI converter so that read responses cannot be interleaved with write responses.


hcook:
lgtm

PR #158, opened 2016-07-08
Refactor groundtest configuration

zhemao:
This allows us to instantiate both Memtest and NastiConverterTest together in the same design and have them operate simultaneously on the same blocks. This mocks the kind of traffic that will be generated in an SoC with PCIe bus mastering.


hcook:
lgtm

PR #160, opened 2016-07-09
Merge in rocket, junctions and uncore using "git subtree"

palmer-dabbelt:
We talked about this a month ago and I was supposed to do it, but forgot while I was at ISCA.  This is a _huge_ change, so if now isn't a good time (right before the workshop) then just close it -- it only took a few minutes.

You shouldn't squash this PR, as the history will disappear.


zhemao:
Superseded by #178.

PR #162, opened 2016-07-11
Star chip port

terpstra:
Backport AXI input and multiple output patches from star-chip.

### DIFF
--- a/src/main/scala/Configs.scala
+++ b/src/main/scala/Configs.scala
@@ -39,7 +39,7 @@ class BaseConfig extends Config (
     lazy val globalAddrMap = {
       val memBase = 0x80000000L
       val memSize = 0x10000000L
-      val io = AddrMap((AddrMapEntry("int", internalIOAddrMap) +: site(ExtMMIOPorts).entries):_*)
+      val io = new AddrMap(AddrMapEntry("int", internalIOAddrMap) +: site(ExtMMIOPorts).entries)
       val addrMap = AddrMap(
         AddrMapEntry("io", io),
         AddrMapEntry("mem", MemRange(memBase, memSize, MemAttr(AddrMapProt.RWX, true))))
--- a/src/main/scala/Configs.scala
+++ b/src/main/scala/Configs.scala
@@ -225,6 +225,7 @@ class BaseConfig extends Config (
       case NExtMMIOAXIChannels => 0
       case NExtMMIOAHBChannels => 0
       case NExtMMIOTLChannels  => 0
+      case NExtBusAXIChannels => 0
       case PLICKey => PLICConfig(site(NTiles), site(UseVM), site(NExtInterrupts), 0)
       case DMKey => new DefaultDebugModuleConfig(site(NTiles), site(XLen))
       case FDivSqrt => true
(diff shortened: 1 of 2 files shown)


src/main/scala/Configs.scala
```diff
@@ -38,25 +38,10 @@ class BaseConfig extends Config (
...
-      if (nMMIOChannels > 0) {
-        val extIOBase = 0x60000000L
-        val extIOSize = 0x20000000L
-        ioMap += AddrMapEntry("ext", MemRange(extIOBase, extIOSize, MemAttr(AddrMapProt.RWX)))
-        Dump("IO_BASE", extIOBase)
-        Dump("IO_SIZE", extIOSize)
-      }
-
+      val memSize = 0x10000000L
+      val io = AddrMap((AddrMapEntry("int", internalIOAddrMap) +: site(ExtMMIOPorts).entries):_*)
```
zhemao: You can avoid using the splat by just instantiating the class directly.

    val io = new AddrMap(AddrMapEntry("int", internalIOAddrMap) +: site(ExtMMIOPorts).entries)

src/main/scala/Configs.scala
```diff
@@ -231,9 +216,16 @@ class BaseConfig extends Config (
...
+      case ExtMMIOPorts => AddrMap()
+/*
+        AddrMap(
+          AddrMapEntry("cfg", MemRange(0x50000000L, 0x04000000L, MemAttr(AddrMapProt.RW))),
+          AddrMapEntry("ext", MemRange(0x60000000L, 0x20000000L, MemAttr(AddrMapProt.RWX))))
+*/
       case NExtMMIOAXIChannels => 0
       case NExtMMIOAHBChannels => 0
       case NExtMMIOTLChannels  => 0
+      case NExtBusAXIChannels => 0
```
zhemao: Shouldn't this be a derived parameter from ExtMMIOPorts? Also, if we have this, can't we get rid of NExtMMIOAXIChannels?

src/main/scala/RocketChip.scala
```diff
@@ -225,25 +230,29 @@ class Uncore(implicit val p: Parameters) extends Module
...
+      if (mmio_axi_start <= i && i < mmio_axi_end) {
+        TopUtils.connectTilelinkNasti(io.mmio_axi(i-mmio_axi_start), ports(i))
+      } else if (mmio_ahb_start <= i && i < mmio_ahb_end) {
+        val ahbBridge = Module(new AHBBridge(true))
+        io.mmio_ahb(i-mmio_ahb_start) <> ahbBridge.io.ahb
+        ahbBridge.io.tl <> ports(i)
+      } else if (mmio_tl_start <= i && i < mmio_tl_end) {
+        TopUtils.connectTilelink(io.mmio_tl(i-mmio_tl_start), ports(i))
+      } else {
+        TopUtils.connectTilelinkNasti(Module(new NastiErrorSlave).io, ports(i))
```
zhemao: This part shouldn't be necessary. The require above already makes sure that the number of interconnect ports matches the number of external MMIOs.

PR #182, opened 2016-07-29
Update README.md

mwachs5:
- List things that are no longer submodules as subpackages instead
- clean up some formatting issues


aswaterman:
I'm about to do a big PR, so I'll fold these commits into it.

PR #185, opened 2016-08-01
Allow aggregate CONFIG on Command Line

mwachs5:
This allows composing Config classes on the command line. Underscores are translated into the ++ operator.

As a fairly extreme example,

make CONFIG=TinyConfig

is equivalent to 

make CONFIG=WithRV32_WithSmallCores_WithStatelessBridge_BaseConfig

This does not support parameterized Config constructors (e.g. WithNCores).

### DIFF
--- a/src/main/scala/Testing.scala
+++ b/src/main/scala/Testing.scala
@@ -177,14 +177,26 @@ object TestGenerator extends App {
   val projectName = args(0)
   val topModuleName = args(1)
   val configClassName = args(2)
-  val config = try {
-      Class.forName(s"$projectName.$configClassName").newInstance.asInstanceOf[Config]
+
+  val aggregateConfigs = configClassName.split('_').reverse
+
+  var finalConfig = new Config()
+  var ii = 0;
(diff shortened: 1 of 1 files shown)


src/main/scala/Testing.scala
```diff
@@ -177,14 +177,26 @@ object TestGenerator extends App {
...
   val topModuleName = args(1)
   val configClassName = args(2)
-  val config = try {
-      Class.forName(s"$projectName.$configClassName").newInstance.asInstanceOf[Config]
+
+  val aggregateConfigs = configClassName.split('_').reverse
+
+  var finalConfig = new Config()
+  var ii = 0;
+  for (ii <- 0 to (aggregateConfigs.length - 1)) {
```
zhemao: This would probably be better expressed as a foldLeft than as a for loop. Something like

    aggregateConfigs.foldLeft(new Config()) { case (finalConfig, configClass) =>
        val currentConfig = ...
        configClass ++ finalConfig
    }

You could even do a foldRight, in which case you can get rid of the `.reverse`.

PR #190, opened 2016-08-04
Add asynchronous FIFOs as an option for clock domain crossings.

ben-k:
This extends AsyncDecoupledTo and AsyncDecoupledFrom to instantiate
async fifos if requested by setting depth > 0.  The modules use
the standard technique of gray-coding the read and write pointers
for synchronization, and are ported from verilog implementations
previously used for this purpose.

Note that only FIFO depths which are powers of 2 are supported.

### DIFF
--- a/junctions/src/main/scala/crossing.scala
+++ b/junctions/src/main/scala/crossing.scala
@@ -86,6 +86,76 @@ class AsyncHandshakeSink[T <: Data](gen: T, sync: Int, clock: Clock, reset: Bool
   }
 }
 
+class AsyncFIFOSource[T <: Data](gen: T, sync: Int, depth: Int, clock: Clock, reset: Bool)
+    extends Module(_clock = clock, _reset = reset) {
+  val io = new Bundle {
+    // These come from the source clock domain
+    val enq  = Decoupled(gen).flip()
+    // These cross to the sink clock domain
+    val bits = gen.cloneType.asOutput
+    val wptr = UInt(OUTPUT, width = log2Up(depth)+1)
+    val rptr = UInt(INPUT, width = log2Up(depth)+1)
+    val raddr = UInt(INPUT, width = log2Up(depth))
+  }
+  require (depth >= 4) // Required based on how the gray code is currently checked
+  require (isPow2(depth))
+
+  val rptr_sync = ShiftRegister(io.rptr, depth)
+
+  val wbinnext = Wire(UInt(width = log2Up(depth)+1))
+  val wgraynext = Wire(UInt(width = log2Up(depth)+1))
+  val wbin = Reg(wbinnext)
+  val waddr = wbin(log2Up(depth)-1,0)
+
+  val wfull_val = (
+    (wgraynext(log2Up(depth)) =/= rptr_sync(log2Up(depth))) &
+    (wgraynext(log2Up(depth)-1) =/= rptr_sync(log2Up(depth)-1)) &
+    (wgraynext(log2Up(depth)-2,0) === rptr_sync(log2Up(depth)-2,0)) )
+  val wfull = Reg(wfull_val, init=Bool(true))
+
+  wbinnext := wbin + (io.enq.valid & !wfull)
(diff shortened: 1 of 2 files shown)


terpstra:
Hi. I already added support for this at the same time as I wrote the original. Find it here:
https://github.com/ucb-bar/junctions/commit/22115938c9a18836936409e9895b125490433ba1

However, I didn't merge that in for the same reason I won't merge this commit: Mem and SeqMem are unsafe when you use them from two clock domains. We need a way to explicitly required SRAMs that are dual-clock safe before a patch like this will work.

ben-k:
Thanks for explaining - I was unaware of the abandoned commit in junctions.

I don't have a good sense of the intended use of Mem vs. SeqMem when it comes to SRAM mapping - my (apparently incorrect) understanding was that only SeqMem would map to SRAMs as in Chisel2.  This use case should be safe if the memory is implemented in flip-flops as the pointer synchronization latency guards against any ill-timed reads.  (It seems like a dual-ported SRAM should also be safe to use, but I haven't though that case through carefully.)

If the Mem were coded as a Vec of Reg, the intent not to map into an SRAM would be clear.  Would that be an acceptable solution?

terpstra:
On Wed, Aug 3, 2016 at 6:08 PM, Ben Keller <notifications@github.com> wrote:

> I don't have a good sense of the intended use of Mem vs. SeqMem when it
> comes to SRAM mapping - my (apparently incorrect) understanding was that
> only SeqMem would map to SRAMs as in Chisel2.
>
I believe that the intent is that they could both potentially be turned
into an SRAM in an ASIC.
Both can already be inferred as a block memory by FPGA tools from the
verilog output by chisel3 today.

> This use case should be safe if the memory is implemented in flip-flops as
> the pointer synchronization latency guards against any ill-timed reads. (It
> seems like a dual-ported SRAM should also be safe to use, but I haven't
> though that case through carefully.)
>
> I agree that it is logically safe. However, depending on how the SRAM is
implemented, it may not be safe to connect it to two clock domains...
(analog crap)

> If the Mem were coded as a Vec of Reg, the intent not to map into an SRAM
> would be clear. Would that be an acceptable solution?
>
Sadly, no. It might be safe on an ASIC, but on an FPGA it is generally
never safe to take inputs from two clock domains through a LUT. Using a
Vec(Reg) would require a mux that takes the index from one clock domain and
the registers from another clock domain. I wrote the Handshake version for
our VC707 board to cross between PCIe and rocket, so this matters to me.

Mem and SeqMem just need a way for us to be explicit about our
requirements. Once we have that, we'll merge some sort of ASyncFifo.
Fortunately, the users of the From/To API don't need to care about what is
under the hood. At some point it will just get faster.

terpstra:
Oh, I should also mention that both the Handshake and FIFO versions
have another problem: reset.

Currently, chisel3 creates registers with synchronous reset.
If one clock starts before the other, one side might not yet have been reset.

Currently this works for us only because a quirk of FPGA synthesis is
that a register is set on FPGA power-on to its reset value.

The proper fix is that we use asynchronously reset registers, where
the reset is deasserted synchronous to the clock domain.

jwright6323:
Is there actually a plan to add asynchronous resets to Chisel?

terpstra:
I've been pressuring the people at SiFive to just make all resets
asynchronous... no luck yet.

IMO the danger of asynchronous resets is just that people actually use them
asynchronously. Just deassert them synchronous to the register's clock and
run clock recovery during timing analysis...

ben-k:
Thanks for clarifying - I definitely agree that this code is not FPGA-safe and shouldn't be merged.  It would be really nice to have Chisel support for asynchronous resets as an option.

We may use still use this (or ucb-bar/junctions@2211593, which looks cleaner) for our fork for ASIC design, though.  We ensure that:

* Asserting chip reset causes every clock domain to have a toggling clock
* Reset is held for a long time in every clock domain before it is deasserted

Given these constraints, I think synchronous reset should be safe for this circuit.

PR #193, opened 2016-08-05
Generate BootROM contents from assembly code

zhemao:
The current way of hardcoding the BootROM contents with hand-assembled machine code is probably not the best. This change makes it so that we can build the bootloader in the boot ROM from assembly and load it into the Chisel generator as a hex file. The config string is then tacked onto the end of the code and the config string address is patched into the proper place.


bootrom/Makefile
```diff
@@ -0,0 +1,15 @@
...
+bootrom_ld = bootrom-link.ld
+bootrom_elf = bootrom.elf
+bootrom_hex = bootrom.hex
+bootrom_depth = 16
+
+GCC=riscv64-unknown-elf-gcc
+ELF2HEX=elf2hex
+
+$(bootrom_hex): $(bootrom_elf)
+	$(ELF2HEX) 4 $(bootrom_depth) $(bootrom_elf) > $@
```
aswaterman: Can you see if Tim's approach using objcopy and xxd (https://github.com/riscv/riscv-isa-sim/blob/master/debug_rom/Makefile) will work here?  I'd rather not increase the use of elf2hex since it's a nonstandard tool, and this Makefile already depends on binutils being installed.

src/main/scala/RocketChip.scala
```diff
@@ -271,25 +274,19 @@ class Uncore(implicit val p: Parameters) extends Module
...
-    rom.putInt(0x0000006f)                        // loop forever
-    rom.putInt(0)                                 // reserved
-    rom.putInt(0)                                 // reserved
-    rom.putInt(configStringAddr)                  // pointer to config string
-    rom.putInt(0)                                 // default trap vector
-    rom.putInt(0)                                 //   ...
-    rom.putInt(0)                                 //   ...
-    rom.putInt(0)                                 //   ...
-
+    rom.putInt(12, configStringAddr)
```
aswaterman: Maybe we should assert that (romdata.size < 4 || romdata(3) == 0) here, to make sure the boot ROM writer doesn't accidentally put some useful data where the config string address ends up?

src/main/scala/Configs.scala
```diff
@@ -315,6 +315,7 @@ class BaseConfig extends Config (
           dataBeats = innerDataBeats,
           dataBits = site(CacheBlockBytes) * 8)
       }
+      case BootROMFile => "./bootrom/bootrom.hex"
```
aswaterman: Does this assume that Chisel is invoked in the root directory of the repo?  Maybe that's OK...

bootrom/default-bootrom.S
```diff
@@ -0,0 +1,13 @@
...
+  // This boot ROM doesn't know about any boot devices, so it just spins,
+  // waiting for the debugger to load a program and change the PC.
+  j _start // reset vector
+  .dword 0 // reserved
+  .dword 0 // reserved
+  .dword 0 // pointer to config string
+  .dword 0 // default trap vector
+  .dword 0
+  .dword 0
+  .dword 0
```
aswaterman: These should all be .word, not .dword

PR #194, opened 2016-08-05
Add firrtl build dependencies

palmer-dabbelt:
Without this when I update firrtl the new version doesn't get built, so
my build is constantly failing.


aswaterman:
You should merge once it passes.

We want to build FIRRTL as part of rocket-chip's sbt build, as we do with chisel3, but this is a good intermediate step.

PR #195, opened 2016-08-08
Remove empty Verilog from FIRRTL

palmer-dabbelt:
FIRRTL generates empty Verilog files when it fails.  This is a huge
pain, since it proceeds to screw up the build process (make doesn't know
those empty Verilog files are invalid).  This patch removes FIRRTL's
output when FIRRTL fails.


palmer-dabbelt:
This is obsoleted by https://github.com/ucb-bar/firrtl/pull/223

PR #196, opened 2016-08-08
Add JTAG DTM and test support in simulation

mwachs5:
This adds the Verilog for the JTAG Debug Transport Module & Asynchronous FIFO, and infrastructure to test it. The testing infrastructure is intended to be driven by an OpenOCD instance which connects to the VCS simulation over a socket interface. The riscv-tests/debug/gdbserver.py automates the communication between all the different processes.

The Client JTAG VPI code is from @fjullien https://github.com/fjullien/jtag_vpi.

Questions to be addressed before accepting this PR:
- how to include openOCD install/download as part of the make flow
- what small subset of tests to be run automatically
- Do we need emulator support before adding this
- Are we happy with this factoring of the Makeflow (adding "TB=rocketDTMTestHarness" is the way to indicate you want to build the sim this way)
--------------------------------------------------------------------
To compile the VCS simulation which includes the JTAG DTM:

cd vsim
make CONFIG=<AnyRV32Config> TB=rocketDTMTestHarness

to run (simulation will be listening for incoming connections):
./simv-DTM-Top-<AnyRV32Config> +jtag_vpi_enable

--------------------------------------------------------------------
To run an actual test using GDB->OpenOCD->JTAG VPI -> Sim:

Install openocd with --enable-jtag_vpi flag at the ./configure step.

riscv-tools/riscv-tests/debug/gdbserver.py --freedom-e300-sim --cmd "/path/to/openocd/install/bin/openocd --s /path/to/openocd/install/share/openocd/scripts" --run "path/to/simv-DTM-Top-<AnyRV32Config>" SimpleRegisterTest.test_s0

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -22,3 +22,6 @@
 [submodule "firrtl"]
 	path = firrtl
 	url = https://github.com/ucb-bar/firrtl.git
+[submodule "openocd"]
+	path = openocd
+	url = http://github.com/sifive/openocd.git
--- a/csrc/jtag_vpi.c
+++ b/csrc/jtag_vpi.c
@@ -1,0 +1,419 @@
+/*
+ * TCP/IP controlled VPI JTAG Interface.
+ * Based on Julius Baxter's work on jp_vpi.c
+ *
+ * Copyright (C) 2012 Franck Jullien, <franck.jullien@gmail.com>
+ * All rights reserved.
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions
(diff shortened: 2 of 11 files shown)


mwachs5:
My plan to bring this up to date is to make a new PR which:

-creates a new TestHarness class (vs messing with the internals of the current one) that can drive JTAG

-adds a configuration option to Top which will instantiate the jtagDTM (and optional synchronizers) and bring out the JTAG pins instead of the debug bus.

mwachs5:
This is replaced by PR #213

PR #197, opened 2016-08-09
Add sbt pack plugin

zhemao:
The sbt pack plugin allows us to dump out jar files for all of the rocketchip libraries (chisel, cde, uncore, junctions, rocket, groundtest, and rocketchip), making it easier for us to use them in external projects.


hcook:
lgtm

edcote:
@zhemao I have been having problems with this plugin and dumping out source jar files. Have you considered this use case?

PR #199, opened 2016-08-10
Refactor top-level to separate coreplex from platform

zhemao:
This commit refactors things so that the coreplex is a distinct module.

The tiles, L2 agents, and tilelink networks, and internal devices (PLIC, PRCI, BootROM, Debug) reside in the coreplex.

All converters, external-facing devices, and clock crossers now reside outside.

### DIFF
--- a/src/main/scala/Configs.scala
+++ b/src/main/scala/Configs.scala
@@ -312,7 +312,7 @@ class BaseConfig extends Config (
         TileLinkParameters(
           coherencePolicy = new MICoherence(
             new NullRepresentation(site(NBanksPerMemoryChannel))),
-          nManagers = globalAddrMap.subMap("io").flatten.size,
+          nManagers = globalAddrMap.subMap("io").numSlaves,
           nCachingClients = 0,
           nCachelessClients = 1,
           maxClientXacts = 4,
--- a/src/main/scala/Configs.scala
+++ b/src/main/scala/Configs.scala
@@ -4,27 +4,18 @@ package rocketchip
 
 import Chisel._
 import junctions._
-import uncore.tilelink._
-import uncore.coherence._
+import rocket._
 import uncore.agents._
+import uncore.tilelink._
 import uncore.devices._
 import uncore.converters._
-import rocket._
-import rocket.Util._
-import groundtest._
+import coreplex._
 import scala.math.max
 import scala.collection.mutable.{LinkedHashSet, ListBuffer}
 import DefaultTestSuites._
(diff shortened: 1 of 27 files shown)


src/main/scala/Configs.scala
```diff
@@ -34,15 +34,18 @@ class BaseConfig extends Config (
...
-      entries ++= site(ExtraDevices).map(_.addrMapEntry)
       new AddrMap(entries)
     }
     lazy val globalAddrMap = {
       val memBase = 0x80000000L
       val memSize = 0x10000000L
-      val io = new AddrMap(AddrMapEntry("int", internalIOAddrMap) +: site(ExtMMIOPorts).entries)
+
+      val intern = AddrMapEntry("int", internalIOAddrMap)
+      val extern = AddrMapEntry("ext", MemRange(0x50000000L, 0x30000000L, MemAttr(AddrMapProt.RWX)))
```
aswaterman: I concede the additional address map is probably necessary, but let's not duplicate magic constants (50000000 appears below).

aswaterman:
high-level observation: it seems strange that the coreplex and platform are configured with the same Config.

src/main/scala/Configs.scala
zhemao: We could probably derive this entry from the ExtAddrMap.

zhemao:
But they have largely separate configuration parameters. And the few things that are shared need to be the same (ExportMMIOPort and ExportBusPort). 

You could say the same things about the Tile or the L2Agent.

src/main/scala/Configs.scala
zhemao: Also, I think not having separate address maps would in fact be better, since it would allow finer-grain permissions in the Tile's TLB.

What if we added an optional boolean argument to the AddrMap like "collapse". Then we make the interconnect generator not recurse into it (treat it like a MemRange) if it's used as a submap.

aswaterman:
I would say that about Tile, etc.

There is no utility in separating the coreplex RTL unless it is separately configurable. It doesn't make sense to have a hierarchy for the RTL while having a flat namespace.

mwachs5:
One of the main motivations behind the commit that allowed merging of Configs on the command line was to allow this sort of separation between platform and coreplex. The idea was you could have a config something like DefaultZynqConfig_DefaultRV32Config.

aswaterman:
It's good the configs are less entangled.  It's still flat, though, just separated out into another file.  We should merge this and move on, but continue thinking about how to improve hierarchical configurations.

PR #200, opened 2016-08-13
Add STOP_COND guard during simulation reset

ben-k:
The generated verilog now includes a STOP_COND guard on conditions
that can cause the test to exit.  Enabling this flag during reset
prevents inadvertant exit due to spurious glitching near the start
of execution, before reset has had time to propagate everywhere.

I don't know how to add this in verilator, but presumably that
would be worth doing as well.


aswaterman:
thanks, i'm reworking some testbench stuff right now, so i'll incorporate this (and a fix to verilator) in the process

PR #201, opened 2016-08-16
Strip DMA and RoCC CSRs out of rocket and uncore

zhemao:
With all the recent refactoring of rocket chip, I think it's finally time to get rid of the old DMA/copy accelerator code in favor of a cleaner approach. This gets rid of all the additions I put in before, including the changes to the RoCC interface. I'm in the process of refactoring and moving to a separate package.


aswaterman:
I concur; that should be quite a bit easier to do with the recent changes

PR #202, opened 2016-08-16
Write test harness in Chisel

aswaterman:
This is an unavoidably invasive commit, because it affects the unit tests
(which formerly exited using stop()), the test harness Verilog generator
(since it is no longer necessary), and the DRAM model (since it is no
longer connected).  However, this should substantially reduce the effort of
building test harnesses in the future, since manual or semi-automatic
Verilog writing should no longer be necessary.  Furthermore, there is now
very little duplication of effort between the Verilator and VCS test
harnesses.

This commit removes support for DRAMsim, which is a bit of an unfortunate
consequence.  The main blocker is the lack of Verilog parameterization for
BlackBox.  It would be straightforward to revive DRAMsim once support for
that feature is added to Chisel and FIRRTL.  But that might not even be
necessary, as we move towards synthesizable DRAM models and FAME-1
transformations.


README.md
```diff
@@ -117,8 +117,8 @@ submodule.
 Here's a look at all the git submodules that are currently tracked in
 the rocket-chip repository:
 
-* **chisel2**
-([https://github.com/ucb-bar/chisel](https://github.com/ucb-bar/chisel)):
+* **chisel3**
+([https://github.com/ucb-bar/chisel3](https://github.com/ucb-bar/chisel3)):
 At Berkeley, we write RTL in Chisel. For those who are not familiar
 with Chisel, please go take a look at
 [http://chisel.eecs.berkeley.edu](http://chisel.eecs.berkeley.edu). We
```
mwachs5: Does this link apply to chisel3? One of the main things you find at that link is the chisel-tutorial, is the chisel-tutorial now in chisel3?

ucbjrl: > On Aug 16, 2016, at 9:20 AM, mwachs5 <notifications@github.com> wrote:
> 
> In README.md:
> 
> > @@ -117,8 +117,8 @@ submodule.
> >  Here's a look at all the git submodules that are currently tracked in
> >  the rocket-chip repository:
> >  
> > -* **chisel2**
> > -([https://github.com/ucb-bar/chisel](https://github.com/ucb-bar/chisel)):
> > +* **chisel3**
> > +([https://github.com/ucb-bar/chisel3](https://github.com/ucb-bar/chisel3)):
> >  At Berkeley, we write RTL in Chisel. For those who are not familiar
> >  with Chisel, please go take a look at
> >  [http://chisel.eecs.berkeley.edu](http://chisel.eecs.berkeley.edu). We
> 
> Does this link apply to chisel3? One of the main things you find at that link is the chisel-tutorial, is the chisel-tutorial now in chisel3?
> 
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub, or mute the thread.
> 

There is a Chisel3 version of the tutorials in the chisel3 branch of chisel-tutorial. Once we have an official Chisel3 release, we'll make an announcement on the web site, and that branch will become master.

zhemao:
Wait, so why would using dramsim require BlackBox parameterization? Each instantiation of the mm_t should be identical, no?

aswaterman:
@zhemao Things like address and data width need to make it into the Verilog, and the parameter-dump system's a bad way to do that.

aswaterman:
@mwachs5 I didn't read that very carefully, but I agree linking to the outdated documentation and wrong branch of the tutorials could result in some confusion... as @ucbjrl said, that'll be fixed soon, but maybe in the mean time I should add a note saying the info on the web site is out of date.

zhemao:
I was thinking it was that. I guess dumping things out to a file, transforming it into a verilog header and then including that in the blackbox verilog is a rather convoluted way of doing it.

src/main/scala/TestHarness.scala
```diff
@@ -0,0 +1,116 @@
...
+  val io = new Bundle {
+    val success = Bool(OUTPUT)
+  }
+  val dut = Module(new Top(p))
+
+  // This test harness isn't especially flexible yet
+  require(dut.io.mem_clk.isEmpty)
+  require(dut.io.mem_rst.isEmpty)
+  require(dut.io.mem_ahb.isEmpty)
+  require(dut.io.mem_tl.isEmpty)
```
zhemao: There are TileLink and AHB test RAMs in uncore.devices. Perhaps we can use those so that we don't have to require mem_ahb and mem_tl be empty. Might be good to move the Nasti RAM there as well.

aswaterman:
@zhemao agree, let's move toward that goal (in a separate commit :-))

RSpliet:
@aswaterman 
> This commit removes support for DRAMsim, which is a bit of an unfortunate
consequence. The main blocker is the lack of Verilog parameterization for
BlackBox. It would be straightforward to revive DRAMsim once support for
that feature is added to Chisel and FIRRTL. But that might not even be
necessary, as we move towards synthesizable DRAM models and FAME-1
transformations.

Could you perhaps elaborate a bit more on the exact limitations with BlackBox that prevent DRAMSim integration? Are these limitations still in place?

colinschmidt:
Its still a work in progress but @zhemao has been working with dramsim in rocket-chip via testchipip see https://github.com/ucb-bar/testchipip/tree/dramsim

PR #205, opened 2016-08-17
make memsize configurable

davidbiancolin:
Needed for rocketchip emulations that have access to more memory. 

I don't know if this is how ya'll want to do this in the long run but maybe it's ok for now? 

[PR this for real this time]


zhemao:
Looks good to me. Go ahead and merge it once the tests pass.

PR #206, opened 2016-08-18
Allow coreplex to take in more than 1 bus port

zhemao:
In the coreplex refactoring, I only had one bus port (TileLink master) exported out of the coreplex. This was fine because we only had the AXI bus ports, and those could be collected with an AXI arbiter. But I now realize that we might want to add different kinds of bus masters. Instead of having to instantiate another TileLink arbiter in the top-level, it would be better just to extend the l1tol2 network.


hcook:
lgtm

PR #208, opened 2016-08-18
connect top level interrupts to coreplex

colinschmidt:
seems like a simple omission as it is connected to by both the test harness and the coreplex itself


zhemao:
Whoops. Thanks for fixing this.

PR #209, opened 2016-08-19
Clean up our use of UIntIsOneOf

zhemao:
I got a bit annoyed at how the UIntIsOneOf and AugmentedSeq implicits were duplicated in rocket and uncore. Also how there was both a uncore.Util and uncore.util package.

This merges the two packages together and gets rid of the repeated definition in rocket.

I also replaced everywhere we use (state === for) || (state === bar) ... This touches a lot of files, so the diff is a bit large.


hcook:
lgtm

PR #211, opened 2016-08-19
Un-flatten some Rocket configuration parameters

aswaterman:
... in the FPU, MulDiv, D$.

Also, remove parameters for immutable ISA constants.


colinschmidt:
Looks fine to me. Thanks for the heads up.

PR #213, opened 2016-08-19
Adds Logic & test support for JTAG implementation of Debug Transport Module.

mwachs5:
This adds synthesizable JTAG DebugTransportModule functionality in Top, and test support (currently VCS only). 
 
    - The DebugTransportModuleJtag is written in Verilog. It probably could be rewritten in
      Chisel except for one negative edge clocking requirement.
    - For real implementations, the AsyncDebugBusTo/From is insufficient. This commit
      includes code to wire them up in cases where they are used, but because they are 
      not reset asynchronously, a Verilog 'AsyncMailbox' is used when p(AsyncDebug) is false.
    - This commit differs significantly from the earlier attempt. Now, the
      DTM and synchronizer is instantiated within Top, as it is a real piece of
      hardware (vs. test infrastructure).
    -TestHarness takes a parameter vs. creating an entirely new TestHarness class.
    It does not make sense to instantiate the previous TestHarness when p(IncludeJtagDTM) 
    is true.
    
**To build Verilog which includes the JtagDTM within Top:**

`make CONFIG=WithJtagDTM_...`
    
**To test using gdb->OpenOCD->jtag_vpi->Verilog:**

First, install openocd (included in this commit)
    
```
    cd openocd
    ./bootstrap
    ./configure --prefix=$OPENOCD --enable-jtag-vpi
    make
    make install
```    

Then to run a simulation:
    
 ```
   $(ROCKETCHIP)/riscv-tools/riscv-tests/debug/gdbserver.py \
      --run ./simv-TestHarness-WithJtagDTM_... \
      --cmd="$OPENOCD/bin/openocd --s $OPENOCD/share/openocd/scripts/" \
      --freedom-e300-sim \
      SimpleRegisterTest.test_s0 
```   

On a 64-bit core:
    
```
    $(ROCKETCHIP)/riscv-tools/riscv-tests/debug/gdbserver.py \
      --run ./simv-TestHarness-WithJtagDTM_... \
      --cmd="$OPENOCD/bin/openocd --s $OPENOCD/share/openocd/scripts/" \
      --freedom-u500-sim \
      SimpleRegisterTest.test_s0

```


mwachs5:
With @zhemao 's change that moved everything around, the scala files I added will need to be moved. I wasn't sure if they were in the right place anyway (jtag.scala in junctions, and DebugTransport.scala in src/main/scala/)

aswaterman:
This looks good to me.  I agree with the TestHarness strategy; it should be 1:1 with the module it's testing, and so should adhere to the same parameterization.

mwachs5:
Cool. Do you want me to do something to add a  test into the Travis regression? Or is this OK for now?

aswaterman:
I'd rather not put openocd into the travis testing path; maybe we should do that with internal jenkins regression testing?

mwachs5:
I added a target to the regression/Makefile which would allow someone/other automated testing environments  to easily run these tests. I won't add it to the Travis regression since it depends on VCS. 

Assuming this version passes Travis, I will merge.

PR #216, opened 2016-08-20
Refactor extra devices builder

zhemao:
Before we had a sequence of builders, one for each device.

But really we just want one builder for each of them, since we have to define all of them in the same place anyway.

This makes it possible to connect the devices to each other.


zhemao:
@mwachs5 @aswaterman Can I merge this once the Travis tests pass again? I want to get this in before classes start again.

aswaterman:
likewise

PR #217, opened 2016-08-20
Add a queue for timing QoR between L2->MMIO network

mwachs5:


aswaterman:
lgtm.  hopefully this will obviate future timing closure problems on this very-un-critical path.

colinschmidt:
Is it worth using the TileLinkDepths and TileLinkEnqueuer? Its how we've added buffering other places in this file

aswaterman:
I thought those only existed for cached TL (i.e., all 5 channels).  Probably worth creating a utility for this, following the next revision of TL.

PR #220, opened 2016-08-23
Work-around for current Scala compiler "structural type loses implici…

ucbjrl:
…ts".

Running rocket-chip through the chisel3 gsdt branch which supports stricter connection checks and uses implicit definitions to deal with "old" direction overrides, exposed a possible bug in the Scala compiler.

    [error] .../src/main/scala/uncore/devices/Prci.scala:27: value asOutput is not a member of uncore.devices.PRCIInterrupts{val mtip: chisel3.core.Bool; val msip: chisel3.core.Bool}
    [error] possible cause: maybe a semicolon is missing before `value asOutput'?
    [error]   }.asOutput
    [error]     ^
    [error] one error found
    [error] (uncore/compile:compileIncremental) Compilation failed

This change isn't strictly required for current chisel3 code, but is being submitted in anticipation of an eventual merge of the gsdt branch prior to a compiler fix.


aswaterman:
I'm merging this as part of another PR; thanks, Jim.

PR #224, opened 2016-08-25
add commit hash for openocd to fix issue

colinschmidt:
The previous addition of this didn't have a commit hash for the openocd submodule and so git was very confused when I tried to submodule init. This fixes that issue by adding the commit hash but I'm not sure if you guys had a specific commit in mind that works. @mwachs5


mwachs5:
My intention was to remove the submodule. I added a test to the regression which just pulls the latest from github/sifive/openocd if you actually want to run the test.

mwachs5:
Ah, I see that it is sticking around in .gitmodules. I think the correct fix is to remove it from .gitmodules.

colinschmidt:
Ah ok. I saw the .gitmoudle listing and thought it was just added strangely. I'll close this and open the one to delete it.

PR #225, opened 2016-08-25
Remove openocd from .gitmodules

mwachs5:
I don't believe we actually want to include openocd as a submodule. I had added commands to install it as part of the regression.

@aswaterman correct?


aswaterman:
sounds good; it's consistent with verilator

PR #227, opened 2016-08-25
Streamline the Generator App and associated utilities. 

hcook:
Remove deprecated call to chiselMain and useless Chisel2 args. Update arguments to sbt run.

### DIFF
--- a/emulator/Makefrag-verilator
+++ b/emulator/Makefrag-verilator
@@ -11,13 +11,11 @@ verilog_debug = $(generated_dir_debug)/$(MODEL).$(CONFIG).v
 
 $(generated_dir)/%.$(CONFIG).fir $(generated_dir)/%.$(CONFIG).prm $(generated_dir)/%.$(CONFIG).d: $(chisel_srcs) $(bootrom_img)
 	mkdir -p $(dir $@)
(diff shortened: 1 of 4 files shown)


davidbiancolin:
How do we feel about moving some of the generation utilities into CDE instead of keeping them in rocketchip? Much of this is useful outside of the context of rocketchip and it is closely coupled with the parameterization system.

hcook:
I guess we could add some functionality to CDE to dump out files in a certain place with a certain name, but I don't really see how that would save use more than a few lines of code. The rest is pretty tightly coupled to our particular sbt run argument parsing and file naming requirements.

davidbiancolin:
It's not so much the number of lines of code that i'm concerned about, it's the product of the number of generators and those lines. I feel like i end up implementing similar functionality across a basically all of my standalone generators. I'd assume that when SiFive has multiple chip/SoC generators, you guys might want a canonical way to elaborate and emit the design with accompanying test suites. 

In any case, this is much cleaner than what was there.

hcook:
Yeah we are only just beginning to revamp that process; I was planning to make the rocketchip.Generator the common baseline and then specialize on top of that. Maybe it should actually become cde.Generator, but I wasn't confident enough about the standardization of what is being dumped across CDE-using projects to push it that far.

PR #229, opened 2016-08-25
Use a better iterator inside the DCache

ben-k:
`p` is already used for the implicit parameters, so we should really avoid using it for other things.


zhemao:
Looks good to me. I've rebased it on master and will merge it in once the Travis push passes again.

PR #231, opened 2016-08-27
Get in line with FIRRTL randomization flag changes

zhemao:
Basically the same as the previous PR that got dropped, except this turns on all the randomization options by default. Necessary to get up to date with the latest FIRRTL changes.


aswaterman:
I see... thanks!

PR #234, opened 2016-08-30
Add address map entries for exported mmio port.

zizztux:
There is a problem that external address map is not included when it is configured with ExtMMIOPorts and without ExtraDevices.
I fixed it.


zhemao:
Whoops. That was an oversight on my part. Thanks for finding this.

PR #237, opened 2016-09-01
Allow configuration to be in separate project from test harness

zhemao:
It may be the case that you want the configuration project to be different from the test harness project. This change allows that to happen by specifying a separate variable CFG_PROJECT, which is by default the same as PROJECT.


ben-k:
Can you give an example of the intended usage of this?

PR #240, opened 2016-09-03
flatten Coreplex module hierarchy

yunsup:
The Coreplex/Uncore/OuterMemorySystem module hierarchy was mostly getting in the way, the code is much cleaner this way.

### DIFF
--- a/src/main/scala/coreplex/Coreplex.scala
+++ b/src/main/scala/coreplex/Coreplex.scala
@@ -42,8 +42,6 @@ case object ExtraCoreplexPorts extends Field[Parameters => Bundle]
 trait HasCoreplexParameters {
   implicit val p: Parameters
   lazy val nTiles = p(NTiles)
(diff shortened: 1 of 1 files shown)


aswaterman:
Looks great

zhemao:
Yeah, this is much cleaner. But it looks like there's a FIRRTL error for the GroundTestCoreplex.

zhemao:
OK, I fixed the problem. The assignment of the incoherent signal was incorrect for >1 tiles. You'd think Chisel would detect that issue though.

aswaterman:
i thought the CI hooks prevented merging when the build fails... any idea why github's offering me the option to merge?

zhemao:
I think because you're an administrator you have special privileges to bypass the CI checks. Anyway, I've fixed it now, so we can merge this.

colinschmidt:
I think the button should be warning you about the failed build despite letting you push it.

PR #241, opened 2016-09-04
Hurricane2 TileLink IP

zhemao:
This PR removes the AXI-tunneled-over-Serial logic we added for the previous Hurricane tapeout and replaces it with a new solution that tunnels TileLink over a serial channel.

It also replaces the old TileLinkMemorySelector with new, more flexible Client*TileLinkIOSwitcher modules.

This IP will be necessary for the upcoming tapeout, but I understand if people think RocketChip is not the best place for them.


aswaterman:
I like the general strategy, but I think this should be maintained elsewhere/on a branch, as it's test-chip-specific.

zhemao:
Sure, that's fine with me. A few of the commits here should go into rocket chip though. I'll see if I can factor them out into a separate PR.

aswaterman:
Agreed, the AXI serialization stuff should be removed, regardless.

PR #242, opened 2016-09-04
Clean up old Hurricane code

zhemao:
This removes the old code written for the previous hurricane tapeouts. Their replacements for the hurricane2 tapeout will be maintained on a separate branch for now.

There's also a few unrelated fixes for the TileLink wrapper and unwrappers and the JunctionsPeekingArbiter.


aswaterman:
This looks good

PR #243, opened 2016-09-04
Have Travis cache the entire verilator directory

zhemao:
It turns out the makefile will try to rebuild the installed verilator binary if the source is missing, so we should cache the entire verilator directory (both source and installation). Hopefully this will speed up future Travis runs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ cache:
   directories:
     $HOME/.ivy2
     regression/install
-    emulator/verilator/install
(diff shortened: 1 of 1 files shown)


colinschmidt:
Looks good. You might want to manually clear the cache after this makes it to master to make sure the cache gets updated. Travis has been funky about this kind of stuff in the past.

zhemao:
Yeah, I've already noticed some of that weirdness. I'll clear the caches now.

PR #248, opened 2016-09-07
relax contraint on adding AddrMapEntry to AddrMap

yunsup:
now you can add them in any order.  there's an explicit check at the end to figure out whether there are overlapping regions.

### DIFF
--- a/src/main/scala/junctions/addrmap.scala
+++ b/src/main/scala/junctions/addrmap.scala
@@ -77,8 +77,6 @@ class AddrMap(
     var cacheable = true
     for (AddrMapEntry(name, r) <- entriesIn) {
       if (r.start != 0) {
-        val align = BigInt(1) << log2Ceil(r.size)
-        require(r.start >= base, s"region $name base address 0x${r.start.toString(16)} overlaps previous base 0x${base.toString(16)}")
         base = r.start
       } else {
         base = (base + r.size - 1) / r.size * r.size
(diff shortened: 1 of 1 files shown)


yunsup:
@aswaterman can you take a look at this?

zhemao:
I can take a look.

yunsup:
great, thanks @zhemao.

src/main/scala/junctions/addrmap.scala
```diff
@@ -121,6 +119,12 @@ class AddrMap(
     }.flatten.sortBy(_.region.start)
   }
 
+  flatten.combinations(2) foreach {
+    case (Seq(AddrMapEntry(an, ar), AddrMapEntry(bn, br))) =>
+      require(!(ar.start < br.start + br.size && br.start < ar.start + ar.size),
```
zhemao: After giving this a lot of thought, I finally convinced myself that this expression was correct. Perhaps it would be helpful to add a comment explaining why this expression works.

src/main/scala/junctions/addrmap.scala
```diff
@@ -121,6 +119,12 @@ class AddrMap(
     }.flatten.sortBy(_.region.start)
   }
 
+  flatten.combinations(2) foreach {
+    case (Seq(AddrMapEntry(an, ar), AddrMapEntry(bn, br))) =>
```
zhemao: This is sort of nitpicky. But if you aren't using `an` and `bn`, you can just match on `_`.

src/main/scala/junctions/addrmap.scala
zhemao: Alternatively, you could use the inverse expression.

    ar.start >= br.start + br.size || br.start >= ar.start + ar.size

yunsup: Let me break this out a bit.

src/main/scala/junctions/addrmap.scala
```diff
@@ -121,6 +119,16 @@ class AddrMap(
     }.flatten.sortBy(_.region.start)
   }
 
+  // checks to see whether any MemRange overlaps within this AddrMap
+  flatten.combinations(2) foreach {
+    case (Seq(AddrMapEntry(an, ar), AddrMapEntry(bn, br))) =>
+      val ar_end = ar.start + ar.size
+      val br_end = br.start + br.size
+      val ar_overlaps_br = ar.start < br_end && br.start < ar_end
+      require(!ar_overlaps_br,
```
zhemao: I still think it's more understandable to just have this be

    ar.start >= br_end || br.start >=ar_end

But I guess it's fine as is.

aswaterman: Favor camel case over snake case

yunsup:
@aswaterman @zhemao I've incorporated both of your comments.

PR #254, opened 2016-09-07
Turn on the inferRW Firrtl pass

ben-k:
Without this, all of the memories wind up as two-ported.


mwachs5:
@shunshou do you expect that we would need this flag?

shunshou:
Yes. I purposefully decided you need to separately use --inferRW. I could also enforce that my FIRRTL transform always runs this first. Ofc not doing so bloats the number of arguments, but I thought that this would allow some possibly useful flexibility.

zhemao:
Seems fine to me.

PR #257, opened 2016-09-08
Fix routing in non-contiguous MMIO regions

yunsup:
This is a temporary fix, which can generate more hardware than necessary, but this is OK for now, since this code will soon be replaced with tilelink2 code.

### DIFF
--- a/src/main/scala/junctions/addrmap.scala
+++ b/src/main/scala/junctions/addrmap.scala
@@ -153,4 +153,8 @@ class AddrMap(
     }
     new AddrMapProt().fromBits(protForRegion.reduce(_|_))
   }
+
+  override def containsAddress(x: UInt) = {
+    flatten.map(_.region.containsAddress(x)).reduce(_||_)
+  }
 }


mwachs5:
works for me

zhemao:
Wait, so why is this necessary now? Is it because the memory regions can be defined out of order so `start` may not be the actual start? I think instead we should make sure that `start` and `size` are set properly. I don't think it's okay to generate more hardware than necessary because we will likely still use this code for routing in tilelink2.

yunsup:
The problem always existed.  Think about this case:

io:int:device0 start:0x0, end:0xfff
io:int:device1 start:0x8000_0000, end: 0x8fff_ffff

io:ext:device0 start:0x5000_0000, end: 0x5000_ffff

Now, think about the internal MMIO crossbar, which has two output ports, one for io:int, and the other for io:ext.  With the current AddrMap implementation, io:int will be reduced to start:0x0, end: 0x8fff_ffff and io:ext will be reduced to start:0x5000_0000, end: 0x5000_ffff.  With the current implementation, both routeSel bits of ClientUncachedTileLinkIORouter will be turned on for address 0x5000_0000.

TileLink2 has its own infrastructure for address bookkeeping and crossbar implementation, so this code will not be used in the TL2 codebase.

zhemao:
Oh, I see. The problem didn't exist before because you couldn't configure the address map in that way. That is, you couldn't have the address range in ext be in-between two ranges in int. I think the solution then is to flatten the interconnect generator so we have a single router/crossbar for all the leaf nodes, instead of going hierarchically down and instantiating redundant comparators at the higher nodes.

yunsup:
Again, flattening the router/crossbar is what TL2 makes it easy to do.  I am not going to spend more time dealing with this problem in TL1 land, since it is throw-away code.  I just want to point out that the current master is *broken*.

zhemao:
Sure. I've updated the branch. Once Travis passes again, feel free to merge.

PR #259, opened 2016-09-08
Refactor Periphery

yunsup:
This commit changes the Periphery to use tilelink2 LazyModules, and splits Peripheries into traits, which can later be chained together using "with"s on top of the BaseTop.  This makes rocket-chip to be used more as a library, therefore I changed the top-level module name to "ExampleTop" to clearly denote it's an example.  Users of rocket-chip will add their own Peripheries to add more interesting stuff between the Coreplex and the I/Os of a chip.  This change is intended to supersede the "extra" pattern.

For those who are curious about this programming pattern, search for the "cake pattern".  h/t @hcook 

Don't merge this PR yet.  I need to do more testing and turn a couple more devices into traits before we merge this upstream.  Once I change all devices which use the device builder infrastructure, I'll respond to this PR, so that the code can be reviewed further and merged upstream.

Please let me know if you have feedback in the mean time.

h/t @terpstra


yunsup:
Looking into why ComparatorConfig failed.  Any tips would be appreciated.

yunsup:
It's because ComparatorConfig uses WithTestRAM, which was the next thing I was going to refactor to use Periphery traits.

yunsup:
Found a couple problems with the second commit.  I'm going to force push my last commit with my change.

yunsup:
Found another problem.  Forced push my fixes, so this PR should be good to go.

zhemao:
@hcook Do these changes look okay? If so, I'll merge this in.

yunsup:
I'm also waiting for @aswaterman to take a look.

zhemao:
I guess I should look up cake pattern and try to understand this too.

mwachs5:
@zhemao if you find anything on the etymology of "cake" pattern, would love to hear it. My guess is something like having a recipe and mixing in the required ingredients? The only google-able explanation is a broken link to a comment on a 404-ed blog post.

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -24,88 +26,74 @@ case object BankIdLSB extends Field[Int]
...
+abstract class Coreplex(implicit val p: Parameters, implicit val c: CoreplexConfig) extends Module
     with HasCoreplexParameters {
-  class CoreplexIO(implicit val p: Parameters) extends Bundle {
-    val mem  = Vec(nMemChannels, new ClientUncachedTileLinkIO()(outermostParams))
-    val ext_clients = Vec(nExtClients, new ClientUncachedTileLinkIO()(innerParams)).flip
-    val mmio = p(ExportMMIOPort).option(new ClientUncachedTileLinkIO()(outermostMMIOParams))
-    val interrupts = Vec(p(NExtInterrupts), Bool()).asInput
+  class CoreplexIO(implicit val p: Parameters, implicit val c: CoreplexConfig) extends Bundle {
+    val master = new Bundle {
+      val mem = Vec(nMemChannels, new ClientUncachedTileLinkIO()(outermostParams))
```
zhemao: This might be nit-picky, but is there a reason `nMemChannels` doesn't come from the CoreplexConfig when all the other parameters in the CoreplexIO bundle do?

zhemao: Yeah, I really think this should be changed to be consistent with the others. Move the NMemoryChannels parameter out of coreplex and into rocketchip, then pass it in through the CoreplexConfig.

zhemao: Oh, now I see why, you still need NMemoryChannels for the TLKey. I guess you'll still have to keep the declaration of NMemoryChannels in coreplex. But I still think the actual value should be passed in through the CoreplexConfig.

aswaterman:
lgtm

aswaterman:
@zhemao will address the nMemChannels thing in a separate commit

src/main/scala/coreplex/Coreplex.scala
yunsup: Yes, you're on the right track, and that should be the next thing that gets refactored.  I wanted to maintain the same functionality so that all the things that depended on it still worked.

zhemao: The only things that depended on it are in the Coreplex itself and thus have access to CoreplexConfig. I've done the refactor in a separate PR #276. Could you take a look.

PR #261, opened 2016-09-08
Use sbt-launch.jar 0.13.12.

richardxia:
The current version of `sbt-launch.jar` packaged with rocket-chip is 0.13.0-RC2, which is over three years old. The 0.13.0-0.13.5 changelog is [here](http://www.scala-sbt.org/0.13.5/docs/Community/Changes.html), and the 0.13.5-0.13.12 changelog is [here](http://www.scala-sbt.org/0.13/docs/sbt-0.13-Tech-Previews.html). According to the latter:

> These releases maintain binary compatibility with plugins that are published against sbt 0.13.0, but add new features in preparation for sbt 1.0.

So theoretically, this should not cause any problems.

I'm requesting that we bump the version because I was running into errors using the `sbt-launch.jar` packaged here that are not present when using sbt 0.13.11.


hcook:
@palmer-dabbelt @zhemao do you guys know why travis isn't running on this PR? is there a way to manually trigger it?

richardxia:
I think the Travis job didn't run because I submitted this as a branch on a personal fork. I just got added to `ucb-bar`, so I'll close this and submit another Pull Request directly from rocket-chip repo.

PR #263, opened 2016-09-09
TL2 Interrupts

terpstra:
This PR creates an interrupt map dynamically in the same way we create TL2 metadata.


aswaterman:
I'll let Yunsup deal with merging this one

yunsup:
lgtm

PR #265, opened 2016-09-09
Make it possible to have Async Reset Flops

mwachs5:
This PR gives us some way of using Asynchronous reset flops in Chisel, by blackboxing the verilog for one flop then giving some Chisel wrappers for using them.

This is a workaround, I'm happy if someone comes up with something better, but I need this now.

I also renamed @terpstra 's TL2 GPIO to 'Example' because it's less likely to conflict with a real TL2 device name.


aswaterman:
I think this is OK as a temporary workaround.

mwachs5:
@terpstra are you OK with me renaming your GPIO to Example? I merged your recent changes to GPIO into Example.

PR #268, opened 2016-09-09
add STOP_COND to emulator & match vsim PRINTF_COND

### LINKED ISSUE #267: emulator doesn't set STOP_COND
Currently when you build your design in the emulator it does not set the STOP_COND variable so any asserts that occur during reset will cause the simulation to fail.

In vsim STOP_COND and PRINTF_COND are both set to values from the TestDriver, which does not exist in emulator.

colinschmidt:
This should fix #267

### DIFF
--- a/csrc/emulator.cc
+++ b/csrc/emulator.cc
@@ -15,6 +15,7 @@
 extern dtm_t* dtm;
 static uint64_t trace_count = 0;
 bool verbose;
+bool done_reset;
 
 void handle_sigterm(int sig)
 {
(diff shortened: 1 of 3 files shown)


csrc/emulator.cc
```diff
@@ -82,12 +83,14 @@ int main(int argc, char** argv)
...
   for (int i = 0; i < 10; i++) {
+    done_reset = false;
     tile->reset = 1;
     tile->clk = 0;
     tile->eval();
     tile->clk = 1;
     tile->eval();
     tile->reset = 0;
+    done_reset = true;
   }
```
aswaterman: you should be able to remove the done_reset = false, as the global will be so initialized, and then you can move done_reset = true to the outside of the loop, which makes the intent clearer.  looks good otherwise!

aswaterman:
Actually, I'll just slide this into my PR to save on travis bandwidth...

colinschmidt:
Ok thanks.

PR #269, opened 2016-09-09
Two tweaks to Rocket perf & QoR

aswaterman:


src/main/scala/rocket/fpu.scala
```diff
@@ -12,8 +12,8 @@ import cde.{Parameters, Field}
 
 case class FPUConfig(
   divSqrt: Boolean = true,
-  sfmaLatency: Int = 2,
-  dfmaLatency: Int = 3
+  sfmaLatency: Int = 3,
+  dfmaLatency: Int = 4
```
colinschmidt: Last time you changed the defaults for the muldiv unit, I bet you would do it this way but instead you changed the entry in Configs.scala instead. Any reason for the difference this time?

aswaterman:
I don't have a great reason, but they are qualitatively different.  In the FMA case, I think these are sensible defaults for cycle time, yet they don't have much impact on area or CPI.  The integer multiplier/divider changes don't have significant impacts on cycle time, but do have a significant impact on both CPI and area.

We don't have a clear policy on what kind of configuration the default values in these configuration objects should correspond to.  A tiny design?  A moderately powerful one, more like what Rocket was originally designed for?  This would inform how to set defaults for the objects in the integer pipeline.

ben-k:
I agree that it would be nice to define what "class" of core is implemented by DefaultConfig - and for this to be written down somewhere.

colinschmidt:
Sounds reasonable. This PR looks good to me. I'm looking forward to having it in master.

PR #270, opened 2016-09-09
Allow configuration of simulation memory latency

zhemao:
Since the reorganization of the test harness, we've lost the ability to use dramsim. This makes it difficult to get reasonable benchmark results in simulation. This doesn't really restore that functionality, but it gets you part of the way there. Basically, you can set it up so that responses from the SimAXIMem have a fixed latency. I'm not really sure what the most sensible default for `SimMemLatency` is, however.

### DIFF
--- a/src/main/scala/rocketchip/Configs.scala
+++ b/src/main/scala/rocketchip/Configs.scala
@@ -74,6 +74,7 @@ class BasePlatformConfig extends Config(
         case RTCPeriod => 100 // gives 10 MHz RTC assuming 1 GHz uncore clock
         case BuildExampleTop =>
           (p: Parameters) => uncore.tilelink2.LazyModule(new ExampleTop(p))
+        case SimMemLatency => 0
         case _ => throw new CDEMatchError
       }
     }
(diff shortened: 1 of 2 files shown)


aswaterman:
Most sensible default is probably 1 (or 0).  Slowing down the regression tests to simulate a fixed-latency pipe doesn't make much sense, and you're likely going to override it for benchmarking, anyway.

aswaterman:
Also, you could just instantiate N 2-entry queues, rather than building a new data structure (since the QoR obviously doesn't matter here).

zhemao:
Oh yeah, I didn't think of that. I guess what you'd really want are N 1-entry queues with pipe=true. I'll try doing that.

aswaterman:
yeah, good point

zhemao:
Okay, I've made the changes. Does this commit look okay to you Andrew?

aswaterman:
Yep. Mind holding off merging until Yunsup merges in his mondo PR?

zhemao:
If you're talking about #259, seems like his branch is out-of-date, so he'll need to update and re-run the Travis tests anyway. Could I merge mine first and then he updates his branch?

yunsup:
I'll take care of merging this after #259 is merged.

zhemao:
Sure

PR #274, opened 2016-09-12
fix warnings in verilog source

colinschmidt:
@mwachs5 I ran the jtag-dtm-regression and the default asm tests and things still work, so I'm assuming there is no need for these empty assigns.


mwachs5:
lgtm

PR #276, opened 2016-09-12
Pass nMemChannels to coreplex through CoreplexConfig

zhemao:
This is the comment I made on the Periphery refactor. To keep things consistent, we should pass nMemChannels to the Coreplex through the CoreplexConfig instead of the cde Parameters.


zhemao:
The only thing I wasn't able to fix is that the coreplex configuration still needs the NMemoryChannels parameter in order to determine the number of TileLink client_xact_id bits. But that will go away once we move to Tilelink2.

PR #277, opened 2016-09-12
Make BootROM & PRCI Optional in Coreplex

mwachs5:
This is an intermediate parameterization to disable the BootROM and/or PRCI instantiation within the Coreplex, since it's now possible to easily instantiate different versions of them in the Periphery instead.


aswaterman:
We're contemplating moving BootROM and PRCI into the periphery for rocket-chip as well, so that this won't have to be a parameter.

yunsup:
#286 addresses the PRCI instantiation, and I'm going to take a stab at making the bootrom a peripheral tonight.

PR #278, opened 2016-09-13
Allow Coreplex, L2, Tiles, and RoCC to be on separate clock domains

zhemao:
This is sort of a backport of stuff we were using for the hurricane tapeout. We want the ability to put certain pieces of the chip in separate clock domains. The main parts of this are

1. RoCC accelerators on separate clock domains from rest of Tile
2. Tiles and L2 agent on separate clock domains from rest of Coreplex
3. Coreplex on separate clock domain from Periphery

 Of course, in the example design, everything is still on a single clock domain.


mwachs5:
Why did you use 'null' instead of 'Option'?

yunsup:
taking a look at this now.

zhemao:
@mwachs5 The signals eventually get wrapped in an Option when they are passed to the Module constructor. The result of `Option(null)` is `None`, so it works out.

yunsup:
I have some stylistic changes I suggest you make, which I'll make a comment on the code.  Also, I have some suggestions for the signal names to make it more consistent through out your changes.

But the ultimate question is: how does one use multiple clock domains?  Your PR doesn't let the user specify a different clock at all (also doesn't actually instantiate any async clock crossing elements such as an async fifo).

src/main/scala/coreplex/Configs.scala
```diff
@@ -100,8 +101,8 @@ class BaseCoreplexConfig extends Config (
         TestGeneration.addSuites(rvi.map(_("p")))
         TestGeneration.addSuites((if(site(UseVM)) List("v") else List()).flatMap(env => rvu.map(_(env))))
         TestGeneration.addSuite(benchmarks)
-        List.tabulate(site(NTiles)){ i => (r: Bool, p: Parameters) =>
-          Module(new RocketTile(resetSignal = r)(p.alterPartial({
+        List.tabulate(site(NTiles)){ i => (c: Clock, r: Bool, p: Parameters) =>
+          Module(new RocketTile(clockSignal = c, resetSignal = r)(p.alterPartial({
```
yunsup: Sometimes you use clock/reset, c/r, or clockSignal/resetSignal.  I'm assuming the reason why you've avoided using clock/reset is because those names are used in chisel?  I've used c as an implicit val for configuration case classes (e.g. CoreplexConfig), so I'd avoid that as well.  clockSignal and resetSignal are long, so that's not good.  How about using _clock and _reset?  This is something that Chisel3 used as function arguments.  Whatever you pick, you should be consistent throughout the codebase.

src/main/scala/coreplex/Configs.scala
```diff
@@ -300,10 +303,11 @@ class WithBufferlessBroadcastHub extends Config(
...
   topDefinitions = (pname, site, here) => pname match {
-    case BuildL2CoherenceManager => (id: Int, p: Parameters) =>
-      Module(new ManagerToClientStatelessBridge()(p.alterPartial({
-        case InnerTLId => "L1toL2"
-        case OuterTLId => "L2toMC" })))
+    case BuildL2CoherenceManager =>
+      (c: Clock, r: Bool, id: Int, p: Parameters) =>
+        Module(new ManagerToClientStatelessBridge(c, r)(p.alterPartial({
+          case InnerTLId => "L1toL2"
+          case OuterTLId => "L2toMC" })))
```
yunsup: again, inconsistent use of clock/reset signal names here.

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -49,7 +50,9 @@ case class CoreplexConfig(
   val plicKey = PLICConfig(nTiles, hasSupervisor, nExtInterrupts, 0)
 }
 
-abstract class Coreplex(implicit val p: Parameters, implicit val c: CoreplexConfig) extends Module
+abstract class Coreplex(clockSignal: Clock = null, resetSignal: Bool = null)
+    (implicit val p: Parameters, implicit val c: CoreplexConfig)
+    extends Module(Option(clockSignal), Option(resetSignal))
```
yunsup: I'm torn about both line breaks here.  Maybe using _clock and _reset instead would let you put the (implicit val ...) on the same line as Coreplex?

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -67,11 +70,13 @@ abstract class Coreplex(implicit val p: Parameters, implicit val c: CoreplexConf
   val io = new CoreplexIO
 }
 
-class DefaultCoreplex(tp: Parameters, tc: CoreplexConfig) extends Coreplex()(tp, tc) {
+class DefaultCoreplex(tp: Parameters, tc: CoreplexConfig,
+    clockSignal: Clock = null, resetSignal: Bool = null)
+    extends Coreplex(clockSignal, resetSignal)(tp, tc) {
```
yunsup: Maybe shortening _clock and _reset could also let all the tp, tc, _clock, _reset arguments be on the same line?

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -198,7 +203,26 @@ class DefaultCoreplex(tp: Parameters, tc: CoreplexConfig) extends Coreplex()(tp,
   }
 }
 
-class GroundTestCoreplex(tp: Parameters, tc: CoreplexConfig) extends DefaultCoreplex(tp, tc) {
+class GroundTestCoreplex(tp: Parameters, tc: CoreplexConfig,
+    clockSignal: Clock = null, resetSignal: Bool = null)
+    extends DefaultCoreplex(tp, tc, clockSignal, resetSignal) {
```
yunsup: same here about the line breaks

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -198,7 +203,26 @@ class DefaultCoreplex(tp: Parameters, tc: CoreplexConfig) extends Coreplex()(tp,
...
   override def hasSuccessFlag = true
   io.success.get := tileList.flatMap(_.io.elements get "success").map(_.asInstanceOf[Bool]).reduce(_&&_)
 }
+
+class UnitTestCoreplex(tp: Parameters, tc: CoreplexConfig,
+    clockSignal: Clock = null, resetSignal: Bool = null)
+    extends Coreplex(clockSignal, resetSignal)(tp, tc) {
```
yunsup: same here about the line breaks

zhemao:
Okay, I'll try to change things to use _clock and _reset everywhere.

As for how we will actually use multiple clocks, presumably that will be done by passing optional additional clock signals through the io bundles.

@terpstra Now that we have the FIRRTL replseqmem pass, is it safe to use FIFO clock crossers?

src/main/scala/coreplex/Coreplex.scala
zhemao: I tried using `_clock` and `_reset` but even then the line is too long. `implicit` is a very long word. I've changed it for the coreplex classes though.

terpstra:
On Mon, Sep 12, 2016 at 11:10 PM, Howard Mao <notifications@github.com>
wrote:

> @terpstra <https://github.com/terpstra> Now that we have the FIRRTL
> replseqmem pass, is it safe to use FIFO clock crossers?
>
No.

However, I have found an alternative. According to Xilinx and Altera
documentation, a LUT does not glitch if only a single input changes at a
time. Therefore a solution using Vec is possible.

colinschmidt:
@yunsup The plan to use this infrastructure is to use a blackboxed module outside of the coreplex to generate our different clock signals, and pass these into BuildCoreplex, with some other changes to route the different clocks to the right blocks. So far we have just been moving forward with this downstream without worrying about how to upstream the changes. This PR, which is pretty simple, helps a little but is clearly not the whole solution. As our downstream solution evolves/matures and upstream continues to be refactored we can hopefully submit another more complete multiclock PR.

zhemao:
By single input they mean single bit? Does that mean we can only have a FIFO with 2 entries?

I think it might not be worthwhile to force ASICs and FPGAs to use the same clock-crossing solution. Perhaps it would be better if we left that as a configuration option. Then the FPGA design could blackbox in the multiclock FIFOs provided by the FPGA vendor.

terpstra:
On Tue, Sep 13, 2016 at 9:04 AM, Howard Mao <notifications@github.com>
wrote:

> By single input they mean single bit? Does that mean we can only have a
> FIFO with 2 entries?
>
No. I have already implemented an AsyncQueue safe for both FPGA and ASIC
that supports 2^n entries.

The point is that the enqueuer only updates one entry in the Vec at a time,
so the reduction Mux in the dequerer side only has one bit flip
combinationally at a time. The address bit path is analysed by timing tools
since it goes from a dequeuer register to a dequeuer register. Thus, only
one of the input data bits to the Mux changes asynchronously at a time.

> I think it might not be worthwhile to force ASICs and FPGAs to use the
> same clock-crossing solution
>
I don't agree. There is a real verification benefit to using the same code
in both.

zhemao:
OK. If it's possible and doesn't impose a performance penalty on the ASIC design, then I'm all for a single solution.

zhemao:
Yunsup, do you have any other changes you want made? Otherwise, I'd like to merge this in.

yunsup:
looking at this

yunsup:
After thinking about this for a while, I'm not really sure how this change helps with separating clock domains for Tiles, L2, and RoCC.  If the stitching is done after the chisel generation step, this commit is not necessary, because all modules already have its clock and reset.  I'd like to see the example blackbox module that wires up different clocks, before we go ahead and merge in this PR.

zhemao:
We aren't going to do the stitching post-elaboration. The blackbox is just to generate the clock and reset signals. They will be passed into the Top and Coreplex through optional IOs. The actual clock and reset synchronization will presumably be done with the chisel clock-crossers that Wes just committed. If you'd like, I can add those to this PR.

yunsup:
Yes, it would be great if we could first see how everything stitches together with clock-crossers before we go ahead with this PR.

zhemao:
Sure. I'll work on that tomorrow.

terpstra:
Wouldn't it be easier to just assign the .clock and .reset on the coreplex, rather than trying to pass around the clock parameters everywhere?

terpstra:
See, for example, <https://github.com/ucb-bar/rocket-chip/blob/master/src/main/scala/uncore/tilelink2/Fuzzer.scala#L237>. I just shoved the RAM model into the domain I wanted from outside. You could do the same to the entire Coreplex Module after you created it.

zhemao:
Yeah, it's definitely easier to not thread _clock and _reset through the class hierarchy. I restarted this with a different approach. I'll make a separate PR.

PR #284, opened 2016-09-13
First cut at refactoring synthesizeable unit testing into a top-level utility

hcook:
Individual tests are lexically co-located with their DUTs.
Touches lots of files but makes no functional changes other than to what is contained in which packages; just moving source code around for now.


src/main/scala/junctions/util.scala
```diff
@@ -244,128 +244,3 @@ class DecoupledHelper(val rvs: Seq[Bool]) {
     (rvs.filter(_ ne exclude) ++ includes).reduce(_ && _)
   }
 }
-
-class MultiWidthFifo(inW: Int, outW: Int, n: Int) extends Module {
```
zhemao: Since you've already created a separate top-level util package, we might want to move everything from this file into that package. There's nothing here that's specific to junctions. Some of the stuff from uncore.util could be moved as well.

zhemao:
I've added some commits to move more stuff into the utils package. I'll merge this once it passes the tests.

PR #285, opened 2016-09-14
Move PRCI from Coreplex to always-on block, where it belongs

aswaterman:


aswaterman:
I'll push this again in a sec

PR #286, opened 2016-09-14
Move PRCI from Coreplex to always-on block, where it belongs

aswaterman:


yunsup:
I'm going to take a look at this tonight.

aswaterman:
Overall, I think this test rig is too brittle.  It's not testing either the coreplex or the periphery, but it depends on how the coreplex and the periphery are connected.

PR #289, opened 2016-09-14
support for BSD sed (GNU sed still works)

sbeamer:
BSD's version of sed uses -E instead of -r for extended regular expressions. Fortunately, GNU sed is more permissive and will also accept -E.

This fix allows rocket chip to be built on OS X.


aswaterman:
lgtm

aswaterman:
will merge manually

PR #290, opened 2016-09-14
fix clang support for emulator-debug

sbeamer:
Clang has difficulty resolving the overloaded function dump().

This fix allows clang to build emulator-debug.


aswaterman:
lgtm

aswaterman:
will merge manually

PR #292, opened 2016-09-14
Multi-Clock Changes Redux

zhemao:
This provides the ability to put the Tiles and L2 on separate clock domains from the rest of the core. Instead of passing things through the builder functions to the Module constructor, I just wrap BuildTiles and BuildL2CoherenceManager in a wrapper that changes the clocks and instantiates the clock crossers.

@yunsup Does this look satisfactory?

@terpstra I added level synchronizers junctions/crossings.scala to synchronize the reset and interrupt signals. Is this the proper way to do it? It's just a register in the `from` domain connected to two or more registers in the `to` domain. Is it safe to do this for multi-bit signals? There might be glitching, but eventually they'll settle to the final value, right?

### DIFF
--- a/src/main/scala/coreplex/Coreplex.scala
+++ b/src/main/scala/coreplex/Coreplex.scala
@@ -24,6 +24,8 @@ case object BuildL2CoherenceManager extends Field[(Int, Parameters) => Coherence
 case object BuildTiles extends Field[Seq[(Bool, Parameters) => Tile]]
 /** The file to read the BootROM contents from */
 case object BootROMFile extends Field[String]
+/** Separate clocks for various coreplex domains? */
+case object CoreplexMultiClock extends Field[CoreplexClockConfig]
 
 trait HasCoreplexParameters {
   implicit val p: Parameters
@@ -33,7 +35,11 @@ trait HasCoreplexParameters {
   lazy val outermostParams = p.alterPartial({ case TLId => "Outermost" })
   lazy val outermostMMIOParams = p.alterPartial({ case TLId => "MMIO_Outermost" })
   lazy val globalAddrMap = p(rocketchip.GlobalAddrMap)
+  lazy val multiclockL2 = p(CoreplexMultiClock).l2agent
+  lazy val multiclockTiles = p(CoreplexMultiClock).tiles
 }
+
+case class CoreplexClockConfig(l2agent: Boolean, tiles: Boolean)
 
 case class CoreplexConfig(
     nTiles: Int,
@@ -46,6 +52,14 @@ case class CoreplexConfig(
   val plicKey = PLICConfig(nTiles, hasSupervisor, nExtInterrupts, 0)
 }
 
+class MultiClockBundle(clockConfig: CoreplexClockConfig, coreplexConfig: CoreplexConfig) extends Bundle {
+  val l2Clock = clockConfig.l2agent.option(Clock())
+  val l2Reset = clockConfig.l2agent.option(Bool())
+  val tileClocks = clockConfig.tiles.option(Vec(coreplexConfig.nTiles, Clock()))
+
+  override def cloneType = new MultiClockBundle(clockConfig, coreplexConfig).asInstanceOf[this.type]
+}
+
 abstract class Coreplex(implicit val p: Parameters, implicit val c: CoreplexConfig) extends Module
     with HasCoreplexParameters {
   class CoreplexIO(implicit val p: Parameters, implicit val c: CoreplexConfig) extends Bundle {
@@ -54,6 +68,7 @@ abstract class Coreplex(implicit val p: Parameters, implicit val c: CoreplexConf
       val mmio = c.hasExtMMIOPort.option(new ClientUncachedTileLinkIO()(outermostMMIOParams))
     }
     val slave = Vec(c.nSlaves, new ClientUncachedTileLinkIO()(innerParams)).flip
+    val clocks = new MultiClockBundle(p(CoreplexMultiClock), c).asInput
     val interrupts = Vec(c.nExtInterrupts, Bool()).asInput
     val debug = new DebugBusIO()(p).flip
     val prci = Vec(c.nTiles, new PRCITileIO).flip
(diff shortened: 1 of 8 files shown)


colinschmidt:
Cool, this seems like it will help hurricane move forward. Looking forward to the merge :)

zhemao:
I haven't moved the RoCC accelerators to separate clock domains yet because the RoCC interface isn't completely safe. Colin will backport some changes from Hwacha in order to make that possible.

yunsup:
I'll take a look at this commit tomorrow.  Sorry about the delay.

src/main/scala/junctions/crossing.scala
```diff
@@ -91,3 +90,46 @@ object ClockToSignal {
     c2s.io.signal_out
   }
 }
+
+object LevelCrossing {
```
mwachs5: What is this for? "LevelCrossing" is a vague name to me... do you mean "Synchronizer for level signals" or do you mean "Power Domain Level Crossing"?

In general, for "library-esque" things like this, please write some good description of what this is and how to use it.

src/main/scala/junctions/crossing.scala
```diff
@@ -91,3 +90,46 @@ object ClockToSignal {
...
+  class SynchronizerBackend[T <: Data](typ: T, sync: Int, _clock: Clock) extends Module(Some(_clock)) {
```
mwachs5: For this kind of synchronizer, you should restrict the Data type, to Bool or Vec[Bool]. It does not make sense,  and is misleading, to synchronize a e.g. Bundle or UInt in this way. Force the user to convert it to Bools if that's what they think they want.

src/main/scala/junctions/crossing.scala
```diff
@@ -91,3 +90,46 @@ object ClockToSignal {
...
+    val front = Module(new SynchronizerFrontend(in, from_clock))
+    val back = Module(new SynchronizerBackend(in, sync - 1, to_clock))
+
+    front.io.in := in
+    back.io.in := front.io.out
+    back.io.out
+  }
+}
+
+object LevelTo {
```
mwachs5: "LevelTo" is way too vague a name. "LevelSyncTo" makes more sense what it is actually doing.

src/main/scala/uncore/tilelink/Crossings.scala
```diff
@@ -0,0 +1,122 @@
+package uncore.tilelink
+
+import Chisel._
+import junctions._
+
+object AsyncClientUncachedTileLinkCrossing {
```
mwachs5: It was suggested to me by @terpstra , and I think it's good practice, that the Crossing types should go with the thing that they are crossing, not in this generic Crossing library. For example, the DebugCrossing is in the same file as the DebugModule.  The analogy is that each class can describe a toString() method, you don't put every class' toString method with the String functions.

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -65,21 +80,82 @@ abstract class Coreplex(implicit val p: Parameters, implicit val c: CoreplexConf
 }
 
 class DefaultCoreplex(tp: Parameters, tc: CoreplexConfig) extends Coreplex()(tp, tc) {
+  class TileWrapper(tileBuilder: (Bool, Parameters) => Tile, _clock: Clock, _reset: Bool)
```
mwachs5: I haven't fully wrapped my head around this Wrapper pattern. Could this be done more cleanly in a  different subclass of Coreplex? Why is this in DefaultCoreplex?

src/main/scala/junctions/crossing.scala
```diff
@@ -91,3 +90,46 @@ object ClockToSignal {
...
+  class SynchronizerFrontend[T <: Data](typ: T, _clock: Clock) extends Module(Some(_clock)) {
+    val io = new Bundle {
+      val in = typ.asInput
+      val out = typ.asOutput
+    }
+
+    io.out := RegNext(io.in)
+  }
+
+  def apply[T <: Data](from_clock: Clock, to_clock: Clock, in: T, sync: Int = 3): T = {
```
mwachs5: If the user says they want 4 synchronizer stages, give them 4 stages in the target clock domain. The first one in the source clock domain doesn't count.

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -65,21 +80,82 @@ abstract class Coreplex(implicit val p: Parameters, implicit val c: CoreplexConf
...
+    val wrappedTile = tileBuilder(reset, p)
+    val io = new Bundle {
+      val tile = wrappedTile.io.cloneType
+      val ext_clock = Clock(INPUT)
+      val ext_reset = Bool(INPUT)
+    }
+
+    io.tile.uncached <> wrappedTile.io.uncached.map(AsyncClientUncachedTileLinkTo(io.ext_clock, io.ext_reset, _))
+    io.tile.cached <> wrappedTile.io.cached.map(AsyncClientTileLinkTo(io.ext_clock, io.ext_reset, _))
+    wrappedTile.io.interrupts := LevelFrom(io.ext_clock, io.tile.interrupts)
```
mwachs5: This makes me pretty nervous. There is a lot to consider here.

There are already Gateways for the interrupts. This sort of thing may be better handled there, like "AsynchronousLevelGateway", which does its own simple synchronization internally. Actually maybe it already is by the LevelGateway, should go look at what kinds of assumptions it is making there. Interrupts that come from a top-level pin should be synchronized by the Gateway anyway, so this seems like something better handled there.

Also, we need to consider what it means now if the CPU services an interrupt (indirecly making this 'interrupt' signal go low) then signals 'complete' to the PLIC. If the low level takes a long time to get through this synchronizer, the PLIC may see the interrupt again. Not saying these things are definitely problems, but you need to carefully consider this and convince yourself they AREN'T. I think that is hard to do in general.

src/main/scala/uncore/tilelink/Crossings.scala
mwachs5: Oh, I realize now you actually did that, see that this was a different crossings file.

src/main/scala/junctions/crossing.scala
zhemao: Yeah, I wasn't really sure what to call this that would make sense. It is, in-fact a "synchronizer for level signals".

src/main/scala/coreplex/Coreplex.scala
zhemao: Are you suggesting that I leave the instantiations of the tile and L2 as methods and then override those in a MultiClockCoreplex subclass? What if I want to multiclock the tiles, but not the L2 or vice-versa?

src/main/scala/coreplex/Coreplex.scala
zhemao: Yeah, I should definitely check what the gateway is doing.

I thought the protocol was that the interrupt signal stays high until the gateway sees "complete".

src/main/scala/junctions/crossing.scala
zhemao: What is the purpose of that besides making the user recognize that the individual bits won't arrive at the same time? The one place we use this for a UInt (for tile hartid), the value isn't expected to change during normal operation.

src/main/scala/coreplex/Coreplex.scala
zhemao: Also, the Gateway is synchronizing the external interrupts into the PLIC's clock domain. The interrupts going into the tile from the PLIC and PRCI were not synchronized until now. Since different tiles can potentially be on different clock domains, we can't make a single gateway.

src/main/scala/junctions/crossing.scala
mwachs5: Yep, that's exactly the purpose. It's to prevent this component that you are creating from being abused. If someone wants to say "in this case, I understand the implications, and I'm going to build a wrapper around this", they can. Or you can provide a wrapper that does what you're doing here, but give it some appropriately Warning-ful/ descriptive name that indicates what it should be used for.

src/main/scala/coreplex/Coreplex.scala
mwachs5: Ah, I see, these are synchronizing between PLIC and the Tile. Ok, then the case we need to worry about is that when you claim the interrupt, the PLIC/TL doesn't return from the 'complete write before the interrupt that you see in the Tile post-synchronizer is cleared. *Probably* this is true, but you need to convince yourself and put appropriate requires (like the synchronizer depths should match, have safety margin, or whatever).

src/main/scala/junctions/crossing.scala
zhemao: As for bundles, what if it's a bundle of bools (which is what we use for the tile interrupts)? I don't think there's a way to restrict this that would make it both safe and not annoying.

mwachs5: Hmm, yes I see that could be annoying. Or if you had a UInt of width 1 it would also yell at you even though it's technically fine.

src/main/scala/coreplex/Coreplex.scala
zhemao: I've taken a look at the PLIC and PRCI to see if synchronization can cause an unintended interrupt trigger.

For the PLIC at least, I'm satisfied that it most likely won't happen. The sequence is

* gateway gets interrupt, sets valid
* PLIC sees valid, sets ready low, sets interrupt high
* Tile sees interrupt, disables interrupts, starts running ISR
* ISR claims interrupt (through TL read)
* PLIC sets interrupt low, sends claim response
* ISR services interrupt, causing device sending interrupt to set interrupt low
* ISR sends completion to PLIC, which forwards to gateway. Tile re-enables interrupts and continues normal execution

The gateway won't send another interrupt to the PLIC until it gets the interrupt completion, so that's definitely safe. The only issue might be if the PLIC interrupt going low after the claim takes longer to settle on the Tile's domain than it takes to send the claim response, process it, and send the interrupt completion. I think that's very unlikely, since the interrupt level will settle after three or four clock cycles (of the Tile's clock), whereas processing the interrupt would involve several instructions to execute on the CPU and a potentially very long round trip through MMIO to the device that triggered the interrupt.

For the PRCI, the sequence is.

* time counter exceeds the timecmp value set by the tile, interrupt triggers
* timer interrupt routine runs, increases the timecmp value, interrupt goes low (on PRCI side)
* timer interrupt routine exits, re-enables interrupts

Here, it is possible that the interrupt level won't go low until after the timer interrupt routine exits. This depends on how long the timer routine continues to run after bumping the timecmp value.

src/main/scala/coreplex/Coreplex.scala
zhemao: And if did this in a subclass, we would still need the wrapper. What it's doing is changing the clock domain (using _clock and _reset), then putting in synchronizers from the tile clock domain to the external clock domain (ext_clock and ext_reset). I think that's the cleanest way of doing it.

src/main/scala/junctions/crossing.scala
zhemao: Fixed, thanks.

src/main/scala/junctions/crossing.scala
zhemao: I renamed everything to LevelSync*. I also added a documentation comment.

src/main/scala/coreplex/Coreplex.scala
zhemao: Okay, so for the PRCI, if you fence before exiting the interrupt routine so that you know the update of the timecmp register has been acknowledged, you can be sure that the interrupt has settled.

If you want to be 100% sure of this, you could have an acknowledgement system in which the interrupt gets passed into the tile domain, registered, and then passed back out to the PLIC/PRCI. Then the PLIC/PRCI could check the returned value to make sure it has updated before sending the acknowledgement.

aswaterman: Expectation at ISA level is that FENCE suffices.  Doesn't matter how this is accomplished... if the mtimecmp write response takes at least as long to propagate as the interrupt itself, then it's correct automatically.  The current implementation happens to have that property, provided the same clock crossing latencies for the write response and the interrupt.  No need for a handshake in this case.

zhemao: In this case, the memory crossing latency is set to be longer than the interrupt crossing latency, so we should be safe.

yunsup:
Setting aside the way the code was expressed (I'll comment below), we need to first talk about the actual design point we're building, because this PR implements a bad design point IMO.

This PR wraps around the Tile and each L2CoherenceManager with a separate clock and reset, which were given to the respective modules, and in the process puts down clock crossing elements.  That means, when the tile talks to an L2 coherence manager, it will do 2 clock crossings instead of 1 -- from the tile clock domain to the coreplex clock domain to the l2 clock domain!

The design we want here is all tiles in their own clock domain, directly talking to one L2 clock domain, which includes the l1-to-l2 clock domain as well as all L2 coherence managers.  That means, if you are implementing an L2 cache with n banks, all banks will reside in one clock domain instead of n clock domains.

On to the way the code is expressed.  After putting in some thought with @terpstra, it seems like a better way to implement this code is with traits using the cake pattern again (you can see a recurring theme here).  Have two traits for the coreplex, one for the I/O requirements and another one to express the connectivity of the tiles and the coherence manager.  There will be a set of  synchronous crossing traits and a set of asynchronous crossing traits for the I/O and connectivity.  To be more concrete, something like

trait SyncCoreplexConnectIO {
  // nothing
}

class ClockReset {
  val clk = Clock()
  val rst = Bool()
}

trait AsyncCoreplexConnectIO {
  val tiles = Vec(nTiles, new ClockReset().asInputs)
  val l2 = new ClockReset().asInputs
}

class SyncCoreplexConn {
  // connects all tile cached & uncached TileLinkIOs to l1-to-l2 net synchronously
  // connect interrupts synchronously
  // connect all tile slaves synchronously
  // connect all coreplex slave ports to l1-to-l2 net synchronously
}

class AsyncCoreplexConn {
  // connects all tile cached & uncached TileLinkIOs to l1-to-l2 net asynchronously
  // connect interrupts asynchronously
  // connect all tile slaves asynchronously
  // connect all coreplex slave ports to l1-to-l2 net asynchronously
}

Then when you instantiate the coreplex, you mix-in the synchronous trait or the asynchronous trait.  This way you can keep the common code agnostic to how you're connecting the tiles to the L2.

Finally, one thing to note here is that there's no reason to connect the hartid through a synchronizer.  It's just constant values in the current code base.

Given all of this, I suggest you close this PR and start from scratch.

terpstra:
I think you meant for SyncCoreplexConn and AsyncCoreplexConn to be traits,
not classes..?

yunsup:
Yes, I meant traits.  I updated my original comment.

zhemao:
Makes sense, I'll work on this and submit a new PR on Monday.

PR #293, opened 2016-09-14
Add some async/clock utilities

mwachs5:
This provides some common "utilities" (black box hacks) to do some clock/reset related things that aren't possible in Chisel.

### DIFF
--- a/vsim/Makefrag
+++ b/vsim/Makefrag
@@ -9,6 +9,9 @@ bb_vsrcs = $(base_dir)/vsrc/DebugTransportModuleJtag.v \
 	$(base_dir)/vsrc/AsyncMailbox.v \
 	$(base_dir)/vsrc/AsyncResetReg.v \
 	$(base_dir)/vsrc/ClockDivider.v \
+        $(base_dir)/vsrc/ClockToSignal.v \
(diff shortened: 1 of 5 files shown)


aswaterman:
thanks!

vsim/Makefrag
```diff
@@ -9,6 +9,9 @@ bb_vsrcs = $(base_dir)/vsrc/DebugTransportModuleJtag.v \
 	$(base_dir)/vsrc/AsyncMailbox.v \
 	$(base_dir)/vsrc/AsyncResetReg.v \
 	$(base_dir)/vsrc/ClockDivider.v \
+        $(base_dir)/vsrc/ClockToSignal.v \
+        $(base_dir)/vsrc/SignalToClock.v \
```
yunsup: Please use tabs within a Makefile

zhemao:
I think this is something we should just add to Chisel.

vsim/Makefrag
mwachs5: sorry...  need to set emacs to recognize "Makefrag" as makefile-mode.

mwachs5:
@zhemao absolutely. But I got tired of not having it at all, so this is the current workaround.

PR #294, opened 2016-09-15
Add firrtl and verilog Makefile targets to vsim

jackkoenig:
Helps create just the Firrtl or Verilog target files. Not sure if these are really wanted but I find them useful.


yunsup:
There's already a verilog target in vsim/Makefrag.  Can you drop the verilog target in your PR?

PR #296, opened 2016-09-15
refactor unittest framework

yunsup:
as a result, there's another SUITE that needs to run


yunsup:
Also takes care of @aswaterman's comment on #286 regarding the unit testing framework.

hcook:
lgtm

packages rocketchip.utest and unittest should be merged into a single package at some point but I will take care of that in a future PR.

zhemao:
I think it would be better to keep the unit test coreplex in rocketchip. That way we don't have to override the project in the regression makefile.

zhemao:
I guess the issue is that you still need to call the TestHarness
TestHarness.

PR #298, opened 2016-09-15
Allow MODEL to be something other than TestHarness

zhemao:
This uses C and Verilog pre-processor macros to set the name of the top-level module so that MODEL doesn't have to be TestHarness in order to simulate. This also moves the UnitTest stuff from rocketchip.utest back up to rocketchip and renames rocketchip.utest.TestHarness to rocketchip.UnitTestHarness.

### DIFF
--- a/csrc/emulator.cc
+++ b/csrc/emulator.cc
@@ -65,7 +65,7 @@ int main(int argc, char** argv)
   srand48(random_seed);
 
   Verilated::randReset(2);
-  VTestHarness *tile = new VTestHarness;
+  MODEL *tile = new MODEL;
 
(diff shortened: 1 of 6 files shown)


yunsup:
I was actually on the fence about this change, hence the refactoring of unit test reflected that.  @zhemao you should solicit feedback before merging PRs in.

zhemao:
I figured this was relatively uncontroversial since we already allow MODEL to be changed, but not really because it doesn't work if you make it anything else.

I didn't get rid of PROJECT in case we later want to add a test suite that actually changes PROJECT.

PR #302, opened 2016-09-15
Tl2 mmio

terpstra:
Refactor rocket-chip to use TL2


aswaterman:
I didn't perform a detailed review, but at a glance, this is looks good.  Thanks for refactoring PRCI.

terpstra:
My travis runs were Killed? Does this mean it ran out of memory building them?

aswaterman:
Looks to me that GCC exhausted memory building the Verilated code.

zhemao:
What the heck? How big of a FIRRTL file is this generating? If it's substantially larger than before, something must be wrong.

terpstra:
Well, there are a ton of new assertions. However, the problem which caused it to run out of memory was that the PRCI was a sparse register mapped device. This has been resolved by adding an address compressor.

PR #303, opened 2016-09-15
TestHarness refactoring [WIP]

hcook:
!!! WIP DO NOT MERGE YET (tho it does compile :D) !!!

Steps towards automatically generating test harnesses from the DUT peripheral traits.

- rocketchip, groundtest, and unittest are now separate peer packages.
  - the PROJECT flag is used to select the scala package containing the platform
  - the CONFIG flag is used to select a particular specific DUT to build
- All build targets use a Module subclass called TestHarness as their point of entry
  - TestHarness has a single IO port, a boolean indicating simulation completion
  - this is analogous to calling main() in a c program
  - failures are indicated by throwing Chisel assertions


zhemao:
Is there a good reason to disallow changing MODEL to something other than TestHarness? I'm fine with all of the test harnesses in the three default packages being called the same thing. But if there's a straightforward way to allow it to be called something different, I think we should keep it. I don't think the use of a define macro is that confusing.

zhemao:
If we do want to disallow it, we should change `MODEL ?=` to `MODEL :=` and not put the model name in the simulation executable. So instead of `emulator-TestHarness-DefaultConfig` it would be `emulator-DefaultConfig`.

zhemao:
Well, actually I take that back, we still want to be able to change MODEL in order to build the verilog for different models. However, we'll want to change the name of the executable to include project name instead of the model name. So we'd have something like `emulator-rocketchip-DefaultConfig` or `emulator-groundtest-MemtestConfig`.

hcook:
Executables now use PROJECT in their name

zhemao:
I think the verilog files will still need the MODEL name. We may still need to generate verilog for stuff other than the TestHarness even if we don't compile it for simulation.

hcook:
Shouldn't any such non-TestHarness verilog targets be in a backend-specific Makefrag, not the one in v**sim**?

zhemao:
That's true, but I'm mainly talking about the change to Generator.scala. The backend-specific Makefrags will extend the Generator trait. I think the solution to this, then is to move the definition of `longName` and the `writeOutputFiles` function out of the Generator trait and into the RocketChipGenerator object.

hcook:
Oh I see, and agree. Will do.

zhemao:
Also, I've already cherry-picked one of these commits into my multiclock2 branch, which will probably screw with things. I think at this point it would be best if I just rebased these commits into a separate branch after merging in my multi-clock changes to master.

hcook:
Actually, since Generator is just a trait, wouldn't it be easier to override the definition of longName in any subclasses or traits that want to recognize different MODELs? I don't think WriteOutputFiles actually belongs in RocketChipGenerator; which files would not be written out in your use case?

zhemao:
FPGA-zynq doesn't use any of the files spit out by writeOutputFiles. It only needs the FIRRTL file. I'd imagine that code using rocketchip as a library would really only need to spit out the FIRRTL file, config string, and possibly the parameter dump. I suspect most projects will want to specify each of the possible outputs individually.

zhemao:
Yeah, that looks good. Why don't we go ahead and resolve the conflicts with master and merge this in.

colinschmidt:
@hcook What was the benefit of not including MODEL in longName? Are you trying to have constructive aliasing or something?

This just adds another step to solving #359 where we also now have to have a subclass of the generator making changes in the Makefrag and build.scala.

hcook:
Being forced to sub-class util.GeneratorApp as an additional step is intentional. Rather than forcing any possible combination of top module and backend to fit through a single generator, we're attempting to force all projects to define their own generator(s), while holding the IO and name of the top module constant across projects. This gives our tools a uniform entry point across many projects; I'd consider this analogous to the fact that main() is always called main() and has the same type signature regardless of what C program you're writing. Finally, because the contents of TestHarness and ExampleTop are constructed based on CONFIG, PROJECT+CONFIG should be sufficient to uniquely identify any frontend target .v file.

jerryhethatday:
> !!! WIP DO NOT MERGE YET (tho it does compile :D) !!!
> 
> Steps towards automatically generating test harnesses from the DUT peripheral traits.
> 
> * rocketchip, groundtest, and unittest are now separate peer packages.
>   
>   * the PROJECT flag is used to select the scala package containing the platform
>   * the CONFIG flag is used to select a particular specific DUT to build
> * All build targets use a Module subclass called TestHarness as their point of entry
>   
>   * TestHarness has a single IO port, a boolean indicating simulation completion
>   * this is analogous to calling main() in a c program
>   * failures are indicated by throwing Chisel assertions

I woner why there are 3 entry points? what is the groundTest and the unittest package for? The readme gives the following :

1. unittest This utility package contains a framework for generateing synthesizable hardware testers of individual modules.
2. groundtest This RTL package generates synthesizable hardware testers that emit randomized memory access streams in order to stress-tests the uncore memory hierarchy.
I wonder when these package will be used? When I run '''make verilog", the TestHarness inside the system will be provoked right？

PR #305, opened 2016-09-16
util: Do BlackBox Async Set/Reset Registers more properly

mwachs5:
The way BlackBox "init" registers were coded before was
not really kosher verilog for most synthesis tools.
Also, the enable logic wasn't really pushed down into the flop.

This change is more explicit about set/reset flops.

Again this is only a temporary fix that would go away
with parametrizable blackboxes (or general async reset support).

### DIFF
--- a/src/main/scala/uncore/util/BlackBoxRegs.scala
+++ b/src/main/scala/uncore/util/BlackBoxRegs.scala
@@ -5,12 +5,16 @@ import Chisel._
 import cde.{Parameters}
 
 /** This black-boxes an Async Reset
-  * Reg.
+  *  (or Set)
+  * Register.
   *  
   * Because Chisel doesn't support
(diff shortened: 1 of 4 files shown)


src/main/scala/uncore/util/BlackBoxRegs.scala
```diff
@@ -22,29 +26,26 @@ import cde.{Parameters}
...
-  *  @param init Value to write at Reset. 
-  *  This is a constant, 
-  *  but this construction
-  *  will likely make backend flows
-  *  and lint tools unhappy.
-  * 
   */
 
-class AsyncResetReg  extends BlackBox {
+abstract class AbstractBBReg extends BlackBox {
```
yunsup: rocket-chip codebase doesn't put an extra line break after the opening curly brace.

PR #308, opened 2016-09-16
Split the TestHarness into its own Verilog file

palmer-dabbelt:
The last commit has the rationale for this (post-syn simulation).  @aswaterman @yunsup I'm not sure how you're running simulations, if there's a better way I'm OK with using that.

@jackkoenig Some of this stuff might want to be in FIRRTL proper, I'm leaving that up to you.

### DIFF
--- a/src/main/scala/rocketchip/Generator.scala
+++ b/src/main/scala/rocketchip/Generator.scala
@@ -8,7 +8,10 @@ import cde._
 import coreplex._
 import firrtl._
 import firrtl.Annotations._
-import java.io.{File, FileWriter, Writer}
+import firrtl.passes.Pass
+import java.io.{File, FileWriter, Writer, PrintWriter}
+
+case object SynTopName extends Field[Option[String]]
 
 case class ParsedInputNames(
     targetDir: String,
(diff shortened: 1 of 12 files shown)


zhemao:
Wait, are you just trying to separate the synthesizable part from the test bench? I don't think this is the right way to go about it.

You should define a separate testbench that makes the dut a BlackBox. Then the post-synth sim scripts could run the generator with MODEL=BlackBoxedTestHarness and MODEL=Top to generate the two verilog files.

palmer-dabbelt:
I'd prefer it done this way: we just merged everything into one invocation
and it seems a lot cleaner, I don't want to regress on that.

palmer-dabbelt:
OK, so this should be fixed now.  If you don't want it I can do this in PLSI, either way is OK for me.

colinschmidt:
I agree with Palmer. Moving the testharness generation into Chisel was a good thing and to me it just seemed like not separating out the harness from the top module was a regression from previous behavior. Having a single invocation generate both in separate files makes plenty of sense to me. I haven't reviewed Palmer's changes yet (will do it now) but I agree with the intended functionality.

src/main/scala/rocketchip/Generator.scala
```diff
@@ -86,10 +91,80 @@ trait Generator extends App with HasGeneratorUtilities {
...
+          new EmitHarnessVerilog(synTopName),
+          Parser.UseInfo
+        )
+
+        firrtl.Driver.compile(
+          s"${names.targetDir}/$longName.fir",
+          s"${names.targetDir}/$longName.v",
+          new EmitTopVerilog(synTopName),
+          Parser.UseInfo,
+          AnnotationMap(Seq(
```
colinschmidt: I assume all these firrtl invocations came from the firrtl repo and are therefore correct.

src/main/scala/rocketchip/Generator.scala
```diff
@@ -86,10 +91,80 @@ trait Generator extends App with HasGeneratorUtilities {
...
+      firrtl.Driver.compile(
+          s"${names.targetDir}/$longName.fir",
+          s"${names.targetDir}/$longName.v",
+          new RocketChipPassManager,
+          Parser.UseInfo
+        )
+      }
+
+      new PrintWriter(new File(s"${names.targetDir}/$longName.TestHarness.v")).close
+      new PrintWriter(new File(s"${names.targetDir}/$longName.conf")).close
```
colinschmidt: Why are these two file close actions here?

src/main/scala/rocketchip/PassManager.scala
```diff
@@ -0,0 +1,61 @@
...
+  // This is some sugar to users don't need to actually know ther's a tagged
+  // union in here to allow them to transparently provide either Pass or
+  // Transform instances to the pass manager.
+  implicit def l2Either[L, R](l: L): Either[L, R] = Left(l)
+  implicit def r2Either[L, R](r: R): Either[L, R] = Right(r)
+
+  // These functions are designed to be overridden by users.  They return a
+  // sequence of transformations that will be applied at various FIRRTL levels.
+  // A transform that operates at a FIRRTL level is expected to both produce
+  // and consume FIRRTL at that level.
```
colinschmidt: I think this makes sense but I also I think people want to write passes that consume low/mid/high and produce high, for convenience, but I'm not sure if the HighToMid And MidToLow are idempotent. I'm sure I read somewhere that they are expected to be eventually but I don't know the current status.

src/main/scala/rocketchip/Passes.scala
```diff
@@ -0,0 +1,78 @@
...
+  def name = "Re-Parent Circuit"
+
+  def run(c: Circuit): Circuit = {
+    Circuit(c.info, c.modules, newTopName)
+  }
+}
+
+// Removes all the unused modules in a circuit by recursing through every
+// instance (starting at the main module)
+class RemoveUnusedModules extends Pass {
```
colinschmidt: I'm not familiar enough with firrtl passes to review this, @azidar @jackkoenig can you check this out?

vsim/Makefrag-verilog
```diff
@@ -9,14 +9,10 @@ firrtl: $(generated_dir)/$(MODEL).$(CONFIG).fir
 
 .PHONY: firrtl
 
-$(generated_dir)/%.$(CONFIG).fir $(generated_dir)/%.$(CONFIG).d $(generated_dir)/%.prm: $(chisel_srcs) $(bootrom_img)
+$(generated_dir)/%.$(CONFIG).v $(generated_dir)/%.$(CONFIG).TestHarness.v $(generated_dir)/%.$(CONFIG).conf $(generated_dir)/%.$(CONFIG).fir $(generated_dir)/%.$(CONFIG).d $(generated_dir)/%.$(CONFIG).prm: $(chisel_srcs) $(bootrom_img)
```
colinschmidt: Nice you also fixed the %.prm bug that has recently been causing a double invocation bug.

zhemao:
In that case, I think these changes should go in PLSI. Putting in all this extra stuff that isn't actually used by rocket-chip itself in the generator doesn't seem right to me. The reason we refactored everything to be the way it is is so that we don't have to throw everything into rocketchip.

palmer-dabbelt:
OK, if you don't want it can you just close the PR?

zhemao:
Sure

jerryhethatday:
> OK, so this should be fixed now. If you don't want it I can do this in PLSI, either way is OK for me.

can you tell me what the PLSI is? I dunt know what its short for?

PR #311, opened 2016-09-17
Rom executable

terpstra:
TL2 ROM needs to be executable!


terpstra:
@mwachs5 This PR should fix the Guava infinite loop bug.

PR #313, opened 2016-09-18
Multiclock2

zhemao:
Here's attempt no. 2. Basically, I've factored it this way.

BaseCoreplex extends Coreplex and adds methods for building the three major blocks:

1. The tiles
2. The L2, l1tol2net, and MMIO manager
3. The MMIO interconnect and coreplex internal devices (PLIC and Debug unit)

The DefaultCoreplex extends Coreplex and connects the three together directly or through queues.

The MultiClockCoreplex extends Coreplex and connects the parts together using asynchronous queues and level synchronizers.


src/main/scala/coreplex/Coreplex.scala
```diff
@@ -46,124 +50,271 @@ case class CoreplexConfig(
...
+  val tileUncachedPorts = tileList.flatMap(tile => tile.io.uncached)
+
+  val l2block = buildL2Block(clock, reset,
+    tileCachedPorts.size, tileUncachedPorts.size)
+  l2block.io.tiles_cached <> tileCachedPorts
+  l2block.io.tiles_uncached <> tileUncachedPorts
+  l2block.io.slave <> io.slave
+  io.master.mem <> l2block.io.mem
+
+  val nTileSlaves = tileList.flatMap(_.io.slave).size
```
colinschmidt: you could keep this flatmap result around as tileSlaves and then use .size in the line below. You end up making this flatmap again later (line 234)

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -46,124 +50,271 @@ case class CoreplexConfig(
...
+    tile.io.cached.map(AsyncClientTileLinkCrossing(
+      tile.clock, tile.reset, _,
+      l2block.clock, l2block.reset))
+  }
+  l2block.io.tiles_uncached <> tileList.flatMap { tile =>
+    tile.io.uncached.map(AsyncClientUncachedTileLinkCrossing(
+      tile.clock, tile.reset, _,
+      l2block.clock, l2block.reset))
   }
+  l2block.io.slave <> io.slave
```
colinschmidt: This doesn't need a clock crossing?

zhemao: Whoops

colinschmidt:
@yunsup Do you have time to look at this?

yunsup:
Yes I do.  The design point that this PR implements makes more sense, however, this PR doesn't implement the synchronous/asynchronous connections via traits.  Even though most of the code is factored out as build* functions, there's still some code duplication.  I'd like to take a stab at touching up the code a little bit, if that's OK with you @zhemao.

colinschmidt:
All these special case crossing objects seem like they could fall back to a single function that takes two bundles...I played around with this a bit for my RoCC decoupler you can check out a gist here: https://gist.github.com/colinschmidt/bc0b15ec90422a87574498e9c9c72ba4

zhemao:
I'll see what I can do about factoring things into traits.

zhemao:
@colinschmidt I agree that we would probably want something more general-purpose. Let's handle that in a later PR.

zhemao:
I've refactored it so that all of the code that's different between DefaultCoreplex and MultiClock coreplex are now in abstract methods of BaseCoreplex that get overridden. This greatly reduces the amount of code duplication.

@yunsup does this look okay?

yunsup:
I probably didn't do a good job describing the trait approach.  Let me take a stab at this.

zhemao:
Let me resolve the conflicts with master before you do that.

zhemao:
@yunsup Are you going to be making your changes today? If it's going to take you a while to get around to it, could we just merge this in and you send a separate PR to fix it up later?

colinschmidt:
I think @yunsup should just fix up the trait things in another PR. I'm basically blocking on this PR for downstream work. I vote that after @zhemao does the rebase to fix conflicts with master he merges it.

yunsup:
Don't merge this PR.  I'll deal with conflicts as well.

The downstream work can always point to a branch to get things going.  We also do this when things are moving slowly on rocket-chip.

yunsup:
I put a bit more thought into this while I was working on this, and I've arrived at a slightly different design point, where the coreplex internal devices (which are on the cBus, named coreplex bus) are on the same clock domain as the l2 block, and the external devices (which are on the pBus, named the peripheral bus) are running off the same (or divided) clock of the l2 block.  I'll probably be able to push my changes to this branch by tomorrow.

zhemao:
Wait, do you mean the periphery devices are running off a clock based on the L2 clock? I don't think that can work, since the L2 clock will be controlled by devices in the periphery.

zhemao:
If the problem is that there is an inefficiency involved in crossing from the L2 clock to the periphery clock and then from the periphery clock to the external IO clock, we'll have to solve that on a case-by-case basis.

yunsup:
The issue you mentioned is orthogonal to the clock crossing between the L2 and the periphery.  The intention of the new design point is that the cBus should be on the same clock domain as the L2, and separately you could put the entire Coreplex on its own clock domain, which is different than the periphery.

yunsup:
I've started a new PR on #321.

colinschmidt:
overwrote by #321

PR #316, opened 2016-09-19
Allow reset vector to be set dynamically

aswaterman:
A chip's power-up sequence, or awake-from-sleep sequence, may wish to set
the reset PC based upon dynamic properties, e.g., the settings of external
pins.  Support this by passing the reset vector to the Coreplex. ExampleTop
simply hard-wires the reset vector, as was the case before.

Additionally, allow MTVEC to *not* be reset.  In most cases, including
riscv-tests, pk, and bbl, overriding MTVEC is one of the first things that
the boot sequence does.  So the reset value is superfluous.


yunsup:
looks good to me

Issue #317, opened 2016-09-21
eliminate hardcoded debugrom contents from Debug.scala 


ekrimer:
There are hardcoded debugrom contents in src/main/scala/uncore/devices/Debug.scala coming from riscv-tools/.... 

I am thinking how to have an easier update cycle for debugrom. I could set it up to pull content from appropriate img files from $RISCV... I think, but then I lose make dependency.

Any thoughts on the best way to organize this? Maybe to have a local copy of debug-rom .S and process it similarly to bootrom? Then there is an issue of possible diverging from spike debug rom, but on the other hand, anyway spike is a little different, and it is up to the user to keep these in sync if that is important.

What are your thoughts?

aswaterman:
Incidentally, we're already working on doing something along those lines.

mwachs5:
Have they already fallen out of sync, or is this a hypothetical?

ekrimer:
@mwachs5 I don't think I noticed them being out of sync, although currently with hardcoded content nothing prevents from getting out of sync. I mainly want to have an easy way to update debugrom. 

@aswaterman :+1:  Glad I asked. I'll then hold off and wait for your changes. Let me know if I can help.

aswaterman:
@ekrimer just to expound, we're working on doing this for the boot ROM, which will serve as a template for how to do it for the debug ROM.  We'll generalize it & port it to the debug ROM thereafter.

ekrimer:
Doesn't bootrom already pull content from file configured by `case BootROMFile => "./bootrom/bootrom.img"` ?

aswaterman:
Only partially.  There's also some boot ROM contents that are spat out of the RTL generator, which are merged together awkwardly with the contents of the boot ROM.

We wanted to clean up that approach before fixing the debug ROM, because in general, the debug ROM contents should be linked based upon output of the RTL generator.

aswaterman:
The net effect is that an ELF will be linked after (and as a result of) the RTL generator executing, in both cases.

ekrimer:
Thanks for clarification.

Actually to followup on divergence point, spike and rocket are actually using different binaries of debug rom. Spike uses only the "debug_rom_raw" version which is compiled with DRV32 and DRV64. Rocket does have that image as "defaultRomContents" but never uses it. Rocket uses either 32- or 64-bit versions that are never used by spike.

[...]

PR #318, opened 2016-09-21
Update Debug.scala

ekrimer:
getting rid of 'defaultRomContents' that is never used by rocket
related to https://github.com/ucb-bar/rocket-chip/issues/317


mwachs5:
See my comment on #317

mwachs5:
In #419 I've renamed the "default" to "xlenAny" to avoid future confusion. I do agree these should still be automatically generated.

PR #319, opened 2016-09-21
[unittest] Parallelize UnitTestSuite

hcook:
All tests now each have their own Timer/timeout, and the suite runs them in parallel until all finish or any timeout.

Also adds SimpleTimer, which tracks a single inflight event and is tolerant to being stopped repeatedly.


zhemao:
Wait, but the reason I serialized the unit tests was because it was very confusing to figure out which one was causing an assertion failure, since they all has the same assertion messages. Have we solved *that* issue yet?

PR #321, opened 2016-09-21
add multiclock support to Coreplex

yunsup:


colinschmidt:
Per in-person code review.

PR #323, opened 2016-09-22
Factor testsuite makefrag generation out of Configs and into Generator

hcook:


zhemao:
What was the reason for getting rid of RegressionTestNames? I think that's what's causing the Travis build to fail.

hcook:
All testsuite makefrag generation is now handled in GeneratorApp and rocketchip.Generator, instead of being threaded haphazardly through the Configs. 

- Sensible default empty make targets are provided for PROJECTs that do not contain Rocket coreplexes.
- Custom GeneratorApps are included in each client PROJECT, and launched using sbt run-main. These will output only the necessary file types.

zhemao:
Henry, there was still one typo. I've pushed a fix for you.

zhemao:
I'll merge once Travis passes.

PR #325, opened 2016-09-22
Rename Coreplex/Peripheral bus, some Coreplex bundle names

yunsup:
These were bothering me while I was adding multiclock support to the Coreplex, but split into a separate PR.  This commit renames the coreplex/peripheral bus to cbus and pbus, and changes the Coreplex bundle names.


zhemao:
Looks good to me.

PR #326, opened 2016-09-22
Make clock divider configurable

zhemao:
This is a change to the ClockDivider blackbox to allow its period to be configured in RTL. Basically, we add a `ctrl` field, which is just a UInt. The output clock will flip on every `ctrl + 1` cycles of the input clock.

I've also fixed the UnitTestConfig so that it runs both the junctions and uncore unit tests.


mwachs5:
Why don't we git rid of this now that we have good clock support? This is easy to write in Chisel.

zhemao:
Oh you're right. I'll try doing that instead.

zhemao:
Okay, so now we have a configurable clock divider implemented in Chisel.

src/main/scala/util/ClockDivider.scala
```diff
@@ -0,0 +1,41 @@
...
+  clock_reg := !clock_reg
+
+  // Hold reset_out high until positive edge of clock_out
+  val reset_reg = RegEnable(Bool(false), Bool(true), clock_reg)
+
+  io.clock_out := clock_reg.asClock
+  io.reset_out := reset_reg
+}
+
+class ClockDivider(n: Int) extends Module {
```
mwachs5: Why don't you just call this "Power of 2 Clock Divider" and then take po2 as the input. Rather than checking that it is a power of 2 and then taking the log of the number.

src/main/scala/util/ClockDivider.scala
```diff
@@ -0,0 +1,41 @@
...
+  val io = new Bundle {
+    val reset_out = Bool(OUTPUT)
+    val clock_out = Clock(OUTPUT)
+  }
+
+  require(isPow2(n), "Clock divider period must be power of 2")
+  require(n >= 2, "Clock divider must divide by at least 2")
```
mwachs5: You could easly support divide-by-1 by just assigning the output to the input

src/main/scala/util/ClockDivider.scala
```diff
@@ -0,0 +1,41 @@
...
+  val io = new Bundle {
+    val reset_out = Bool(OUTPUT)
+    val clock_out = Clock(OUTPUT)
+  }
+
+  val clock_reg = Reg(Bool())
+  clock_reg := !clock_reg
+
+  // Hold reset_out high until positive edge of clock_out
+  val reset_reg = RegEnable(Bool(false), Bool(true), clock_reg)
```
mwachs5: Not sure I really get what reset_out is doing here... at any rate it is not the same as what the verilog model was doing. Could you get the same functionality by just putting this on the last flop in the chain?

src/main/scala/util/ClockDivider.scala
zhemao: Fair enough

src/main/scala/util/ClockDivider.scala
zhemao: It's making sure that reset is held high until the rising edge of the outer clock. I guess it would save registers to just put it at the end.

mwachs5:
I still don't get the reset. @terpstra  you wrote the original... what was the point of the reset output of your ClockDivider? If you want some sort of catcher for the reset, seems like that should be seperate from the divider itself.

zhemao:
Since registers are synchronously reset, you need to make sure the circuit getting the divided clock "sees" the reset. Consider what happens if you assert reset high for a few cycles of the input clock, but there has not been a rising edge on the output clock. The registers connected to the divided clock will not reset.

Maybe the correct thing to do is to require the user to hold the reset high for enough cycles themselves.

zhemao:
Also, Wesley's original verilog clock divider just connected the input reset to the output reset. I'm fine with keeping that behavior if you think it's not worth it to have the divider handle reset synchronization itself.

mwachs5:
I think you should pull the reset stuff out. Make the user do whatever reset synchronization they need, it doesn't really have anything to do with the clock divider.

Also @terpstra pointed out this should flagged with comments that it's only for simulation/ bad for FPGA or the like.

zhemao:
Oh yeah, this definitely wouldn't be correct for the FPGA. On the FPGA you should use the built-in PLLs.

zhemao:
Okay, this is hopefully the last one. The clock divider only handles the clocks now. The TL unit tests for multiple clock domains just use the regular reset for everything.

mwachs5:
lgtm

PR #328, opened 2016-09-22
Atomics

terpstra:
Add atomics to TL2 and re-enable Comparator atomic testing.

### DIFF
--- a/src/main/scala/groundtest/Comparator.scala
+++ b/src/main/scala/groundtest/Comparator.scala
@@ -336,10 +336,17 @@ class ComparatorSink(implicit val p: Parameters) extends Module
       base.addr_beat === g.addr_beat || !g.hasData(),
       base.data === g.data || !g.hasData())
 
+    // TL1 likes to duplicate 32-bits into both halves of a 64-bit value
+    // TL2 doesn't do this, so they compare differently when they are the same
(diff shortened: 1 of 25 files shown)


aswaterman:
I don't have time to CR this tonight, but since it's TL2 stuff, I'm happy to leave merging it to your discretion.

terpstra:
It's not ready for merging just yet. The build seems to run out of RAM
again with Verilator. Also, it's turned up some sort of deadlock in the TL1
L2 cache. :-/ The TL2 port of the Comparator seems to have been working,
though.

terpstra:
Rebased against newest master.
Now includes @hcook 's fix for Legacy not-full-beat atomics.
Now includes firrtl+verilator work-around code (Vec[Bool]=>UInt) for faster builds.

terpstra:
It only does single-beat atomics.

Sai-Manish:
Hello @terpstra,
I wanted to use Boom without atomics support, just disabling in the config doesn't seem to be removing the logic completely :
I have disabled this parameter : https://github.com/riscv-boom/riscv-boom/blob/3d305f2ff5dab9371d9c95754888d4b49baea556/src/main/scala/v3/common/parameters.scala#L88

Is there a better way to remove the support?
Any help would be appreciated.

PR #330, opened 2016-09-22
Move Coreplex to separate clock in ExampleMultiClockTop

zhemao:
This moves the coreplex as a whole to a separate clock and adds synchronizers from the coreplex to the periphery.

It also uses clock dividers to simulate that things work even if coreplex clock, tile clock and main clock are all different.

The one unsettled question is whether we need synchronizers for the interrupts and coreplex-local interrupts. Currently, the coreplex-local interrupts remain high until the tile intervenes to set them low, so perhaps it's okay? There might still be metastability events though.

### DIFF
--- a/src/main/scala/coreplex/BaseCoreplex.scala
+++ b/src/main/scala/coreplex/BaseCoreplex.scala
@@ -74,6 +74,11 @@ abstract class BaseCoreplexModule[+L <: BaseCoreplex, +B <: BaseCoreplexBundle](
   val nUncachedPorts = tiles.map(tile => tile.io.uncached.size).reduce(_ + _)
   val nBanks = c.nMemChannels * nBanksPerMemChannel
 
+  val uncoreExtMem = Wire(Vec(c.nMemChannels, new ClientUncachedTileLinkIO()(outermostParams)))
+  val uncoreExtMMIO = Wire(new ClientUncachedTileLinkIO()(outermostMMIOParams))
+  val uncoreExtSlave = Wire(Vec(c.nSlaves, new ClientUncachedTileLinkIO()(innerParams)))
+  val uncoreExtDebug = Wire(new DebugBusIO)
+
   // Build an uncore backing the Tiles
   buildUncore(p.alterPartial({
     case HastiId => "TL"
@@ -106,7 +111,7 @@ abstract class BaseCoreplexModule[+L <: BaseCoreplex, +B <: BaseCoreplexBundle](
     // Wire the tiles to the TileLink client ports of the L1toL2 network,
     // and coherence manager(s) to the other side
     l1tol2net.io.clients_cached <> uncoreTileIOs.map(_.cached).flatten
-    l1tol2net.io.clients_uncached <> uncoreTileIOs.map(_.uncached).flatten ++ io.slave
+    l1tol2net.io.clients_uncached <> uncoreTileIOs.map(_.uncached).flatten ++ uncoreExtSlave
     l1tol2net.io.managers <> managerEndpoints.map(_.innerTL) :+ mmioManager.io.inner
 
     // Create a converter between TileLinkIO and MemIO for each channel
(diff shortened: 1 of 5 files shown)


zhemao:
Okay, so it seems that the clint interrupts were actually getting synchronized anyway, just from the wrong clock. I'll add a commit to fix that.

yunsup:
The additional logic to put the coreplex in its own clock domain should live outside the coreplex (i.e., in BaseTop).  I suggest you start over.

zhemao:
What do you mean? The coreplex is moved into a different clock in ExampleMultiClockExampleTop. The extcr clock and reset are the clock and reset for the periphery. It's needed by coreplex in order to synchronize its outputs. 

Are you suggesting I put the clock crossers in ExampleMultiClockExampleTop? How would that work for CoreplexLocalInterrupts, which need to cross from the external clock domain to the tile clock domain (and not the coreplex clock domain)?

zhemao:
I guess I could have that clock crosser on the outside as well, since the periphery has access to the tile clock.

zhemao:
Wait, it would actually be really annoying to do it that way, since you would need to override all one of the periphery traits (debug, mem, mmio, slave, and clint) instead of just one.

colinschmidt:
I see both of your points. Moving the clock crossings out of coreplex makes sense but it does  require overriding all periphery traits that communicate with the coreplex.

The periphery traits already use some parameters to make more clock crossings (AsyncMemChannels, etc.) so maybe a parameter AsyncCoreplexCrossing or w/e could be used.

You could also attempt to check that coreplex.clock and top.clock are the same at elaboration time but I don't know if anyone has done that in chisel before. This would allow you to change the traits without introducing more parameters so it just kind of automagically works...

aswaterman:
I don't think checking whether the clocks are different then instantiating async FIFOs is a good solution, either.  Probably better to error than automagically instantiate clock crossings.  Additionally, a truly asynchronous split between the coreplex and the periphery bus is not performant; you'd always either use the same clock, or synchronously divide it down.  It kind of sounds like a tapeout hack.

zhemao:
It's not exactly that simple, since you would also need to know what the tile clock is in order to properly synchronize the coreplex-local interrupts. I agree with Andrew that automatically instantiating clock-crossings is not the way to go.

zhemao:
It also does seem rather silly to cross the clock from the coreplex to the periphery domain for the converters and then cross again to the external domain. Maybe there's no avoiding adding separate asynchronous traits for the IOs.

colinschmidt:
Ok so you're going to make another async trait for all the periphery stuff? @yunsup will that be ok with you?

aswaterman:
@zhemao yeah - it makes sense that various peripherals would be asynchronous with the coreplex, but it's hard to see a good reason for the peripheral bus itself to be asynchronous with the coreplex.

yunsup:
No.  Having another async trait for all periphery is not the way to go.  Take a closer look at the Coreplex, and see how I exposed the wires to the buildUncore and buildMMIONetwork functions.  The Periphery *does not* need to know about the async crossings.  Expose the wires to the Periphery, and have one trait (either direct connection or async connection) that does the connection between the Coreplex and the Periphery.

zhemao:
Okay, I'll try it.

zhemao:
Wait, but that's still messy for the CoreplexLocalInterrupts though.

So basically in the periphery trait, we would have a clock crosser from Periphery clock domain to tile clock domain. Then in the coreplex we would have to be aware that the periphery is doing the clock-crossing and not do any clock crossing itself. Doesn't that break the abstraction? I think it really is cleaner this way, in which the periphery sees the coreplex IOs as all being in the periphery's clock domain.

yunsup:
Let me take a look at it.

colinschmidt:
@yunsup do you have an update?

yunsup:
Yes, you're right that the CoreplexLocalInterrupts bundle will be synchronized twice, and that's not good.  However, the clint device will soon be pushed into the coreplex (we're waiting to turn the cbus into TL2), and this will be a non-problem.

There's still a question of how we should handle two (or more) clock-crossing through the module hierarchy that can be optimized away.  My general answer is that this information needs to get negotiated in scala land first (in the LazyModule), before Chisel code is elaborated.  The current case where you can optimize the synchronizers by having both clock crossings within the Coreplex is a degenerate case of optimizing clock-crossing elements.

That said, I think the way to go for now is to instantiate all clock crossings outside the coreplex (to put the coreplex and the periphery on different clock domains), and wait for the clint to be pushed into the coreplex, and cross the bridge later when we need to optimize further.

I'll take a crack at implementing the clock crossings outside the Coreplex.

zhemao:
I'll just go ahead and move the clock crossings into the periphery then.

zhemao:
The other PR https://github.com/ucb-bar/rocket-chip/pull/348

PR #334, opened 2016-09-23
fix some incorrect behavior on enabling external mmio port.

zizztux:
When I config ExtMMIOPorts with MIFDataBits == 128, there two issue.
1. undefined TLId error
2. MSB of DataBeats TL signal is unconnected.

This commit fix them.


zhemao:
I guess this is the consequence of us moving a bunch of devices out into the periphery. I'll take a look and make sure this is the right thing to do.

zhemao:
I'll take care of this in PR #335

PR #335, opened 2016-09-23
Fixup Tracegen

zhemao:
There have been some changes to AXE which break our current implementation of TraceGen. These fixes help us get back on track. Basically, we have to make sure we store some known value to all of the addresses we are using before we do any other operation.

### DIFF
--- a/src/main/scala/coreplex/BaseCoreplex.scala
+++ b/src/main/scala/coreplex/BaseCoreplex.scala
@@ -115,9 +115,9 @@ abstract class BaseCoreplexModule[+L <: BaseCoreplex, +B <: BaseCoreplexBundle](
     val outerTLParams = p.alterPartial({ case TLId => "L2toMC" })
     val backendBuffering = TileLinkDepths(0,0,0,0,0)
     for ((bank, icPort) <- managerEndpoints zip mem_ic.io.in) {
-      val unwrap = Module(new ClientTileLinkIOUnwrapper()(outerTLParams))
-      unwrap.io.in <> TileLinkEnqueuer(bank.outerTL, backendBuffering)(outerTLParams)
-      TileLinkWidthAdapter(icPort, unwrap.io.out)
+      val enqueued = TileLinkEnqueuer(bank.outerTL, backendBuffering)
+      val unwrapped = TileLinkIOUnwrapper(enqueued)
+      TileLinkWidthAdapter(icPort, unwrapped)
     }
 
     io.master.mem <> mem_ic.io.out
--- a/src/main/scala/coreplex/BaseCoreplex.scala
+++ b/src/main/scala/coreplex/BaseCoreplex.scala
@@ -32,7 +32,7 @@ trait HasCoreplexParameters {
   lazy val lsb = p(BankIdLSB)
   lazy val innerParams = p.alterPartial({ case TLId => "L1toL2" })
   lazy val outermostParams = p.alterPartial({ case TLId => "Outermost" })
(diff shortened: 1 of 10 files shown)


zizztux:
Hi.
Thanks to your merging.

but, it seems to miss my another commit.
I send PR with 2 commit.
please, check it.

zhemao:
@zizztux The second commit shouldn't be necessary. If it is, then there's something wrong with the TileLinkWidthAdapter.

zhemao:
I've tested our MIF128BitConfig with these changes and it seems to pass.

zhemao:
I'm going to push a commit that makes the TileLinkWidthAdapter etc. less error prone.

zhemao:
Is this okay to merge in? It's definitely more correct than what we have now.

hcook:
lgtm. 

49a88f6 was definitely an anti-pattern....

src/main/scala/coreplex/BaseCoreplex.scala
```diff
@@ -32,7 +32,7 @@ trait HasCoreplexParameters {
   lazy val lsb = p(BankIdLSB)
   lazy val innerParams = p.alterPartial({ case TLId => "L1toL2" })
   lazy val outermostParams = p.alterPartial({ case TLId => "Outermost" })
-  lazy val outermostMMIOParams = p.alterPartial({ case TLId => "MMIO_Outermost" })
+  lazy val outerMMIOParams = p.alterPartial({ case TLId => "L2toMMIO" })
```
yunsup: We should be consistent with how we name these things.  Currently, in rocketchip/Periphery.scala, we call these Parameters as:

  lazy val innerParams = p.alterPartial({ case TLId => "L1toL2" })
  lazy val innerMMIOParams = p.alterPartial({ case TLId => "L2toMMIO" })
  lazy val outermostParams = p.alterPartial({ case TLId => "Outermost" })
  lazy val outermostMMIOParams = p.alterPartial({ case TLId => "MMIO_Outermost" })

zhemao: Yeah, I noticed that later on. I think they should both be outerMMIOParams, since they are at the same level as outerParams (with TLKey "L2toMC").

yunsup: It really doesn't matter (with TL2 rollout), but I think innerMMIOParams is better than outerMMIOParams.  IMO, it's more confusing to have outerMMIOParams vs. outermostMMIOParams.

src/main/scala/rocketchip/Periphery.scala
```diff
@@ -213,7 +213,7 @@ trait PeripheryMasterMMIOModule extends HasPeripheryParameters {
   val pBus: TileLinkRecursiveInterconnect
 
   val mmio_ports = p(ExtMMIOPorts) map { port =>
-    TileLinkWidthAdapter(pBus.port(port.name), "MMIO_Outermost")
+    TileLinkWidthAdapter(pBus.port(port.name), innerMMIOParams)
```
yunsup: Howie, is this the right change?  This diff changes the actual code.  Was this a bug before?

zhemao: Whoops. That should be outermostMMIOParams.

zhemao: That's odd because I tested MIF32BitConfig and MIF128BitConfig and it seemed to work.

zhemao: Oh wait, those configs don't actually instantiate an external MMIO port, that's why.

PR #336, opened 2016-09-23
Async rst async queue

mwachs5:
@terpstra @yunsup 

In general, it's a good idea to have asynchronous reset on the crossing FIFOs, since the blocks may come out of reset at different times, clocks may not have started yet, etc.

This PR achieves that and blows away the non-async-reset option.

Are there any other low-level crossing primitives that need this treatment?

### DIFF
--- a/src/main/scala/junctions/asyncqueue.scala
+++ b/src/main/scala/junctions/asyncqueue.scala
@@ -76,7 +76,10 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int, clockIn: Clock, r
   // On an FPGA, only one input changes at a time => mem updates don't cause glitches
   // The register only latches when the selected valued is not being written
   val index = if (depth == 1) UInt(0) else ridx(bits-1, 0) ^ (ridx(bits, bits) << (bits-1))
-  io.deq.bits  := RegEnable(io.mem(index), valid && !reset) // This does NOT need to be reset.
+  // This register does not NEED to be reset, as its contents will not
+  // be considered unless deq_reg is true.
+  io.deq.bits  := RegEnable(io.mem(index), valid) 
+    
 
   val deq_reg = AsyncResetReg(valid, 0)
   io.deq.valid := deq_reg
--- a/src/main/scala/junctions/asyncqueue.scala
+++ b/src/main/scala/junctions/asyncqueue.scala
@@ -5,19 +5,24 @@ import Chisel._
 
 object GrayCounter {
   def apply(bits: Int, increment: Bool = Bool(true)): UInt = {
-    val binary = RegInit(UInt(0, width = bits))
-    val incremented = binary + increment.asUInt()
-    binary := incremented
+    val binary = Module (new AsyncRegVec(w=bits, 0))//RegInit(UInt(0, width = bits))
+    val incremented = binary.io.q + increment.asUInt()
+    binary.io.d := incremented
+    binary.io.en := Bool(true)
     incremented ^ (incremented >> UInt(1))
   }
 }
 
 object AsyncGrayCounter {
   def apply(in: UInt, sync: Int): UInt = {
-    val syncv = RegInit(Vec.fill(sync){UInt(0, width = in.getWidth)})
-    syncv.last := in
-    (syncv.init zip syncv.tail).foreach { case (sink, source) => sink := source }
-    syncv(0)
+    val syncv = Vec.fill(sync)(new AsyncResetRegVec(w = in.getWidth, 0))//RegInit(Vec.fill(sync){UInt(0, width = in.getWidth)})
+    syncv.last.io.d := in
+      (syncv.init zip syncv.tail).foreach { case (sink, source) => {
+        sink.io.d := source.io.q
+        sink.io.en := Bool(true)
+      }
+      }
+    syncv(0).io.d
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/main/scala/junctions/asyncqueue.scala
```diff
@@ -66,9 +75,13 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int, clockIn: Clock, r
   // On an FPGA, only one input changes at a time => mem updates don't cause glitches
   // The register only latches when the selected valued is not being written
   val index = if (depth == 1) UInt(0) else ridx(bits-1, 0) ^ (ridx(bits, bits) << (bits-1))
-  io.deq.bits  := RegEnable(io.mem(index), valid && !reset)
-  io.deq.valid := RegNext(valid, Bool(false))
-  io.ridx := RegNext(ridx, UInt(0))
+  io.deq.bits  := RegEnable(io.mem(index), valid && !reset) // This does NOT need to be reset.
```
mwachs5: Actually @terpstra why is *this* reset included in the enable logic?

mwachs5: @terpstra I removed this reset. Let me know if there was something I was missing there.

terpstra:
How urgently do you need this merged? I'd like to go over it very carefully on Monday.

src/main/scala/junctions/asyncqueue.scala
```diff
@@ -2,22 +2,28 @@
...
 import Chisel._
+import uncore.util.{AsyncResetRegVec, AsyncResetReg}
 
 object GrayCounter {
   def apply(bits: Int, increment: Bool = Bool(true)): UInt = {
-    val binary = RegInit(UInt(0, width = bits))
-    val incremented = binary + increment.asUInt()
-    binary := incremented
+    val incremented = Wire(UInt(width=bits))
+    val binary = AsyncResetReg(incremented, 0, Bool(true))
```
terpstra: Wouldn't
  val binary = AsyncResetReg(incremented)
be enough?

src/main/scala/junctions/asyncqueue.scala
```diff
@@ -33,15 +39,19 @@ class AsyncQueueSource[T <: Data](gen: T, depth: Int, sync: Int, clockIn: Clock,
...
   val index = if (depth == 1) UInt(0) else io.widx(bits-1, 0) ^ (io.widx(bits, bits) << (bits-1))
   when (io.enq.fire() && !reset) { mem(index) := io.enq.bits }
-  io.enq.ready := RegNext(ready, Bool(false))
-  io.widx := RegNext(widx, UInt(0))
+  val ready_reg = AsyncResetReg(ready, 0)
+  io.enq.ready := ready_reg
+
+  val widx_reg = AsyncResetReg(widx, 0)
+  io.widx := widx_reg
+
```
terpstra: Why did you add intermediate variable names? :-(

src/main/scala/junctions/asyncqueue.scala
```diff
@@ -66,9 +76,16 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int, clockIn: Clock, r
   // On an FPGA, only one input changes at a time => mem updates don't cause glitches
   // The register only latches when the selected valued is not being written
   val index = if (depth == 1) UInt(0) else ridx(bits-1, 0) ^ (ridx(bits, bits) << (bits-1))
-  io.deq.bits  := RegEnable(io.mem(index), valid && !reset)
-  io.deq.valid := RegNext(valid, Bool(false))
-  io.ridx := RegNext(ridx, UInt(0))
+  // This register does not NEED to be reset, as its contents will not
+  // be considered unless deq_reg is true.
+  io.deq.bits  := RegEnable(io.mem(index), valid)
```
terpstra: The !reset was there as a work-around for the fact that I did not have async reset. I didn't want it latching garbage. You were right to remove it.

src/main/scala/junctions/asyncqueue.scala
```diff
@@ -66,9 +76,16 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int, clockIn: Clock, r
...
+  // This register does not NEED to be reset, as its contents will not
+  // be considered unless deq_reg is true.
+  io.deq.bits  := RegEnable(io.mem(index), valid) 
+    
+
+  val deq_reg = AsyncResetReg(valid, 0)
+  io.deq.valid := deq_reg
+
+  val ridx_reg = AsyncResetReg(ridx, 0)
+  io.ridx := ridx_reg
```
terpstra: You just like making my pretty code ugly, don't you?

src/main/scala/junctions/asyncqueue.scala
```diff
@@ -2,22 +2,28 @@
...
-    syncv(0)
+    val syncv = List.fill(sync)(Module (new AsyncResetRegVec(w = in.getWidth, 0)))
+    syncv.last.io.d := in
+    syncv.last.io.en := Bool(true)
+      (syncv.init zip syncv.tail).foreach { case (sink, source) => {
+        sink.io.d := source.io.q
+        sink.io.en := Bool(true)
+      }
+      }
+    syncv(0).io.d
```
terpstra: This is rather hideous. Oh well...

mwachs5: Yeah. I tried to think of some Scala fanciness, then decided sometimes you should just write what you mean.

src/main/scala/junctions/asyncqueue.scala
mwachs5: ah, this was the remnant of my having actual module instantiations before I realized there were nice objects now. I'll tidy up.

src/main/scala/junctions/asyncqueue.scala
mwachs5: I like to be clear about the reset value. But you're right about the Bool(true). I removed that.

src/main/scala/junctions/asyncqueue.scala
mwachs5: tidied.

PR #337, opened 2016-09-23
Periphery: make bus width and arithmetic atomics configurable

terpstra:
This removes the hard-coded constants that had started pervading the design and leverages the new RegisterRouter's flexibility of bus width.


terpstra:
@yunsup Please review these changes. I have changes in another repo that leverage them.

PR #338, opened 2016-09-23
tilelink2 Xbar: decouple ready from valid

terpstra:
This moves the Xbar from using custom code to using the Arbiter.
The arbiter has better ready-valid decoupling.


terpstra:
@aswaterman This switches the TL2 Xbar to decouple ready from valid, as discussed.

aswaterman:
cool - it's hard to give a quick CR because most of the changes are lines moving around, but I certainly approve it on technical grounds

PR #339, opened 2016-09-23
AXI Fragmenter

zhemao:
Not sure what happened to Wes' AXI fragmenter for the RISCV workshop demo. We need something like it for our CRAFT verification work, so I figured I would add my own.

For multibeat aw, we wait for the input w request before sending both an aw and w out.

For multibeat ar, we take in the input request and then send some number of output requests. This could be optimized for the case of single beat requests, but I don't think the extra logic is worth it.

We don't forward b responses except for the last one. For r responses, we just mess with the "last" field.

Only one set of transactions with the same ID is allowed to be inflight at a time. We save the lens for each transaction in registers index by the ID.

Since we have the fragmenter, we can also avoid handling multibeat requests in AXI -> TL converter.

### DIFF
--- a/src/main/scala/rocketchip/Periphery.scala
+++ b/src/main/scala/rocketchip/Periphery.scala
@@ -76,6 +76,13 @@ object PeripheryUtils {
     bridge.io.tl <> tl
     bridge.io.ahb
   }
+  def convertAXItoTL(axi: NastiIO)(implicit p: Parameters) = {
+    val frag = Module(new NastiFragmenter)
+    val bridge = Module(new TileLinkIONastiIOConverter)
+    frag.io.in <> axi
+    bridge.io.nasti <> addQueueAXI(frag.io.out)
+    bridge.io.tl
+  }
 }
 
 /** Utility trait for quick access to some relevant parameters */
@@ -281,12 +288,11 @@ trait PeripherySlaveModule extends HasPeripheryParameters {
         else AsyncNastiFrom(io.bus_clk.get(idx), io.bus_rst.get(idx), bus)
       )
     }
-    val conv = Module(new TileLinkIONastiIOConverter()(edgeSlaveParams))
-    conv.io.nasti <> arb.io.slave
 
     val (r_start, r_end) = outer.pBusMasters.range("ext")
     require(r_end - r_start == 1, "RangeManager should return 1 slot")
-    TileLinkWidthAdapter(coreplexIO.slave(r_start), conv.io.tl)
+    val tl = PeripheryUtils.convertAXItoTL(arb.io.slave)(edgeSlaveParams)
+    TileLinkWidthAdapter(coreplexIO.slave(r_start), tl)
   }
 }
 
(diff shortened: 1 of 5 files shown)


zhemao:
@terpstra Is there some AXI protocol requirement that I'm forgetting? Would be good to make sure.

terpstra:
It seems reasonable. I'd make sure your test cases include a narrow burst,
though.

src/main/scala/rocketchip/Periphery.scala
```diff
@@ -274,12 +281,10 @@ trait PeripherySlaveModule extends HasPeripheryParameters {
         else AsyncNastiFrom(io.bus_clk.get(idx), io.bus_rst.get(idx), bus)
       )
     }
-    val conv = Module(new TileLinkIONastiIOConverter()(innerParams))
-    conv.io.nasti <> arb.io.slave
 
     val r = outer.pBusMasters.range("ext")
     require(r._2 - r._1 == 1, "RangeManager should return 1 slot")
-    coreplexIO.slave(r._1) <> conv.io.tl
+    coreplexIO.slave(r._1) <> PeripheryUtils.convertAXItoTL(arb.io.slave)(innerParams)
```
terpstra: I would ask Henry Styles to run a regression check on the PCIe dev board before you switch this over. AFAIK, he is the only one affected.

zhemao:
The test case is a narrow burst. It's using 4 beats of 32 bits each.

terpstra:
Have you run this by Henry yet? I'd like to reuse some of this code in the TL2-AXI4 converter pipeline.

terpstra:
I note that you have the same busy register used for both AR and AW... those are actually distinct namespaces. It's also not super efficient to only allow a single transfer per ID.

terpstra:
You also don't handle burst types other than increment. There's wrapping and fixed to consider.

terpstra:
You also don't correctly update the address for a misaligned start access. Please consult section A3.4.1 of the AXI4 specification.

Aligned_Address = (INT(Start_Address / Number_Bytes) ) x Number_Bytes
Address_N = Aligned_Address + (N – 1) × Number_Bytes

terpstra:
Hmm. I think you would also need to modify the size of the first beat in a Fragmenter.

terpstra:
Err, no, cancel that. The first transfer will still be small even if size is unchanged, so that is fine. You definitely still need to increment the address properly, though. Otherwise you would have the following:

address = 2, size = 2 (2^2=4 bytes), len = 4:
CORRECT: write(2, 2), write(4, 4), write(8, 4), write(12, 4)
YOURS: write(2,2), write(6,2), write(10,2), write(14, 2)
for a function write(offset, numBytes)

zhemao:
Henry said there's some other bug that broke PCIe he needs to resolve before testing this. 

The reason I'm not supporting more than one ID in-flight is because that will greatly expand the amount of state I need to keep. Instead of a few registers per ID I would need a queue per ID.

You're right that I do need to fix the support for FIXED_BURST among other things. I wasn't aware that start addresses could be misaligned. Thanks for pointing that out.

zhemao:
It seems that we are adding AXI fragmentation a different way, so I'm closing this one. We actually don't need this anymore for CRAFT, since we are verifying the actual AXI (NASTI) crossbar.

PR #340, opened 2016-09-25
rocketchip: rename identically names devices with _%d

terpstra:
If you connect two devices with the same name in TL2 (totally ok there),
when they get put into the TL1 addrmap, one gets silently overwritten.
This renames the second occurance as _1, third as _2, and so on.


terpstra:
@zhemao Perhaps this fix should be moved into the AddrMap itself. At least the AddrMap should fail a requirement if you double-add the same name?

zhemao:
Yeah, there should probably be a requirement in the address map. I'm not sure if you want to implicitly rename devices.

terpstra:
I agree. A requirement failure is probably the right fix.

PR #341, opened 2016-09-26
rocket-chip: makes some parameters configurable.

zizztux:
I need to configure these parameters. And other users of rocket-chip also, I think.
- external memory base address
- number of external interrupt
- async io port

### DIFF
--- a/src/main/scala/rocketchip/Configs.scala
+++ b/src/main/scala/rocketchip/Configs.scala
@@ -65,6 +65,7 @@ class BasePlatformConfig extends Config(
         case AsyncMemChannels => false
         case NMemoryChannels => Dump("N_MEM_CHANNELS", 1)
         case TMemoryChannels => BusType.AXI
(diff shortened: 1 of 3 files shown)


zizztux:
travis-ci build result
https://travis-ci.org/zizztux/rocket-chip/builds/162664885

I don't know how upstream github system receives build result from forked repository's travis-ci.
so, I attach the link, above.

zhemao:
Maybe @aswaterman can comment more on this, but I'm not sure we want the memory base to be a configurable parameter. It's specified in the platform spec that it must start at 0x80000000 and all of our software is written with this in mind.

I'm fine with the other changes.

terpstra:
WithNExtInterrupts exists in current master.
I'm not sure if ExtMemBase should be configurable or not.

We need to re-add the capability to cross the AXI ports to a foreign clock domain.

hcook:
Closing this pending further discussion in #465 and #466

PR #343, opened 2016-09-26
Tl2 graphml

terpstra:
This entertaining little PR adds support for generating fancy TL2 bus diagrams.
Attached is the ExampleTop of rocket chip as it stands right now, as rendered by yEd using hierarchical layout. More sophisticated bus topologies look even cooler.

rocket

### DIFF
--- a/src/main/scala/rocketchip/BaseTop.scala
+++ b/src/main/scala/rocketchip/BaseTop.scala
@@ -8,7 +8,7 @@ import junctions._
 import uncore.tilelink._
 import uncore.tilelink2._
 import uncore.devices._
-import util.{ParameterizedBundle, ConfigStringOutput}
+import util.{ParameterizedBundle, ConfigStringOutput, GraphMLOutput}
 import rocket._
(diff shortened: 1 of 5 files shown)


terpstra:
rocket2
Radial layout

terpstra:
Rebased against master.

yunsup:
setting aside Wes' previous commit message "rocketchip: globals are for sissies", this PR looks good to me.

colinschmidt:
Will all documentation generation tools live inside rocket-chip going forward?

terpstra:
I don't think there is a concrete plan to do this. However, if something
integrates directly with codegen, I guess it should go in rocket-chip.


So, for example, you won't find the latex for the platform spec in
rocket-chip, but something that extracts topology and/or register layouts
from the scala code, I guess needs to go where the code goes..?

terpstra:
@yunsup Globals ARE for sissies. However, sometimes the easy thing to do is also the right thing. =)

PR #344, opened 2016-09-26
Get rid of NastiDemux and MemIO

zhemao:
This removes some code that we no longer use. As far as I'm aware, we are no longer using anything related to MemIO. The only reason we still had the file around was for the MIF* config fields. I've just moved these into rocketchip.

We are also no longer using the NastiMemoryDemux or NastiMemorySelector. I think we added these for a previous tapeout, but we've now switched to doing everything in TileLink and put the extra code in a separate addon module.


yunsup:
Wow!  MemIO is finally going away.  MemIO served its purpose very well for the past 6 years.  :)

src/main/scala/rocketchip/Periphery.scala
```diff
@@ -17,6 +17,11 @@ import rocket.XLen
 import scala.math.max
 import coreplex._
 
+case object MIFAddrBits extends Field[Int]
+case object MIFDataBits extends Field[Int]
+case object MIFTagBits extends Field[Int]
+case object MIFDataBeats extends Field[Int]
+
```
hcook: I propose taking it a step further and actually eliminating all the MIF* parameter fields entirely. Keeping them around when the chip now supports multiple external interfaces seems like it can only lead to confusion down the road. They're basically just feeding the creation of the outermost NastiKey and nothing else. Ideally NastiKey, HastiKey and TLKey would all be handled in the same way when used to parameterize external interfaces.

zhemao: I think their main use is to change MIFDataBits to fit whatever interface width you have on the outside.

Probably the best way to do this is to change them to be ExtMemWidth and ExtMMIOWidth.

src/main/scala/groundtest/NastiTest.scala
```diff
@@ -19,7 +18,7 @@ class NastiGenerator(id: Int)(implicit val p: Parameters) extends Module
     val mem = new NastiIO
   }
 
-  val mifDataBytes = mifDataBits / 8
+  val nastiBytes = nastiXDataBits / 8
```
zhemao: It's really the same as nastiXStrobeBits. But I guess we could also have a nastiXDataBytes.

hcook: Yeah i think it makes the code clearer to have both.

src/main/scala/rocketchip/Periphery.scala
hcook: As usual there's a chicken and egg question about whether the outside drives the inside or the inside drives the outside. I agree with splitting MIFDataBits into two parameters (and eliminate the other ones?). However, I think ExtMMIOWidth should just be forced to always be p(PeripheryBusKey).beatBytes...

zhemao: Yeah, that makes sense.

zhemao: So I think we actually still need an ExtMMIODataBits. This is just for the MMIO ports that we export off chip. It's actually possible that we will need different widths for each exported MMIO port. But for now, let's just have one.

hcook: I guess my point was, given we'll need different widths for each exported MMIO port, wouldn't it make sense to just use the .beatBytes field of whatever ParameterKey we have for that port? This is still progress though.

src/main/scala/rocketchip/Configs.scala
```diff
@@ -27,17 +27,14 @@ class BasePlatformConfig extends Config(
       lazy val innerDataBeats = (8 * site(CacheBlockBytes)) / innerDataBits
       pname match {
         //Memory Parameters
-        case MIFTagBits => Dump("MIF_TAG_BITS", 5)
-        case MIFDataBits => Dump("MIF_DATA_BITS", 64)
-        case MIFAddrBits => Dump("MIF_ADDR_BITS",
```
hcook: I think you aren't dumping MIF_ADDR_BITS anymore... Do we actually still need to export all the MEM_XX_BITS things via the dump? Isn't the test harness that uses them now defunct?

src/main/scala/rocketchip/Periphery.scala
zhemao: I thought the PeripheryBusKey was for the TL2 periphery bus? The external MMIOs are still coming from the legacy TL1 bus.

src/main/scala/rocketchip/Configs.scala
zhemao: I don't think we need to dump anything anymore.

zhemao:
I'll just start this over again. There's going to be conflicts with the TLId changes.

hcook:
Yeah, I think aside from resolving conflicts, the main question is how the following will be factored:

`case TLKey("EdgetoSlave") => site(TLKey("L1toL2")).copy(dataBeats = site(MIFDataBeats))
case TLKey("MCtoEdge") => site(TLKey("L2toMC")).copy(dataBeats = site(MIFDataBeats))
case TLKey("MMIOtoEdge") => site(TLKey("L2toMMIO")).copy(dataBeats = site(MIFDataBeats))`

Seems like these should be three separate parameters; but I also don't think it's worth naming new fields for each? Maybe `WithMIFDataBits(n: Int)` should just override the TLKeys directly?

src/main/scala/rocketchip/Configs.scala
hcook: Let's try removing at least all the Dumps related to MIF_ stuff.

zhemao:
They are basically all based on the NastiKey, so maybe the best thing to do would be to get derive them from the NastiKey.

zhemao:
On the other hand, the memory could go to AHB.

zhemao:
Actually, I think the current "Ext" HastiKey is incorrect. It's using XLen for its width instead of being the same width as the MCtoEdge TileLink key.

yunsup:
Howie, I think you're right.  Adding @terpstra

yunsup:
You might be able to get rid of HastiKey("Ext"), change always do HastiId => "TL" in rocketchip/Configs.scala (and get rid of HastiId in coreplex/Configs.scala), and just do "val mem_ahb = Vec(nMemAHBChannels, new HastiMasterIO()(edgeMemParams))" instead.

zhemao:
Yeah, that could work.

Maybe we should do that for the NastiKey as well. Have it based off of the TileLink key.

yunsup:
Actually, I don't think that's a bad idea.  @hcook?

zhemao:
So Henry and I talked and we think that could be very error prone, since we have to remember to change the TLId for all the Nasti and Hasti bundles and converters.

We've decided to just have a WithMemDataBits config that just changes the NastiKey, "Ext" HastiKey, and relevant TLKeys.

PR #346, opened 2016-09-26
Tl2 addr width0

terpstra:
Refactor use of the address into accessor methods addr_{lo,hi}.
This should solve the problem @a0u had with beatBytes==1 on SPI.
This should also make it easier when we bite the bullet and have address (lo+hi) on ABC.


a0u:
I can confirm that the misaligned address assertion no longer triggers with the SPI controller.

PR #348, opened 2016-09-27
Multiclocked Coreplex attempt #2

zhemao:
The clock crossers have now been moved into Periphery traits. I'm not sure this is better, I had to mess around with the type system a lot to make this work. I suppose one advantage is that we need not instantiate clock crossers for the debug bus if the debug bus is disabled. I'm leaving the other PR up to in case we want to compare.


ben-k:
It looks like you didn't actually merge in either version - is that your intent?

zhemao:
Yeah, I'm PR'ing a third version after talking with Yunsup.

PR #349, opened 2016-09-27
Give TileLink keys more sensible names

zhemao:
The "Outermost" and "MMIO_Outermost" keys are renamed to "MCtoEdge"
and "MMIOtoEdge". They are only used after width conversion to adapt to
external interfaces in the periphery.

That means there is no longer any width adapting done in the Coreplex,
the width adapter is moved to the Periphery just before conversion
to AXI/AHB or sending off chip.

We also add an additional TLKey "EdgetoSlave" for the bus_axi interface.
This means there is also a width adapter from the AXI -> TL converter
to the coreplex slave port.

This also fixes a bug introduced previously in which the wrong
parameters object is given to the width adapter for external MMIO.

### DIFF
--- a/src/main/scala/coreplex/BaseCoreplex.scala
+++ b/src/main/scala/coreplex/BaseCoreplex.scala
@@ -31,7 +31,7 @@ trait HasCoreplexParameters {
   lazy val nBanksPerMemChannel = p(NBanksPerMemoryChannel)
   lazy val lsb = p(BankIdLSB)
   lazy val innerParams = p.alterPartial({ case TLId => "L1toL2" })
-  lazy val outerParams = p.alterPartial({ case TLId => "L2toMC" })
+  lazy val outerMemParams = p.alterPartial({ case TLId => "L2toMC" })
   lazy val outerMMIOParams = p.alterPartial({ case TLId => "L2toMMIO" })
   lazy val globalAddrMap = p(rocketchip.GlobalAddrMap)
 }
@@ -50,7 +50,7 @@ abstract class BaseCoreplex(c: CoreplexConfig)(implicit p: Parameters) extends L
 
 abstract class BaseCoreplexBundle(val c: CoreplexConfig)(implicit val p: Parameters) extends Bundle with HasCoreplexParameters {
   val master = new Bundle {
-    val mem = Vec(c.nMemChannels, new ClientUncachedTileLinkIO()(outerParams))
+    val mem = Vec(c.nMemChannels, new ClientUncachedTileLinkIO()(outerMemParams))
     val mmio = new ClientUncachedTileLinkIO()(outerMMIOParams)
   }
   val slave = Vec(c.nSlaves, new ClientUncachedTileLinkIO()(innerParams)).flip
@@ -112,7 +112,7 @@ abstract class BaseCoreplexModule[+L <: BaseCoreplex, +B <: BaseCoreplexBundle](
     l1tol2net.io.managers <> managerEndpoints.map(_.innerTL) :+ mmioManager.io.inner
 
     // Create a converter between TileLinkIO and MemIO for each channel
-    val mem_ic = Module(new TileLinkMemoryInterconnect(nBanksPerMemChannel, c.nMemChannels)(outerParams))
+    val mem_ic = Module(new TileLinkMemoryInterconnect(nBanksPerMemChannel, c.nMemChannels)(outerMemParams))
 
     val backendBuffering = TileLinkDepths(0,0,0,0,0)
     for ((bank, icPort) <- managerEndpoints zip mem_ic.io.in) {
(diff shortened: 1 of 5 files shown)


src/main/scala/coreplex/BaseCoreplex.scala
```diff
@@ -50,7 +50,7 @@ abstract class BaseCoreplex(c: CoreplexConfig)(implicit p: Parameters) extends L
 
 abstract class BaseCoreplexBundle(val c: CoreplexConfig)(implicit val p: Parameters) extends Bundle with HasCoreplexParameters {
   val master = new Bundle {
-    val mem = Vec(c.nMemChannels, new ClientUncachedTileLinkIO()(outermostParams))
+    val mem = Vec(c.nMemChannels, new ClientUncachedTileLinkIO()(outerMemParams))
```
yunsup: This is changing behavior.  Was this a bug before?

zhemao: No, before the width adapter was still inside the coreplex. Now the width adapter is in the periphery. That way there are no edge* parameters in the coreplex. I think that's the most sensible and consistent way.

src/main/scala/coreplex/BaseCoreplex.scala
```diff
@@ -110,14 +110,12 @@ abstract class BaseCoreplexModule[+L <: BaseCoreplex, +B <: BaseCoreplexBundle](
...
-    val mem_ic = Module(new TileLinkMemoryInterconnect(nBanksPerMemChannel, c.nMemChannels)(outermostParams))
+    val mem_ic = Module(new TileLinkMemoryInterconnect(nBanksPerMemChannel, c.nMemChannels)(outerMemParams))
 
-    val outerTLParams = p.alterPartial({ case TLId => "L2toMC" })
     val backendBuffering = TileLinkDepths(0,0,0,0,0)
     for ((bank, icPort) <- managerEndpoints zip mem_ic.io.in) {
       val enqueued = TileLinkEnqueuer(bank.outerTL, backendBuffering)
-      val unwrapped = TileLinkIOUnwrapper(enqueued)
-      TileLinkWidthAdapter(icPort, unwrapped)
+      icPort <> TileLinkIOUnwrapper(enqueued)
```
yunsup: I see what you're doing here now.  You've moved the width conversion out to the Periphery.  I support this change.

src/main/scala/rocketchip/Periphery.scala
```diff
@@ -173,7 +175,8 @@ trait PeripheryMasterMemModule extends HasPeripheryParameters {
 
   // Abuse the fact that zip takes the shorter of the two lists
   ((io.mem_axi zip coreplexIO.master.mem) zipWithIndex) foreach { case ((axi, mem), idx) =>
-    val axi_sync = PeripheryUtils.convertTLtoAXI(mem)
+    val mem_edge = TileLinkWidthAdapter(mem, edgeMemParams)
```
yunsup: Why don't you factor mem_edge out of the specific axi, ahb, tl connection routines?

yunsup:
This commit has 3 different things going on.  I suggest you split them out to separate commits:

1) bug fix
2) making coreplex use outerMemParams instead of edgeMemParams
3) rename parameters

yunsup:
Howie, can you also take a look at why the regression failed?

zhemao:
Yeah, the regressions failed because I forgot that the old Outermost TLKey was also increasing the size of the client_xact_id. So now the FancyMemtestConfig is failing because ID bits are getting cut off. I'm fixing that now by just making the client_xact_id for L2toMC large enough to get through the arbiters.

src/main/scala/rocketchip/Periphery.scala
zhemao: Good point

zhemao: Done

zhemao:
I'll try to split the commits.

src/main/scala/coreplex/BaseCoreplex.scala
hcook: Yes this seems more straightforward

src/main/scala/coreplex/Configs.scala
```diff
@@ -132,15 +132,11 @@ class BaseCoreplexConfig extends Config (
...
-          maxClientsPerPort = site(NAcquireTransactors) + 2,
+          maxClientXacts = site(NAcquireTransactors) + 2,
+          maxClientsPerPort = site(NBanksPerMemoryChannel),
           maxManagerXacts = 1,
           dataBeats = innerDataBeats,
           dataBits = site(CacheBlockBytes)*8)
-      case TLKey("Outermost") => site(TLKey("L2toMC")).copy(
-        maxClientXacts = site(NAcquireTransactors) + 2,
-        maxClientsPerPort = site(NBanksPerMemoryChannel),
-        dataBeats = site(MIFDataBeats))
```
hcook: Coreplex not knowing about "outermost"/"edgeX" also feels correct.

yunsup:
This is great!

PR #350, opened 2016-09-27
Factor coreplex IO connection into separate trait

zhemao:
This would allow, for instance, putting the coreplex on a separate clock
domain and crossing the IOs over through asynchronous queues.

The ExampleMultiClockTop* classes are removed since they no longer fit
into the class hierarchy.

### DIFF
--- a/src/main/scala/rocketchip/BaseTop.scala
+++ b/src/main/scala/rocketchip/BaseTop.scala
@@ -67,7 +67,7 @@ abstract class BaseTopModule[+L <: BaseTop, +B <: BaseTopBundle](
   val io: B = b
 
   val coreplex = p(BuildCoreplex)(outer.c, p)
-  val coreplexIO = coreplex.io
+  val coreplexIO = Wire(coreplex.io)
 
   val pBus =
     Module(new TileLinkRecursiveInterconnect(1, p(GlobalAddrMap).subMap("io:pbus"))(
@@ -94,4 +94,11 @@ abstract class BaseTopModule[+L <: BaseTop, +B <: BaseTopBundle](
   GraphMLOutput.contents = Some(outer.graphML)
 
   io.success := coreplexIO.success
+}
+
+trait DirectConnection {
+  val coreplexIO: BaseCoreplexBundle
+  val coreplex: BaseCoreplexModule[BaseCoreplex, BaseCoreplexBundle]
+
+  coreplexIO <> coreplex.io
 }
--- a/src/main/scala/coreplex/BaseCoreplex.scala
+++ b/src/main/scala/coreplex/BaseCoreplex.scala
@@ -59,6 +59,8 @@ abstract class BaseCoreplexBundle(val c: CoreplexConfig)(implicit val p: Paramet
   val clint = Vec(c.nTiles, new CoreplexLocalInterrupts).asInput
   val resetVector = UInt(INPUT, p(XLen))
   val success = Bool(OUTPUT) // used for testing
+
+  override def cloneType = this.getClass.getConstructors.head.newInstance(c, p).asInstanceOf[this.type]
 }
 
 abstract class BaseCoreplexModule[+L <: BaseCoreplex, +B <: BaseCoreplexBundle](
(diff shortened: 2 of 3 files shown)


src/main/scala/rocketchip/BaseTop.scala
```diff
@@ -67,7 +67,8 @@ abstract class BaseTopModule[+L <: BaseTop, +B <: BaseTopBundle](
   val io: B = b
 
   val coreplex = p(BuildCoreplex)(outer.c, p)
-  val coreplexIO = coreplex.io
+  val coreplexIO = Wire(new BaseCoreplexBundle(outer.c)(p){})
```
yunsup: This is bad, since if the user extended the type, you won't get the same wire.  You're better off doing "val coreplexIO = Wire(coreplex.io)" and implicitly calling the clone method.

src/main/scala/rocketchip/BaseTop.scala
```diff
@@ -67,7 +67,8 @@ abstract class BaseTopModule[+L <: BaseTop, +B <: BaseTopBundle](
...
+  val baseCoreplexIO = coreplex.io
```
yunsup: I would come up with a better name than baseCoreplexIO.  There's already a BaseCoreplex module, but in this case, you're not referring to that.

src/main/scala/rocketchip/BaseTop.scala
zhemao: I tried doing that at first, but that requires every *CoreplexBundle class to explicitly define a cloneType method. My thought process was that if you have extra ports, you should handle those yourself. I guess it's fine to require a cloneType method for every CoreplexBundle.

src/main/scala/rocketchip/BaseTop.scala
zhemao: That's true. Maybe unsynchronizedCoreplexIO or asyncCoreplexIO since it's the version before it gets synchronized?

zhemao: I decided to just avoid this problem entirely by using coreplex.io directly in the DirectConnection trait.

yunsup: I second your decision.

src/main/scala/rocketchip/BaseTop.scala
yunsup: Actually, similar to ParameterizedBundle, you can do this in the BaseCoreplexBundle, and get rid of all cloneTypes in the extended classes:
```
  override def cloneType: BaseCoreplexBundle.this.type =
    this.getClass.getConstructors.head.newInstance(c, p).asInstanceOf[this.type]
```

zhemao: Yeah, I thought of doing something like that bus wasn't sure how. I'll try it, thanks.

zhemao: Updated

src/main/scala/coreplex/Coreplex.scala
```diff
@@ -76,9 +77,9 @@ class MultiClockCoreplex(c: CoreplexConfig)(implicit p: Parameters) extends Base
...
-    with TileClockResetBundle
+class MultiClockCoreplexBundle(c: CoreplexConfig)(implicit p: Parameters)
+    extends BaseCoreplexBundle(c)(p) with TileClockResetBundle
 
-class MultiClockCoreplexModule[+L <: MultiClockCoreplex, +B <: MultiClockCoreplexBundle](
-    c: CoreplexConfig, l: L, b: => B)(implicit p: Parameters) extends BaseCoreplexModule(c, l, b)(p)
-    with AsyncConnection
+class MultiClockCoreplexModule[+L <: MultiClockCoreplex, +B <: MultiClockCoreplexBundle]
+    (c: CoreplexConfig, l: L, b: => B)(implicit p: Parameters)
+    extends BaseCoreplexModule(c, l, b)(p) with AsyncConnection
```
yunsup: Can you not touch the lines above?  This is not relevant to this PR.

zhemao: Yeah, sorry. I'll figure out a way to revert it.

zhemao: Fixed

PR #352, opened 2016-09-27
JTAG DTM & Debug Module fixes

### LINKED ISSUE #347: Jtag DTM Synchronizer Broken
My last commit #342 screwed up the JTAG DTM.  My tests didn't have good clean targets. Fixing the issue now.

mwachs5:
This partially resolves #347 . I added the missing connection between the synchronizer and the DTM, and fixed some issues in the DebugModule where the debugRAM address was aliasing if debugRAM size was less than 64 bytes.

I started bringing the DebugTransportModuleJtag in line with @timsifive's latest RISC-V spec (adding sticky bits and additional fields in dtminfo). OpenOCD doesn't yet support the sticky bits however.


mwachs5:
This is fixed in the HW, but I am still working out some software kinks before merging this in.

mwachs5:
The latest openOCD now works with this version. I'll merge after @aswaterman merges #360 .

aswaterman:
cool

aswaterman:
oops, i killed the wrong travis job, i'll fix it

PR #354, opened 2016-09-27
crossing: Remove reset from the logic in Register Crossing 

mwachs5:
It is no longer needed when the underlying crossings are asynchronously reset. Update the order of operations.

@terpstra  can you take a look at this?


zhemao:
I went ahead and merged it, since I want to get my PR through as well.

PR #355, opened 2016-09-27
tilelink2: don't use chisel3 namespace

terpstra:
@aswaterman This removes all uses of chisel3._ as requested


aswaterman:
thanks, go ahead and merge it once it passes travis

PR #361, opened 2016-09-29
Add TL2 synthesizeable UnitTests for many adapters

hcook:
These tests mostly use the Fuzzer and RAMModel to check that adapters correctly handle-randomly generated legal traffic, without deadlocking or corrupting data.

These tests are added to the UnittestSuite target in regression/Makefile. They should fit within the existing timeframe of the other suites but might have to be adjusted if they start taking too long to elaborate.

### DIFF
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -52,7 +52,7 @@ endif
 
 ifeq ($(SUITE),UnittestSuite)
 PROJECT=unittest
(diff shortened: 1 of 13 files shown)


hcook:
This particular assortment of UnitTests took 14m in Travis, as opposed to 23m for the other Suites

PR #362, opened 2016-09-29
Deterministic interrupt

yunsup:
the interrupt ranges are now assigned deterministically.

### DIFF
--- a/src/main/scala/rocketchip/Utils.scala
+++ b/src/main/scala/rocketchip/Utils.scala
@@ -15,19 +15,19 @@ import java.nio.{ByteBuffer, ByteOrder}
 
 class RangeManager {
   private var finalized = false
-  private val l = collection.mutable.HashMap[String, Int]()
+  private val l = collection.mutable.ListBuffer[(String, Int)]()
   def add(name: String, element: Int) = { require(!finalized); l += (name -> element) }
   def rangeMap = {
     finalized = true
-    l map {
+    (l map {
       var sum = 0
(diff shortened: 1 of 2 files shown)


src/main/scala/rocketchip/Utils.scala
```diff
@@ -15,19 +15,19 @@ import java.nio.{ByteBuffer, ByteOrder}
...
       var sum = 0
       x => { sum += x._2; (x._1 -> (sum-x._2, sum)) }
-    }
+    }).toMap
   }
   def range(name: String) = rangeMap(name)
   def print = {
-    rangeMap map { case (name, (start, end)) =>
-      println(s"${name} on port ${start}-${end-1}")
+    rangeMap.toSeq.sortBy(_._2).map { case (name, (start, end)) =>
```
zhemao: Couldn't you just iterate through the ListBuffer again here? It will save you from having to call sortBy. Also, you should definitely be using foreach and not map. Something like

    var start = 0
    l.foreach { case (name, size) =>
        val end = start + size
        println(...)
        start = end
    }

yunsup: Wouldn't it be unnecessarily risky to recalculate the interrupt numbers?  sortBy was added to print the assigned interrupt numbers in a ordered fashion.  This code will go away as soon as the cbus and its devices are turned into TL2.

zhemao: Sure. You should still probably be using foreach instead of map, though.

yunsup: You're right.

PR #363, opened 2016-09-29
bump torture pointer after sim name change

colinschmidt:
Just to keep things in sink.  Feel free to merge as part of some other PR if you want. Its just the master branch of torture


yunsup:
Thanks Colin for taking care of this!

PR #365, opened 2016-09-30
Fix up SlowIO and add receiver + unit test

zhemao:
Changing up the SlowIO module for hurricane.

We need both the transmitter (driving the slow clock) and receiver (watching the slow clock) side. I think this is actually a bit tricky, so I'll explain my thinking so that we can convince ourselves that it's correct.

The transmitter holds its outputs constant for some (configurable) period before and after the rising edge and takes its inputs (just before) the rising edge.

The receiver doesn't know when the clock edges will occur, so it holds everything constant from the falling to (right after) the next rising edge. It takes its inputs right after the rising edge of the slow clock.

This way, the signals going in each direction will all be constant for some period (even if it is just one fast cycle) around the rising edge of the slow clock. So both transmitter and receiver see the correct input at each rising edge.

### DIFF
--- a/src/main/scala/junctions/slowio.scala
+++ b/src/main/scala/junctions/slowio.scala
@@ -3,68 +3,179 @@
 package junctions
 import Chisel._
 
-class SlowIO[T <: Data](val divisor_max: Int)(data: => T) extends Module
-{
+class SlowIOClockDivider(val divisor_max: Int) extends Module {
   val io = new Bundle {
-    val out_fast = Decoupled(data).flip
-    val out_slow = Decoupled(data)
-    val in_fast = Decoupled(data)
-    val in_slow = Decoupled(data).flip
-    val clk_slow = Bool(OUTPUT)
-    val set_divisor = Valid(Bits(width = 32)).flip
-    val divisor = Bits(OUTPUT, 32)
+    val divisor = UInt(INPUT, log2Up(divisor_max))
+    val slow_clk = Bool(OUTPUT)
+    val slow_clk_rising = Bool(OUTPUT)
+    val slow_clk_falling = Bool(OUTPUT)
   }
 
-  require(divisor_max >= 8 && divisor_max <= 65536 && isPow2(divisor_max))
-  val divisor = Reg(init=UInt(divisor_max-1))
-  val d_shadow = Reg(init=UInt(divisor_max-1))
-  val hold = Reg(init=UInt(divisor_max/4-1))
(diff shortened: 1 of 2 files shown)


zhemao:
Oh crap, I did not mean to change the torture submodule. Let me fix that.

aswaterman:
Hopefully this really is just a cleanup and maintains identical behavior -- if so, it will definitely work.

aswaterman:
Your reason doesn't seem to match my original design.  "This way, the signals going in each direction will all be constant for some period (even if it is just one fast cycle)" isn't how it used to work -- there was a programmable hold margin, because one fast cycle isn't necessarily enough to meet the hold time requirements of the slow side.

zhemao:
The transmitter is still holding it for a configurable number of cycles before and after the rising edge. It's the receiver that might change its inputs one cycle later. I'm not entirely sure what to do about that, since the receiver doesn't know how many cycles there are until the next rising edge.

I think one issue is that there is a period during which the outputs can change at will. I realize now that, in your original design, the outputs could only change once per cycle (some configurable hold time away from the rising edge).

What if we did it this way: Both transmitter and receiver take the input on the rising edge and only change their output on the falling edge. Then we wouldn't have to configure the hold time. And both sides would be consistent with each other.

aswaterman:
I think that, for both expediency and a reasonable guarantee you'll get a working chip, you should stick with my design.  It has worked without problem every time and it seems risky to change it (I think you already identified one bug in the new design).

Refactoring it's fine, but I really think you should be cautious here.

zhemao:
We still need the receiver though. So there will still be an unknown element even if we take your transmitter design. What did we do for the receiver before?

colinschmidt:
I agree with Andrew this is one of the places we need to be conservative.

zhemao:
Okay. I've pushed a different design. It follows the same behavior as Andrew's original design. Except instead of changing the outputs at a configurable hold time from the rising edge, it changes the outputs at the falling edge. The receiver follows the same behavior.

aswaterman:
Is the receiver the thing that runs on the FPGA?  The reason we didn't have any code for it in rocket-chip is that it's completely trivial.  You just use the received clock (piped through an IBUFG) to clock the registers directly.

It's also not critical to get right, because you can change it later.

aswaterman:
Dude, I'm telling you, leave it alone.

zhemao:
Sure, let's just abandon this then.

PR #366, opened 2016-09-30
Change the configuration interface of SlowIO

zhemao:
This is a much more minimal change than before.

Previously, the SlowIO module had a set_divisor valid interface. Sends over it would update two shadow registers, which then get picked up on the falling edge of the clock.

In hurricane2, the host and divisor value will already be held in SCRs, so it doesn't really make sense to use a valid interface into another shadow register.

Instead, we just let the SCRs be the shadow registers and pass their value in directly. Also, I'm separating them out into two different inputs instead of a single combined 32-bit value from which we pick out two 16-bit values. It's less confusing this way. If we want to make sure we set them at the same time, we just merge them into a single status register and break them out accordingly.


aswaterman:
Makes sense to me.

PR #369, opened 2016-09-30
Suggest sane names for common objects

mwachs5:
I find that I end up with a bunch of "AsyncResetRegVec_N" and other commonly used structures in my hierarchy. I'd prefer to give them names like "myReg", "yourReg", "myReg_rcrossing", etc.

I'm not sure this is the right way to go about this, happy for feedback and suggestions.

This is an example of where https://github.com/ucb-bar/chisel3/issues/303 would make a difference.

This PR also resolves the difference in behavior between instantiating a single AsyncResetReg (with a Bool) and a AsyncResetReg (UInt).

### DIFF
--- a/src/main/scala/uncore/tilelink2/RegisterCrossing.scala
+++ b/src/main/scala/uncore/tilelink2/RegisterCrossing.scala
@@ -147,15 +147,18 @@ object AsyncRWSlaveRegField {
     slave_reset: Bool,
     width:  Int,
     init: Int,
+    name: Option[String] = None,
     master_allow: Bool = Bool(true),
     slave_allow: Bool = Bool(true)
   ): (UInt, RegField) = {
 
     val async_slave_reg = Module(new AsyncResetRegVec(width, init))
+    if (name.isDefined) {async_slave_reg.suggestName(name.get)}
     async_slave_reg.reset := slave_reset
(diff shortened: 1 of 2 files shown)


src/main/scala/uncore/tilelink2/RegisterCrossing.scala
```diff
@@ -147,15 +147,18 @@ object AsyncRWSlaveRegField {
     slave_reset: Bool,
     width:  Int,
     init: Int,
+    name: Option[String] = None,
     master_allow: Bool = Bool(true),
     slave_allow: Bool = Bool(true)
   ): (UInt, RegField) = {
 
     val async_slave_reg = Module(new AsyncResetRegVec(width, init))
+    if (name.isDefined) {async_slave_reg.suggestName(name.get)}
```
hcook: You're welcome to reject this as too-cute but I'd usually express something like the above as:
`name.foreach(async_slave_reg.suggestName(_))`

Basically you can treat the option as a (possibly empty) collection. See http://blog.tmorris.net/posts/scalaoption-cheat-sheet/ for more...

mwachs5: That was exactly the kind of feedback I was looking for, knew there had to be a more scala-y way.

PR #372, opened 2016-10-02
Tl2 rr irrevocable fix

terpstra:
The RegisterRouter fundamentally uses a Decoupled interface for arbitration amongst RegFields. Furthermore, a device might update a RegField while it is being read. This is Decoupled behaviour. This PR accepts the Decoupled nature of the RegisterRouter and uses a bypassable Queue to make the behaviour Irrevocable as required by TL2. The test case and API of RegField have been relaxed accordingly.

### DIFF
--- a/src/main/scala/uncore/tilelink2/RegField.scala
+++ b/src/main/scala/uncore/tilelink2/RegField.scala
@@ -3,7 +3,7 @@
 package uncore.tilelink2
 
 import Chisel._
-import chisel3.util.{Irrevocable, IrrevocableIO}
+import chisel3.util.{ReadyValidIO}
 import util.{SimpleRegIO}
 
 case class RegReadFn private(combinational: Boolean, fn: (Bool, Bool) => (Bool, Bool, UInt))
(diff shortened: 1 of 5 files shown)


terpstra:
@hcook It seems we got the Irrevocable fix for the RegisterRouter wrong last time. It is in fact Decoupled, and we should test for that.

terpstra:
@aswaterman The actual fix was adding the Queue.irrevocable call in RegisterRouter.scala. The rest of the PR is just documenting and testing what this change means.

terpstra:
This breaks another repo, where the fix is to remove the Irrevocable and PostQueueIrrevocablize calls.

terpstra:
Argh. Having to put a register on the output data path of every RegisterRouter device makes me question the wisdom of using Irrevocable in the first place. The whole reason to do it was to save datapath registers in things like the Fragmenter and WidthWidget.

terpstra:
The register between the clint and the WidthWidget is unavoidable. The only question is whose job it is to put down a register. There are less WidthWidgets than RegisterRouters, so I think that choice is clear.

The question then becomes: how much would it cost to make the Fragmenter accept Decoupled? I suspect it would only require registers for the 'address', 'size', and 'source' on the A channel.

aswaterman:
Decoupled vs. Irrevocable to be postponed; merging as-is for now.

PR #374, opened 2016-10-02
correctly initialize $random function with seed

yunsup:


yunsup:
don't merge this in.  there's a better fix than this.

PR #376, opened 2016-10-03
correctly initialize $random with seed

yunsup:
I've gone back and forth on this, and regardless of whether we use $urandom or $random, this code seems necessary anyways, so we should merge this in.

### DIFF
--- a/vsrc/TestDriver.v
+++ b/vsrc/TestDriver.v
@@ -21,6 +21,12 @@ module TestDriver;
   reg [1023:0] vcdfile = 0;
   initial
   begin
+    // do not delete the line below
+    // $random function needs to be called with the seed once
+    // to affect all the downstream $random functions within the
+    // Chisel-generated Verilog code
+    $fdisplay(stderr, "seed %0d, testing $random %0x", unsigned'($get_initial_random_seed), $random($get_initial_random_seed));
+
     $value$plusargs("max-cycles=%d", max_cycles);
     verbose = $test$plusargs("verbose");
 `ifdef DEBUG


vsrc/TestDriver.v
```diff
@@ -21,6 +21,12 @@ module TestDriver;
   reg [1023:0] vcdfile = 0;
   initial
   begin
+    // do not delete the line below
+    // $random function needs to be called with the seed once
+    // to affect all the downstream $random functions within the
+    // Chisel-generated Verilog code
+    $fdisplay(stderr, "seed %0d, testing $random %0x", unsigned'($get_initial_random_seed), $random($get_initial_random_seed));
```
aswaterman: I agree we should print the seed, but the "testing $random" seems like debugging code that doesn't belong upstream.  Can just call $random after the $fdisplay.

yunsup: I think it's defensible to print out the first $random call.  One would quickly know when the $random is not changing for different versions of simulators (as well as simulators from other vendors).

aswaterman: Isn't get_initial_random_seed vcs-specific anyway?  You'd need to port this code to another simulator manually...

aswaterman: I guess I don't really care, just thought I'd point it out.

yunsup: You're right about get_initial_random_seed.  Just wanted to note that we would have noticed this problem 2 years ago when this problem got introduced, if $random was printed out.  I still vote for keeping this display statement.

aswaterman: makes sense

PR #377, opened 2016-10-03
use $urandom as seed for $random

yunsup:
for better interoperability between different Verilog simulators.  Turns out $get_initial_random_seed is VCS only.  h/t @scottj97

### DIFF
--- a/vsrc/TestDriver.v
+++ b/vsrc/TestDriver.v
@@ -19,16 +19,23 @@ module TestDriver;
   reg [63:0] trace_count = 0;
   reg [1023:0] vcdplusfile = 0;
   reg [1023:0] vcdfile = 0;
+  int unsigned rand_value;
   initial
   begin
-    // do not delete the line below
-    // $random function needs to be called with the seed once
-    // to affect all the downstream $random functions within the
-    // Chisel-generated Verilog code
-    $fdisplay(stderr, "seed %0d, testing $random %0x", unsigned'($get_initial_random_seed), $random($get_initial_random_seed));
-
     $value$plusargs("max-cycles=%d", max_cycles);
     verbose = $test$plusargs("verbose");
+
+    // do not delete the line below.
+    // $random function needs to be called with the seed once to affect all
+    // the downstream $random functions within the Chisel-generated Verilog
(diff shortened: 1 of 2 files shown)


vsrc/TestDriver.v
```diff
@@ -25,7 +25,7 @@ module TestDriver;
     // $random function needs to be called with the seed once
     // to affect all the downstream $random functions within the
     // Chisel-generated Verilog code
-    $fdisplay(stderr, "seed %0d, testing $random %0x", unsigned'($get_initial_random_seed), $random($get_initial_random_seed));
+    $fdisplay(stderr, "testing $random %0x", $random($urandom));
```
aswaterman: Please remove the $fdisplay

aswaterman: Better yet, guard it by +verbose.  You do have a point that this is a useful sanity check.

yunsup: pushed a change, thanks.

yunsup:
The test failing is not my fault.  My change doesn't even affect source code that Verilator accepts.  I'm going to run VCS with random seeds for 1000 times, and call it a day if I can't find the problem.

vsrc/TestDriver.v
```diff
@@ -19,16 +19,21 @@ module TestDriver;
...
   begin
+    $value$plusargs("max-cycles=%d", max_cycles);
+    verbose = $test$plusargs("verbose");
+
     // do not delete the line below
     // $random function needs to be called with the seed once
     // to affect all the downstream $random functions within the
     // Chisel-generated Verilog code
-    $fdisplay(stderr, "seed %0d, testing $random %0x", unsigned'($get_initial_random_seed), $random($get_initial_random_seed));
+    rand_value = $random($urandom);
```
scottj97: I would add an explanation here of why $urandom is used. How about:
// $urandom is seeded via cmdline (+ntb_random_seed in VCS) but that doesn't seed $random

yunsup: thanks Scott.  I'll add the comment once #380 goes through, and force push to this branch.

PR #380, opened 2016-10-03
correctly initialize the active flag

yunsup:
I've reproduced the problem in #377, and this is the fix.  The SimpleTimer wasn't reseting the active flag, and depending on the random value, the timer was raising an error.


yunsup:
I'm just going to merge this with #377

PR #382, opened 2016-10-05
axi4: diplomacy capable AXI4 with TL2 bridge

terpstra:
Diplomacy style AXI4 now ready for bridging to PCIe.


terpstra:
@henrystyles We should try this on the VC707 as soon as it passes Travis.

terpstra:
Rebased against newest master

PR #387, opened 2016-10-07
Safer crossings

terpstra:
This change makes AsyncQueue and TLAsync{Source,Sink} automatically clear the FIFO counters and safely block ready/valid when one half of the crossing is reset.

@jchang0 Here is the branch to test

### DIFF
--- a/src/main/scala/regmapper/RegisterCrossing.scala
+++ b/src/main/scala/regmapper/RegisterCrossing.scala
@@ -35,6 +35,15 @@ class BusyRegisterCrossing extends Module {
   io.master_response_valid  := (bypass    || io.crossing_request_ready) &&  busy
 }
 
+class RegisterCrossingAssertion extends Module {
+  val io = new Bundle {
+    val master_bypass = Bool(INPUT)
+    val slave_reset = Bool(INPUT)
+  }
+
+  assert (io.master_bypass || !io.slave_reset)
+}
+
 // RegField should support connecting to one of these
 class RegisterWriteIO[T <: Data](gen: T) extends Bundle {
   val request  = Decoupled(gen).flip
@@ -101,6 +110,12 @@ class RegisterWriteCrossing[T <: Data](gen: T, sync: Int = 3) extends Module {
   crossing.io.deq.ready := Bool(true)
   io.slave_valid := crossing.io.deq.valid
   io.slave_register := crossing.io.deq.bits
+
+  val assertion = Module(new RegisterCrossingAssertion)
+  assertion.clock := io.master_clock
+  assertion.reset := io.master_reset
+  assertion.io.master_bypass := io.master_bypass
+  assertion.io.slave_reset := io.slave_reset
 }
 
 // RegField should support connecting to one of these
@@ -147,6 +162,12 @@ class RegisterReadCrossing[T <: Data](gen: T, sync: Int = 3) extends Module {
 
   crossing.io.enq.valid := Bool(true)
   crossing.io.enq.bits := io.slave_register
+
+  val assertion = Module(new RegisterCrossingAssertion)
+  assertion.clock := io.master_clock
+  assertion.reset := io.master_reset
+  assertion.io.master_bypass := io.master_bypass
+  assertion.io.slave_reset := io.slave_reset
 }
 
 /** Wrapper to create an
--- a/src/main/scala/regmapper/RegisterCrossing.scala
+++ b/src/main/scala/regmapper/RegisterCrossing.scala
@@ -25,7 +25,7 @@ class BusyRegisterCrossing(clock: Clock, reset: Bool)
 
 // RegField should support connecting to one of these
 class RegisterWriteIO[T <: Data](gen: T) extends Bundle {
-  val request  = Decoupled(gen).flip()
+  val request  = Decoupled(gen).flip
   val response = Irrevocable(Bool()) // ignore .bits
 }
 
@@ -85,7 +85,7 @@ class RegisterWriteCrossing[T <: Data](gen: T, sync: Int = 3) extends Module {
 
 // RegField should support connecting to one of these
 class RegisterReadIO[T <: Data](gen: T) extends Bundle {
-  val request  = Decoupled(Bool()).flip() // ignore .bits
+  val request  = Decoupled(Bool()).flip // ignore .bits
   val response = Irrevocable(gen)
 }
 
--- a/src/main/scala/regmapper/RegisterCrossing.scala
+++ b/src/main/scala/regmapper/RegisterCrossing.scala
@@ -7,20 +7,32 @@ import chisel3.util.{Irrevocable}
 import util.{AsyncQueue,AsyncResetRegVec}
 
 // A very simple flow control state machine, run in the specified clock domain
-class BusyRegisterCrossing(clock: Clock, reset: Bool)
-    extends Module(_clock = clock, _reset = reset) {
+class BusyRegisterCrossing extends Module {
   val io = new Bundle {
-    val progress       = Bool(INPUT)
-    val request_valid  = Bool(INPUT)
-    val response_ready = Bool(INPUT)
-    val busy           = Bool(OUTPUT)
+    val bypass                  = Bool(INPUT)
+    val master_request_valid    = Bool(INPUT)
+    val master_request_ready    = Bool(OUTPUT)
+    val master_response_valid   = Bool(OUTPUT)
+    val master_response_ready   = Bool(INPUT)
+    val crossing_request_valid  = Bool(OUTPUT)
+    val crossing_request_ready  = Bool(INPUT)
+    // ... no crossing_response_ready; we are always ready
   }
 
-  val busy = RegInit(Bool(false))
-  when (io.progress) {
-    busy := Mux(busy, !io.response_ready, io.request_valid)
+  val busy   = RegInit(Bool(false))
+  val bypass = Reg(Bool())
+
+  when (io.crossing_request_ready || Mux(busy, bypass, io.bypass)) {
+    busy := Mux(busy, !io.master_response_ready, io.master_request_valid)
   }
-  io.busy := busy
+
+  when (io.master_request_valid && io.master_request_ready) {
+    bypass := io.bypass
+  }
+
+  io.crossing_request_valid := io.master_request_valid && !io.bypass && !busy
+  io.master_request_ready   := (io.bypass || io.crossing_request_ready) && !busy
+  io.master_response_valid  := (bypass    || io.crossing_request_ready) &&  busy
 }
 
 // RegField should support connecting to one of these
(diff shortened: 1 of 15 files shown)


terpstra:
Rebased to remove the isolation commit that Megan already merged.

src/main/scala/util/AsyncQueue.scala
```diff
@@ -61,9 +72,16 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int) extends Module {
...
     val mem  = Vec(depth, gen).asInput
+    // Reset for the other side
+    val source_reset_n = Bool().flip
   }
 
-  val ridx = GrayCounter(bits+1, io.deq.fire())
+  // extend the source reset to a full cycle (assertion latency <= 1 cycle)
+  val catch_reset_n = AsyncResetReg(Bool(true), clock, !io.source_reset_n)
+  // reset_n has a 1 cycle shorter path to valid than widx does
+  val reset_n = AsyncGrayCounter(catch_reset_n.asUInt, sync)(0)
```
mwachs5: I think you need more synchronization on your reset catcher (for the de-assertion). In general, you're going to have to draw me a diagram of what's going on here before I'd be comfortable using this instead of doing my own reset synchronization outside.

src/main/scala/util/AsyncQueue.scala
```diff
@@ -74,11 +92,18 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int) extends Module {
...
-  io.deq.bits  := RegEnable(io.mem(index), valid) 
-    
-  io.deq.valid := AsyncResetReg(valid, 0)
+  // It is possible that bits latches when the source domain is reset / has power cut
+  // This is safe, because isolation gates brought mem low before the zeroed widx reached us
+  io.deq.bits  := RegEnable(io.mem(index), valid)
 
-  io.ridx := AsyncResetReg(ridx, 0)
+  val valid_reg = AsyncResetReg(valid.asUInt)(0)
+  io.deq.valid := valid_reg && reset_n
```
mwachs5: Is the reset in the logic now? Really dislike. Or is this some other thing means something other than reset. Is this like "fifo_reset" which is known to be some combinational function?

src/main/scala/util/AsyncQueue.scala
```diff
@@ -74,11 +92,18 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int) extends Module {
...
+
+  val ridx_reg = AsyncResetReg(ridx)
+  io.ridx := ridx_reg
+
+  // It is a fatal error to reset half a Queue while it still has data
+  assert (reset_n || widx === ridx)
```
mwachs5: Good thing to check. Wondering if it's always going to be true in such a low-level primitive.

src/main/scala/regmapper/RegisterCrossing.scala
```diff
@@ -7,129 +7,167 @@ import chisel3.util.{Irrevocable}
...
-//  2. wait for inflight traffic to resolve
-//  3. assert reset in the domain
-//  4. turn off the domain
-//  5. turn on the domain
-//  6. deassert reset in the domain
-//  7. raise allow on the other side
+// To turn off=>on a domain:
+//  A. To turn disable the master domain
+//    1. wait for all inflight traffic to resolve
+//    2. assert master reset
```
terpstra: You don't have to. The instructions were for how to disable a domain, which involves using bypass. You can set bypass and not reset if you choose.

src/main/scala/uncore/tilelink2/Crossing.scala
```diff
@@ -15,16 +15,24 @@ class TLAsyncCrossingSource(sync: Int = 3) extends LazyModule
...
       val depth = edgeOut.manager.depth
 
       out.a <> ToAsyncBundle(in.a, depth, sync)
       in.d <> FromAsyncBundle(out.d, sync)
 
+      assert (io.bypass || reset_n, "Unsafe to reset sink manager without first asserting source bypass")
+
       if (bce) {
+        val up = RegInit(Bool(false))
+        when (reset_n) { up := Bool(true) }
```
terpstra: You need to pass reset_n, because isolation brings the wires to zero. We want isolation to appear the same as reset to the other side.

src/main/scala/uncore/tilelink2/Crossing.scala
```diff
@@ -51,12 +59,19 @@ class TLAsyncCrossingSink(depth: Int = 8, sync: Int = 3) extends LazyModule
...
       val bce = edgeOut.manager.anySupportAcquire && edgeOut.client.anySupportProbe
 
       out.a <> FromAsyncBundle(in.a, sync)
       in.d <> ToAsyncBundle(out.d, depth, sync)
 
+      assert (!out.d.valid || reset_n, "D channel respose sent to missing client")
+
       if (bce) {
+        val up = RegInit(Bool(false))
+        when (reset_n) { up := Bool(true) }
```
terpstra: I suppose one could rename it to sink_reset_n and source_reset_n. I'll make a change to that effect now.

src/main/scala/uncore/tilelink2/Isolation.scala
```diff
@@ -55,9 +56,10 @@ class TLIsolation(fOut: (Bool, UInt) => UInt, fIn: (Bool, UInt) => UInt) extends
 object TLIsolation
 {
   // applied to the TL source node; y.node := TLIsolation(fOut, fIn)(x.node)
-  // f should insert an isolation gate between the input UInt and its result
-  // fOut is applied for data flowing from client to manager
-  // fIn  is applied for data flowing from manager to client
+  // f* should insert an isolation gate between the input UInt and its result
+  // fOut is applied to data flowing from client to manager
+  // fIn  is applied to data flowing from manager to client
+  // >>>> WARNING: the isolation functions must bring the values to 0 <<<<
```
terpstra: The isolation needs to use '0' so that you know the counters are 0 and that the reset_n is 0.

src/main/scala/util/AsyncQueue.scala
```diff
@@ -29,27 +29,38 @@ class AsyncQueueSource[T <: Data](gen: T, depth: Int, sync: Int) extends Module
...
   val io = new Bundle {
     // These come from the source domain
-    val enq  = Decoupled(gen).flip()
+    val enq  = Decoupled(gen).flip
     // These cross to the sink clock domain
     val ridx = UInt(INPUT,  width = bits+1)
     val widx = UInt(OUTPUT, width = bits+1)
     val mem  = Vec(depth, gen).asOutput
+    // Reset for the other side
+    val sink_reset_n = Bool().flip
```
terpstra: I am receiving the reset form the far side of the clock crossing. So it must be an input...

src/main/scala/util/AsyncQueue.scala
terpstra: No, I don't. Feel free to draw it, but I am perfectly comfortable that this is safe.
The 'catcher' extends the pulse to the end of the next cycle. The sequence of flops after it eliminates the metastability.

src/main/scala/util/AsyncQueue.scala
mwachs5: I meant the same "reset_n" question. The isolation answer makes sense.

src/main/scala/util/AsyncQueue.scala
mwachs5: Yeah, I think the name "reset_n" is what is confusing here. IN my mind, reset_n should be exactly the same as "~reset". If that's not the case, give it a different name.

terpstra:
I've included a signal renaming commit and rebased it against master.
@mwachs5 Please add your 'suggestName' hack to the tip of this rebased branch.

src/main/scala/util/AsyncQueue.scala
```diff
@@ -29,27 +29,38 @@ class AsyncQueueSource[T <: Data](gen: T, depth: Int, sync: Int) extends Module
...
   }
 
+  // extend the sink reset to a full cycle (assertion latency <= 1 cycle)
+  val catch_reset_n = AsyncResetReg(Bool(true), clock, !io.sink_reset_n)
+  // reset_n has a 1 cycle shorter path to ready than ridx does
+  val sink_reset_n = AsyncGrayCounter(catch_reset_n.asUInt, sync)(0)
```
mwachs5: Isn't "AsyncGrayCounter" just a ShiftRegister? 

Even if it was more than that, why are you using "AsyncGrayCounter" for the purpose of synchronizing the reset?

mwachs5: I suppose the fact that it's asynchronously reset is a major difference between the usual ShiftRegister. If we had a primitive that was "AsynchronouslyResetShiftRegister", would both of these things use it (not saying we should create such a monstrosity, just trying to understand what you're doing here).

terpstra:
No more computer for 4 hours... Yes it just makes a sequence of UInt flops.
Please rename the object to UIntSyncChain. The reset is just being synced.

terpstra:
I've rebased this against current master, resolving all conflicts.

terpstra:
Rebased and tidied up the commit order a bit.

src/main/scala/util/AsyncQueue.scala
```diff
@@ -96,10 +98,10 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int) extends Module {
   // This is safe, because isolation gates brought mem low before the zeroed widx reached us
   io.deq.bits  := RegEnable(io.mem(index), valid)
 
-  val valid_reg = AsyncResetReg(valid.asUInt)(0)
+  val valid_reg = AsyncResetReg(valid.asUInt, "valid_reg")(0)
```
mwachs5: You could just use the Bool form of this like you did above for catch_reset_n...

val valid_reg = AsyncResetReg(valid, clock, reset, "valid_reg")

mwachs5: Or we should just add the missing form so you don't have to specify clock and reset, like the UInt form below.

terpstra: If you want to add a Bool form to your AsyncResetReg stuff, I would use it.
However, I don't find the cast to be particularly evil.

src/main/scala/util/AsyncQueue.scala
terpstra: Every implementation of an AsyncQueue assumes that independent 1-bit signals sent from one side will arrive no more than one cycle apart in the target side. It's the whole reason why Gray codes are a safe way to transport a counter... So, yes, I am assuming that reset_n will not arrive more than 1 cycle later than widx. Then, I ensure the path to valid/ready is one cycle shorter, guaranteeing that there can be no spurious valid/ready.

terpstra:
Rebased to current master and moved the Isolation change between its dependency and dependent commit.

src/main/scala/regmapper/RegisterCrossing.scala
```diff
@@ -7,129 +7,167 @@ import chisel3.util.{Irrevocable}
...
+//    3. (optional) stop the master clock
+//    --- YOU MAY NOT TURN OFF POWER ---
+//    4. re-enable the clock
+//    5. deassert reset
+//  B. To turn off the slave domain
+//    1. assert bypass
+//    2. wait for inflight traffic to resolve
+//    3. assert slave reset
+//    4. (optional) stop the slave clock
+//    --- YOU MAT NOT TURN OFF POWER ---
```
mwachs5: comment typo

src/main/scala/regmapper/RegisterCrossing.scala
```diff
@@ -7,129 +7,167 @@ import chisel3.util.{Irrevocable}
...
+// If you need to cut power, use something that support isolation gates.
 
 class RegisterWriteCrossingIO[T <: Data](gen: T) extends Bundle {
   // Master clock domain
   val master_clock   = Clock(INPUT)
   val master_reset   = Bool(INPUT)
-  val master_allow   = Bool(INPUT) // actually wait for the slave
   val master_port    = new RegisterWriteIO(gen)
+  // Bypass requests from the master to be noops
+  val master_bypass  = Bool(INPUT)
```
mwachs5: One thing that might be good to comment...

If I do a write while the master_bypass is asserted, does it have any effect when master_bypass is de-asserted? 

In other words, on the slave side there is a register with value 0. I assert master_bypass, then do a write with value 0xFF. If I do a read with master_bypass, I just get 0 (right?) And if I de-assert master_bypass, the register still is 0, the value 0xFF is completely forgotten.

I'm pretty sure this is what you coded & intended, but might not be immediately obvious to someone deciding to use this.

terpstra: You are correct.

PR #389, opened 2016-10-07
Corrected csr address of misa register

trosenkranz:
Done to comply with
'The RISC-V Instruction Set Manual - Volume II: Privileged Architecture'
'Privileged Architecture Version 1.9'
'July 8, 2016'

### DIFF
--- a/src/main/scala/rocket/instructions.scala
+++ b/src/main/scala/rocket/instructions.scala
@@ -289,7 +289,6 @@ object CSRs {
   val sip = 0x144
   val sptbr = 0x180
   val mstatus = 0x300
-  val misa = 0x301
   val medeleg = 0x302
   val mideleg = 0x303
   val mie = 0x304
@@ -368,6 +367,7 @@ object CSRs {
   val mhpmevent29 = 0x33d
   val mhpmevent30 = 0x33e
   val mhpmevent31 = 0x33f
+  val misa = 0xf10
   val mvendorid = 0xf11
   val marchid = 0xf12
   val mimpid = 0xf13


trosenkranz:
Before the patch the register was located at address 0x301 which is in a "Standard read/write" address space of the csrs.

According to the specification the misa csr reports the "ISA supported by the hart" which is implementation dependent and thus cannot be change during runtime.

aswaterman:
Sorry for the confusion here.  In what will become version 1.9.1 of the privilege architecture, misa is redefined to be writable register, to optionally support disabling ISA extensions at runtime.  The version of riscv-tools that rocket-chip points to is consistent with this feature.

trosenkranz:
Thanks for pointing that out.

Is there a location, where one can check the latest working draft of the specification? Or a mailing list, where changes are discussed?

aswaterman:
@trosenkranz we've been meaning to put our proposals on arxiv on a regular basis, but have been busy with another project.  I will try to set that up next week.

In the mean time, the only other important change is the handling of the performance counters at higher privileged levels.

sorear:
@aswaterman given that arXiv takes at least a day to publish anything, I think it'd still frequently be out of sync with the code?

aswaterman:
@sorear That's true, but it doesn't seem to me to be too onerous, provided we're careful to note in commit messages when changes are API-breaking.  There have only been two of these since the last official publication in July.  We're open to suggestions, though.

sorear:
https://github.com/arunthomas/riscv-rfcs was a neat idea, but I'm not sure if it can be made to work well at this time.  I still think it'd be a good fit for hashing out the details of "B", "P", "L"

sorear:
@aswaterman I think a public bug tracker for the spec would be a good idea even if it were just for editorial issues (but ideally proposals would be tracked as well); I'd like something less ephemeral than isa-dev to use for tracking inconsistencies, vagueness, and known inaccuracies.

PR #390, opened 2016-10-08
async_queue: Give names to all the registers which show up in the queue

mwachs5:
This is to aid debugging but even more so for backend constraint writers, who generally
need predictable names for registers to set false paths, etc.

I also picked up @terpstra 's change which renamed 'reg' to 'control' to avoid use of Verilog keyword. 

I see @terpstra already did some stuff in #387 to help with predictable names, I can work on merging this with his later today.


colinschmidt:
Awesome. I made this change downstream to deal with MIM stuff. Thanks for putting it upstream. :)

mwachs5:
I will work on the conflicts with #387.

PR #391, opened 2016-10-09
Simplify AsyncResetReg

aswaterman:
No need for AsyncSetReg, as AsyncResetReg can be used exclusively with
inverted inputs.


yunsup:
loos good to me

PR #395, opened 2016-10-13
AXI4 fragmenter

terpstra:
This AXI4 fragmenter slices all AXI accesses into simple power-of-two sized and aligned transfers of the largest size supported by the slave.  This makes it suitable as a first stage transformation to apply before an AXI4=>TL bridge. It also makes it suitable for placing after TL=>AXI4 bridge driving an AXI-lite slave.

### DIFF
--- a/src/main/scala/uncore/axi4/Fragmenter.scala
+++ b/src/main/scala/uncore/axi4/Fragmenter.scala
@@ -53,10 +53,10 @@ class AXI4Fragmenter(lite: Boolean = false, maxInFlight: Int = 32, combinational
      *   FIXED => each beat is a new request
      *   WRAP/INCR => take xfr up to next power of two, capped by max size of target
      *
-     * On AR and AW, we fragment the requests
-     * On W we insert 'last' to match
-     * On R we surpress 'last'
-     * On B we surpress 'valid'
+     * On AR and AW, we fragment one request into many
+     * On W we set 'last' on beats which are fragment boundaries
+     * On R we clear 'last' on the fragments being reassembled
+     * On B we clear 'valid' on the responses for the injected fragments
      *
      * AR=>R and AW+W=>B are completely independent state machines.
      */
@@ -208,6 +208,8 @@ class AXI4Fragmenter(lite: Boolean = false, maxInFlight: Int = 32, combinational
     in_w.ready := out_w.ready && (!wbeats_ready || wbeats_valid)
     out_w.bits := in_w.bits
     out_w.bits.last := w_last
+    // We should also recreate the last last
+    assert (!out_w.valid || !in_w.bits.last || w_last)
 
     // R flow control
     val r_last = out_r.bits.last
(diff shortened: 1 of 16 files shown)


terpstra:
Rebased the PR against current master.

terpstra:
@henrystyles We should try this out on the VC707.

terpstra:
@hcook Any chance of a quick code review?

src/main/scala/uncore/axi4/Fragmenter.scala
```diff
@@ -0,0 +1,268 @@
...
+      val mux_addr = Wire(init = align_addr)
+      when (a.bits.burst === AXI4Parameters.BURST_WRAP) {
+        mux_addr := (align_addr & wrapMask) | ~(~a.bits.addr | wrapMask)
+      }
+      when (a.bits.burst === AXI4Parameters.BURST_FIXED) {
+        mux_addr := a.bits.addr
+      }
+
+      val last = beats1 === len
+      a.ready := out.ready && last
```
hcook: Correction - AXI is and stay irrevocable, so this is fine.

src/main/scala/uncore/axi4/Fragmenter.scala
```diff
@@ -0,0 +1,268 @@
...
+
+    // B flow control
+    val b_last = sideband_aw_b.deq.bits
+    in_b.valid := out_b.valid && sideband_aw_b.deq.valid && b_last
+    out_b.ready := sideband_aw_b.deq.valid && (!b_last || in_b.ready)
+    sideband_aw_b.deq.ready := out_b.valid && (!b_last || in_b.ready)
+    in_b.bits := out_b.bits
+
+    // Merge errors from dropped B responses
+    val r_resp = Reg(UInt(width = AXI4Parameters.respBits))
```
hcook: initialize to 0

src/main/scala/uncore/axi4/Test.scala
```diff
@@ -0,0 +1,61 @@
...
+
+import Chisel._
+import diplomacy._
+import uncore.tilelink2._
+import unittest._
+
+class RRTest0(address: BigInt) extends AXI4RegisterRouter(address, 0, 32, 0, 4)(
+  new AXI4RegBundle((), _)    with RRTest0Bundle)(
+  new AXI4RegModule((), _, _) with RRTest0Module)
+
```
hcook: nice reuse

src/main/scala/uncore/axi4/Fragmenter.scala
```diff
@@ -0,0 +1,268 @@
...
+      val wipeHigh = ~leftOR(~len)       // clear all bits in position  >= a cleared bit
+      val remain1  = fillLow | wipeHigh  // MSB(a.len+1)-1
+      val align1   = ~leftOR(alignment)  // transfer size limited by address alignment
+      val support1 = if (allSame) fixed1 else dynamic1 // maximum supported size-1 based on target address
+      val maxSupported1 = remain1 & align1 & support1 // Take the minimum of all the limits
+
+      // Things that cause us to degenerate to a single beat
+      val fixed = a.bits.burst === AXI4Parameters.BURST_FIXED
+      val narrow = a.bits.size =/= UInt(lgBytes)
+      val misaligned = lo =/= UInt(0)
```
terpstra: Not only can it be handled this way, but on reflection, it must be handled this way.
Otherwise, the fragmenter can pass through a misaligned access, which will cause a TL2 Monitor assertion failure.

terpstra:
@hcook I've addressed your concerns. However, I had to modify the sideband state machine quite a bit. Please take a peak if you could. I'm running some stress tests against it.

PR #397, opened 2016-10-14
tilelink2: replace addr_hi with address

terpstra:
When faced with ambiguous routing of wmask=0, we decided to include
all the address bits. Hopefully in most cases the low bits will be
optimized away anyway.


terpstra:
Ugh. Getting the full address correct out of Legacy was a lot harder than expected.

mwachs5:
Anyone looking at waveforms thanks you.

hcook:
Going to go ahead and approve this despite the potential impact on A and B channel address field width. It simplifies the code base and the specification, makes debugging easier, and resolves the problem of ambiguous PutPartials with empty write masks. Hopefully savings can be clawed back through detecting and pruning un-driven address bits.

PR #399, opened 2016-10-14
Diplomatic Names

terpstra:
Create interesting names for lazy modules.

Prefer to use the name from a val: 'val foo = LazyModule(new ...)'
If that can't be determined, use the outermost common ancestor's val name.
If that can't be determined, fall back on the class name.


terpstra:
Merging this will require changing the names in any constraint files used in other repositories...

mwachs5:
Ok, I am playing around with this a bit more, and one thing that is a little weird is that the instance name is appended to the module name, so that every module gets explicitly uniquified.

e.g. if I do something like:

`val foo = LazyModule(new FooModule())
`
Then the resulting verilog is:

`FooModule_foo foo_FooModule
`

Is it really necessary to uniquify the modules by name? Is a TLFragmenter_foo always internally different than a TLFragmenter_bar?

mwachs5:
To answer my own question... TLFragmenter and the like were getting uniquified anyway, so it is kind of nice to have the name in front of it instead of some random number.

mwachs5:
I made a slight modification so that the leaf nodes, which the user has explictly given a label, don't get the module name appended. This means 'clint' is just called 'clint', but keeps the nice labeling on clint_TLFragmenter, etc.

PR #400, opened 2016-10-15
Better crossing reset handling

terpstra:
The source must wait for the sink's clock after the source was reset, and vice versa.
Also, be more precise about what exactly is an critical crossing failure.


terpstra:
@jchang0 This is the PR that should hopefully meet the Formal assumptions.

PR #402, opened 2016-10-17
AXI4=>TL2 converter

terpstra:
This bridge when combined with an AXI4Fragmenter should support all of AXI4.
It will report DECERR if the address is invalid and SLVERR if the TL2 slave reports error.
It enforces FIFO by only allowing one of each AXI ID inflight at a time, which should be enough throughput when combined with an L2 cache since it transfers a cache line per round trip.


terpstra:
@aswaterman This does not yet issue a Hint while waiting.
@hcook We need to define what a prefetch Hint looks like!

PR #407, opened 2016-10-20
add verilator vcd command line option

vegaluisjose:
I have created another pull-request for firrtl and chisel3 that uses this option


colinschmidt:
I don't think this is necessary. You can already specify the name of the vcdfile on the command line see: https://github.com/ucb-bar/rocket-chip/blob/master/csrc/emulator.cc#L49

vegaluisjose:
Hi Colin,

I see... I did it because I found the following in chisel3 and also firrtl

https://github.com/ucb-bar/chisel3/blob/master/src/main/resources/top.cpp#L22-L35

https://github.com/ucb-bar/firrtl/blob/master/src/test/resources/top.cpp#L22-L35

where the filename is constant "dump.vcd"

PR #409, opened 2016-10-20
Bump firrtl and update vsim Makefrag-verilog

jackkoenig:
We still should move to invoking Firrtl programmatically rather than via the command-line, but this commit updates Firrtl for now.


colinschmidt:
This is what I've been using downstream, and it works fine for us.

PR #410, opened 2016-10-20
Support new execution harness and options.

ucbjrl:
Manually copy firrtl jar into unmanaged chisel3 lib so it replaces anything we'd download via sbt.
Update option spelling.

** DON'T MERGE THIS **.

This pr is largely to demonstrate the changes required to support the new chisel/firrtl execute harness, and provide us with something that should track master and allow us to continue to test rocket-chip with newer versions of chisel3 and firrtl.


aswaterman:
I'm taking this approach (but will merge in separately from this PR).

PR #411, opened 2016-10-22
axi4: ensure we accept AR before reporting R

terpstra:
This fixes a specification violation detected by VIP.


terpstra:
@scottj97 Here is the PR for that bug you found. Thanks again!

PR #412, opened 2016-10-24
Print out seed if we can

colinschmidt:
Now that we have ifdef VCS in here lets use it for something more than compatibility

This is a little diff I keep around for debugging non-determinism, thought others might find it useful. The reason I can't just use the stdout where VCS prints the seed is sometimes the non-determinism is rare and so I want to run a bunch of tests in parallel and correlating the stdout to the output file can be difficult.

### DIFF
--- a/vsrc/TestDriver.v
+++ b/vsrc/TestDriver.v
@@ -34,7 +34,11 @@ module TestDriver;
     rand_value = $urandom;
     rand_value = $random(rand_value);
     if (verbose) begin
+`ifdef VCS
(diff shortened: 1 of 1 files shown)


aswaterman:
I think we should change emulator.cc and TestDriver.v to print the same thing (perhaps "using random seed X, first random value is Y")

colinschmidt:
I tried to reply via email but it seems like that didn't work so I'll paste it here:

That sounds reasonable but I'm not sure when I'll get around to that.

aswaterman:
ok, i'll merge anyway

colinschmidt:
That sounds reasonable but I'm not sure when I'll get around to that.

PR #413, opened 2016-10-24
Add PutAtomic support to width adapters

zhemao:
The TileLink widener and narrower did not support PutAtomic operations, so they were not passing through correctly. This commit fixes that issue and adds a unittest.

### DIFF
--- a/src/main/scala/uncore/converters/Tilelink.scala
+++ b/src/main/scala/uncore/converters/Tilelink.scala
@@ -6,6 +6,8 @@ import junctions.PAddrBits
 import uncore.tilelink._
 import uncore.util._
 import uncore.constants._
+import uncore.devices.TileLinkTestRAM
+import unittest.UnitTest
 import cde.Parameters
 
 /** Utilities for safely wrapping a *UncachedTileLink by pinning probe.ready and release.valid low */
@@ -222,6 +224,8 @@ class TileLinkIOWidener(innerTLId: String, outerTLId: String)
   val stretch = ognt.g_type === Grant.getDataBlockType
   val smallget = iacq.a_type === Acquire.getType
   val smallput = iacq.a_type === Acquire.putType
+  val atomic = iacq.a_type === Acquire.putAtomicType
+  val wideget = iacq.a_type === Acquire.getBlockType
   val smallgnt = ognt.g_type === Grant.getDataBeatType
 
   val sending_put = Reg(init = Bool(false))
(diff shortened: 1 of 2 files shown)


src/main/scala/uncore/converters/Tilelink.scala
```diff
@@ -278,11 +282,22 @@ class TileLinkIOWidener(innerTLId: String, outerTLId: String)
...
+    addr_block = out_addr_block,
+    addr_beat = out_addr_beat,
+    addr_byte = out_addr_byte,
+    atomic_opcode = iacq.op_code(),
+    operand_size = iacq.op_size(),
+    data = align_data(switch_addr, iacq.data))(outerConfig)
+
   io.out.acquire.valid := sending_put || (!shrink && io.in.acquire.valid)
-  io.out.acquire.bits := MuxCase(get_block_acquire, Seq(
+  io.out.acquire.bits := MuxCase(Wire(io.out.acquire.bits, init=iacq), Seq(
```
colinschmidt: Is there some way to have the default be an assertion or something? It seems like that would have made this problem easier to find.

src/main/scala/uncore/converters/Tilelink.scala
```diff
@@ -466,15 +482,33 @@ class TileLinkIONarrower(innerTLId: String, outerTLId: String)
...
+    addr_byte = atomic_addr_byte,
+    atomic_opcode = iacq.op_code(),
+    operand_size = iacq.op_size(),
+    data = atomic_data_vec(atomic_data_sel))
+
   val sending_put = Reg(init = Bool(false))
 
   val pass_valid = io.in.acquire.valid && !stretch
 
   io.out.acquire.bits := MuxCase(Wire(io.out.acquire.bits, init=iacq), Seq(
```
colinschmidt: Same as above.

src/main/scala/uncore/converters/Tilelink.scala
```diff
@@ -466,15 +482,33 @@ class TileLinkIONarrower(innerTLId: String, outerTLId: String)
...
-    (sending_put, put_block_acquire),
-    (shrink, get_block_acquire),
-    (smallput, put_acquire),
-    (smallget, get_acquire)))
+    sending_put -> put_block_acquire,
+    shrink -> get_block_acquire,
```
colinschmidt: In the previous mux_case shrink was named wideget, consistency seems nice here. I think wideget seems like a better name but don't have a strong preference.

src/main/scala/uncore/converters/Tilelink.scala
zhemao: Yeah, I would like that too, but I don't think that's possible with MuxCase. I can add the assertion manually by OR'ing across all the conditions.

src/main/scala/uncore/converters/Tilelink.scala
```diff
@@ -398,6 +413,7 @@ class TileLinkIONarrower(innerTLId: String, outerTLId: String)
   val shrink = iacq.a_type === Acquire.getBlockType
   val smallput = iacq.a_type === Acquire.putType
   val smallget = iacq.a_type === Acquire.getType
+  val atomic = iacq.a_type === Acquire.putAtomicType
```
hcook: do you need assertions on the outer size similar to the ones on smallput and smallget below (line 446)?

zhemao: Good catch. Yeah, I'll add that.

hcook:
Thanks for adding a test. I'm worried about what happens when you narrow a 64b atomic onto a 32b bus. Seems like that is just disallowed for gets and puts anyways?

zhemao:
Yeah, in general a sub-block operation larger than the physical width is just disallowed. I will add an assertion.

zhemao:
@hcook I've made some changes if you care to take a look. Seems like Wesley is pushing a lot of stuff right now. I'll merge this in after he's done.

terpstra:
I just updated the PR against master; all my commits have been merged. Thanks!

colinschmidt:
LGTM I'll merge this once travis passes.

PR #415, opened 2016-10-24
rocketchip: include an socBus between l1tol2 and periphery

terpstra:
Sometimes we have high performance devices that go inbetween.


zhemao:
Wait, I'm a bit confused. So now instead of legacy -> peripheryBus it's legacy -> socBus -> peripheryBus?

What exactly does this gain and which devices are hanging off of which bus now?

terpstra:
The periphery bus is for simple slaves. Like 32-bit wide, FIFO, etc. So we put Atomics infront of them and a WidthWidget.

The socBus stays nice and wide. Suitable for attaching PCIe or other high-speed slaves.

PR #419, opened 2016-10-25
TL2 Debug Module

mwachs5:
This replaces the Debug Module's System Bus interface with a generic RegMap interface. It creates a TL2 version and hooks it up inside BaseCoreplex using a Legacy converter. This is part of the process of moving TL2 inside the coreplex.


terpstra:
I'm closing this PR, because we need to include several more changes and then merge them all at once. The actual PR will be the "remove all TL1 remnant bus remnants".

PR #421, opened 2016-10-25
Fixed AsyncFifo with reset messaging

jchang0:
Without the fix, could cause spurious dequeue when no item is in the queue when reset on sending side.


src/main/scala/util/AsyncQueue.scala
```diff
@@ -137,7 +143,14 @@ class AsyncQueueSink[T <: Data](gen: T, depth: Int, sync: Int, safe: Boolean = t
     source_valid.io.in := source_extend.io.out
     source_ready := source_valid.io.out
 
-    assert (io.source_reset_n || widx === ridx, "Source reset while AsyncQueueSink not empty")
+    val reset_and_extend = !source_ready || !io.source_reset_n || reset
+    val reset_and_extend_prev = Reg(Bool(), reset_and_extend, Bool(true))
+    val reset_rise = !reset_and_extend_prev && reset_and_extend
+    val prev_idx_match = AsyncResetReg(updateData=(io.widx===io.ridx), resetData=0)
```
terpstra: Why are there these lines here if the assertion is commented out?

jchang0: This is why we need lint check for dead code.

PR #423, opened 2016-10-26
Bump chisel/firrtl

aswaterman:
Chisel now depends on firrtl and will look for a locally published copy.  The workaround is to publish firrtl locally.  This could cause headaches if multiple firrtl versions are in play.  A possible resolution is to have firrtl snapshots versioned by the commit hash, per @jackkoenig.

@richardxia this presumably requires some downstream changes, or perhaps I should take an alternate approach altogether.

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -19,9 +19,12 @@ SHELL := /bin/bash
 FIRRTL_JAR ?= $(base_dir)/firrtl/utils/bin/firrtl.jar
 FIRRTL ?= java -Xmx2G -Xss8M -XX:MaxPermSize=256M -cp $(FIRRTL_JAR) firrtl.Driver
 
+# Build firrtl.jar and put it where chisel3 can find it.
 $(FIRRTL_JAR): $(shell find $(base_dir)/firrtl/src/main/scala -iname "*.scala")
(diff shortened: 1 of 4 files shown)


ucbjrl:
The work-around from #410 (the Makefrag part) in combination with ucb-bar/chisel3#334 doesn't work for you?

richardxia:
If we're going to publish to a local Ivy repository, then we really need to properly version the packages. Otherwise, the SiFive internal CI builds may clobber each other's firrtl packages if more than one is running concurrently with different versions of firrtl. Alternatively, if there's a way to publish to a project-specific directory instead of a user-specific directory, that would work, too.

aswaterman:
@ucbjrl I ignored that PR because it said "do not merge"... that looks like it would work, too.

@richardxia which approach is preferable?

aswaterman:
Sounds like we should go with @ucbjrl's workaround instead.

richardxia:
Yes, I think I prefer @ucbjrl's approach. Although is manually copying `firrtl.jar` necessary? If firrtl is a subproject of chisel, then shouldn't SBT be able to resolve the path somehow?

aswaterman:
@richardxia see https://github.com/sbt/sbt/issues/2777 -- it doesn't seem like it can handle both cases, but I'd certainly like a fix.

PR #424, opened 2016-10-27
Coreplex cake

terpstra:
This rather sweeping PR moves TL2 all the way throughout the Coreplex.
The only remains of TL1 are now the rocket L1 interface and the coherence manager.


terpstra:
I expect it to hang in TinyConfig, and I will get Andrew to help me figure out why it's failing tomorrow. Besides that, I think it's ready to be reviewed. If there is a good PLIC unittest, we need to run it against this as well.

src/main/scala/rocket/dcache.scala
```diff
@@ -495,60 +497,83 @@ class DCache(implicit p: Parameters) extends L1HellaCacheModule()(p) {
...
+
+    // place AMO data in correct word lane
+    val minAMOBytes = 4
+    val grantData = Mux(io.dmem.resp.valid, io.dmem.resp.bits.data, acq.data)
+    val alignedGrantData = Mux(acq.size <= log2Ceil(minAMOBytes), Fill(coreDataBytes/minAMOBytes, grantData(8*minAMOBytes-1, 0)), grantData)
+
+    tl_in.d.valid := io.dmem.resp.valid || state === s_grant
+    tl_in.d.bits := Mux(isRead,
+      edge.AccessAck(acq, UInt(0), alignedGrantData),
+      edge.AccessAck(acq, UInt(0)))
```
aswaterman: alignedGrantData should be selected for AMOs as well as reads.  As is, it looks like isRead will be false in that case, so AMO data won't be returned.

Can we fix the bug by deleting the Mux and always return edge.AccessAck(acq, UInt(0), alignedGrantData)?

aswaterman: Yeah, AFAICT, this is the only thing within the implementation that would prevent it from working.  Need this capability for the case of a multicore where one of the cores has a scratchpad.  (Probably wouldn't ever be used on a uniprocessor, as the only other master in the system is likely to be a DMA engine, which won't issue AMOs.)

terpstra: You would presumably need to decode the the AMO opcode as well. Anyway, I'll take a look at it tomorrow morning.

aswaterman: Hmm, yes.

src/main/scala/rocket/dcache.scala
```diff
@@ -495,60 +497,83 @@ class DCache(implicit p: Parameters) extends L1HellaCacheModule()(p) {
...
-    g_type = acq.getBuiltInGrantType(),
-    client_xact_id = acq.client_xact_id,
-    manager_xact_id = UInt(0),
-    addr_beat = acq.addr_beat,
-    data = alignedGrantData)
+    val ready = state === s_ready || tl_in.d.fire()
+    io.dmem.req.valid := (tl_in.a.valid && ready) || state === s_replay
+    tl_in.a.ready := io.dmem.req.ready && ready
+    io.dmem.req.bits := formCacheReq(Mux(state === s_replay, acq, tl_in.a.bits))
+    // this blows.  the TL data is already in the correct byte lane, but the D$
```
mwachs5: Let's not be a brogrammers, can we keep the comments helpful/PG...

src/main/scala/rocketchip/Periphery.scala
```diff
@@ -204,55 +198,40 @@ trait PeripheryMasterMemModule extends HasPeripheryParameters {
...
-      val idx = i-mmio_ahb_start
-      io.mmio_ahb(idx) <> PeripheryUtils.convertTLtoAHB(mmio_ports(i), atomics = true)
-    } else if (mmio_tl_start <= i && i < mmio_tl_end) {
-      val idx = i-mmio_tl_start
-      io.mmio_tl(idx) <> TileLinkEnqueuer(mmio_ports(i), 2)
-    } else {
-      require(false, "Unconnected external MMIO port")
-    }
-  }
+  // nothing to do
```
mwachs5: This only makes sense in the context off all the obsolete code...

terpstra:
This now passes not only rocket chip regressions but also all regressions in a different actual SoC. Please review and/or approve.

src/main/scala/coreplex/BaseCoreplex.scala
```diff
@@ -35,58 +35,145 @@ trait HasCoreplexParameters {
...
+  val mmio = outer.mmio.bundleOut
+  val interrupts = outer.mmioInt.bundleIn
+}
+
+trait CoreplexNetworkModule extends HasCoreplexParameters {
+    this: BareCoreplexModule[BareCoreplex, BareCoreplexBundle[BareCoreplex]] =>
+  implicit val p = outer.p
+}
+
+trait CoreplexRISCV {
```
aswaterman: Why this name, rather than just Coreplex?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -195,26 +201,54 @@ class IdentityNode[PO, PI, EO, EI, B <: Data](imp: NodeImp[PO, PI, EO, EI, B])
...
 }
 
 class SinkNode[PO, PI, EO, EI, B <: Data](imp: NodeImp[PO, PI, EO, EI, B])(pi: PI, num: Range.Inclusive = 1 to 1)
   extends SimpleNode(imp)({case (0, _) => Seq()}, {case (n, Seq()) => Seq.fill(n)(pi)}, 0 to 0, num)
 {
   require (num.end >= 1, s"${name} is a sink which does not accept inputs${lazyModule.line}")
+  override lazy val bundleOut = { require(false, s"${name} has no bundleOut; try bundleIn?"); bundleIn }
+}
+
+class BlindOutputNode[PO, PI, EO, EI, B <: Data](imp: NodeImp[PO, PI, EO, EI, B])(pi: PI)
```
aswaterman: Can you add one-line comments explaining what these classes do, since it's not immediately obvious from the name?

aswaterman:
Looks good to me.

src/main/scala/diplomacy/Nodes.scala
terpstra: There's a paragraph explaining them at the top of the file already. You just don't see it in the diff.

src/main/scala/coreplex/BaseCoreplex.scala
terpstra: We have BareCoreplex and BaseCoreplex, the classes. Then the traits are CoreplexNetwork and CoreplexRISCV. Calling one just Coreplex makes it ambiguous...

aswaterman: How about CoreplexTiles?

terpstra:
Rebased against current master

hcook:
lgtm. impressive that overall LoC delta is negative :)

colinschmidt:
Looks good to me. The part I care most about is the l1tol2net and rocket/rocc. Are they next on the TL2 chopping block?

terpstra:
Yes, at some point in the near future we will switch the L2 to TL2 and then swallow the input slaves (PCIe in). The very last step is converting rocket.

terpstra:
FYI, this PR has turned up a bug in the AutomicAutomata that appears when devices respond on the same cycle as their request. I'm working to fix it as part of this PR.

terpstra:
FPGA build with PCIe works. Ready to be merged!

PR #427, opened 2016-10-29
Fix issue with PutBlock and Release in BroadcastHub

zhemao:
There is logic in the broadcast hub to skip the outer acquire if there
is an outgoing release, since the data will be written out through the
release channel. However, this will cause an issue in the case of
PutBlock requests. If the tail beats of the PutBlock show up after the
outer release has already been sent, the data will be corrupted.

The fix is to make the outer release block if there are pending
inner PutBlock beats.

I've also added a regression test to the broadcast regression test suite that catches this behavior.

### DIFF
--- a/src/main/scala/groundtest/Regression.scala
+++ b/src/main/scala/groundtest/Regression.scala
@@ -681,6 +681,54 @@ class MergedPutRegression(implicit p: Parameters) extends Regression()(p)
   io.errored := data_mismatch
 }
 
+class PutAfterReleaseRegression(implicit p: Parameters) extends Regression()(p) {
+  val (s_idle :: s_cache_req :: s_cache_resp ::
+       s_write_first_req :: s_delay :: s_write_remaining_req :: s_write_resp ::
+       s_read_req :: s_read_resp :: s_finished :: Nil) = Enum(Bits(), 10)
+  val state = Reg(init = s_idle)
+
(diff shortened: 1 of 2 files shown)


src/main/scala/groundtest/Regression.scala
```diff
@@ -625,7 +625,7 @@ class MergedPutRegression(implicit p: Parameters) extends Regression()(p)
 
   io.cache.req.valid := (state === s_cache_req)
   io.cache.req.bits.cmd := M_XWR
-  io.cache.req.bits.typ := UInt(log2Ceil(64 / 8))
+  io.cache.req.bits.typ := MT_D
```
aswaterman: This is the only use of MT_ outside of Rocket (MT_ is an ISA constant, not a TileLink constant).  Perhaps it should stay that way.

aswaterman:
This seems reasonable, but @hcook should approve

src/main/scala/groundtest/Regression.scala
zhemao: But this is the HellaCache interface and not the TileLink interface. The L1 cache does consider the "typ" fields to be one of the MT_ constants. I'm fine with it either way, since the behavior doesn't really change. But I think using the MT_ constants is more consistent.

aswaterman:
Right, it makes sense when interfacing with the L1 (at least for now).

hcook:
lgtm

PR #428, opened 2016-10-29
Move the BOOM configs to the boom submodule, default to BOOMConfig

palmer-dabbelt:
With this commit you'll no longer be able to run upstream Rocket Chip
configs by just running

  make CONFIG=SomeRocketChipConfig

instead you'll need to run

  make CONFIG=SomeRocketChipConfig CFG_PROJECT=rocketchip

the advantage is that the boom branch will no longer require any diff
against rocket-chip aside from some default Makefile variables and a
submodule pointer.

Be sure to look at

  https://github.com/ucb-bar/riscv-boom/pull/15

before merging this.


ccelio:
Thanks, this looks awesome. Glad this ability was added to rocket-chip. 

Although FYI: I slightly diff on a few things, like I've added some HPM events to the rocket core, and I may be adding some "chicken bits" which modify csr.scala.  If that's a problem let me know, but those things are very low-churn so the merge updates have been trivial.

ccelio:
It looks like the regression failed on not-BOOM tests due to CFG_PROJECT not being set for them. It looks like that's just a matter of changing regression/Makefile.  I don't mind merging this and fixing that later since BOOM seems to be working.

PR #439, opened 2016-11-18
Add various granular configs.

richardxia:
Here I'm upstreaming some of the configs that we created internally at SiFive but that only configure parameters related to the coreplex/rocket chip. Let me know if these are the right kinds of configs to make available in rocketchip or if there are changes we should make before merging.


aswaterman:
Looks fine to me.

Tangentially, I suspect that once all of the legacy TileLink parameterization is removed from the parameters system, we'll end up refactoring some of it.  That job will probably be easier if these generic configuration objects are in rocket-chip, so they can be refactored atomically.

richardxia:
Thanks for taking a look!

PR #440, opened 2016-11-19
TL2 take-over of the L1

terpstra:
This PR injects TL2 into the tile and L1 data caches. A TL2 Broadcast hub is use to provide multi-core coherence between data caches (passes TraceGen WMO). A streamlined config library is adopted in place of cde. Icaches and RoCCs are still using TL1 behind a tilelink2.Legacy adapter.

Known API-breaking changes:
- No more BuildTiles or common Tile base class/IO. Make a custom Tile for your core, then make a Coreplex mix-in that adds some number of instances of the new Tile.
- package cde is being retired in favor of package config. Their Parameters and Config classes are API compatible but you must change the import (config offers improved functionality for Parameters and Config, but removes Knobs and Constraints). 

Known functionality regressions:
- No AXI pbus mastering (#437) -- will fix next week.
- No more cde.Knob, cde.Constraint, or cde.World -- will not fix.
- No more L2 Cache -- will fix in next two weeks.

PR #441, opened 2016-11-19
[Rocket] Send correct type for iomshr reqs

colinschmidt:
If you configure rocket-chip to use a larger tilelink data size, say 128 bits rather than 64, you will end up using the bottom three bits of operand_size rather than just 2 as rocket expects. This eventually leads to a TL_Monitor assertion firing because the request address is not aligned with this large size.
StoreGen already correctly handles this so we should use it.

I know this might be a throwaway fix given the rapid TL2 transfer but that PR doesn't touch these lines yet.

### DIFF
--- a/src/main/scala/rocket/nbdcache.scala
+++ b/src/main/scala/rocket/nbdcache.scala
@@ -202,7 +202,7 @@ class IOMSHR(id: Int)(implicit p: Parameters) extends L1HellaCacheModule()(p) {
     addr_block = addr_block,
     addr_beat = addr_beat,
     addr_byte = addr_byte,
-    operand_size = req.typ,
+    operand_size = storegen.size,
     alloc = Bool(false))
 
   val put_acquire = Put(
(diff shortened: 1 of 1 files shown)


hcook:
Good catch! I'll cherry-pick this into #440.

colinschmidt:
Thanks

ccelio:
@colinschmidt what do you mean by "larger tilelink data size?" Are you referencing the "innerDataBits = site(XLen)"?

colinschmidt:
In configurations of rocket-chip that include hwacha we widen the tilelink network to have 128bits of data rather than 64 which became the default about a year ago.

We do this by setting dataBeats ala

```
case TLKey("L1toL2") => site(TLKey("DefaultL1toL2")).copy(dataBeats = 4)
```

ccelio:
Thanks, I'm wondering if this was the issue found in https://github.com/ucb-bar/riscv-boom/issues/13.

PR #442, opened 2016-11-21
Per ABI, only x1 and x5 should be treated as function returns

aswaterman:
We were also doing so for x3 and x7, as well, which could reduce performance for compilers that happen to perform indirect jumps via t2 (x7).


aswaterman:
@yunsup poke

Issue #331, opened 2016-09-22
Update documentation in README.md


zhemao:
After the cake pattern refactor, the documentation on "The Top Level Module" is completely outdated. After things settle down, we should take the time to fix it so that it accurately reflects the state of the codebase. Probably instead of talking about what we've done specifically, it would be more helpful to instruct users on how to piece together their own top-level using traits.

mwachs5:
I started working on this, but it's a quite a diff. My basic plan of attack:

* Rewrite "The Top Level Module" explanation assuming that you just did "make verilog CONFIG=DefaultConfig", and just describing what the verilog I/Os mean.

* There are now two real ways to "how can I parameterize my rocket-chip": (1) The old way using CONFIG, and (2) the new way, mixing traits to make your own BaseTop (or ExampleTop) subclass.

*  Basic explanation of the 3-piece cake pattern repeated all over the place now (lazy module definition, Module I/O definitition, module elaboration), explain some part of ExampleTop that shows how that 3-piece pattern works.

zhemao:
If it helps, I wrote an example of using the 3-piece cake pattern to create a custom top. It's in the README of the [project template](https://github.com/ucb-bar/project-template) repository.

mwachs5:
That is super helpful @zhemao, we should definitely point to that.

MarcoGeuna:
Totally agree with @mwachs5: your example is really helpful @zhemao, thank you!!

PR #444, opened 2016-11-22
rocketchip: bump all submodules (and remove cde)

terpstra:
I would like to try using Chisel PR 366 to make memory channels flexible again. First, we need to move the submodules to the respective heads of the submodules.


terpstra:
FYI, I could not fully bump hardfloat. It doesn't compile currently AFAICT.

aswaterman:
Yeah, that's a known issue.  Some APIs have changed, and it needs to be massaged for chisel3 purposes.

hcook:
Closes #448 #331

Issue #437, opened 2016-11-16
Cannot mix in traits that set pBusMasters > 0


davidbiancolin:
In fpga-zynq we use a serial adapter/tether to write into the L2 of rocketchip. 

When i mix the serial adapter traits into my cake i get an exception (below). The same error can also be exercised by setting NExtBusAXIChannels > 0 in the ExampleTop, which should be handled appropriately by the PeripherySlave cake mix in. 

```
[error] (run-main-0) java.lang.reflect.InvocationTargetException
java.lang.reflect.InvocationTargetException
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
        at util.HasGeneratorUtilities$$anonfun$2.apply(GeneratorUtils.scala:47)
        at util.HasGeneratorUtilities$$anonfun$2.apply(GeneratorUtils.scala:44)
        at chisel3.core.Module$.do_apply(Module.scala:29)
        at chisel3.Driver$$anonfun$elaborate$1.apply(Driver.scala:190)
        at chisel3.Driver$$anonfun$elaborate$1.apply(Driver.scala:190)
        at chisel3.internal.Builder$$anonfun$build$1.apply(Builder.scala:194)
        at chisel3.internal.Builder$$anonfun$build$1.apply(Builder.scala:192)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
        at chisel3.internal.Builder$.build(Builder.scala:192)
        at chisel3.Driver$.elaborate(Driver.scala:190)
Caused by: cde.ParameterUndefinedException: Parameter NCoreplexExtClients undefined.
        at cde.World$TopLookup$1.apply(Parameters.scala:156)
        at cde.World$_View.sym(Parameters.scala:114)
        at cde.Parameters$$anonfun$makeMask$3.apply(Parameters.scala:300)
        at cde.Parameters$$anonfun$makeMask$3.apply(Parameters.scala:295)
        at cde.Parameters$KidLookup$1.apply(Parameters.scala:337)
        at cde.World$_View.apply(Parameters.scala:111)
        at cde.View.apply(Parameters.scala:38)
[...]

PR #447, opened 2016-11-23
Axi4 master

terpstra:
This PR re-adds AXI4 bus mastering to rocket.
Thus resolving #437.


src/main/scala/coreplex/RocketCore.scala
```diff
@@ -8,8 +8,8 @@ import uncore.coherence._
 import rocket._
 import uncore.devices.NTiles
 
-trait RocketPlex extends CoreplexRISCVPlatform {
-  val module: RocketPlexModule
+trait RocketCore extends CoreplexRISCVPlatform {
```
aswaterman: I don't think core is the right word here.  Tiles contain cores, not the other way around.

terpstra: Yeah. RocketTile was taken, or I'd have used that.

PR #451, opened 2016-11-24
More configs

terpstra:
A collection of small improvements needed to bootstrap our larger SoC project.

### DIFF
--- a/src/main/scala/rocketchip/Configs.scala
+++ b/src/main/scala/rocketchip/Configs.scala
@@ -28,8 +28,9 @@ class BasePlatformConfig extends Config(
     case PeripheryBusArithmetic => true
     // Note that PLIC asserts that this is > 0.
     case IncludeJtagDTM => false
-    case ExtMem => AXIMasterConfig(0x80000000L, 0x10000000L, 8, 4)
-    case ExtBus => AXIMasterConfig(0x60000000L, 0x20000000L, 8, 4)
+    case ExtMem => MasterConfig(base=0x80000000L, size=0x10000000L, beatBytes=8, idBits=4)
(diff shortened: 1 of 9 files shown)


src/main/scala/rocketchip/Configs.scala
```diff
@@ -28,8 +28,9 @@ class BasePlatformConfig extends Config(
     case PeripheryBusArithmetic => true
     // Note that PLIC asserts that this is > 0.
     case IncludeJtagDTM => false
-    case ExtMem => AXIMasterConfig(0x80000000L, 0x10000000L, 8, 4)
-    case ExtBus => AXIMasterConfig(0x60000000L, 0x20000000L, 8, 4)
+    case ExtMem => MasterConfig(0x80000000L, 0x10000000L, 8, 4)
+    case ExtBus => MasterConfig(0x60000000L, 0x20000000L, 8, 4)
+    case ExtIn  => SlaveConfig(8, 8, 2)
     case RTCPeriod => 100 // gives 10 MHz RTC assuming 1 GHz uncore clock
```
yunsup: Can we actually repeat the actual arguments, something like SlaveConfig(beatBytes = 8, idBits = 8, sourceBits = 2)?  Same for MasterConfig.  The user will have to open up the matching case class to see what these numbers mean.

terpstra: Ok. I've changed it. Please actually approve the review, as otherwise I cannot merge it.

terpstra: I've made the changes you requests. If you are happy with them, please approve the PR so I can merge it.

yunsup:
looks good!

PR #452, opened 2016-11-24
Use % in makefrag-verilog to prevent double firrtl execution

### LINKED ISSUE #358: Firrtl is called twice during RTL compilation
This can be observed in `vsim/` with a simple `make -n verilog`.  Chisel is run, and then firrtl is run twice with identical arguments.

Presumably there is some bug in `Makefrag-verilog` that is causing this.

colinschmidt:
Fixes #358


terpstra:
Why does this fix it? Both rules declared that they changed two files.

terpstra:
This fix is definitely not right. I declares a pattern rule that will generate any verilog file requested, all from the same input file.

colinschmidt:
Oops, yeah missed the second part. This still only runs it once, and I fixed the output directory again. Thanks for the help!

Issue #450, opened 2016-11-24
Guidance on Building a Standalone RoCC for Unit Testing


seldridge:
I'd like to be able to build a standalone version of a RoCC and run unit tests on it. This is generally a problem of building anything that mixes in `HasCoreParameters` as this pulls in all sorts of parameters that are only really populated for a full version of Rocket Chip.

I had this working before with a simple top module that instantiated the chosen RoCC, attached a [`TileLinkTestRam`](https://github.com/ucb-bar/rocket-chip/blob/94cc1efadc43a10ad0bdd14b0f672da89a729ffa/src/main/scala/uncore/devices/Bram.scala#L110) to the `autl` line, defined a few missing parameters (`NCoreplexExtClients`), preloaded the RAM using DPI, and drove this with a top-level C++ testbench. (I wasn't using the other ports of the RoCC and tied them off to some "assert if accessed" modules). However, some very recent modifications now require [`TLCacheEdge`](https://github.com/ucb-bar/rocket-chip/blob/master/src/main/scala/rocket/rocket.scala#L42) to be defined and I'm struggling to get something to build while attempting to grok a lot of things I haven't seen before (`LazyModule`, `diplomacy`?).

Is there any guidance on the best approach here? 

I'm currently trying to think about this from a "RoccPlex" approach and attempt to replicate the smallest subset of `RocketTile` / `RocketPlex` and see how far I can get.

There are a couple of alternatives approaches which I think are unsound:
  * Forgo unit testing and always rely on a Rocket + RoCC system. This may be fine, but only if you're running bare metal. Relying on the proxy kernel puts a user in an ~2 minute loop waiting for tests to run.
  * Define some wrapper that extends `CoreModule` and allow the actual RoCC to not extend `CoreModule` but use other parameters which are overridden later.

Alternative phrasing would be: "Does Hwacha have unit tests and how do they work?"

colinschmidt:
Hwacha does not have unit tests. The development loop uses bare-metal assembly tests for testing. In the past for classes where students develop RoCC accelerators we have used the advance tester to send RoCC commands and then expect results on the other interfaces. I haven't used that method in a long while and never on something as complicated as Hwacha. 
Hwacha also hasn't been updated to the Laziness of the Core/Tile yet only the coreplex.
You could also consider looking at how groundtest does things as I know a lot of things outside the core can be tested that way so perhaps things inside the core could also be tested that way.

hcook:
[...]

PR #468, opened 2016-11-30
Make instantiated RoCC use dcacheParams

seldridge:
Get RoccExampleConfig back to building. See: [#450](https://github.com/ucb-bar/rocket-chip/issues/450#issuecomment-263743855)


seldridge:
Great suggestion. The `CacheName` then doesn't look like it's pulled out of thin air... There's an additional commit on the PR with that change. I spot checked it with a build on `rv64ui-p-add` and a RoCC cmd/resp I'm testing and it didn't see any breakages.

colinschmidt:
This looks good to me, I'll put your branch on ucb-bar/rocket-chip since we don't have PRs building on travis (its only branches).

terpstra:
Sorry for introducing this bug. Glad you found the fix ok!

Issue #470, opened 2016-11-30
Travis fails all rocketsuites recently


colinschmidt:
Does anyone know why all of the rocket regression suites have been timing out recently?
Is this just a travis update that we have to deal with?

hcook:
I'm looking into this too but haven't made progress yet. Build #2113 passed and build #2114 failed, which strikes me as very suspicious considering #2114 is just a merge commit...

hcook:
Then UnitTestSuite passes in #2115 before failing again in #2116; none of that code has been touched.

colinschmidt:
Yeah thats why I was suspecting some weird travis change on their side but based on their twitter and update schedule page it doesn't seem like anything should have changed.

Thanks for looking into it.

hcook:
Comparing 2115.6 and 2116.6 (UnitTestSuite passing then failing) it looks like the biggest difference is that "make tools verilator -C regression SUITE=none" [install] takes  0.67s vs 2179.47s!

hcook:
https://www.traviscistatus.com/incidents/nq0sf0srvx8s has been going on for about the right timeframe, but they claim only affects OSX builds...

hcook:
I was thinking it could be because bumping riscv-tools invalidated the travis cache, but that doesn't explain why #2113 passed and #2114 failed...

hcook:
Looking at https://travis-ci.org/ucb-bar/rocket-chip/caches you can see the caches of the failing builds are notable smaller than the successful ones, but I'm not sure why different jobs are using/creating different caches... https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices implies that jobs using the same cache can corrupt it, but I'm not sure how that's applicable here

terpstra:
I looked into this a while ago. So let me explain some of the things I know:
1. The size of the cache you see on the /caching URL is 600MB * # of travis suites. So it is bigger when all the suites pass.
2. When there is a cache, then install takes 0.67s, otherwise it is super slow.
3. The build time for riscv-tools went up considerably recently.
4. In order for Travis to work, both the uncached install + the suite test must fit in 50 minutes.
5. I was unable to merge a PR that bumped all the submodules earlier, because it caused uncached rebuilds to exceed the 50 minute target. You guys managed to merge a bump somehow anyway.

hcook:
I'm looking into using container images stored on DockerHub to cache pre-built toolchains

zhemao:
This has been fixed, right?

[...]

PR #471, opened 2016-11-30
WIP PR to figure out why travis is failing

hcook:


colinschmidt:
I think travis will kill any jobs longer than 50 minutes on travis.org https://docs.travis-ci.com/user/customizing-the-build#Build-Timeouts

I guess we have to split up rocket suite further?

colinschmidt:
Another thought is why are some jobs using the cache while others build them. If we fixed that we could reliably re-run the failing builds.

colinschmidt:
Looking into this...https://docs.travis-ci.com/user/caching/#Caches-and-build-matrices

colinschmidt:
Just read your comment on #470. I was thinking that having separate caches is not what we want, since the tools we build are usable by all builds. If you look at what caches they are fetching its clear they fetch different ones and some of them are correct. By using one cache I'm hoping that once its correct all builds will be able to succeed, but there is this danger they mention of corruption...

hcook:
If you look under "Setting up build cache" in the log, you can see the passing ones say:
```
6.77sattempting to download cache archive
fetching travis-issues/cache-linux-precise-aa5acdf01d867f4243c7dd79e0b55d5791270f4e6bc66147f2a5c6575ffa3419--jdk-default--scala-2.10.4.tgz
found cache
```
whereas the failing ones say:
```
9.00sattempting to download cache archive
fetching travis-issues/cache-linux-precise-8f255974946400da5e81c3fb4382c8e62d686c994835ee7d40aa3eed2c1a3ca0--jdk-default--scala-2.10.4.tgz
fetching travis-issues/cache--jdk-default--scala-2.10.4.tgz
fetching master/cache-linux-precise-8f255974946400da5e81c3fb4382c8e62d686c994835ee7d40aa3eed2c1a3ca0--jdk-default--scala-2.10.4.tgz
found cache
```
and then have to build all of riscv-tools which leads them to time out.

But why one cache contains riscv-tools in regression/install whereas the other does not I cannot figure out.

colinschmidt:
Yeah one finds a cache on this branch already (I guess because it finished the first time) and the others just get master. If we can get them to all use the same cache without corruption then we could have a no-op suite that only builds the tools for caching purposes.

colinschmidt:
Oh goodness...we could also "warm up" the travis caches https://github.com/travis-ci/travis-ci/issues/6410

terpstra:
We shouldn't be depending on the cache being there. The caches are intended only to speed up the build for convenience. We should not be depending on them for correctness. The reason you have SOME of the suites building with a cache is because they already succeeded once. The caches are on a per SUITE basis.

hcook:
We're fundamentally dependent on the caches to supply riscv-gcc and have been for awhile afaict

terpstra:
Before travis started timing out two weeks ago, caches were not working at all. That means every run rebuilt riscv-tools. Not fitting under 50 minutes is a recent regression.

zhemao:
I'm pretty sure the jobs used to take more than 50 minutes if it couldn't find a recent cache. When did Travis start enforcing a 50 min. per-job timeout?

Since it takes around 45 min. to build riscv-tools from scratch, I don't see how we can fix this other than by ensuring that riscv-tools is pre-built.

hcook:
Still rebuilding verilator instead of using the one in the image, but i'm merging this now to start draining the PR queue

PR #476, opened 2016-12-01
Added few access functions for the variables that are needed for formal testbench

jchang0:


src/main/scala/uncore/tilelink2/Parameters.scala
```diff
@@ -7,9 +7,9 @@ import diplomacy._
 import scala.math.max
 
 case class TLManagerParameters(
-  address:            Seq[AddressSet],
+  val address:            Seq[AddressSet],
   regionType:         RegionType.T  = RegionType.GET_EFFECTS,
-  executable:         Boolean       = false, // processor can execute from this memory
+  val executable:         Boolean       = false, // processor can execute from this memory
   nodePath:           Seq[BaseNode] = Seq(),
```
terpstra: Remove these two 'val' changes. "case class" already has val on all arguments.

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -91,7 +91,10 @@ object LazyModule
   protected[diplomacy] var stack = List[LazyModule]()
   private var index = 0
 
+  var module_list = List[LazyModule]()
+
   def apply[T <: LazyModule](bc: T)(implicit sourceInfo: SourceInfo): T = {
+    module_list = bc :: module_list
     // Make sure the user put LazyModule around modules in the correct order
```
terpstra: Unsafe because multiple chisel runs will be munged together.

src/main/scala/uncore/tilelink2/Edges.scala
```diff
@@ -95,6 +95,24 @@ class TLEdge(
...
+  def param(x: TLDataChannel): UInt = {
+    x match {
+      case a: TLBundleA => a.param
+      case b: TLBundleB => b.param
+      case c: TLBundleC => c.param
+      case d: TLBundleD => d.param
+    }
+  }
+
   def size(x: TLDataChannel): UInt = {
```
terpstra: fine!

src/main/scala/uncore/tilelink2/Fragmenter.scala
```diff
@@ -13,7 +13,7 @@ import scala.math.{min,max}
 // Fragmenter modifies: PutFull, PutPartial, LogicalData, Get, Hint
 // Fragmenter passes: ArithmeticData (truncated to minSize if alwaysMin)
 // Fragmenter cannot modify acquire (could livelock); thus it is unsafe to put caches on both sides
-class TLFragmenter(minSize: Int, maxSize: Int, alwaysMin: Boolean = false) extends LazyModule
+class TLFragmenter(val minSize: Int, val maxSize: Int, val alwaysMin: Boolean = false) extends LazyModule
 {
```
terpstra: fine!

src/main/scala/uncore/tilelink2/Monitor.scala
```diff
@@ -415,11 +415,15 @@ class TLMonitor(gen: () => TLBundleSnoop, edge: () => TLEdge, sourceInfo: Source
...
+  
   lazy val module = new LazyModuleImp(this) {
     val io = new Bundle {
       val in = gen().asInput
     }
 
+    code_insertion(io.in, edge())
+    
     legalize(io.in, edge())(sourceInfo)
   }
```
terpstra: We should be creating TLMonitors based on a Parameter. Then you can pick your Monitor, the default Monitor, or NO Monitor...

PR #478, opened 2016-12-05
Parameterize diplomatic connections

terpstra:
This feeds Parameters all the way through diplomacy so that you can use Parameters for things like controlling the monitors.


terpstra:
I'm not 100% sure we want to do this, but it makes it much cleaner for @jchang0 to toggle stuff on/off for formal.

terpstra:
Rebased against current master

PR #480, opened 2016-12-05
clint: undefined registers must be zero

terpstra:
This is needed so that SMP-safe boot loaders can safely
read/write to the IPI register of non-existent harts.


aswaterman:
OK with me, but I'm wondering why the code doesn't examine the config string to only IPI harts that exist?

terpstra:
This works even without needing to parse the config string. So you can use it when chaining loaders before bbl, for example if you need to boot from spi/sd.

aswaterman:
ok

PR #481, opened 2016-12-06
More travis job re-balancing

hcook:


aswaterman:
@marmbrus committing to rocket-chip?!

marmbrus:
Lol, I'm branching (get it? you guys are into those, right?) out!  Lookout hardware world!

hcook:
Seems like firrtl/chisel is still getting rebuilt, but this is good enough to meet time limits as-is

seldridge:
Thanks for taking care of this @hcook. I had hit similar build timeouts on GitHub/Travis enterprise. I was able to complain and get these increased, but it's great for me to be able to run my own, fast riscv-tests-like regressions piggy backing off of what you have.

One thing: you may want to consider using `pwd` when mounting the volume instead of the hard coded ucb-bar path if someone has this forked.

PR #485, opened 2016-12-08
Diplomatic AHB

### LINKED ISSUE #383: AHB spec signal hreadyout missing in Hasti
According to the AHB 3 spec, AHB slaves should have a signal called hreadyout to indicate that a transfer has finished on the bus. I don't see this signal (or a similar one) in the hasti implementation. Is this intentional or masked by a different signal?
@zhemao @terpstra

terpstra:
This adds a hopefully more reliable AHB adapter. This closes #383.


terpstra:
@scottj97 @henrystyles Ok, I think this is good enough to start testing.

Issue #467, opened 2016-11-30
RoCCExampleConfig make error


leishangwen:
hi all,when i build RoCCExampleConfig emulator, i meet some exception, hope anyone can help me,thank you!

here is what i type:

> cd emulator
> make CONFIG=RoccExampleConfig

and the error is 

> mkdir -p /home/leisl/riscv/rocket-chip/emulator/generated-src/
> cd /home/leisl/riscv/rocket-chip && java -Xmx2G -Xss8M -XX:MaxPermSize=256M -jar /home/leisl/riscv/rocket-chip/sbt-launch.jar "run-main rocketchip.Generator /home/leisl/riscv/rocket-chip/emulator/generated-src rocketchip TestHarness rocketchip RoccExampleConfig"
> [info] Loading project definition from /home/leisl/riscv/rocket-chip/project
> Using addons: 
> [info] Set current project to rocketchip (in build file:/home/leisl/riscv/rocket-chip/)
> [info] Compiling 1 Scala source to /home/leisl/riscv/rocket-chip/chisel3/target/scala-2.11/classes...
> [info] Running rocketchip.Generator /home/leisl/riscv/rocket-chip/emulator/generated-src rocketchip TestHarness rocketchip RoccExampleConfig
> [info] [0.003] Elaborating design...
> 
> Interrupt map:
>   [1, 2] => ExampleRocketTop
> [error] (run-main-0) java.lang.reflect.InvocationTargetException
> java.lang.reflect.InvocationTargetException
> 	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
> 	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
> 	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
> 	at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
> 	at util.HasGeneratorUtilities$$anonfun$2.apply(GeneratorUtils.scala:47)
> 	at util.HasGeneratorUtilities$$anonfun$2.apply(GeneratorUtils.scala:44)
> 	at chisel3.core.Module$.do_apply(Module.scala:29)
> 	at chisel3.Driver$$anonfun$elaborate$1.apply(Driver.scala:190)
> 	at chisel3.Driver$$anonfun$elaborate$1.apply(Driver.scala:190)
[...]

PR #486, opened 2016-12-09
Convert frontend and icache to diplomacy/tl2

hcook:
ICache generates tilelink2.Get messages. Frontend and ICache are now both LazyModules. Adds a tilelink2.Xbar to the Tile.


src/main/scala/rocket/Tile.scala
```diff
@@ -129,11 +131,14 @@ class RocketTile(tileId: Int)(implicit p: Parameters) extends LazyModule {
...
-    val uncachedArb = Module(new ClientUncachedTileLinkIOArbiter(uncachedArbPorts.size)(icacheParams))
-    uncachedArb.io.in <> uncachedArbPorts
-    ucLegacy.module.io.legacy <> uncachedArb.io.out
+      val uncachedArb = Module(new ClientUncachedTileLinkIOArbiter(uncachedArbPorts.size)(icacheParams))
+      uncachedArb.io.in <> uncachedArbPorts
+      ucLegacy.module.io.legacy <> uncachedArb.io.out
+    } else {
+      ucLegacy.module.io.legacy.acquire.valid := Bool(false)
+      ucLegacy.module.io.legacy.grant.ready := Bool(false)
+    }
```
aswaterman: Is anything besides RoCC using the legacy port?  I think it's approaching time to mandate that RoCC accelerators move over to TL2.

src/main/scala/rocket/HellaCache.scala
```diff
@@ -51,8 +40,8 @@ trait HasL1HellaCacheParameters extends HasCacheParameters with HasCoreParameter
 
   require(isPow2(nSets))
   require(rowBits >= coreDataBits)
-  require(rowBits <= outerDataBits)
-  require(xLen <= outerDataBits) // would need offset addr for puts if data width < xlen
+  require(rowBits == cacheDataBits) // TODO should rowBits even be seperably specifiable?
```
aswaterman: I think we want to keep this option, since the RAM dimensions can serve as an energy/performance tuning parameter.

src/main/scala/rocket/Tile.scala
hcook: Nope, RoCC is now the only thing. I was going to make a separate PR that provides a LegacyRoCC diplomatic wrapper around old RoCC instances. 

Going forward, I agree it's time revisit what we actually want the RoCC interfaces to look like.  I think at the moment things are in particular disarray because of the non-upstreaming of the custom instruction stuff (e.g. #467?) so it might be a good time to re-spec out that side of the interface as well as the memory side? We probably should also cake-ify the Tile to make it easier to add arbitrary numbers and mixes of RoCCs.

src/main/scala/rocket/HellaCache.scala
hcook: I think we can agree that we always want to be able to sink an entire beat of refilled data per cycle, so it seems like TileLink width is at least a lower bound on rowBits. I'd go further and claim that the refill width is the broader architectural parameter; picking a set of RAMs of whatever optimal width so as to fulfill the requirement is backend flow decision that shouldn't even be expressed in Chisel?

src/main/scala/rocket/Tile.scala
aswaterman: Agreed, we should take the time to rethink RoCC entirely.  Probably need @yunsup @colinschmidt on board.

src/main/scala/rocket/HellaCache.scala
aswaterman: I agree that TL width and coreInstBits are both lower bounds on rowBits.  But we might want it to be wider, to amortize the other costs in the I$ readout, like the tag check, over multiple accesses.  Since that decision is exposed to the RTL, it can't be handled in the backend.

If it's really a pain, we can ditch it, though.

hcook: Keeping support for wider seems reasonable.

PR #487, opened 2016-12-11
Cleanup emulator.cc, use getopt, add help text

seldridge:
This changes the emulator to conform to POSIX-style options (e.g.,
short/'-' and long/'--') while preserving legacy option parsing (i.e.,
'+'). Options are read from the user until the first
non-option (either POSIX or legacy) is encountered. This and
everything following is assumed to be the binary and any arguments the
user wants to run on the emulator. All non-options are passed directly
to the DTM. This allows for the same option to be passed, safely, to
both the emulator and the binary, e.g., "+verbose".

This introduces a dependency on <getopt.h>.

One usage of `goto`. It's warranted...

This needs to have regressions run on it.


aswaterman:
thanks!

PR #489, opened 2016-12-13
Changes made to diplomacy for feature support + bug fix in Fragmenter and fuzzer + support infinite traffic in fuzzer

jchang0:


jchang0:
wasn't intentional, but did submodule update, so not sure what happended

PR #493, opened 2016-12-15
Perform integer division when parsing rocketchip.DefaultConfig.conf

GuzTech:
I got an error when performing `make verilog` in the vsim folder:

```
Traceback (most recent call last):
  File "/home/oguz286/Projects/rocket-chip/vsim/vlsi_mem_gen", line 177, in <module>
    main()
  File "/home/oguz286/Projects/rocket-chip/vsim/vlsi_mem_gen", line 174, in main
    print(gen_mem(*parse_line(line)))
  File "/home/oguz286/Projects/rocket-chip/vsim/vlsi_mem_gen", line 151, in gen_mem
    for i in range(mask_seg):
TypeError: 'float' object cannot be interpreted as an integer
```

This was due to the `parse_line` function performing a floating point division, and I changed it to an integer division.

### DIFF
--- a/vsim/vlsi_mem_gen
+++ b/vsim/vlsi_mem_gen
@@ -30,7 +30,7 @@ def parse_line(line):
       mask_gran = int(tokens[i+1])
     else:
       sys.exit('%s: unknown argument %s' % (sys.argv[0], a))
-  return (name, width, depth, mask_gran, width/mask_gran, ports)
+  return (name, width, depth, mask_gran, width//mask_gran, ports)
 
 def gen_mem(name, width, depth, mask_gran, mask_seg, ports):
   addr_width = max(math.ceil(math.log(depth)/math.log(2)),1)


aswaterman:
@GuzTech are you sure this is doing the right thing?  Does vlsi_mem_gen do something sensible if the width is not divisible by the mask granularity?

GuzTech:
@aswaterman I know that it doesn't give an error because the types are wrong, so that's already better than before :)

Since it's an integer division, it most likely would round down the answer, possibly leading to problems. IMHO the best way would be to do a floating point division like before, round up the answer, cast it to an integer, and in the for loop where the value is used, handle the case where the width is not divisible by the mask granularity.

aswaterman:
If the script's output is incorrect, not-erroring is not an improvement, though! @a0u can you check into the assumptions made within the script?

GuzTech:
Very true, but the issue here is that the script expects that `width/mask_gran` returns an integer by performing an integer division. In Python 2, the `/` operator performs floating point division if one of the operands is a float, and an integer division otherwise (like in this case). In Python 3 however, `/` always performs floating point division regardless of the types of the operands, and `//` performs integer division.

Using `//` in both Python 2 and 3 is correct in the sense that it yields the same result. Now if for some reason there are situations where width is **not** divisible by mask_gran, then that's an issue with how the script handles that, but that has nothing to do with my patch.

Yes it's convoluted, and I didn't know this since I don't use Python. You learn something new everyday :)

a0u:
@GuzTech Please post the contents of `vsim/generated-src/rocketchip.*.conf`.  Given how `SeqMems` are instantiated (i.e., `SeqMem(depth, Vec(width/mask_gran, Bits(width=mask_gran)))`), Chisel should not be emitting any SRAM configurations where `width` is not evenly divisible by `mask_gran`.

The `vlsi_mem_gen` script would not produce correct Verilog anyway if this assumption was violated, so ignoring it does not fix anything.  We should instead exit with a proper error message or, alternatively, introduce an explicit option to either pad the SRAM width or map the remainder to flip-flops.

GuzTech:
Contents of rocketchip.DefaultConfig.conf:

```
name _T_47_ext depth 512 width 64 ports mrw mask_gran 8 
name _T_146_ext depth 64 width 88 ports mwrite,read mask_gran 22 
name tag_array_ext depth 64 width 80 ports mrw mask_gran 20 
name _T_850_ext depth 512 width 64 ports rw  
name mem_ext depth 33554432 width 64 ports mwrite,read mask_gran 8 
name mem_0_ext depth 512 width 64 ports mwrite,read mask_gran 8 
```

So yes, they are divisible. But again, like I said, my patch has nothing to do with the behavior of the `vlsi_mem_gen` script. It behaves in the same way as it did before, but now also works with Python 3, and I have verified this.

hcook:
Closes #510

Issue #484, opened 2016-12-08
Better emulator.cc command line argument processing (i.e., getopt_long)


seldridge:
I've never been a fan of the way that [`emulator.cc`](https://github.com/ucb-bar/rocket-chip/blob/master/csrc/emulator.cc#L48) does command line processing. There's no help text, the normal version will happily read options that it has no business understanding (the user can specify a VCD file and a start cycles to the non-debug emulator), the command line processing is bespoke, and all the `+` options just seem like an odd choice. I'd like to propose cleaning this up with the following depending on what will get merged:
  * Add help text to `emulator.cc`
  * Have the non-debug version error out if passed command line arguments that it can't handle
  * Have the emulator dump help text if the user doesn't specify at least one non-command line argument
  * Refactor `emulator.cc` to use `getopt_long`

The first three are trivial, but if I do the latter, the cleanest way is to convert all existing `+` arguments to short or long `-` equivalents and just rely on `getopt_long` to deal with the processing. This is a pretty big (superficial) change to the longstanding way of running the emulator and has some far reaching (though again, superficial) consequences to all the Makefiles that call the emulator. This similarly has the potential to break build flows for people that may be using these arguments.

I can get together a pull request (in my free time), but what here sounds good enough to merge?

seldridge:
Okay, there's more interactions here than expected. The real way that the emulator should be processing the options is dependent on what htif_t expects. The first command line argument that doesn't begin with '-' or '+' is assumed to be the start of whatever program is going to be run. There's an underlying problem here where `emulator.cc` will use arguments to the binary that it shouldn't. I used to run into this all the time where I'd have a `-v` flag for some program run through the proxy kernel and the emulator would try to consume it.

I'll align `emulator.cc` with how hitf_t expects the command line arguments to look. This then should be done without needing to change anything (e.g., Makefiles).

aswaterman:
Hold up, doesn't getopt_long infect with GPL?

seldridge:
I am referring to the glibc version here, however, there does exist a [bsd variant](https://github.com/freebsd/freebsd/blob/386ddae58459341ec567604707805814a2128a57/crypto/openssh/openbsd-compat/getopt.h). Is it enough that you don't specify which version is being linked against or is this something that should just be avoided?

aswaterman:
Conceivably, we could snag the BSD version into rocket-chip/csrc and unconditionally build that.

[...]

PR #498, opened 2016-12-23
Fix emulator argument processing for unknown DTM arguments

minux:
Revision 7e79421 (issue #484) makes emulator.cc rejecting any unrecognized
+arg "legacy" arguments, however, this breaks riscv-torture emulator tests
as it needs to pass +signature to the DTM.

I think it is actually impossible to check for unknown argument here
unless we hardcode a list of all arguments recognized by fesrv. Fix this
issue by passing all arguments starting with the first unknown argument
to DTM.

Updates #484.

### DIFF
--- a/csrc/emulator.cc
+++ b/csrc/emulator.cc
@@ -118,10 +118,6 @@ int main(int argc, char** argv)
       // we've hit the binary. The binary is expected to be a
       // non-option and not start with '-' or '+'.
       case 1: {
-        if (optarg[0] != '+') {
-          to_dtm.push_back(optarg);
-          goto done_processing;
-        }
         std::string arg = optarg;
         if (arg == "+verbose")
           verbose = true;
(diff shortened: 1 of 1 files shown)


seldridge:
Ah, sorry about this. I expected that this was going to knick something...

From torture, it looks like `+signature` is passed before the binary which means this needs something. However, killing the processing once we hit the first unrecognized option means that only one of these undocumented fesvr options can be passed.

Since the emulator is linking against fesvr, I'd argue that it does make sense to document and process (by adding them to the to_DTM vec) additional options that fesvr understands (https://github.com/riscv/riscv-fesvr/blob/1f28f0e8dbf1172cd600c35e785cd37cb2cb9651/fesvr/htif.cc#L61) and therefore the emulator implicitly knows about. 

Alternatively, and as you recognize, any misunderstood options before the binary could be passed to the DTM and not kill processing. This, however, makes it impossible to check for invalid arguments to the emulator (and fesvr).

seldridge:
** All other options after the first unrecognized option and before the binary _do get passed here_. I was wrong about that above. 

This does work, but I'd be in favor of making it ransparent to the user what options are supported even if those options are really getting handled by fesvr.

aswaterman:
My understanding is the following three issues are separable:

- we should fix this regression to enable torture
- the fesvr +args should be described in the help message
- parsing args in multiple places should probably not be happening

Does this PR properly address the first?

minux:
Yes, this PR fixes C torture tests.

minux:
I think with Verilog's $value$plusargs support, there is actually
no way we can detect invalid plusargs here.

Even with my fix, running emulator without any argument shows
the help messages, which should be good enough, IMHO.

PR #502, opened 2017-01-06
Refactor Tile to use cake pattern

hcook:
Preparing to support heterogenous Tiles and Coreplexes, this PR provides traits for common Tile subsystems. Also adds a legacy TL1 wrapper for RoCC co-processors.


hcook:
@ccelio I've reintroduced parameter `BuildRocketTile` as a preliminary step to re-support BOOM integration; still need to do a little juggling w.r.t what is in BaseTile vs RocketTile and which one BoomTile inherits from.

I'm also wondering if we could maybe get `BuildCore` to work instead, if the DCacheShim got moved inside BoomCore and I added support for counting misses in the caches. Any thoughts?

ccelio:
@hcook Thanks for your work on this. 

I do believe conceptually that `buildCore` may be a better match for BOOM. There's really no difference between BOOM and rocket except the core and the interfaces can be made to match once I move the D$ shim internally into BOOM's core.  It should also allow changes made at the Tile level to be more transparent, which should be an added help to me as I track master.

(For reference: https://github.com/ucb-bar/riscv-boom/blob/master/src/main/scala/tile.scala)

Now to play Devil's Advocate on the other side, is Tile the better quantum of differentiation? Considering the amount of flexibility/parameterization within the Tile, that ZScale connected to the system via another mechanism anyways, and no other cores that I am aware of would prefer to bring in their own Tile subclass, I'm leaning with "no".  I think a `buildTile` world would be one in which designers are bringing in their own cache designs and their own co-processors that don't cleanly fit the RoCC interface. Is that a likely scenario, and one which is better served via a `buildTile` versus trait mix-in at the chip level? I can't say. 

For me, I'm okay with either decision. But I will at least move dcache-shim into the BOOM core, since that was a mistake to expose it as I did.

src/main/scala/coreplex/BaseTile.scala
```diff
@@ -0,0 +1,56 @@
...
+  implicit val p: Parameters
+  val module: TileNetworkModule
+  val l1backend = LazyModule(new TLXbar)
+  val masterNodes = List(TLOutputNode())
+  masterNodes.head := l1backend.node
+}
+
+trait TileNetworkBundle {
+  val outer: TileNetwork
+  val master = outer.masterNodes.head.bundleOut
```
hcook: @terpstra What I really want here is to be able to connect arbitrary numbers of inner masters to a single OutputNode and then have that node generate a matching Vec in the interface Bundle. What's the best way of doing something like that within the current Chisel/FIRRTL constraints?

hcook:
@ccelio Alright, let's try out doing the injection at the Core interface this time around. I think I'm going to get this PR in and then start a new one doing the actual Rocket core parameterization refactor + BuildCore.

src/main/scala/coreplex/BaseTile.scala
terpstra: As discussed offline, either get Chisel PR #366 merged, or use an OutputNode for multiple connections.

PR #504, opened 2017-01-10
Don't special-case power-of-2 replacement policy for BTB

aswaterman:
Also, bump riscv-tools.


ccelio:
Easiest. Code review. Evar.

Issue #506, opened 2017-01-11
Parameters with nested use of site report the outermost key on MatchError


hcook:
`Parameters.empty.alter { (pname, site, here, up) => pname match { case A => site(B) }}`
 `p(A) // returns "scala.MatchError: A"; not B as users probably would expect`

hcook:
Closed by #507

PR #507, opened 2017-01-12
Configs: use a uniform syntax without Match exceptions

terpstra:
The old style of specifying Configs used total functions.  The only way to
indicate that a key was not matched was to throw an exception.  Not only was
this a performance concern, but it also caused confusing error messages
whenever you had a match failure from a lookup within a lookup.  The
exception could get handled by an outer-lookup that then reported the wrong
key as missing.

### DIFF
--- a/src/main/scala/config/Config.scala
+++ b/src/main/scala/config/Config.scala
@@ -3,9 +3,6 @@
 package config
 
 class Field[T]
-class CDEMatchError() extends Exception {
-  override def fillInStackTrace() = this
-}
 
 abstract class View {
   final def apply[T](pname: Field[T]): T = apply(pname, this)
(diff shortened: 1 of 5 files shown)


hcook:
Closes #506

jchang0:
Can we do some kind of depreciation for CDEMatchError instead of removing it altogether for the time being? (not even sure if it's possible to depreciate a class)

terpstra:
We could, but it is only used in Config definitions, which need to be refactored anyway.

hcook:
To expand on @terpstra's comment, all cases of 
`class BasePlatformConfig extends Config(`
 ` (pname,site,here) => pname match {`
 `   case Foo => 2`
 `   case _ => throw new CDEMatchError`
  `})`
become
`class BasePlatformConfig extends Config( { (site,here up) =>`
`  case Foo => 2`
`})`

colinschmidt:
Wait does the config package not use exceptions for control flow?

PR #513, opened 2017-01-17
Add TLBuffers on the L1 backends and blind exit points

hcook:


terpstra:
I think you need to remove this one:
  mmio :=
    TLBuffer()(
    TLWidthWidget(l1tol2_beatBytes)(

... but I'd like to see a graphml image.

PR #518, opened 2017-01-18
jtag_dtm: Update regression to run and pass.

mwachs5:
This reverts some changes to make ExampleRocketChipTop and its TestHarness actually absorb the WithJtagDTM configuration parameter. This allows us to (again) test the JTAG DTM.

Is it possible to add the test in regression to the suite? As of now I manually run it when I'm updating the JTAG DTM, but I think it would be good to automate this so it doesn't keep getting broken. Is there a limit on resources or time that we're in danger of hitting if I add this?


mwachs5:
I just remembered we can't run this regression because the JTAG VPI doesn't work on Travis. So, I guess for now I'll just keep manually running it when I want it to work.

PR #520, opened 2017-01-19
[WIP] : Use Chisel to generate JTAG Tap

mwachs5:
This is a WIP PR, do not merge.

This uses @ducky64 's Chisel JTAG TAP Generator (https://github.com/ucb-art/chisel-jtag) code to replace the Verilog version of the JTAG DTM. For now I just added it as a submodule for testing/dev purposes.

I was playing around with the various reset sources, they're still rather confused and need cleanup.

I would still like more separation between the JTAG TAP itself and the logic that it drives. Right now the DTM and the TAP logic are still wrapped together.  I talked to @terpstra about using Diplomacy to gather up all the parts of a design that have a few JTAG registers, then just instantiate one TAP to control them.


mwachs5:
This is going to get rolled into the Debug update.

PR #530, opened 2017-01-26
Show how to build riscv-tools in parallel

edwardcwang:
Make it easier to build riscv-tools faster on multicore machines


README.md
```diff
@@ -37,6 +37,7 @@ of riscv-tools:
     $ cd rocket-chip/riscv-tools
     $ git submodule update --init --recursive
     $ export RISCV=/path/to/install/riscv/toolchain
+    $ export MAKEFLAGS="$MAKEFLAGS -jN"
```
aswaterman: To make this user-proof, put a number here instead of N.  If someone follows these directions literally, they'll get an error when they try to run make.

    export MAKEFLAGS="$MAKEFLAGS -j16"

edwardcwang: I could do that, and what do you think about a comment explaining the need to replace N? I noticed that there are 14 other uses of "-jN" in the document and a comment "Assuming you have N cores on your host system".

aswaterman: Good point.

I'll merge as-is, and we can test this theory.

edwardcwang:
@aswaterman Would it be possible to merge this?

aswaterman:
Yeah, I assumed you'd merge it after I approved it, not realizing you don't have commit access.

edwardcwang:
Yeah, I put a request to join the group for rocket write so that I could merge this but it appears that the request hasn't been approved yet...

PR #532, opened 2017-01-27
Ignore the built firrtl.jar.

richardxia:
In #526, we now build firrtl.jar to rocket-chip's `lib/` directory. This adds a gitignore rule for the entire `lib/` directory.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ project/target
 *.swp
 *~
 .addons-dont-touch
+/lib/


richardxia:
Looks like the build failed on that "No output has been received in the last 10m0s" error, which I think we've seen before. Can you retry the build? I don't think I have permission to.

jackkoenig:
I restarted the one that failed

PR #533, opened 2017-01-27
Rational clock crossing

terpstra:
If you know two clocks are related with a N:1 or 1:N relationship, you can cross the clock domains with lower latency than an AsyncQueue. This crossing adds 1 cycle in the target clock domain. For a TileLink round-trip, this means it adds one cycle latency to each domain, equivalent to a TLBuffer. Just like a TLBuffer, all outputs of TLRational are registers (bits, valid, and ready). As opposed to a TLBuffer, the construction costs 3*bits instead of 2*bits, so it is not completely free.

### DIFF
--- a/src/main/scala/uncore/tilelink2/Nodes.scala
+++ b/src/main/scala/uncore/tilelink2/Nodes.scala
@@ -180,3 +180,39 @@ case class TLAsyncSinkNode(depth: Int, sync: Int) extends MixedNode(TLAsyncImp, 
   uFn = { case (1, Seq(p)) => Seq(TLAsyncManagerPortParameters(depth, p)) },
   numPO = 1 to 1,
   numPI = 1 to 1)
+
+object TLRationalImp extends NodeImp[TLClientPortParameters, TLManagerPortParameters, TLEdgeParameters, TLEdgeParameters, TLRationalBundle]
+{
+  def edgeO(pd: TLClientPortParameters, pu: TLManagerPortParameters): TLEdgeParameters = TLEdgeParameters(pd, pu)
+  def edgeI(pd: TLClientPortParameters, pu: TLManagerPortParameters): TLEdgeParameters = TLEdgeParameters(pd, pu)
+
+  def bundleO(eo: Seq[TLEdgeParameters]): Vec[TLRationalBundle] = Vec(eo.size, new TLRationalBundle(TLBundleParameters.union(eo.map(_.bundle))))
+  def bundleI(ei: Seq[TLEdgeParameters]): Vec[TLRationalBundle] = Vec(ei.size, new TLRationalBundle(TLBundleParameters.union(ei.map(_.bundle))))
+
+  def colour = "#00ff00" // green
+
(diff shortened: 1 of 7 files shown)


src/main/scala/uncore/tilelink2/Nodes.scala
```diff
@@ -180,3 +180,39 @@ case class TLAsyncSinkNode(depth: Int, sync: Int) extends MixedNode(TLAsyncImp,
...
+
+object TLRationalImp extends NodeImp[TLClientPortParameters, TLManagerPortParameters, TLEdgeParameters, TLEdgeParameters, TLRationalBundle]
+{
+  def edgeO(pd: TLClientPortParameters, pu: TLManagerPortParameters): TLEdgeParameters = TLEdgeParameters(pd, pu)
+  def edgeI(pd: TLClientPortParameters, pu: TLManagerPortParameters): TLEdgeParameters = TLEdgeParameters(pd, pu)
+
+  def bundleO(eo: Seq[TLEdgeParameters]): Vec[TLRationalBundle] = Vec(eo.size, new TLRationalBundle(TLBundleParameters.union(eo.map(_.bundle))))
+  def bundleI(ei: Seq[TLEdgeParameters]): Vec[TLRationalBundle] = Vec(ei.size, new TLRationalBundle(TLBundleParameters.union(ei.map(_.bundle))))
+
+  def colour = "#00ff00" // green
```
aswaterman: It's spelled "color"

mwachs5:
Are there any assertions that can be added or generated here to make sure you're right about the relationships between clocks?

terpstra:
That's a good idea, but I'm not sure how to do it. Do you have any suggestions? Ultimately, timing analysis should fail if you set it up wrong.

src/main/scala/uncore/tilelink2/RationalCrossing.scala
```diff
@@ -0,0 +1,177 @@
+// See LICENSE.SiFive for license details.
+
+package uncore.tilelink2
+
+import Chisel._
+import chisel3.internal.sourceinfo.SourceInfo
+import config._
+import diplomacy._
+import util._
+
```
mwachs5: To me, "Rational" is not an intuitive name for this, but that's not my complaint. Can you please add a comment explaining what "Rational" means (basically what you said in this PR)

mwachs5:
@terpstra sorry I don't have an easy way to do this right now,  but something to keep in mind when we emit timing constraints / waivers.

src/main/scala/uncore/tilelink2/RationalCrossing.scala
richardxia: A few random ideas for names:

- EvenlyDivisible
- Divisible
- Multiple
- Congruent (because a \congruent b mod min(a, b), though a = k*b is a stronger assertion than saying they have the same remainder)

terpstra: FYI, you can use this crossing to go from N:M by going via N:(N*M):M. So it truly is a Rational crossing...

terpstra: Probably you get slightly better timing closure with N:lcm(N,M):M...

PR #537, opened 2017-01-30
BankedL2Config: use the same LazyModule for all L2 banks

terpstra:
This makes it much easier for banked coherence managers to support
cross-bank functionality, like a common control port, for example.

### DIFF
--- a/src/main/scala/diplomacy/LazyModule.scala
+++ b/src/main/scala/diplomacy/LazyModule.scala
@@ -27,10 +27,15 @@ abstract class LazyModule()(implicit val p: Parameters)
       m.getName != "children"
     }.flatMap { m =>
       if (classOf[LazyModule].isAssignableFrom(m.getReturnType)) {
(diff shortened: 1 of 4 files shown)


src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -27,10 +27,15 @@ abstract class LazyModule()(implicit val p: Parameters)
       m.getName != "children"
     }.flatMap { m =>
       if (classOf[LazyModule].isAssignableFrom(m.getReturnType)) {
-        Seq((m.getName, m.invoke(this)))
+        val obj = m.invoke(this)
+        if (obj eq null) Seq() else Seq((m.getName, obj))
```
hcook: do you want to use Nil (http://www.scala-lang.org/api/2.7.0/scala/Nil$object.html) instead of Seq()?

terpstra: I didn't know about Nil. ;)

PR #538, opened 2017-01-30
[WIP] Blackboxed TLRegisterRouter

hcook:
Uses parameterized Blackbox to create an arbitrarily-sized register file in verilog, while also connecting it to the periphery bus using a TLRegisterRouter. Inserting arbitrary verilog devices into BlackBoxedRegs.v is left to the user for now; module BlackBoxedRegs merely defines the io semantics expected by the register router.

### DIFF
--- a/src/main/scala/uncore/devices/BlackBoxRegisterRouter.scala
+++ b/src/main/scala/uncore/devices/BlackBoxRegisterRouter.scala
@@ -1,0 +1,27 @@
+// See LICENSE.SiFive for license details.
+
+package uncore.devices
+
+import Chisel._
+
+import config._
+import diplomacy._
+import regmapper._
+import rocketchip.{PeripheryBusConfig}
+import uncore.tilelink2._
+import util.BlackBoxedRegsWrapper
+
+case class  BlackBoxedRegMapParams(address: BigInt, numRegs: Int, widthBytes: Int)
+case object BlackBoxedRegMapKey extends Field[Seq[BlackBoxedRegMapParams]]
+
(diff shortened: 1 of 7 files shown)


src/main/scala/uncore/devices/BlackBoxRegisterRouter.scala
```diff
@@ -0,0 +1,27 @@
...
+import util.BlackBoxedRegsWrapper
+
+case class  BlackBoxedRegMapParams(address: BigInt, numRegs: Int, width: Int)
+case object BlackBoxedRegMapKey extends Field[Seq[BlackBoxedRegMapParams]]
+
+trait HasBlackBoxedRegMap extends Module with HasRegMap {
+  val params: BlackBoxedRegMapParams
+  val w = params.width
+  val regs = Module(new BlackBoxedRegsWrapper(params.numRegs, w))
+  regmap((regs.io.v.zipWithIndex.map { case(r, i) => i*(w/8) -> Seq(RegField.rwReg(w,r)) }):_*)
```
terpstra: This is not really safe. If your TL device has width 128 and the BB has width 32, you will be able to write 4 registers at once, which will fail.

Furthermore, your just blithely divide w/8 and assume it will all work out. If w=23, it would round down. If w=24, you have registers that repeat every 3 bytes, meaning a 4-byte write should update two of them...

I think it would have been a better fit to build this using an SRAM-like TL interface than trying to force it into RR. If you want to use the RR for some teaching reason, then add on a requirement that beatBytes*8 == params.width.

PR #540, opened 2017-01-31
Dedup rocket

terpstra:
Make it possible for multiple RocketTiles to share the same module implementation.


colinschmidt:
Just a question: were these literals named "nicely" in firrtl/verilog before this change?

jackkoenig:
A literal has no name

colinschmidt:
oh yeah that makes sense.

PR #541, opened 2017-02-02
Reverse hints

terpstra:
Sometimes you might want to hint that dropping permissions is wise.

### DIFF
--- a/src/main/scala/uncore/tilelink2/Bundles.scala
+++ b/src/main/scala/uncore/tilelink2/Bundles.scala
@@ -104,10 +104,13 @@ object TLAtomics
 
 object TLHints
 {
-  val width = 1
+  val width = 3
 
-  def PREFETCH_READ  = UInt(0, width)
-  def PREFETCH_WRITE = UInt(1, width)
+  def GROW_toB = UInt(0, width)
+  def GROW_toT = UInt(1, width)
+  def SHRINK_toN = UInt(4, width)
+  def SHRINK_toB = UInt(5, width)
+  def SHRINK_toT = UInt(6, width)
 }
 
 sealed trait TLChannel extends TLBundleBase {
--- a/src/main/scala/diplomacy/LazyModule.scala
+++ b/src/main/scala/diplomacy/LazyModule.scala
@@ -32,7 +32,7 @@ abstract class LazyModule()(implicit val p: Parameters)
       } else if (classOf[Seq[LazyModule]].isAssignableFrom(m.getReturnType)) {
         val obj = m.invoke(this)
         if (obj eq null) Seq() else {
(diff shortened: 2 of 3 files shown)


terpstra:
Actually, I'm not sure it's a good idea to make this change until we have something that can actually issue these sorts of Hints.

src/main/scala/uncore/tilelink2/Bundles.scala
```diff
@@ -104,10 +104,13 @@ object TLAtomics
...
 object TLHints
 {
-  val width = 1
+  val width = 3
 
-  def PREFETCH_READ  = UInt(0, width)
-  def PREFETCH_WRITE = UInt(1, width)
+  def GROW_toB = UInt(0, width)
+  def GROW_toT = UInt(1, width)
+  def SHRINK_toN = UInt(4, width)
```
hcook: // flushes the block entirely

src/main/scala/uncore/tilelink2/Bundles.scala
```diff
@@ -104,10 +104,13 @@ object TLAtomics
...
+  def SHRINK_toB = UInt(5, width)
```
hcook: // produces the dirty data, while also ceding write permissions in anticipation of consumption

src/main/scala/uncore/tilelink2/Bundles.scala
```diff
@@ -104,10 +104,13 @@ object TLAtomics
...
+  def SHRINK_toT = UInt(6, width)
```
hcook: // cleans the block by writing back dirty data while retaining all permissions

src/main/scala/uncore/tilelink2/Bundles.scala
```diff
@@ -104,10 +104,13 @@ object TLAtomics
 
 object TLHints
 {
-  val width = 1
+  val width = 3
 
-  def PREFETCH_READ  = UInt(0, width)
-  def PREFETCH_WRITE = UInt(1, width)
+  def GROW_toB = UInt(0, width)
```
hcook: // acquires read permissions

src/main/scala/uncore/tilelink2/Bundles.scala
```diff
@@ -104,10 +104,13 @@ object TLAtomics
...
+  def GROW_toT = UInt(1, width)
```
hcook: // acquires read and write permissions

terpstra:
I agree we should make this change, but the L2 change is non-trivial and I was hoping to have a test case before we go and break it without realizing.

terpstra:
We've decided not to do this, and take a less "you must do this" approach.

hcook:
Should cherry-pick 8c7f1c6

PR #556, opened 2017-02-17
Rational config

terpstra:
Make it possible to specify which side of the crossing is fast and which is slow.

Using this option can make it easier to close timing for the slow side of the crossing; it places the registers all on the slow side. However, if the wrong side is declared fast, the crossing will not be legal! The default remains a symmetric crossing which is safe in all cases and always puts the register on the sink side.

### DIFF
--- a/src/main/scala/rocketchip/RocketTestSuite.scala
+++ b/src/main/scala/rocketchip/RocketTestSuite.scala
@@ -26,6 +26,9 @@ run-$makeTargetName: $$(addprefix $$(output_dir)/, $$(addsuffix .out, $$($makeTa
 
 run-$makeTargetName-debug: $$(addprefix $$(output_dir)/, $$(addsuffix .vpd, $$($makeTargetName)))
 \t@echo; perl -ne 'print "  [$$$$1] $$$$ARGV \\t$$$$2\\n" if( /\\*{3}(.{8})\\*{3}(.*)/ || /ASSERTION (FAILED)/i )' $$(patsubst %.vpd,%.out,$$^) /dev/null | perl -ne 'if(/(.*)/){print "$$$$1\\n\\n"; exit(1) if eof()}'
+
+run-$makeTargetName-fst: $$(addprefix $$(output_dir)/, $$(addsuffix .fst, $$($makeTargetName)))
+\t@echo; perl -ne 'print "  [$$$$1] $$$$ARGV \\t$$$$2\\n" if( /\\*{3}(.{8})\\*{3}(.*)/ || /ASSERTION (FAILED)/i )' $$(patsubst %.fst,%.out,$$^) /dev/null | perl -ne 'if(/(.*)/){print "$$$$1\\n\\n"; exit(1) if eof()}'
 """
 }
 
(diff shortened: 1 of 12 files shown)


terpstra:
This PR also added support for using gtkwave with verilator, for a fully open source work-flow. This was quite helpful for me while I was in Germany and unable to use DVE due to high latency. FYI, I was able to synthesize and view waveforms entirely locally on my MacBook, which made me happy.

vsrc/ClockDivider2.v
```diff
@@ -0,0 +1,21 @@
+// See LICENSE.SiFive for license details.
+
+/** This black-boxes a Clock Divider.
+  *
+  * Because Chisel does not support
+  * blocking assignments, it is impossible
+  * to create a deterministic divided clock.
+  *
```
mwachs5: Still think we should add a note/warning here about constraints, simulation vs synthesis, etc.

src/main/scala/util/RationalCrossing.scala
```diff
@@ -7,6 +7,32 @@
 package util
 import Chisel._
 
+// A rational crossing must put registers on the slow side.
+// This trait covers the options of how/where to put the registers.
+// BEWARE: the source+sink must agree on the direction!
```
mwachs5: Any way to check this when you hook them up?

src/main/scala/util/ClockDivider.scala
```diff
@@ -5,15 +5,11 @@ package util
 import Chisel._
 
 /** Divide the clock by 2 */
```
mwachs5: Can you add a note here similar to the one in the Verilog, in case people don't go read the Verilog... and that this is intended to be used for simulation or with careful application of constraints.

src/main/scala/rocketchip/RocketTestSuite.scala
```diff
@@ -26,6 +26,9 @@ run-$makeTargetName: $$(addprefix $$(output_dir)/, $$(addsuffix .out, $$($makeTa
 
 run-$makeTargetName-debug: $$(addprefix $$(output_dir)/, $$(addsuffix .vpd, $$($makeTargetName)))
 \t@echo; perl -ne 'print "  [$$$$1] $$$$ARGV \\t$$$$2\\n" if( /\\*{3}(.{8})\\*{3}(.*)/ || /ASSERTION (FAILED)/i )' $$(patsubst %.vpd,%.out,$$^) /dev/null | perl -ne 'if(/(.*)/){print "$$$$1\\n\\n"; exit(1) if eof()}'
+
+run-$makeTargetName-fst: $$(addprefix $$(output_dir)/, $$(addsuffix .fst, $$($makeTargetName)))
```
mwachs5: nice.

src/main/scala/util/RationalCrossing.scala
terpstra: Yes. Diplomacy makes sure you don't use it wrong and there are also asserts that detect if the slow is in the wrong side.

src/main/scala/rocketchip/RocketTestSuite.scala
terpstra: I just copy-pasted ... but, yeah, that whole Makefile from Scala is hideous.

PR #561, opened 2017-02-23
diplomacy: use HeterogeneousBag instead of Vec

terpstra:
This makes it possible for the bundles to have different widths.

Previously, we had to widen all the bundles passing through a node
to the widest of all the possibilities.  This would mean that if
you had two source[] fields, they end up the same.


terpstra:
Do not merge this until firrtl and chisel stop conspiring to add a superfluous _ infront of integer-named values in a Record. @jackkoenig

terpstra:
Closing in favor of #562.

PR #562, opened 2017-02-24
Periphery adjustments

hcook:
Changes some external-facing periphery names and types.

### DIFF
--- a/src/main/scala/diplomacy/Nodes.scala
+++ b/src/main/scala/diplomacy/Nodes.scala
@@ -4,6 +4,7 @@ package diplomacy
 
 import Chisel._
 import config._
+import util.HeterogeneousBag
 import scala.collection.mutable.ListBuffer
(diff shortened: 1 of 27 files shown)


src/main/scala/diplomacy/Nodes.scala
```diff
@@ -112,7 +113,7 @@ trait InwardNode[DI, UI, BI <: Data] extends BaseNode with InwardNodeHandle[DI,
   protected[diplomacy] val iStar: Int
   protected[diplomacy] val iPortMapping: Seq[(Int, Int)]
   protected[diplomacy] val iParams: Seq[UI]
-  val bundleIn: Vec[BI]
+  val bundleIn: HeterogeneousBag[BI]
```
jerryhethatday, line 116: now, you guys moves to Seq, why not use the HeterogeneousBag anymore.

Issue #516, opened 2017-01-18
configuration string pattern mismatch between rocketchip and bbl


sashimi-yzh:
[bbl](https://github.com/riscv/riscv-pk/blob/master/machine/configstring.c#L9) uses the pattern `ram{0{addr` to search for memory information. However, the configure string output during elaboration looks like
```
Generated Configuration String
plic {
  priority 0xc000000;
  pending 0xc001000;
  ndevs 2;
};
rtc {
  addr 0x200bff8;
};
core {
  0 {
    0 {
      isa rv64imafds;
      timecmp 0x2004000;
      ipi 0x2000000;
      plic {
        m {
         ie 0xc002000;
         thresh 0xc200000;
         claim 0xc200004;
        };
        s {
         ie 0xc002080;
         thresh 0xc201000;
         claim 0xc201004;
        };
      };
    };
  };
};
ExampleRocketTop {  <= this should be "ram" for bbl to query memory information
  addr 0x80000000;
  size 0x10000000;
}
bootrom {
  addr 0x1000;
  size 0x1000;
}
debug {
  addr 0x0;
  size 0x1000;
}
clint {
  addr 0x2000000;
  size 0x10000;
}
ExampleRocketTop {
  addr 0x60000000;
  size 0x20000000;
}
plic {
  addr 0xc000000;
  size 0x4000000;
}
```

seldridge:
FWIW, this is a duplicate of #463 which was superseded by #474. You can hardcode the ram string inside of the `query_mem` function that you reference of `configstring.c` for the time being. 

This is generally bothersome as `pk` is broken out of the box, but it's the state of the world...

aswaterman:
Thans @seldridge -- it is bothersome, and @terpstra and I will try to sort out the disagreement on the contents of the config string soon.

sashimi-yzh:
Oh, I did not see #474 is going to fix this. Thank you for pointing this out! And it makes sense to do the hardcode. :)

Issue #474, opened 2016-12-01
Config String needs a specification


terpstra:
We need:
1. A definition of what devices should print for themselves.
2. An implementation that gathers the information for devices as viewed from the core/plic/prci and outputs a string matching definition 1.

terpstra:
This is now resolved; we are using DTS and it's been merged in both hardware and software.

davidbiancolin:
yay!

Issue #232, opened 2016-08-30
Heterogeneous core configs cannot generate the correct config string


colinschmidt:
The way the config string is created currently only looks at one of the core's parameters for things like what isa is supported.
rocket-chip should be able to handle the creation of heterogeneous ISA cores including the correct config string.

aswaterman:
This is already on the docket.

colinschmidt:
Cool. It would also be useful if the RoCCParameters included an extension name. Then just add something like the following to the isa string as well.

```
${site(BuildRoCC).map(_.xName).mkString}
```

aswaterman:
Agreed.  We plan to standardize how components register themselves in the config string & address map (it's quite ad hoc now); RoCC should use that mechanism once it exists.

PR #567, opened 2017-03-02
Dts - Device Tree Output

terpstra:
This adds support to rocketchip for generating a compliant device tree description. This should resolve issues #516 #474 #232, once we update spike+bbl+linux.

### DIFF
--- a/src/main/scala/diplomacy/Nodes.scala
+++ b/src/main/scala/diplomacy/Nodes.scala
@@ -302,8 +302,8 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   numPI: Range.Inclusive = 1 to 999)
   extends MixedNode(inner, outer)(numPO, numPI)
 {
-  require (numPO.end >= 1, s"${name} does not accept outputs${lazyModule.line}")
-  require (numPI.end >= 1, s"${name} does not accept inputs${lazyModule.line}")
+//  require (numPO.end >= 1, s"${name} does not accept outputs${lazyModule.line}")
+//  require (numPI.end >= 1, s"${name} does not accept inputs${lazyModule.line}")
 
   val externalIn: Boolean = true
   val externalOut: Boolean = true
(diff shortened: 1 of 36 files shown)


terpstra:
Remaining to be done: sort out the boot ROM contents. The ROM should set a0 to hartid and a1 to a pointer to the compiled and flattened dts. Right now the ROM behaves analogous to the old config string approach.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -302,8 +302,8 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   numPI: Range.Inclusive = 1 to 999)
   extends MixedNode(inner, outer)(numPO, numPI)
 {
-  require (numPO.end >= 1, s"${name} does not accept outputs${lazyModule.line}")
-  require (numPI.end >= 1, s"${name} does not accept inputs${lazyModule.line}")
+//  require (numPO.end >= 1, s"${name} does not accept outputs${lazyModule.line}")
+//  require (numPI.end >= 1, s"${name} does not accept inputs${lazyModule.line}")
```
aswaterman: Remove commented code?

neuschaefer:
Wouldn't DTB/FDT (devicetree binary/blob, aka. flattened devicetree) be a better format than DTS here? Operating systems already have DTB parsers, but AFAIK no DTS parsers. And if a user wants to look at the devicetree in DTS form, they can still easily decompile it from DTB or linux's `/proc/device-tree` using the [devicetree compiler](https://git.kernel.org/cgit/utils/dtc/dtc.git).

In your current plans, will the bootloader or the kernel have to understand DTS? IOW, will the bootloader compile the DTS to DTB so it's easier to parse for the kernel?

terpstra:
This is an intermediate commit. I need to revise how the ROM is generated still. The end goal is DTB and a pointer in a1 on power-up.

neuschaefer:
Ah, ok, I'm totally fine with that. Thanks.

PR #568, opened 2017-03-03
PeripheryExtInterrupts: elide node if NExtTopInterrupts = 0

terpstra:


terpstra:
Without this patch, using the trait with 0 interrupts violates requirements.

PR #574, opened 2017-03-08
Fix in-register representation of fdiv.s/fsqrt.s result

aswaterman:
We were zero-extending it, which is a double-precision zero in the recoded format.  So, when spilled and reloaded with fsd/fld, the original value was destroyed.  Instead, set the MSBs so that it represents sNaN.  When spilled, the single-precision number will be preserved as the NaN payload.


ccelio:
Is there a good way to test this? Like a slightly more comprehensive fdiv_s.s that includes fld/fsd?

aswaterman:
The recoding test is the right place for this, but I'm gonna leave that to someone else.

PR #575, opened 2017-03-08
dcache: we need the bits within the beat to select the right word

terpstra:


terpstra:
This bug was introduced in https://github.com/ucb-bar/rocket-chip/commit/b772edcb1b423457c5430e2eea6f60a98fe1d448

Issue #571, opened 2017-03-07
Hasti can't  handle N masters to one slave 


colin4124:
First of all, let's look at one piece of code in [hasti.scala](https://github.com/ucb-bar/rocket-chip/blob/master/src/main/scala/junctions/hasti.scala#L280):

``` scala
  for (m <- 0 until nMasters) {
    // If the master is connected to a slave, the slave determines hready.
    // However, if no slave is connected, for progress report ready anyway, if:
    //   bad address (swallow request) OR idle (permit stupid masters to move FSM)
    val autoready = nowhereM(m) || masters(m).isIdle()
    val hready = Mux1H(unionGrantMS(m), slaves.map(_.hready ^ autoready)) ^ autoready
    masters(m).hready := hready
    // If we diverted a master, we need to absorb his address phase to replay later
    diversions(m).io.divert := (bubbleM(m) || blockedM(m)) && NSeq(m) && hready
  }
```

When the number of masters is N and slaves is only one,  `unionGrantMS(m)` has only one element, the same as `slaves.map(_.hready ^ autoready)` . As the result, two parameters of `Mux1H` both are the vector that contains only one element.

Then let's look at the implement of `Mux1H` in [Mux.scala](https://github.com/ucb-bar/chisel3/blob/master/src/main/scala/chisel3/util/Mux.scala#L16):

``` scala
/** Builds a Mux tree out of the input signal vector using a one hot encoded
  * select signal. Returns the output of the Mux tree.
  *
  * @note results undefined if multiple select signals are simultaneously high
  */
object Mux1H {
  def apply[T <: Data](sel: Seq[Bool], in: Seq[T]): T =
    apply(sel zip in)
  def apply[T <: Data](in: Iterable[(Bool, T)]): T = SeqUtils.oneHotMux(in)
  def apply[T <: Data](sel: UInt, in: Seq[T]): T =
    apply((0 until in.size).map(sel(_)), in)
  def apply(sel: UInt, in: UInt): Bool = (sel & in).orR
}
```

Mux1H will pass his two parameters as a vector of tuple, witch contains only one tuple, to [SeqUtils.oneHotMux](https://github.com/ucb-bar/chisel3/blob/master/chiselFrontend/src/main/scala/chisel3/core/SeqUtils.scala#L54) : 

``` scala
  /** Returns the data value corresponding to the lone true predicate.
    *
    * @note assumes exactly one true predicate, results undefined otherwise
    */
[...]

PR #576, opened 2017-03-09
Fixed Hasti can't handle N masters to one slave

colin4124:
Fixed Hasti can't handle N masters to one slave #571


terpstra:
This code is obsolete anyway. Is there a reason you don't switch the the uncore.ahb package? At some point we are probably just going to remove hasti/nasti/etc...

colin4124:
I want to tailor rocket chip as MCU, so I want to use AHB-Lite(hasti) only. I recently study how to use hasti by writing some testcase for it.  

I wonder uncore.ahb is implementation of AHB-lite or AHB ? If I want to use uncore.ahb through TileLink2 and its converter， how much additional circuit usage than using uncore.ahb directly?

PR #578, opened 2017-03-09
Privileged architecture updates

aswaterman:
This brings rocket more-or-less in line with priv-1.10-draft.  It also contains some misc. fixes and improvements made along the way.

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -503,4 +503,8 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
       flushing := false
     }
   }
+
+  // performance events
+  io.cpu.acquire := tl_out.a.fire()
+  io.cpu.release := tl_out.c.fire()
 }
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -505,6 +505,6 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   }
 
   // performance events
-  io.cpu.acquire := edge.last(tl_out.a)
-  io.cpu.release := edge.last(tl_out.c)
+  io.cpu.acquire := edge.done(tl_out.a)
+  io.cpu.release := edge.done(tl_out.c)
 }
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -505,6 +505,6 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   }
 
   // performance events
-  io.cpu.acquire := tl_out.a.fire()
-  io.cpu.release := tl_out.c.fire()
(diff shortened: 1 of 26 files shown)


src/main/scala/rocket/NBDcache.scala
```diff
@@ -973,4 +973,8 @@ class NonBlockingDCacheModule(outer: NonBlockingDCache) extends HellaCacheModule
   io.cpu.resp.bits.data_word_bypass := loadgen.wordData
   io.cpu.ordered := mshrs.io.fence_rdy && !s1_valid && !s2_valid
   io.cpu.replay_next := (s1_replay && s1_read) || mshrs.io.replay_next
+
+  // performance events
+  io.cpu.acquire := tl_out.a.fire()
```
hcook: I think this is also going to count IOMSHR requests. Intentional?

src/main/scala/rocket/NBDcache.scala
```diff
@@ -973,4 +973,8 @@ class NonBlockingDCacheModule(outer: NonBlockingDCache) extends HellaCacheModule
...
+  io.cpu.release := tl_out.c.fire()
```
hcook: This will count every beat of the released data individually. If you want to only count the total number of released blocks I'd make it `io.cpu.release := edge.last(tl_out.c)`

src/main/scala/rocket/DCache.scala
```diff
@@ -503,4 +503,8 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
       flushing := false
     }
   }
+
+  // performance events
+  io.cpu.acquire := tl_out.a.fire()
+  io.cpu.release := tl_out.c.fire()
```
hcook: As with NBDCache; you possibly want:
`io.cpu.acquire := edge.last(tl_out.a) && edge.opcode(tl_out.a) === TLMessages.Acquire`
`io.cpu.release := edge.last(tl_out.c)`

src/main/scala/rocket/NBDcache.scala
aswaterman: This one's intentional.

src/main/scala/rocket/NBDcache.scala
aswaterman: ok

src/main/scala/uncore/devices/Debug.scala
```diff
@@ -724,7 +724,7 @@ trait DebugModule extends Module with HasDebugModuleParameters with HasRegMap {
   // This logic assumes only up to 128 components.
   rdHaltnotStatus := Bits(0)
   for (ii <- 0 until numHaltnotStatus) {
-    when (dbReq.addr === UInt(ii)) {
+    when (dbReq.addr(1, 0) === UInt(ii)) {
```
mwachs5: Ah. Thanks. Will look for this in v13 branch too.

PR #580, opened 2017-03-10
TLB: add a helper API to determine homogeneous page permissions

terpstra:
@aswaterman Here's the method. I have not tested it in synthesis; let me know how it goes.


aswaterman:
Thanks. Manually merging into priv-1.10 branch.

PR #582, opened 2017-03-11
Fuzzer enhancements

terpstra:
Add delays and noise into more unit level component tests.

### DIFF
--- a/src/main/scala/uncore/ahb/Test.scala
+++ b/src/main/scala/uncore/ahb/Test.scala
@@ -25,7 +25,7 @@ class AHBFuzzBridge()(implicit p: Parameters) extends LazyModule
   val gpio  = LazyModule(new RRTest0(0x100))
 
   model.node := fuzz.node
-  xbar.node := TLToAHB()(model.node)
+  xbar.node := TLToAHB()(TLDelayer(0.1)(model.node))
   ram.node  := xbar.node
   gpio.node := xbar.node
 
--- a/src/main/scala/uncore/tilelink2/Fuzzer.scala
+++ b/src/main/scala/uncore/tilelink2/Fuzzer.scala
@@ -37,10 +37,11 @@ object LFSR64
   def apply(increment: Bool = Bool(true)): UInt =
   { 
     val wide = 64
-    val undef = Reg(UInt(width = wide)) // random value based on simulation seed
-    val lfsr = RegInit(Mux(undef === UInt(0), UInt(1), undef))
+    val lfsr = Reg(UInt(width = wide)) // random initial value based on simulation seed
     val xor = lfsr(0) ^ lfsr(1) ^ lfsr(3) ^ lfsr(4)
-    when (increment) { lfsr := Cat(xor, lfsr(wide-1,1)) }
+    when (increment) {
+      lfsr := Mux(lfsr === UInt(0), UInt(1), Cat(xor, lfsr(wide-1,1)))
+    }
     lfsr
   }
 }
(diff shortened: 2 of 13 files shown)


terpstra:
@scottj97 @jchang0 Hopefully this will make the AHB violation appear.

src/main/scala/uncore/tilelink2/Fuzzer.scala
```diff
@@ -34,16 +34,11 @@ class IDMapGenerator(numIds: Int) extends Module {
...
-    counter
-  }
-  
-  def apply(increment: Bool = Bool(true), seed: Int = next): UInt =
+  def apply(increment: Bool = Bool(true)): UInt =
   { 
     val wide = 64
-    val lfsr = RegInit(UInt((seed * 0xDEADBEEFCAFEBAB1L) >>> 1, width = wide))
+    val undef = Reg(UInt(width = wide)) // random value based on simulation seed
+    val lfsr = RegInit(Mux(undef === UInt(0), UInt(1), undef))
```
aswaterman: I like this approach for testing purposes.  Have you observed how it synthesizes?  My concern is that it will degenerate to

    val lfsr = RegInit(Mux(dontCare, UInt(1), undef))

and then to 

    val lfsr = RegInit(dontCare)

and will always and forever be zero, even though the simulation model disagrees.

src/main/scala/uncore/ahb/Test.scala
```diff
@@ -25,7 +25,7 @@ class AHBFuzzBridge()(implicit p: Parameters) extends LazyModule
   val gpio  = LazyModule(new RRTest0(0x100))
 
   model.node := fuzz.node
-  xbar.node := TLToAHB()(model.node)
+  xbar.node := TLToAHB()(TLDelayer(0.1)(model.node))
```
terpstra: That's the percentage chance that it delays a message. You probably want it to be specific to the test. Too many delays + too high => slow test.

aswaterman: ack

src/main/scala/uncore/tilelink2/Fuzzer.scala
terpstra: I've inspected the verilog. It does do what was intended:

''''
  `ifdef RANDOMIZE_REG_INIT
  _GEN_1 = {2{$random}};
  _T_7 = _GEN_1[63:0];
  `endif


    if (reset) begin
      if (_T_9) begin
        lfsrs_0 <= 64'h1;
      end else begin
        lfsrs_0 <= _T_7;
      end
''''

terpstra: Nevertheless, I am making a follow-up commit that is probably safer vs. optimization.

aswaterman: My specific concern is what design compiler will do with it, since _T_7 (hence _T_9) are uninitialized for the synthesis run.

terpstra: Who cares what Design Compiler does? This is only for the Fuzzer's LFSR. It's simulation only.

That said, I doubt it is legal to eliminate a logic loop that drives something, even if initialized?

aswaterman: If it's only meant to be used for the fuzzer, then make it a private object or hide it within some other namespace.

terpstra: The new version should be synthesizable also for DC.

PR #583, opened 2017-03-13
Don't connect unused ports

terpstra:
Turn off Monitors and connections for BCE channels that are unused.


terpstra:
I've realised that this is not the right thing to do. If you have a device that supports TL-C, you want to be able to connect it to a TL-UL device. It is legitimate for the TL-C device to expect b.valid false in this case; not floating. The goal should only be to float the .bits.X fields and keep valid low.

PR #584, opened 2017-03-14
AHB master port support

terpstra:
This PR adds the capability to master the TileLink bus from AHB.

### DIFF
--- a/src/main/scala/uncore/tilelink2/Buffer.scala
+++ b/src/main/scala/uncore/tilelink2/Buffer.scala
@@ -8,33 +8,60 @@ import config._
 import diplomacy._
 import scala.math.{min,max}
 
-// pipe is only used if a queue has depth = 1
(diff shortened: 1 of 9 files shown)


src/main/scala/uncore/tilelink2/Buffer.scala
```diff
@@ -8,33 +8,60 @@ import config._
 import diplomacy._
 import scala.math.{min,max}
 
-// pipe is only used if a queue has depth = 1
-class TLBuffer(a: Int = 2, b: Int = 2, c: Int = 2, d: Int = 2, e: Int = 2, pipe: Boolean = true)(implicit p: Parameters) extends LazyModule
+case class TLBufferConfig(depth: Int, flow: Boolean, pipe: Boolean)
```
hcook: Maybe TLBufferParams?

PR #587, opened 2017-03-16
ahb: rewrote TLToAHB to avoid retracting requests on stall

terpstra:


src/main/scala/uncore/tilelink2/ToAHB.scala
```diff
@@ -40,97 +57,140 @@ class TLToAHB(val combinational: Boolean = true)(implicit p: Parameters) extends
...
     }
   }
 }
 
 object TLToAHB
 {
   // applied to the TL source node; y.node := TLToAHB()(x.node)
-  def apply(combinational: Boolean = true)(x: TLOutwardNode)(implicit p: Parameters, sourceInfo: SourceInfo): AHBOutwardNode = {
-    val ahb = LazyModule(new TLToAHB(combinational))
+  def apply(a_pipe: Boolean = true)(x: TLOutwardNode)(implicit p: Parameters, sourceInfo: SourceInfo): AHBOutwardNode = {
```
hcook: can we make this aPipe or aFlow?

PR #597, opened 2017-03-20
Add option checks for HTIF

seldridge:
* Use new `htif.h` macros to add to the help text for the rocket-chip emulator showing what host options are supported. This is broken down into "emulator options", "host options", and "binary / target options".
* Move `filter_argv_for_dtm` checks inside the `htif_t` constructor
* In the event that a "legacy" `+` argument is found, use "normal" `--` processing logic for it

The logic here for option parsing is as follows: there exist three distinct sets of options that are expected to be provided to the emulator in order:
  * emulator options (e.g., `--seed`)
  * host (HTIF/DTM) options (e.g., `+signature`) 
  * target (RISC-V binary) options (e.g., `pk ...`)

Processing of command line arguments continues until the rocket-chip emulator sees the first HTIF argument or a non-argument (assumed to be the binary). Everything from the first HTIF argument (or the binary) is packages up into a `char**` along with the name of the emulator `argv[0]` and given to the `dtm_t` constructor.

This depends on https://github.com/riscv/riscv-fesvr/pull/25.

This should rectify the issues of #498 by making the rocket-chip emulator understand that `+signature` is valid.

Sample emulator help text:
```
Usage: ./emulator-freechips.rocketchip.system-DefaultConfig [EMULATOR OPTION]... [HOST OPTION]... BINARY [TARGET OPTION]...
Run a BINARY on the Rocket Chip emulator.

Mandatory arguments to long options are mandatory for short options too.

EMULATOR OPTIONS
  -c, --cycle-count        Print the cycle count before exiting
       +cycle-count
  -h, --help               Display this help and exit
  -m, --max-cycles=CYCLES  Kill the emulation after CYCLES
       +max-cycles=CYCLES
  -s, --seed=SEED          Use random number seed SEED
  -V, --verbose            Enable all Chisel printfs (cycle-by-cycle info)
       +verbose

EMULATOR OPTIONS (only supported in debug build -- try `make debug`)
  -v, --vcd=FILE,          Write vcd trace to FILE (or '-' for stdout)
  -x, --dump-start=CYCLE   Start VCD tracing at CYCLE
       +dump-start

HOST OPTIONS
      --rfb=DISPLAY        Add new remote frame buffer on display DISPLAY
       +rfb=DISPLAY          to be accessible on 5900 + DISPLAY (default = 0)
      --signature=FILE     Write torture test signature to FILE
       +signature=FILE
      --chroot=PATH        Use PATH as location of syscall-servicing binaries
       +chroot=PATH

HOST OPTIONS (currently unsupported)
      --disk=DISK          Add DISK device. Use a ramdisk since this isn't
       +disk=DISK            supported

TARGET (RISC-V BINARY) OPTIONS
  These are the options passed to the program executing on the emulated RISC-V
  microprocessor.

EXAMPLES
  - run a bare metal test:
    ./emulator-freechips.rocketchip.system-DefaultConfig $RISCV/riscv64-unknown-elf/share/riscv-tests/isa/rv64ui-p-add
  - run a bare metal test showing cycle-by-cycle information:
    ./emulator-freechips.rocketchip.system-DefaultConfig +verbose $RISCV/riscv64-unknown-elf/share/riscv-tests/isa/rv64ui-p-add 2>&1 | spike-dasm
  - run an ELF (you wrote, called 'hello') using the proxy kernel:
    ./emulator-freechips.rocketchip.system-DefaultConfig pk hello
```

### DIFF
--- a/csrc/SimDTM.cc
+++ b/csrc/SimDTM.cc
@@ -6,27 +6,6 @@
 
 dtm_t* dtm;
 
-
-namespace {
-
-  // Remove args that will confuse dtm, such as those that require two tokens, like VCS code coverage "-cm line+cond"
-std::vector<std::string> filter_argv_for_dtm(int argc, char** argv)
-{
-  std::vector<std::string> out;
-  for (int i = 1; i < argc; ++i) { // start with 1 to skip my executable name
-    if (!strncmp(argv[i], "-cm", 3)) {
-      ++i; // skip this one and the next one
-    }
-    else {
-      out.push_back(argv[i]);
-    }
-  }
-  return out;
-}
-
-}
-
-
 extern "C" int debug_tick
 (
   unsigned char* debug_req_valid,
@@ -44,7 +23,7 @@ extern "C" int debug_tick
     s_vpi_vlog_info info;
     if (!vpi_get_vlog_info(&info))
       abort();
-      dtm = new dtm_t(filter_argv_for_dtm(info.argc, info.argv));
+      dtm = new dtm_t(info.argc, info.argv);
   }
 
   dtm_t::resp resp_bits;
--- a/csrc/emulator.cc
+++ b/csrc/emulator.cc
@@ -34,33 +34,39 @@ extern "C" int vpi_get_vlog_info(void* arg)
   return 0;
 }
 
-static void usage(const char * program_name) {
-  printf("Usage: %s [OPTION]... BINARY [BINARY ARGS]\n", program_name);
+static void usage(const char * program_name)
+{
+  printf("Usage: %s [EMULATOR OPTION]... [HOST OPTION]... BINARY [TARGET OPTION]...\n",
+         program_name);
   fputs("\
 Run a BINARY on the Rocket Chip emulator.\n\
 \n\
 Mandatory arguments to long options are mandatory for short options too.\n\
-  -c, --cycle-count          print the cycle count before exiting\n\
+\n\
+EMULATOR OPTIONS\n\
+  -c, --cycle-count          Print the cycle count before exiting\n\
        +cycle-count\n\
-  -h, --help                 display this help and exit\n\
-  -m, --max-cycles=CYCLES    kill the emulation after CYCLES\n\
-       +max-cycles=CYCLES\n\
-  -s, --seed=SEED            use random number seed SEED\n\
-  -V, --verbose              enable all Chisel printfs\n\
+  -h, --help                 Display this help and exit\n\
+  -m, --max-cycles=[CYCLES]  Kill the emulation after CYCLES\n\
+       +max-cycles=[CYCLES]\n\
+  -s, --seed=[SEED]          Use random number seed SEED\n\
+  -V, --verbose              Enable all Chisel printfs\n\
        +verbose\n\
 ", stdout);
 #if VM_TRACE
   fputs("\
-  -v, --vcd=FILE,            write vcd trace to FILE (or '-' for stdout)\n\
-  -x, --dump-start=CYCLE     start VCD tracing at CYCLE\n\
+  -v, --vcd=[FILE],          write vcd trace to FILE (or '-' for stdout)\n\
+  -x, --dump-start=[CYCLE]   start VCD tracing at CYCLE\n\
       +dump-start\n\
+\n\
+EMULATOR OPTIONS (unsupported in non-debug build -- try `make debug`)\n\
 ", stdout);
-#else
+#endif
   fputs("\
 VCD options (e.g., -v, +dump-start) require a debug-enabled emulator.\n\
 Try `make debug`.\n\
 ", stdout);
-#endif
+  fputs(HTIF_USAGE_OPTIONS, stdout);
 }
 
 int main(int argc, char** argv)
@@ -73,8 +79,8 @@ int main(int argc, char** argv)
   FILE * vcdfile = NULL;
   uint64_t start = 0;
 #endif
+  char ** htif_argv = NULL;
 
-  std::vector<std::string> to_dtm;
   while (1) {
     static struct option long_options[] = {
       {"cycle-count", no_argument,       0, 'c' },
@@ -86,7 +92,7 @@ int main(int argc, char** argv)
       {"vcd",         required_argument, 0, 'v' },
       {"dump-start",  required_argument, 0, 'x' },
 #endif
-      {0, 0, 0, 0}
+      HTIF_LONG_OPTIONS
     };
     int option_index = 0;
 #if VM_TRACE
@@ -95,8 +101,9 @@ int main(int argc, char** argv)
     int c = getopt_long(argc, argv, "-chm:s:V", long_options, &option_index);
 #endif
     if (c == -1) break;
+ retry:
     switch (c) {
-      // Process "normal" options with '--' long options or '-' short options
+      // Process long and short EMULATOR options
       case '?': usage(argv[0]);             return 1;
       case 'c': print_cycles = true;        break;
       case 'h': usage(argv[0]);             return 0;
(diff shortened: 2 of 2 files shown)


aswaterman:
The sample output looks great.  I'm under the gun this week, but I plan to review this and the corresponding fesvr PR soon, probably once the priv-1.10 stuff is merged into master.

seldridge:
@aswaterman -- all good! This is, admittedly, low priority. I'm open to suggestions if you have some thoughts on a better or unified strategy to handle the argument parsing.

seldridge:
@aswaterman: FWIW, I revived this---I've been fielding some questions about what the emulator supports, how your run the emulator, etc. and I figured this was worth revisiting.

This will naturally not pass regressions until riscv/riscv-fesvr#25 is merged and riscv-tools is bumped.

aswaterman:
whoops, sorry for letting this fall by the wayside - I'll look at the fesvr PR tomorrow.

edcote:
@seldridge @aswaterman @mwachs5 

This "low priority" issue caused me pain this morning.  Thus the ping.  It would appear that csrc/emulator.cc has broken against riscv-fesvr master for some time.

Please re-review this pull request and communicate any blocking items.

seldridge:
This includes a bump to the newer riscv-tools that includes the `htif_t` API change. Since this is a riscv-tools bump, this will need an admin to hold Travis' hand while it does a cache refresh.

I have not locally tested this riscv-tools bump.

Also, I keep rebasing. If this is ill advised, tell me.

In my understanding nothing is broken here, just that rocket-chip master is currently incompatible with riscv-tools master. I view this as being an allowable state so long as rocket-chip's master riscv-tools submodule is still consistent (which it is).

mwachs5:
This failed because it ran out of space on the device while building riscv-tools. Not sure if this was due to an excessive amount of errors, or if you are increasing the tools that you are downloading and building. Thoughts @seldridge?

seldridge:
That's referencing a `riscv-tools` that passed regressions (https://github.com/riscv/riscv-tools/pull/158), so I had assumed that this was just it running out of space and not something more nefarious. I was actually hoping that you might have some thoughts here... as I expected that Travis has done this before with previous riscv-tools bumps to rocket-chip.

I am not aware of either an excessive amount of errors or any new tools that are being downloaded here.

mwachs5:
well, worth a retry at least.

seldridge:
Thanks, Megan! Your restart prevents me from having to superficially modify files to restart the build.

It looks like it got past the point of building riscv-tools. Travis did, however, run out of space when actually packing everything up for the cache. That may be problematic.

`run-asm-tests` passes on my end, so I don't anticipate any failures for the rest of this.

seldridge:
These appear to be purely timeout failures due to individual tests rebuilding riscv-tools each time. I'm not sure if this is the domain of [README_TRAVIS.md](../blob/master/README_TRAVIS.md) or if this is related to the fact that [Travis appeared to be running out of disks space while it was caching](https://travis-ci.org/freechipsproject/rocket-chip/jobs/318217491#L1691).

mwachs5:
Hmm, strange. I haven't seen this before. Do you now how much isolation Travis jobs have from eachother? If seperate PRs are all banging on it with disk space hogs (I know my Debug Regression PR is a disk space hog at the moment) is that cumulative? Or does each PR get its own disk space?

seldridge:
Operating under the assumption that it is a problem with that bad cache during the riscv-tools build and that's causing the tests to rebuild everything... I found one match: https://github.com/travis-ci/travis-ci/issues/8638

Possibly clearing the cache? Maybe you all can check to see how big the cache has gotten and if that's the issue. Granted, clearing the cache may just cause this problem to migrate to other builds.

mwachs5:
let me just delete the cache for this PR and restart it (and also my debug PR should not be creating such massive logs anymore...)

mwachs5:
most caches are 500MB. This PR's was only 21MB, presumably because the large part of it failed to upload.

mwachs5:
I don't know why this keeps failing. @seldridge can you try comparing the size of the directory and/or cache when you build this locally vs on master?

mwachs5:
Also let me delete all the caches that don't correspond to PRs right now

mwachs5:
I deleted a bunch of caches. Going to hit retry again...

mwachs5:
Looks like this still ran out of space when trying to upload the cache, then the tests timed out rebuilding it.

seldridge:
Sorry, drawn off on other things. Note: this is "building locally" in the sense that I'm running it on my RHEL 6.9 laptop. I never had success getting a Travis VM running on this.

```
> cd rocket-chip
> du -hc --max-depth=1 regression/
32K     regression/stamps
852M    regression/install
852M    regression/
852M    total
```

I'm unable to find documentation on allowable Travis cache sizes.

mwachs5:
Well the "good" news is that #1088 is failing for the same reason now when I bumped riscv-tools to master. So at least we're in the same boat.

mwachs5:
my experiments seem to indicate that it has nothing to do with the version of riscv-tools, just the fact that you can't use the master cache. I think anyone with a PR that wants to update riscv-tools is going to (suddenly) have this problem. I'm not sure why.

mwachs5:
Apparently Travis updated their Trusty images Dec 12. Not sure if this has anything to do with it, but in #1088  I am trying out the fix suggested here: https://blog.travis-ci.com/2017-12-12-new-trusty-images-q4-launch

I also emailed Travis support for help in figuring out what's going on, but understandably things may be delayed until 2018.

mwachs5:
So, I was able to get riscv-tools to compile using the old trusty image. I also emailed with Travis support and they suggested we just try using the `sudo:required` as apparently that will give us more space. If you don't mind, I will push that change to this PR.

mwachs5:
Oh, realized I can't because it's your fork. I'll give it a try on #1165

mwachs5:
You could try cherry-picking 255448d1329ca5017153e3db89a53952fc4ef212 (Note that this also bumps riscv-tools to current master). I also pulled this PR into #1088 to try to make forward progress there.

mwachs5:
@seldridge I ran into problems trying to use this. How is one supposed to pass arbitrary plusargs into the verilog? For example, rocket supports `+max-core-cycles=#`. This is interpreted by the core itself: https://github.com/freechipsproject/rocket-chip/blob/cfa819fc5867e18ba105b16f804eca09a6c60ca4/src/main/scala/rocket/RocketCore.scala#L725

But this no longer works:
```
/scratch/megan/rocket-chip/emulator/emulator-freechips.rocketchip.system-WithJtagDTMSystem_DefaultRV32Config +verbose +max-core-cycles=100 dummybin
/scratch/megan/rocket-chip/emulator/emulator-freechips.rocketchip.system-WithJtagDTMSystem_DefaultRV32Config: invalid HTIF legacy plus-arg "+max-core-cycles=100"
```

There could be any number of plusargs that are understood by the Verilog and don't need to be interpreted by emulator.cc, correct? How does this PR address that?

seldridge:
Thanks for beating Travis into submission and getting this tested.

I was not at all aware of the arbitrary plus arg reader/writer from Rocket, but I was worried about something like this...

The correct way, in my view, would be to have PlusArgs do what HTIF does: provide a header that defines what arguments are supported, like: https://github.com/riscv/riscv-fesvr/blob/master/fesvr/htif.h#L95. Technically it would need to be different from this as plusargs have no meaning sit in that getop long option struct. But, generally, dump a struct of allowable Verilog-consumed plusargs and whether or not they need arguments. This would then enable `emulator.cc` to (finally) parse whatever PlusArgs that anyone dumps into Rocket.

I can get something together that does this, but it may not be until next week. I think I can just have a file writer that latches onto Chisel's build dir to dump the appropriate header in there.

Apologies for the far-reaching nature of this... though, I still think the goal is sane: show all supported arguments to an arbitrary rocket SoC emulator.

mwachs5:
That's fine. In the short term, we can just look for PlusArg in the rocket-chip code base and just hard code them as as supported. It would be fancy pants if it was automatically generated from the ElaborationArtefacts... will think about it.

mwachs5:
Ok 62818fff8a977704dd7757128bf9a26c1fde8e16 was my hacky attempt to resolve this for now.

mwachs5:
Included in #1179

seldridge:
Thanks for merging this from your tree @mwachs5. 

I looked into it a little more and I think the correct way to do this is to make `PlusArgs` generate `ElaborationArtefacts`. Or, likely better, use a singleton that will build up all the available plus args and then emit these along with other artefacts in `BaseCoreplexModule`. This seems almost comically simple... (and I had always assumed that you all had some infrastructure to hook into the elaboration). I can take a stab at it in a new PR (likely slowly) unless there's some big rush.

mwachs5:
Yeah some sort of JSON-emitted thing fits the pattern. I don't think there is a big rush though!

PR #603, opened 2017-03-22
rocket: avoid LinkedHashMap.keys to preserve traversal order

hcook:
Don't trust LinkedHashMap.keys.map to provide a consistent order of traversal across module instantiations.

Even though the documentation says:
"The iterator and all traversal methods of this class visit elements in the order they were inserted."

...Scala!

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -3,6 +3,8 @@
 
 package rocket
 
+import collection.mutable.LinkedHashMap
+
 import Chisel._
 import Instructions._
 import config._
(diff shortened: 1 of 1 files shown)


jackkoenig:
[I feel your pain on this one](https://github.com/ucb-bar/firrtl/pull/303) 👎

sdtwigg:
Ew, so I looked into this and it looks like there is already a Scala issue filed on it (and potentially slated to be fixed in the far-away 2.13 release)

https://issues.scala-lang.org/browse/SI-9594

sdtwigg:
PS: It should still be safe as long as you are calling just map, flatMap, filter, etc. (or in a for-comprehension).

aswaterman:
Yeah, our workaround was to map over the whole map, rather than just over the keys.

PR #605, opened 2017-03-23
Privileged architecture 1.10 work-in-progress: PMPs, SFENCE, etc.

aswaterman:
And some miscellaneous microarchitectural changes made along the way.


src/main/scala/rocket/Instructions.scala
```diff
@@ -302,6 +302,26 @@ object CSRs {
...
+  val pmpaddr6 = 0x3b6
+  val pmpaddr7 = 0x3b7
+  val pmpaddr8 = 0x3b8
+  val pmpaddr9 = 0x3b9
+  val pmpaddr10 = 0x3ba
+  val pmpaddr11 = 0x3bb
+  val pmpaddr12 = 0x3bc
+  val pmpaddr13 = 0x3bd
+  val pmpaddr14 = 0x3be
+  val pmpaddr15 = 0x3bf
```
aswaterman: We can add more later if we are so inclined, just at the cost of CSR space.

PR #611, opened 2017-03-25
rename l2FrontendBus as fsb, expose bsb

yunsup:


terpstra:
Also, the bsb Buffer inside the Coreplex should use BufferParams.flow for the ACE channels and BufferParams.none for the BD channels. While you are at it, please also adjust the TLBuffer connecting the L2 in CoreplexNetwork:133 to use flow only for ACE and none for BD.

PR #614, opened 2017-03-27
TLROM: allow name and compatibility strings to be provided by subclasses

mwachs5:
This allows TLROM devices to specify a name and compatibility string for the Config String other than "rom"


mwachs5:
Ok, then is it correct to just assign `val device = resource`

mwachs5:
Ah, I missed the `.reg`. You mean to just pass it directly to the TLManagerNode as `resources`.

terpstra:
To your second comment, yes. I think .reg is a Seq[Resource], tho, so accept 'resources' not 'resource'.

PR #616, opened 2017-03-27
RocketTile: Create a wrapper for SyncRocketTile as well

mwachs5:
There is no guarantee that debugInterrupt is synchronous
to tlClk, even though it is true in the current implementation.
It will not be true in future implementations, as decoupling
this allows the debugInterrupt to be asserted across tlClk
gating/reset scenarios.

Therefore, even for SyncRocketTile, the debug interrupt needs to be
synchronized to coreClk, and for RationalRocketTile, 1 cycle
of synchronization is not sufficient.

Even though other interrupts may be synchronized already, we just
synchronize them all to simplify the code at the expense of
a few cycles latency.

It could still be nice to use a parameter vs hard coding "3".


terpstra:
Presumably there is a commit where you change coreplex/RocketTiles.scala?

mwachs5:
Hmm, I was trying to find where this was actually used on my branch and was wondering why I didn't find it. Now I know a bit better what I am looking for let me add that to this PR

mwachs5:
Ok, I didn't actually even have this on my branch so let me know if this looks right.

terpstra:
Seems right. How were you using it if you didn't have this?

mwachs5:
It just wasn't getting used (on the debug_v013 branch I had changed debugInterrupt to use diplomacy interrupts, but didn't actually change this line). 

Worked in sim because sims don't verify CDC very well.

PR #617, opened 2017-03-27
bootrom: also setup SBI a0+a1 for when we hang

terpstra:
This way a program started by the debugger still has the DTB handy.


mwachs5:
I'm not sure what you mean by a program started by the debugger, but fine by me.

terpstra:
@mwachs5 if it boots up hung, waiting for a debugger, when you attach the debugger to start a program, this way a1 already has the DTB address for that program. also, you can read it out from the debugger on power-on.

mwachs5:
OK. I was going to ask if the value _dtb is somehow known at Chisel elaboration time (to pass to Debug Module anyway).

terpstra:
Hmm. At the moment it comes the software bootloader you compile. However, I can see why you would want it at chisel elaboration. I'm not sure best how to do that.

mwachs5:
So does _dtb change, based on other stuff in the bootrom? Or do you pick the value (I thought previously there was a standardization of the reset vector, pointer to dtb, etc... is that idea still there?)

Anyway none of this is meant to gate the PR, we can discuss tomorrow

terpstra:
The location of the DTB is just defined by the SBI; a pointer is in a1 when your bootloader starts. I put it at a relatively fixed location in rocketchip, but that's not any sort of standard.

PR #620, opened 2017-03-27
Add Debug V13 Debug ROM

mwachs5:
This PR doesn't change any functionality yet, just adds scripts to generate new code.

I'm trying to break up anything I can from the massive Debug v13 PR. 

This just adds the v13 ROM contents and a script to compile them to DebugRomContents.scala.

Is `scripts` directory an appropriate place for this?


scripts/debug_rom/Makefile
```diff
@@ -0,0 +1,31 @@
+# See LICENSE.SiFive for license details
+# Recursive make is bad, but in this case we're cross compiling which is a
+# pretty unusual use case.
+
+CC = $(RISCV)/bin/riscv64-unknown-elf-gcc
+OBJCOPY = $(RISCV)/bin/riscv64-unknown-elf-objcopy
```
palmer-dabbelt: IIRC we still build riscv32-unknown-elf-gcc for 32-bit targets?  If so, then this should use XLEN.

mwachs5: I don't think we do actually...

Anyway, the compiled contents *must* be XLEN-independent for the DebugModule to work (I did check that). This script is meant to be run once and then the code checked in, not as part of our flow.

mwachs5:
I'm going to turn this into the overall Debug v13 PR.

PR #623, opened 2017-03-28
Create README_TRAVIS.md

mwachs5:


aswaterman:
Incorporating into #622

PR #624, opened 2017-03-28
Debug v013 [WIP]

mwachs5:
This PR implements the v13 version of the RISC-V Debug Spec, with a few known incompatibilities as listed below. 

Before this PR can pass, riscv-fesvr must be updated to use a slightly modified protocol. I will file an issue there with some thoughts.

Known issues even once this PR is passing:
* ~~ABSTRACTCS.cmderr is speced to be write-1-to-clear. It is currently write-0-to-clear because that is what the latest version of OpenOCD is doing.~~
* There should be a mapping between HARTID and DMCONTROL.hartsel. Right now it is assumed 1:1. 
* AccessRegisterCommand.transfer bit is ignored as OpenOCD is not currently setting it.
* Some registers in DCSR are no longer in the spec (nb debugInterrupt). 
* Some top level signals are not exported all the way out using the Periphery traits (ndreset, dmactive, and unavail)


mwachs5:
This is failing because it's too slow and timing out. Will look into tightening up some inefficiencies.

src/main/scala/rocket/CSR.scala
```diff
@@ -58,9 +57,9 @@ class DCSR extends Bundle {
   val stopcycle = Bool()
   val stoptime = Bool()
   val cause = UInt(width = 3)
+  // TODO: debugint is not in the Debug Spec v13
```
aswaterman: As in, debugint should now be wired to 0?  I can fix that once this is merged.

src/main/scala/uncore/tilelink2/Parameters.scala
```diff
@@ -56,7 +56,7 @@ case class TLManagerParameters(
 
   // The device had better not support a transfer larger than it's alignment
   val minAlignment = address.map(_.alignment).min
-  require (minAlignment >= maxTransfer)
+  require (minAlignment >= maxTransfer, "minAlignment (" + minAlignment + ") must be >= maxTransfer (" + maxTransfer + ")")
```
aswaterman: For future ref, these kind of strings can be less annoyingly written with string interpolation, by writing an ```s``` before the open-quotes, as in

    require (minAlignment >= maxTransfer, s"minAlignment ($minAlignment) must be >= maxTransfer ($maxTransfer)")

src/main/scala/rocket/CSR.scala
mwachs5: Yeah. I had removed it, then put it back because I had a thought of how to use it in the Debug ROM, then realized that idea wouldn't work... so I left it in rather than removing it again in case I thought of another great idea for it. But I think the idea if we did want to use it in the ROM would be to put it in a different implementation specific CSR.

aswaterman: ack

mwachs5: Is that ack as in syn-ack or ack as in OMG no more CSRs Plz

aswaterman: hahaha, the former

mwachs5:
Think this is just still too slow. I profiled and 2/3 of the runtime is loading the program, so I'm going to make the write_chunk in dtm.cc more efficient

hcook:
i can bump the timeout ifneedbe

mwachs5:
ok well the latest PR should cut the program load time down significantly, if that doesn't work then we can bump the timeout.

PR #627, opened 2017-03-28
Fix MMIO/cache refill concurrency bug in DCache

aswaterman:
There's a structural hazard on s2_req, so disallow cache refill initiation while any MMIO loads are in flight.

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -167,7 +167,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   val (s2_hit, s2_grow_param, s2_new_hit_state) = s2_hit_state.onAccess(s2_req.cmd)
   val s2_valid_hit = s2_valid_masked && s2_readwrite && s2_hit
   val s2_valid_miss = s2_valid_masked && s2_readwrite && !s2_hit && !(pstore1_valid || pstore2_valid) && !release_ack_wait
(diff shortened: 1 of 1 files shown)


aswaterman:
@sols1

ccelio:
What does this failure look like? A dropped store? A load that never returns data?

sols1:
Back-to-back uncached and cached ops cause structural hazard:

rv64ud-v-fsgnj_4

sols1:
test runs correctly after #https://github.com/ucb-bar/rocket-chip/pull/627/commits/4452d6c50cac9d99d933d853dff6b1157c35be9d

ccelio:
Thanks @sols1. I'm on a different branch of rocket-chip -- so sorry for being obtuse --  I'm wondering how this bug would manifest from the point-of-view of the software/user.  Am I understanding correctly that the cached refill request is overwritten by the MMIO request on leaving the dcache? So a cached load (that misses in the L1) would never return if following a cycle behind an MMIO?

hcook:
@ccelio probably was introduced/enabled to occur in #605 just 4 days ago. Most likely outcome is a load returning data for the wrong address after refilling into the cache.

ccelio:
@hcook Ah okay, thanks.

sols1:
AFAIU, the test program is loaded through a debug backdoor into dcache and then dcache is flushed to propagate changes to the main memory. B/c of this bug 0 is written flushed from dcache instead of instruction. When core starts executing, it gets illegal instruction exception.

PR #629, opened 2017-03-29
Pass MODEL variable to emulator.cc

seldridge:
This enables hot-swapping of the top-level test harness by specifying
`MODEL=MyTestHarness` when building the emulator.


aswaterman:
Folding this into another PR to save on travis bandwidth.

PR #630, opened 2017-03-30
Interrupt wiring

hcook:
Get per-tile local interrupts out of the tile and coreplex.

Need to do further examination of InterruptXing.


aswaterman:
I'm merging this into my next PR

PR #632, opened 2017-03-30
Do not allow make to remove .vpd files on Ctrl-C

solomatnikov:


aswaterman:
@solomatnikov thanks, I am cherry-picking this commit into my PR to save on travis bandwidth

PR #633, opened 2017-03-30
DTS linux

terpstra:
This makes the DTS format more compatible with stock linux.


aswaterman:
lgtm, will merge into my next PR

PR #634, opened 2017-03-30
New PMP encoding + potpourri

aswaterman:
The main purpose of this PR is to represent the new PMP encoding, and to fix one minor bug (the wrong exception cause was reported when page-table walks failed a PMP check).

This PR incorporates a few unrelated changes to help clean the Travis pipes:

- DTS changes (@terpstra)
- Local interrupt wiring in the coreplex (@hcook)
- Don't delete VPDs after SIGINT (@solomatnikov)
- Bump Firrtl (@jackkoenig)

### DIFF
--- a/src/main/scala/tile/Interrupts.scala
+++ b/src/main/scala/tile/Interrupts.scala
@@ -45,7 +45,7 @@ trait HasExternalInterruptsModule {
 
   // go from flat diplomatic Interrupts to bundled TileInterrupts
   def decodeCoreInterrupts(core: TileInterrupts) {
-    val core_ips = Vec(
+    val core_ips = Seq(
       core.debug,
       core.msip,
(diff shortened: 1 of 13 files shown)


aswaterman:
BBL doesn't currently delegate them to S-mode, so Linux won't even know they exist.

src/main/scala/tile/Interrupts.scala
```diff
@@ -45,7 +45,7 @@ trait HasExternalInterruptsModule {
 
   // go from flat diplomatic Interrupts to bundled TileInterrupts
   def decodeCoreInterrupts(core: TileInterrupts) {
-    val core_ips = Vec(
+    val core_ips = Seq(
```
hcook: :'(

PR #635, opened 2017-03-30
Added --trace-underscore to make verilator generated waveform more useful

solomatnikov:


colinschmidt:
I believe this was an intentional omission to improve verilator performance. That being said it does seem reasonable to allow this if the user really wants _all_ signals. Perhaps an option?

solomatnikov:
It can be an option.

However, I am sure I understand the point of waveform, missing most of the signals in the design (?). A few days ago I was to trying to understand TLMonitor assertions and could not see anything useful in the waveform.

jackkoenig:
@colinschmidt is right, we intentionally changed the way temporaries are emitted so that they would be omitted without this option and thus make Verilator debug much more efficient.

@solomatnikov The **hope** is that you should be able to debug based only on the named signals. This is obviously problematic for signals that don't get names because they are defined within a function. This is what @chiselName was intended to fix, but rocket-chip does not use it yet. Might be time to start.

PR #646, opened 2017-04-03
I$ improvements

aswaterman:
- Revive I$ parity option; pipeline it into the ID stage to keep it off SRAM critical path
- Get I$ s1_kill signal off the critical path
- Add a 1-entry flow queue between I$ and pipeline, cutting I$ -> decode -> stall signal -> I$ path

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -68,7 +68,8 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
   val stall = !io.resp.ready
 
   val refill_addr = Reg(UInt(width = paddrBits))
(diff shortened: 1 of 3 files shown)


aswaterman:
@palmer-dabbelt can you verify this doesn't affect benchmark results?

palmer-dabbelt:
@aswaterman I'm running regressions.

aswaterman:
This still boots Linux.

palmer-dabbelt:
I think something is probably just screwed up on my end, but I won't have a chance to look at it today.

aswaterman:
This doesn't affect Dhrystone.  Working on test failures now.  @hcook CR when you get a chance?

PR #647, opened 2017-04-03
debug: Put DebugROM back inside the overall Debug Module

mwachs5:
DebugROM was pulled out of DebugModule for CDC reasons which no longer apply. Unify these modules again and update the DTS description of DebugModule.


mwachs5:
@terpstra rubber stamp?

PR #648, opened 2017-04-03
Only set id_reg_fence when AMO/FENCE is actually executed

aswaterman:
This is a performance bug, not a correctness bug.  But randomly stalling
because of garbage bits coming out of the I$ should be avoided.

h/t @solomatnikov

### DIFF
--- a/src/main/scala/rocket/Rocket.scala
+++ b/src/main/scala/rocket/Rocket.scala
@@ -213,10 +213,10 @@ class Rocket(implicit p: Parameters) extends CoreModule()(p)
   val id_amo_rl = id_inst(0)(25)
   val id_fence_next = id_ctrl.fence || id_ctrl.amo && id_amo_rl
   val id_mem_busy = !io.dmem.ordered || io.dmem.req.valid
+  when (!id_mem_busy) { id_reg_fence := false }
   val id_rocc_busy = Bool(usingRoCC) &&
     (io.rocc.busy || ex_reg_valid && ex_ctrl.rocc ||
      mem_reg_valid && mem_ctrl.rocc || wb_reg_valid && wb_ctrl.rocc)
-  id_reg_fence := id_fence_next || id_reg_fence && id_mem_busy
   val id_do_fence = id_rocc_busy && id_ctrl.fence ||
     id_mem_busy && (id_ctrl.amo && id_amo_aq || id_ctrl.fence_i || id_reg_fence && (id_ctrl.mem || id_ctrl.rocc))
 
@@ -294,6 +294,7 @@ class Rocket(implicit p: Parameters) extends CoreModule()(p)
     ex_ctrl := id_ctrl
     ex_reg_rvc := ibuf.io.inst(0).bits.rvc
     ex_ctrl.csr := id_csr
+    when (id_fence_next) { id_reg_fence := true }
     when (id_xcpt) { // pass PC down ALU writeback pipeline for badaddr
       ex_ctrl.alu_fn := ALU.FN_ADD
       ex_ctrl.alu_dw := DW_XPR


src/main/scala/rocket/Rocket.scala
```diff
@@ -213,10 +213,10 @@ class Rocket(implicit p: Parameters) extends CoreModule()(p)
   val id_amo_rl = id_inst(0)(25)
   val id_fence_next = id_ctrl.fence || id_ctrl.amo && id_amo_rl
   val id_mem_busy = !io.dmem.ordered || io.dmem.req.valid
+  when (!id_mem_busy) { id_reg_fence := false }
   val id_rocc_busy = Bool(usingRoCC) &&
     (io.rocc.busy || ex_reg_valid && ex_ctrl.rocc ||
      mem_reg_valid && mem_ctrl.rocc || wb_reg_valid && wb_ctrl.rocc)
-  id_reg_fence := id_fence_next || id_reg_fence && id_mem_busy
   val id_do_fence = id_rocc_busy && id_ctrl.fence ||
     id_mem_busy && (id_ctrl.amo && id_amo_aq || id_ctrl.fence_i || id_reg_fence && (id_ctrl.mem || id_ctrl.rocc))
```
solomatnikov: Is `id_ctrl.amo` (like `id_ctrl.mem`) asserted when there is a valid instruction in decode stage?

aswaterman: All the id_ctrl signals are don't-cares when the instruction's invalid.

solomatnikov: What's the right logical expression to figure out if fence should happen or not?

`id_do_fence` does not exist in generated Verilog.

From the expression above it seems fence happens when `id_ctrl.amo && id_amo_aq == 1`, however, both are don't-cares when the instruction's invalid.

aswaterman: The performance of illegal instruction exceptions doesn't really matter, so it's OK to stall the pipeline in the ```!id_ctrl.valid``` case.

This is distinct from the I$-provided-no-instruction case, which can happen on I$ miss or even branch misprediction.  Branch misprediction causing fences is much worse :)

solomatnikov: I didn't write the question properly - of course, if illegal instruction is executed and exception is taken, then stalls don't matter.

What I am trying to figure out is when fence is supposed to happen according to `aq` and `rl` modifiers.

Here is a waveform before this PR:
id_ctrl_amo and id_ctrl_mem

At C1 there is no illegal instruction exception, `id_ctrl.legal == 0`, `id_ctrl.mem == 0`, however, `id_ctrl.amo == 1`, `id_amo_aq == 1`, `id_amo_rl == 1` and `id_fence_next == 1`. And if `id_mem_busy` is asserted, then fence should happen according to the logic above.

aswaterman:
Ah, I see.  Yeah, ```id_amo_rl && id_mem_busy``` should induce a stall.  On the other hand, ```id_amo_rl``` should only matter if the instruction enters the next pipeline stage, causing ```id_reg_fence``` to be set.

solomatnikov:
Could you make `id_do_fence` a wire as @jackkoenig suggested to preserve it in generated Verilog?

Then, I can connect to it directly w/o re-implementing the logic.

aswaterman:
If the following patch works for you, feel free to apply it and pull request it:

diff --git a/src/main/scala/rocket/Rocket.scala b/src/main/scala/rocket/Rocket.scala
index 98e527d..d860e6f 100644
--- a/src/main/scala/rocket/Rocket.scala
+++ b/src/main/scala/rocket/Rocket.scala
@@ -217,8 +217,8 @@ class Rocket(implicit p: Parameters) extends CoreModule()(p)
   val id_rocc_busy = Bool(usingRoCC) &&
     (io.rocc.busy || ex_reg_valid && ex_ctrl.rocc ||
      mem_reg_valid && mem_ctrl.rocc || wb_reg_valid && wb_ctrl.rocc)
-  val id_do_fence = id_rocc_busy && id_ctrl.fence ||
-    id_mem_busy && (id_ctrl.amo && id_amo_aq || id_ctrl.fence_i || id_reg_fence && (id_ctrl.mem || id_ctrl.rocc))
+  val id_do_fence = Wire(init = id_rocc_busy && id_ctrl.fence ||
+    id_mem_busy && (id_ctrl.amo && id_amo_aq || id_ctrl.fence_i || id_reg_fence && (id_ctrl.mem || id_ctrl.rocc)))
 
   val bpu = Module(new BreakpointUnit(nBreakpoints))
   bpu.io.status := csr.io.status

src/main/scala/rocket/Rocket.scala
solomatnikov: The logic still doesn't look correct (from perf point of view):

id_reg_fence

At M1 there is an AMO w/ rl in the decode stage and id_reg_fence is set in the next cycle.

At M2 there is a regular mem op in the decode stage but id_reg_fence is still set.

aswaterman: That looks right to me.  Once id_reg_fence is high, it should remain high until id_mem_busy falls - so that the AMO.RL is guaranteed to be visible before the regular mem op.

solomatnikov:
I think it would be easier if aq/rl modifiers are passed to DCache along w/ cmd, even if DCache does need them. The scoreboard can pick them up when the memory operation is actually executed, i.e. after s1_kill and s2_nack.

src/main/scala/rocket/Rocket.scala
solomatnikov: In this case AMO.RL is squashed in the decode stage and another instruction is fetched in M2

aswaterman:
Yeah, that makes sense.  I will do that soon, along with addressing the performance issue.

solomatnikov:
Also, for the same reasons it would be useful to send fence instruction itself even if it is a complete nop from DCache point of view.

PR #654, opened 2017-04-06
Handle single-step with a pipeline stall, not a flush

aswaterman:
The pipeline flush approach broke when I changed the pipeline stage
the flush happens from


mwachs5:
will test this shortly

PR #655, opened 2017-04-06
debug: Make it easier to override parts of the Default Debug Config

mwachs5:
Still learning Scala...

### DIFF
--- a/src/main/scala/coreplex/Configs.scala
+++ b/src/main/scala/coreplex/Configs.scala
@@ -22,7 +22,7 @@ class BaseCoreplexConfig extends Config ((site, here, up) => {
   case BuildCore => (p: Parameters) => new Rocket()(p)
   case RocketCrossing => Synchronous
   case RocketTilesKey =>  Nil
-  case DMKey => new DefaultDebugModuleConfig(site(XLen))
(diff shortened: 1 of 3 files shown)


src/main/scala/coreplex/Configs.scala
```diff
@@ -22,7 +22,7 @@ class BaseCoreplexConfig extends Config ((site, here, up) => {
   case BuildCore => (p: Parameters) => new Rocket()(p)
   case RocketCrossing => Synchronous
   case RocketTilesKey =>  Nil
-  case DMKey => new DefaultDebugModuleConfig(site(XLen))
+  case DMKey => DefaultDebugModuleConfig(site(XLen))
```
hcook: i'd use named parameters here `DefaultDebugModuleConfig(nDMIAddrSize = site(XLen))` only because this class takes so many Ints in a row and you might want to swap their order later

mwachs5: That's not quite what the DefaultDebugModuleConfig is doing, it actually takes in XLEN as (currently the only) parameter and then turns it into nDMIAddrSize. Point well taken though

PR #658, opened 2017-04-06
Fix typo in Debug code generation

mwachs5:
abstractGeneratedI should be abstractGeneratedS when pulling out the opcode.
This doesn't actually break anything, but fix it for clarity.


mwachs5:
included in #655

PR #663, opened 2017-04-08
Add vectored interrupt support

aswaterman:
https://github.com/riscv/riscv-isa-manual/commit/4dcaa944ba40e074d25516a157fc37f7491b71cc

I also added a test, which does indeed pass, but I don't want to bump
riscv-tools for that alone:

https://github.com/riscv/riscv-tests/commit/ba6d88466a96ec3147b96a57f746a137085764f9

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -475,7 +475,15 @@ class CSRFile(perfEventSets: EventSets = new EventSets(Seq()))(implicit p: Param
   val trapToDebug = Bool(usingDebug) && (reg_singleStepped || causeIsDebugInt || causeIsDebugTrigger || causeIsDebugBreak || reg_debug)
   val debugTVec = Mux(reg_debug, Mux(insn_break, UInt(0x800), UInt(0x808)), UInt(0x800))
   val delegate = Bool(usingVM) && reg_mstatus.prv <= PRV.S && Mux(cause(xLen-1), reg_mideleg(cause_lsbs), reg_medeleg(cause_lsbs))
-  val tvec = Mux(trapToDebug, debugTVec, Mux(delegate, reg_stvec.sextTo(vaddrBitsExtended), reg_mtvec))
+  val notDebugTVec = {
+    val base = Mux(delegate, reg_stvec.sextTo(vaddrBitsExtended), reg_mtvec)
+    val baseAlign = 2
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/CSR.scala
```diff
@@ -475,7 +475,15 @@ class CSRFile(perfEventSets: EventSets = new EventSets(Seq()))(implicit p: Param
   val trapToDebug = Bool(usingDebug) && (reg_singleStepped || causeIsDebugInt || causeIsDebugTrigger || causeIsDebugBreak || reg_debug)
   val debugTVec = Mux(reg_debug, Mux(insn_break, UInt(0x800), UInt(0x808)), UInt(0x800))
   val delegate = Bool(usingVM) && reg_mstatus.prv <= PRV.S && Mux(cause(xLen-1), reg_mideleg(cause_lsbs), reg_medeleg(cause_lsbs))
-  val tvec = Mux(trapToDebug, debugTVec, Mux(delegate, reg_stvec.sextTo(vaddrBitsExtended), reg_mtvec))
+  val notDebugTVec = {
+    val base = Mux(delegate, reg_stvec.sextTo(vaddrBitsExtended), reg_mtvec)
+    val baseAlign = 2
+    val interruptAlign = log2Ceil(new MIP().getWidth)
+    val interruptOffset = cause(log2Ceil(new MIP().getWidth)-1, 0) << baseAlign
```
terpstra: You don't like to do CSE by hand now and then?

Is 4 bytes really enough space for the trap handler for each interrupt? Do you intend that they do a relative jump and that's it?

aswaterman: oops, that is an editing error... I'll fix that in a subsequent PR.

aswaterman: Re: the architectural decision, yeah, we're pretty sure we made the least-bad decision.  There's some desire to have more space per handler, but that's really only useful if you can fit a crucial basic block in that space.  In our experience, you need to allocate a lot of space (say 32/64 bytes) to make that likely in practice.  Otherwise, there will be a jump instruction anyway...

terpstra: Sure. But these are the people who complained about the sequence: "load, shift, add, jump". You only reduced it by *3* instructions when you could have saved *4*. Drams!

PR #666, opened 2017-04-11
Implement new FP encoding proposal

aswaterman:
https://groups.google.com/a/groups.riscv.org/d/msg/isa-dev/_r7hBlzsEd8/cWPyJKMzCQAJ

### DIFF
--- a/src/main/scala/tile/FPU.scala
+++ b/src/main/scala/tile/FPU.scala
@@ -20,32 +20,11 @@ case class FPUParams(
 
 object FPConstants
 {
-  def FCMD_ADD =    BitPat("b0??00")
-  def FCMD_SUB =    BitPat("b0??01")
-  def FCMD_MUL =    BitPat("b0??10")
-  def FCMD_MADD =   BitPat("b1??00")
-  def FCMD_MSUB =   BitPat("b1??01")
-  def FCMD_NMSUB =  BitPat("b1??10")
-  def FCMD_NMADD =  BitPat("b1??11")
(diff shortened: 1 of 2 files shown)


ccelio:
Naturally, this is the pull-request that would get the mark of the Devil.

aswaterman:
rofl

aswaterman:
@ccelio repent for your sins

src/main/scala/tile/FPU.scala
```diff
@@ -221,132 +202,222 @@ object ClassifyRecFN {
...
 
-/** Format conversion without rounding or NaN handling */
-object RecFNToRecFN_noncompliant {
-  def apply(in: UInt, inExpWidth: Int, inSigWidth: Int, outExpWidth: Int, outSigWidth: Int) = {
-    val sign = in(inSigWidth + inExpWidth)
-    val fractIn = in(inSigWidth - 2, 0)
-    val expIn = in(inSigWidth + inExpWidth - 1, inSigWidth - 1)
-    val fractOut = fractIn << outSigWidth >> inSigWidth
+  // convert between formats, ignoring rounding, range, NaN
+	def unsafeConvert(x: UInt, to: FType) = if (this == to) x else {
```
ccelio: Is there a reason the indention here is whack?

aswaterman: Yeah, for some reason a tab ended up in there?  Not sure how, since I have expandtab on in vim... but I didn't see it, since I have ts=2.

PR #673, opened 2017-04-12
Unconditionally write badaddr, possibly to zero

aswaterman:
https://github.com/riscv/riscv-isa-manual/commit/59d33f6b833d6499edfa825a204933098345d53d


aswaterman:
And to make it easy to set badaddr on more exceptions in future; software can branch on zero value to see if feature is probably unimplemented.

PR #676, opened 2017-04-13
only replicate scratch slave d-channel resp when AMO req

yunsup:


src/main/scala/rocket/ScratchpadSlavePort.scala
```diff
@@ -91,7 +92,8 @@ class ScratchpadSlavePort(address: AddressSet)(implicit p: Parameters) extends L
     // place AMO data in correct word lane
     val minAMOBytes = 4
     val grantData = Mux(io.dmem.resp.valid, io.dmem.resp.bits.data, acq.data)
-    val alignedGrantData = Mux(acq.size <= log2Ceil(minAMOBytes), Fill(coreDataBytes/minAMOBytes, grantData(8*minAMOBytes-1, 0)), grantData)
+    val alignedGrantData =
+      Mux(isAMO && (acq.size <= log2Ceil(minAMOBytes)), Fill(coreDataBytes/minAMOBytes, grantData(8*minAMOBytes-1, 0)), grantData)
```
aswaterman: To be consistent with the rest of the code, use ```edge.hasData(acq)``` instead of ```isAMO```.

PR #678, opened 2017-04-13
RAMModel atomics

terpstra:


terpstra:
Do not merge yet; I want to add a CRC to the atomic flight tracker to confirm they return the expected result.

PR #679, opened 2017-04-13
jtag: make it easier to assign MFR ID externally

mwachs5:
This pulls the Manufacturer ID out to the top level of Rocket Chip Example Top.


aswaterman:
To save on travis bandwidth I'll absorb this into my PR

PR #680, opened 2017-04-14
TLB functionality improvements

aswaterman:
- Check that AMOLogical/AMOArithmetic is supported on uncacheable regions.
- When virtual addresses are wider than physical addresses, but virtual memory is turned off, verify that physical-address MSBs are zero, or raise an access exception.
- Two bonus unrelated commits: improve Seq UInt indexing QoR; fix a whitespace error.

### DIFF
--- a/src/main/scala/rocket/TLB.scala
+++ b/src/main/scala/rocket/TLB.scala
@@ -10,6 +10,7 @@ import diplomacy._
 import coreplex.CacheBlockBytes
 import tile.{XLen, CoreModule, CoreBundle}
 import uncore.tilelink2._
+import uncore.constants._
 import util._
 
 case object PAddrBits extends Field[Int]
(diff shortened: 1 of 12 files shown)


src/main/scala/rocket/TLB.scala
```diff
@@ -166,11 +175,18 @@ class TLB(lgMaxSize: Int, nEntries: Int)(implicit edge: TLEdgeOut, p: Parameters
   val pr_array = Cat(Fill(2, prot_r), entries.init.map(_.pr).asUInt)
   val pw_array = Cat(Fill(2, prot_w), entries.init.map(_.pw).asUInt)
   val px_array = Cat(Fill(2, prot_x), entries.init.map(_.px).asUInt)
+  val paa_array = Cat(Fill(2, prot_aa), entries.init.map(_.paa).asUInt)
+  val pal_array = Cat(Fill(2, prot_al), entries.init.map(_.pal).asUInt)
   val c_array = Cat(Fill(2, cacheable), entries.init.map(_.c).asUInt)
+  val ae_st_array = ~pw_array | Mux(isAMOLogical(io.req.bits.cmd), ~pal_array, 0.U) | Mux(isAMOArithmetic(io.req.bits.cmd), ~paa_array, 0.U)
```
terpstra: it's a bit confusing that this one has inverse polarity compared to the others, but whatever.

aswaterman: I kind of agree, but couldn't think of the right name for the opposite polarity.  (It's not really store_ok, because it needs to indicate that the store is ok with respect to access exceptions specifically... maybe the verbose store_physical_access_ok?)

PR #683, opened 2017-04-14
Treat exceptions as steps for the purposes of single-stepping

aswaterman:


mwachs5:
I will merge this into #675

PR #684, opened 2017-04-18
Grab bag of memory system improvements

aswaterman:
- Mitigate critical path for D$ exception reporting
- Send D$ grant acks early
- Cope with early grant acks in broadcast hub
- Cope with early release acks in D$
- Generate access exceptions for LR/SC on uncacheable memory (including scratchpad)
- Guarantee eventual probe forward progress for pessimal LR/SC sequences

### DIFF
--- a/src/main/scala/rocket/NBDcache.scala
+++ b/src/main/scala/rocket/NBDcache.scala
@@ -975,8 +975,9 @@ class NonBlockingDCacheModule(outer: NonBlockingDCache) extends HellaCacheModule
   io.cpu.ordered := mshrs.io.fence_rdy && !s1_valid && !s2_valid
   io.cpu.replay_next := (s1_replay && s1_read) || mshrs.io.replay_next
 
-  val s1_xcpt = Mux(s1_nack || !dtlb.io.req.valid, 0.U.asTypeOf(dtlb.io.resp), dtlb.io.resp)
-  io.cpu.s2_xcpt := RegEnable(s1_xcpt, s1_clk_en)
+  val s1_xcpt_valid = dtlb.io.req.valid && !s1_nack
+  val s1_xcpt = dtlb.io.resp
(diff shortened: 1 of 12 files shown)


src/main/scala/rocket/TLBPermissions.scala
```diff
@@ -18,6 +18,7 @@ case class TLBPermissions(
 object TLBPageLookup
 {
   private case class TLBFixedPermissions(
+    t: RegionType.T,
```
terpstra: I think you should have made this into 'e: Boolean = RegionType.PUT_EFFECTS, RegionType.GET_EFFECTS) contains x'. The way it is now it will not combine/optimize cases that actually end up with the same boolean value.

src/main/scala/rocket/NBDcache.scala
```diff
@@ -794,18 +794,18 @@ class NonBlockingDCacheModule(outer: NonBlockingDCache) extends HellaCacheModule
...
-  when (lrsc_valid) { lrsc_count := lrsc_count - 1 }
+  when (lrsc_count > 0) { lrsc_count := lrsc_count - 1 }
   when (s2_valid_masked && s2_hit || s2_replay) {
     when (s2_lr) {
       lrsc_count := lrscCycles - 1
       lrsc_addr := s2_req.addr >> blockOffBits
     }
-    when (lrsc_valid) {
+    when (lrsc_count > 0) {
       lrsc_count := 0
```
terpstra: Do we still have adequate coverage tests for NBDCache? We keep refactoring it.

aswaterman: Hardly adequate, but nonzero.  I've only changed it in ways that match the blocking D$, though...

src/main/scala/rocket/TLBPermissions.scala
aswaterman: Ah, that makes sense.

PR #687, opened 2017-04-19
debug: Make DMI NOPs really NOPs.

mwachs5:
This simplifies SW design and CDC issues.


mwachs5:
Ah. I forgot that FESVR uses NOPs too. Hooray for assertions.

PR #694, opened 2017-04-20
Allow RawModule as the top level for the Generator

mwachs5:
If the thing you want to generate is a RawModule (especially if you want more control over the top level signal names), this allows you to do it by changing the class that the object is cast to.

### DIFF
--- a/src/main/scala/util/GeneratorUtils.scala
+++ b/src/main/scala/util/GeneratorUtils.scala
@@ -3,6 +3,7 @@
 package util
 
 import Chisel._
+import chisel3.experimental.{RawModule}
 import config._
 import diplomacy.LazyModule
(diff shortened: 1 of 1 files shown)


aswaterman:
@jackkoenig is it safe to assume this won't change (modulo the package name)?

jackkoenig:
I'm a little iffy about RawModule being a super-type of ImplicitModule, I had thought there would be UserModule which is the parent of both, but apparently RawModule == UserModule. I'm going to file a Chisel issue because things don't look like what I thought they were supposed to.

mwachs5:
This is not blocking me, just something I ran into.

mwachs5:
I guess i don't care about this too much. Going go close until this is out of the "Experimental" phase.

PR #695, opened 2017-04-21
Pipeline AXI4 MMIO

terpstra:
This PR makes it possible to send multiple AXI4 requests into/out-of TileLink which preserve FIFO semantics.


terpstra:
Last piece is in place. Just needs DV to sign off!

terpstra:
Rebased and ITIM earlyAck enabled

terpstra:
It looks you guys did not remove AXI4BridgeTest from regression. An unlucky seed may cause the test to fail if it writes and then reads from the Error device. Could you either disable or fix the test.

hcook:
I'm adding the ability to subset address ranges to the TLFuzzer

PR #697, opened 2017-04-21
Add narrowData option to AsyncQueue.

ben-k:
This option reduces the number of wires that cross the clock boundary.
This can be a useful feature if the clock boundary coincides with
a voltage boundary, in which case the number of level shifters is reduced.
However, this introduces a path that crosses from sink->source->sink domain,
so the option is disabled by default.

The attached image should clarify the design choice.

narrowdata


mwachs5:
This is not a real comment on this PR, but where can we put these diagrams for easy reference...

I think it would be nice to have a place for such diagrams in rocket-chip somewhere.

PR #701, opened 2017-04-25
dcache: put a flow Q to absorb back-pressure without restarting pipeline

terpstra:
When used with a RationalCrossing, pipelined MMIO does not come out cleanly.
The first beat works, but if the second beat gets stalled, the pipeline is
restarted. This is a quick hacky test to absorb the beats. Perhaps a better
fix can be made to achieve the same effect.

@aswaterman Amusingly, when this patch is combined with a 2:1 crossing, we
achieve the same exact waveform as with a certain other bad configuration, but
without the need for any hacks. MMIOs come out pipelined and there is only
a 5 cycle delay from load to dependent load. No fastLoadWord needed.

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -57,6 +57,9 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   data.io.req <> dataArb.io.out
   dataArb.io.out.ready := true
 
+  val tl_out_a = Wire(tl_out.a)
+  tl_out.a <> Queue(tl_out_a, 3, flow = true)
+
   val s1_valid = Reg(next=io.cpu.req.fire(), init=Bool(false))
   val s1_probe = Reg(next=tl_out.b.fire(), init=Bool(false))
   val probe_bits = RegEnable(tl_out.b.bits, tl_out.b.fire()) // TODO has data now :(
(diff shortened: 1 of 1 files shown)


terpstra:
In case it's not obvious, I am not seriously proposing to merge this PR, just demonstrate the improvement. Maybe we can avoid the pipeline restart another way?

aswaterman:
With a rational crossing, having a tiny buffer on the processor side actually makes sense, right?

terpstra:
Perhaps. I take it you don't want back-pressure from the bus to affect the processor pipeline?

Then, to enable it to sink back-to-back MMIO w/o restart we would need something like this:
1:1 no Q
2:1 Q depth = nMMIOs/2
3:1 Q depth = nMMIOs*2/3
m:1 Q depth = nMMIOs*(m-1)/m
...

However, at some point the processor pipeline restart is faster than the time to drain the Q. If your processor pipeline is 6 stages and it runs at 2:1, draining 3 entries in the Q is enough time for the core to be refilling it again after restart.

So depth = min((6+m-1)/m, nMMIOs*(m-1)/m) ?

terpstra:
After writing that formula down, I wonder if maybe we should just set depth = 3, like I did by chance in the PR. That works for all m.

aswaterman:
I don't think you want it to be a constant, as that costs something like 200-300 flops, which will be of little utility for systems that don't rely on MMIO burst performance.

terpstra:
depth = min(2, nMMIOs-1) ? I'm not sure we want to wire 'm' into the design if Yunsup wants to have variable ratios.

```
AA0
B.1
CB1
D.2
EC2
buffer exhausted, pipeline killed
1D1
2.1
3E0
4.0
5AA
```

aswaterman:
So, non-existent when nMMIOs=1, and also when not using the rational crossing?  Makes sense to me.

aswaterman:
Just to declutter the dcache code a bit, can you instantiate the queue on the outside instead?

src/main/scala/rocket/DCache.scala
```diff
@@ -57,6 +58,19 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   data.io.req <> dataArb.io.out
   dataArb.io.out.ready := true
 
+  val rational = p(coreplex.RocketCrossing) match {
+    case coreplex.SynchronousCrossing(_) => true
+    case _ => false
+  }
+
+  val tl_out_a = Wire(tl_out.a)
+  val q_depth = if (rational) min(2, maxUncachedInFlight-1) else 0
```
aswaterman: Not a request to change, just FYI, you can do ```(2 min (maxUncachedInFlight-1))``` and avoid having to import scala.math.min

terpstra:
I was planning on using a TLBuffer outside until I realized you may need a different policy for the non-blocking case. Not to mention, you only want this on channel A.

aswaterman:
True, the non-blocking case probably doesn't need it at all.  It just seems like really weird separation of concerns for the DCache to be aware of the crossing used outside of it... the queue goes with the crossing.

terpstra:
Well, where do you want to put it? It's the intersection of the blocking dcache + crossing.

aswaterman:
It's fine how it is

terpstra:
I'm also not 100% convinced it's the crossing's fault. It's the blocking dcache's fault that it can't tolerate a few stalls gracefully.

PR #704, opened 2017-04-25
Miscellaneous uncore cleanups

hcook:
* More verbosity
* Make AHB SRAM response fuzzing off by default

### DIFF
--- a/src/main/scala/uncore/ahb/SRAM.scala
+++ b/src/main/scala/uncore/ahb/SRAM.scala
@@ -7,7 +7,7 @@ import config._
 import diplomacy._
 import util._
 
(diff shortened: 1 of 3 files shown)


src/main/scala/uncore/ahb/SRAM.scala
```diff
@@ -95,7 +95,7 @@ class AHBRAM(address: AddressSet, executable: Boolean = true, beatBytes: Int = 4
     when (a_request)  { d_request := Bool(true) }
 
     // Finally, the outputs
-    in.hreadyout := !d_request || LFSR16(Bool(true))(0) // Bool(true)
+    in.hreadyout := (if(fuzzHreadyout) { !d_request || LFSR16(Bool(true))(0) } else { Bool(true) })
     in.hresp     := AHBParameters.RESP_OKAY
```
terpstra: Good idea.

PR #707, opened 2017-04-26
ITIM

aswaterman:
Tightly integrated instruction memory


src/main/scala/rocket/ICache.scala
```diff
@@ -111,72 +162,127 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
...
-    val wen = tl_out.d.valid && repl_way === UInt(i)
-    when (wen) {
-      val e_d = code.encode(tl_out.d.bits.data)
-      data_array.write((refill_idx << log2Ceil(refillCycles)) | refill_cnt, e_d)
-    }
-    val s0_raddr = s0_vaddr(untagBits-1,blockOffBits-log2Ceil(refillCycles))
-    s1_dout(i) := data_array.read(s0_raddr, !wen && s0_valid)
-  }
-*/
-
```
hcook: 👻

src/main/scala/rocket/Frontend.scala
```diff
@@ -42,19 +42,26 @@ class FrontendIO(implicit p: Parameters) extends CoreBundle()(p) {
...
+class Frontend(hartid: Int)(implicit p: Parameters) extends LazyModule {
   lazy val module = new FrontendModule(this)
-  val icache = LazyModule(new ICache(latency = 2))
+  val icache = LazyModule(new ICache(latency = 2, hartid))
   val node = TLOutputNode()
+  val slaveNode = icache.slaveNode.map { n =>
+    val res = TLInputNode()
+    n := res
+    res
+  }
```
hcook: this should not be necessary but let me play around with it myself

aswaterman: Yeah it's kind of gnarly as written

PR #710, opened 2017-04-27
Fixed DCache response address by simplifying code marked as don't care (2nd attempt)

solomatnikov:


src/main/scala/rocket/DCache.scala
```diff
@@ -357,7 +357,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
         s2_req.cmd := M_XRD
         s2_req.typ := req.typ
         s2_req.tag := req.tag
-        s2_req.addr := Cat(s1_paddr >> beatOffBits /* don't-care */, req.addr(beatOffBits-1, 0))
+        s2_req.addr := req.addr
```
aswaterman: The reason I didn't do this originally is that the I/O MSHRs only needed to store 6 address bits (the upper N-6 would be pruned in synthesis).  With this change, they won't be pruned, so the MSHR becomes 3-4x more expensive.

I think we can accomplish both goals, though.  Let me think about it some more.

aswaterman:
Can you confirm that 8fd5ecdff8c2609f60a93b411d2c43a81f7452c0 does the right thing?

solomatnikov:
#https://github.com/ucb-bar/rocket-chip/commit/8fd5ecdff8c2609f60a93b411d2c43a81f7452c0 works

PR #711, opened 2017-04-28
Some new global parameters

hcook:
Rebase this once #707 gets merged


aswaterman:
I'm folding this into #707

PR #714, opened 2017-04-28
Interrupts: Less Pessimistic Synchronization

mwachs5:
This does a more fine-grained synchronization for the different interrupt types and different synchronization types:

Within the Coreplex:
-- only debug interrupt is always fully asynchronous
-- Local Interrupts are always assumed to have been synchronized to core clock 
-- CLINT interrupt is known to be synchronous to core clock
-- PLIC interrupt is known to be synchronous to periphery clock, treat it accordingly

Additionally, there are now two subtraits of the PeripheryExtInterrupts (which becomes abstract). One which does the synchronization and one which assumes it has already been done.

There may still be some issues in the way the interrupt numbers are being assigned, still verifying this does what we want.

### DIFF
--- a/src/main/scala/coreplex/RocketTiles.scala
+++ b/src/main/scala/coreplex/RocketTiles.scala
@@ -43,16 +43,16 @@ trait HasRocketTiles extends CoreplexRISCVPlatform {
     val periphIntXbar = LazyModule(new IntXbar)
     val coreIntXbar   = LazyModule(new IntXbar)
 
-    // CLINT and Local Interrupts must be synchronized to the core clock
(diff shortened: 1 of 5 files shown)


aswaterman:
It looks like the CLINT is in the periphery clock, though, so it does need to go through a synchronizer before entering the core (same as PLIC outputs I think?).  I didn't intend to change the synchronization of the CLINT or of the PLIC outputs - just the PLIC inputs and the local interrupts.

mwachs5:
Ok I'll double check which clock CLINT is on and adjust accordingly

PR #716, opened 2017-05-01
Fix ITIM loads

aswaterman:
An incorrectly-set ready signal caused bad data to be read from the RAM.


yunsup:
passes failing torture test

PR #720, opened 2017-05-03
local_interrupts: Correct off-by-1 if there is no SEIP

mwachs5:


mwachs5:
You mean #695? peeps are on it.

PR #725, opened 2017-05-04
ITIM supports PutPartial

scottj97:
Enables PutPartial to the new ITIM


yunsup:
this change is put into #727

PR #726, opened 2017-05-04
Interlock to prevent ITIM hazard when tl.a.valid & tl.d.valid & !tl.d.ready

aswaterman:


scottj97:
Trying it out

aswaterman:
great

yunsup:
this change is put into #727

PR #727, opened 2017-05-04
Wes's fix for AXI2TL timeout when writes backed up

scottj97:
Issue lots of outstanding AXI writes and eventually it will stop responding altogether.

Wes fixed today.

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -226,7 +226,8 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
       io.resp.valid := s2_valid && s2_hit && !s2_disparity
 
       tl_in.map { tl =>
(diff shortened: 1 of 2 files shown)


src/main/scala/rocket/ICache.scala
```diff
@@ -47,6 +47,7 @@ class ICache(val latency: Int, val hartid: Int)(implicit p: Parameters) extends
         regionType      = RegionType.UNCACHED,
         executable      = true,
         supportsPutFull = TransferSizes(1, wordBytes),
+        supportsPutPartial = TransferSizes(1, wordBytes),
```
terpstra: Just writing this does not make it true. Is this just a stop-gap measure before we have an actual adapter?

aswaterman: No, it's actually true.  Not adding it to the list of supported operations was an oversight.

PR #734, opened 2017-05-09
tilelink: better address masking for fuzzing

hcook:


src/main/scala/uncore/tilelink2/Fuzzer.scala
```diff
@@ -137,7 +136,8 @@ class TLFuzzer(
     val log_op    = noiseMaker(2, inc, 0)
     val amo_size  = UInt(2) + noiseMaker(1, inc, 0) // word or dword
     val size      = noiseMaker(sizeBits, inc, 0)
-    val addr      = noiseMaker(addressBits, inc, 2) & ~UIntToOH1(size, addressBits)
+    val addrMask  = overrideAddress.map(_.max.U).getOrElse(~UInt(0, addressBits))
+    val addr      = noiseMaker(addressBits, inc, 2) & ~UIntToOH1(size, addressBits) & addrMask
```
terpstra: I don't get it. What are you trying to fix here? Isn't noiseMaker anyway restricted to addressBits?

hcook: If base address is 0x4000 and desired target range size is 0x1000, old version would additionally fuzz range between 0x5000-0x7fff.

hcook: ping

PR #739, opened 2017-05-11
Bump riscv-tools, to get some -mcmodel=medany fixes

palmer-dabbelt:


palmer-dabbelt:
Thanks.  I was worried we'd just entered some sort of time loop!

terpstra:
We still need to wait for it to rebuild again on master to update the cache. So in 2 more hours we can restart these other PRs.

PR #740, opened 2017-05-12
DTS tweaks

terpstra:
As discussed, the updated DTS for our cores.


aswaterman:
Does this match Spike's behavior?

terpstra:
The isa string matches. Spike has no microarchitecture, so no.

aswaterman:
Of course, I meant the ISA string.

PR #750, opened 2017-05-18
verilator: bump to the newest release

terpstra:
If this works, then we can try patching verilator upstream to add the plusargs feature we need.


mwachs5:
Can we do a test with an inout?

terpstra:
It didn't actually build the new verilator. It cheated with the cache. A fix the Makefile is needed.

terpstra:
I take it back. It DID build verilator. It was just hidden inside a closed arrow of the build log.

terpstra:
Closing this PR in favor of one that bumps to 3.903 once it is released.

PR #752, opened 2017-05-18
Plusargs -- tilelink timeout detection from the command line

terpstra:
This PR adds the ability to add a tilelink progress watchdog at runtime.
```./simv ... +tilelink_timeout=100```
will cause the Monitor to assert fail if no progress is made on a channel after 100 cycles.

This feature is currently not supported by verilator.
See: https://www.veripool.org/issues/1165

### DIFF
--- a/src/main/scala/util/PlusArg.scala
+++ b/src/main/scala/util/PlusArg.scala
@@ -1,0 +1,21 @@
+// See LICENSE.SiFive for license details.
+
+package util
+import Chisel._
+
+class plusarg_reader(format: String, default: Int) extends BlackBox(Map(
+    "FORMAT"  -> chisel3.core.StringParam(format),
+    "DEFAULT" -> chisel3.core.IntParam(default))) {
+  val io = new Bundle {
+    val out = UInt(OUTPUT, width = 32)
+  }
+}
+
+object PlusArg
+{
+  // PlusArg("foo") will return 42 if the simulation is run with +foo=42
+  // Do not use this as an initial register value. The value is set in an
+  // initial block and thus accessing it from another initial is racey.
+  def apply(name: String, default: Int = 0): UInt =
+    Module(new plusarg_reader(name + "=%d", default)).io.out
+}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by scottj97** [comment]
Please don't call it "deadlock" detection. This is a watchdog timer, which will detect lots of problems that may have nothing to do with a real deadlock.

May I suggest: +tilelink_timeout=100

That way it's clear what it's really looking at.

File: src/main/scala/util/PlusArg.scala [hunk 1]
```diff
@@ -0,0 +1,20 @@
...
+    "DEFAULT" -> chisel3.core.IntParam(default))) {
+  val io = new Bundle {
+    val out = UInt(OUTPUT, width = 32)
+  }
+}
+
+object PlusArg
+{
+  // PlusArg("foo") will return 42 if the simulation is run with +foo=42
+  // Do not use this as an initial register value.
```
scottj97: Comments should explain why not (it is set in an initial block which would cause a race condition if used in any other initial block).

src/main/scala/util/PlusArg.scala
```diff
@@ -0,0 +1,20 @@
...
+  def apply(name: String, default: Int = 0): UInt =
+    Module(new plusarg_reader(name + "=%d", default)).io.out
```
scottj97: I think the string concatenation of "=%d" better belongs in plusarg_reader.v so that it's clear what the $value$plusargs is doing.

terpstra: I don't agree. I  might want to have a hex version as well from chisel.

src/main/scala/util/PlusArg.scala
terpstra: Done.

terpstra:
I've changed the name to +tilelink_timeout.

vsrc/plusarg_reader.v
```diff
@@ -0,0 +1,19 @@
+// See LICENSE.SiFive for license details.
+
+module plusarg_reader #(string FORMAT="borked", int DEFAULT=0) (
```
scottj97: Add comment: no default parameter values are intended, nor does IEEE 1800-2012 require them (clause A.2.4 param_assignment), but Incisive demands them.

src/main/scala/util/PlusArg.scala
scottj97: That's good then. As long as nobody tries to do a float or string

yunsup:
This is cool.

PR #755, opened 2017-05-19
Verilator plusargs

terpstra:
Once verilator 3.903 is released, util.PlusArgs will support it.

@scottj97 can you confirm the modified verilog still works on all the other tools?

### DIFF
--- a/vsrc/plusarg_reader.v
+++ b/vsrc/plusarg_reader.v
@@ -10,12 +10,10 @@ reg [31:0] myplus;
 assign out = myplus;
 
 initial begin
+`ifdef SYNTHESIS
    myplus = DEFAULT;
-`ifndef SYNTHESIS
-`ifndef verilator
-   // Work-around for https://www.veripool.org/issues/1165
+`else
    if (!$value$plusargs(FORMAT, myplus)) myplus = DEFAULT;
-`endif
 `endif
 end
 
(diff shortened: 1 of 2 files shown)


vsrc/plusarg_reader.v
```diff
@@ -10,12 +10,10 @@ reg [31:0] myplus;
...
 initial begin
+`ifdef SYNTHESIS
    myplus = DEFAULT;
-`ifndef SYNTHESIS
-`ifndef verilator
-   // Work-around for https://www.veripool.org/issues/1165
-   if (!$value$plusargs(FORMAT, myplus)) myplus = DEFAULT;
-`endif
+`else
+   if (~|$value$plusargs(FORMAT, myplus)) myplus = DEFAULT;
```
scottj97: What's wrong with **!**$value$plusargs (versus **~|**)? This function is defined to return an integer, either zero or nonzero, which is what ! tests for. ~| suggests this is a bus with many booleans and you want to know if any of them are true.

scottj97:
I confirmed the new plusarg_reader.v works in VCS (K-2015.09-SP2), Incisive (15.20.012), and Questa (10.5c).

vsrc/plusarg_reader.v
terpstra: The ! complains in verilator that the return result is 32-bit, not 1-bit wide. This causes it to fail.

terpstra: However, this might no longer be necessary as the verilator author, Wilson Snyder, has made a follow-up commit that might resolve this.

PR #756, opened 2017-05-19
Assert wb_wxd only when wb_waddr is not zero

ilmessere:
Check that the destination register address is
not zero when computing wb_wxd.

This should solve the following problem:
instructions like nop sets the wb_wxd signal
to '1' even if the destination register is zero,
this prevents possible writes from the RoCC
interface to be accepted because the ready
signal will be '0' as long as there are other
instructions writing in the register file.

For example:
	custom0_rd_rs1_rs2	a1, a2, a3
	nop
	nop
	...
	nop
	add	a4, a4, a1

In this code the write back to the 'a1'register
won't happen until the value is needed by the add
instruction, causing a 5 cycles penalty, even if
it was possible to accept the write many cycles
before.

### DIFF
--- a/src/main/scala/rocket/Rocket.scala
+++ b/src/main/scala/rocket/Rocket.scala
@@ -448,7 +448,7 @@ class Rocket(implicit p: Parameters) extends CoreModule()(p)
     (wb_reg_valid && wb_ctrl.mem && io.dmem.s2_xcpt.ae.ld, UInt(Causes.load_access))
   ))
 
-  val wb_wxd = wb_reg_valid && wb_ctrl.wxd
+  val wb_wxd = wb_reg_valid && wb_ctrl.wxd && (wb_waddr =/= UInt(0)) //check that the waddr is not zero 
   val wb_set_sboard = wb_ctrl.div || wb_dcache_miss || wb_ctrl.rocc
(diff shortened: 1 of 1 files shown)


aswaterman:
This seems to be optimizing for a case that shouldn't occur in practice.  NOPs should only show up for code-alignment purposes (e.g. in the loop head), during which it's highly unlikely that a RoCC instruction is in flight.

ilmessere:
My idea was to use some independent instructions and a single well-placed NOP
to mask the latency of the coprocessor and, above all, avoid a 5 cycles stall when executing an
instruction that uses the value coming from the RoCC interface.

In particular in the following case :
          custom0_rd            a1                // read some data
          inst1
          inst2
          nop
          inst3
          inst4
          inst5
          add                         a2, a2, a1     // use the data

the NOP would allow to write back the data in 'a1' so that there's just one cycle of penalty.

I understand that this could be rare case, but I thought someone could find this useful.

aswaterman:
Ahh, I see.

As a workaround you could use a different NOP, a never-taken branch (bne x0, x0, ...), which will never induce any stalls and does not write rd.  It will never enter the BTB and so should never be mispredicted.

aswaterman:
Closing for now, given the workaround still seems sensible

PR #766, opened 2017-05-25
minNum -> minimumNumber

aswaterman:
https://github.com/riscv/riscv-isa-manual/pull/75/commits/1ea65632cb08a6c03f9ffc507118e33061289595


aswaterman:
Yeah, both definitions are legal for the C library routines fmin/fmax, too.

Waiting for KA to merge the manual changes before merging this.

PR #768, opened 2017-05-26
jtag_vpi: Attempt to more aggressively flush the simulator output

mwachs5:
The message is needed by other processes, and if the simulator output is not flushed, it can deadlock.


mwachs5:
I think that in the future we will replace the whole socket interface with a named pipe/file input so that we don't need anything fancy in the simulator.

PR #769, opened 2017-05-30
Incorporate new div/sqrt unit

aswaterman:


src/main/scala/tile/FPU.scala
```diff
@@ -241,9 +241,10 @@ trait HasFPUParameters {
   val (dExpWidth, dSigWidth) = (FType.D.exp, FType.D.sig)
   val floatTypes = FType.all.filter(_.ieeeWidth <= fLen)
   val maxType = floatTypes.last
-  def prevType(t: FType) = floatTypes(floatTypes.indexOf(t) - 1)
+  def prevType(t: FType) = floatTypes(typeTag(t) - 1)
   val maxExpWidth = maxType.exp
   val maxSigWidth = maxType.sig
+  def typeTag(t: FType) = floatTypes.indexOf(t)
```
hcook: `typeTag` makes me think of http://docs.scala-lang.org/overviews/reflection/typetags-manifests.html, but it looks like we are no longer using the scala construct anywhere in the code base, so...ymmv.

src/main/scala/tile/FPU.scala
```diff
@@ -764,45 +763,38 @@ class FPU(cfg: FPUParams)(implicit p: Parameters) extends FPUModule()(p) {
...
 
-    when (divSqrt_outValid) {
-      divSqrt_wen := !divSqrt_killed
-      divSqrt_wdata_double := sanitizeNaN(divSqrt.io.out, FType.D)
-      divSqrt_in_flight := false
-      divSqrt_flags_double := divSqrt.io.exceptionFlags
+    makeDivSqrt(FType.S, wb_ctrl.singleOut)
+    fLen match {
+      case 32 =>
+      case 64 => makeDivSqrt(FType.D, !wb_ctrl.singleOut)
```
hcook: is this going to have 128 eventually?

aswaterman: Yeah, for the Q extension.  For now, a match error for anything other than 32/64 is the desired behavior

src/main/scala/tile/FPU.scala
aswaterman: I was thinking of tagged unions when I wrote that

PR #770, opened 2017-05-30
Bump Chisel and Firrtl

jackkoenig:
A couple of bug fixes, 2 fairly large performance bugs fixed in Firrtl


jackkoenig:
This is failing due to bad interoperability of Chisel bundles in chisel3 code, I'm looking into it.

jackkoenig:
It's caused by a Bundle with an implicit output (ie. unspecified) being used as the next in a RegEnable (which is chisel3._ code). Fortunately, it is fixed by https://github.com/ucb-bar/chisel3/pull/617

PR #771, opened 2017-05-30
JTAG VPI: Make it work without debug_pp flag

mwachs5:
If -debug_pp is omitted, need to specify more in the PLI TAB file.

Again would like to get rid of all the VPI stuff entirely in the future.


scottj97:
The acc=rw:* is likely overkill and unnecessarily degrading simulation performance, but if this is going away soon I wouldn't worry about it.

PR #788, opened 2017-06-07
plic: do not output #address-cells

terpstra:
This is only needed for an interrupt-map, not an interrupt-controller.

@palmer-dabbelt This is not used in the kernel code, so only the kernel documentation needs to be updated to match.


terpstra:
@palmer-dabbelt Do we also want to rename riscv,ndev? I originally thought we would need to also change spike and other repositories, but spike has no PLIC.

terpstra:
Subsumed by #827

PR #789, opened 2017-06-08
Fix FMUL sign, again

aswaterman:
XOR XOR XOR


aswaterman:
I'll add this to riscv-tests tomorrow but I wanted to get a fix in before going to bed.  (This is what I had planned on committing before our discussion a few days ago.)

PR #800, opened 2017-06-15
GeneratorUtils: support to elaborate a RawModule

ss2783:
Changes to allow elaborate a top which is RawModule


jackkoenig:
Apologies for the delay in my response. I highly doubt the hierarchy is going to change so I'd say go ahead.

PR #801, opened 2017-06-16
changing SystemVerilog params to Verilog style

ss2783:
vivado-2016.1 synthesis doesn't support SystemVerilog string type parameters


ss2783:
@terpstra Please review

terpstra:
Restarting travis to see if it builds.

@scottj97 wasn't there some problem with one of the tools when you put 'string' in the argument list?

scottj97:
There was no problem with that on the Verilog simulators I tried, but I didn't try Vivado.

terpstra:
I thought you removed the types from the parameters for some tool?

scottj97:
Perhaps you're thinking of when I added default values for the params for
Incisive?

PR #802, opened 2017-06-16
Check for rvc before declaring illegal instruction after an ebreak.

richardxia:
This fixes a problem where an sbreak called in S-mode may cause an illegal instruction exception when the trap vector is a compressed instruction.


aswaterman:
thanks for testing this

PR #803, opened 2017-06-16
travis: remove travis wait call for riscv-tools

mwachs5:
Travis is timing out or failing and we can't see anything without the output.


terpstra:
wrong fix

PR #804, opened 2017-06-16
travis: attempt to make 2 build stages for cache

mwachs5:
First stage builds riscv-tools, next stage builds verilator


mwachs5:
This works now.

PR #807, opened 2017-06-19
Move RoCC interface to Diplomacy and TL2

colinschmidt:
This PR moves the RoCC interface to use Diplomacy and TileLink2.
This allows the memory channel parameters to be negotiated making RoCC configurations significantly simpler.
I've specified another TL Node type that allows the consumer to specify a single client node or use an output node to connect to multiple sub-modules. I think this strikes a nice implementation where the simple RoCC accelerators that do not use TL2 do not need any code in their accelerator relating to it.
I've also updated Hwacha to use this interface to ensure that more complicated uses behave as expected and everything is working on that front.

@hcook please review

@seldridge @zhemao are the others I know of using this interface that might have feedback.

### DIFF
--- a/src/main/scala/tile/LazyRoCC.scala
+++ b/src/main/scala/tile/LazyRoCC.scala
@@ -4,159 +4,18 @@
 package tile
 
 import Chisel._
-import Chisel.ImplicitConversions._
 import config._
 import coreplex._
 import diplomacy._
 import rocket._
-import uncore.constants._
-import uncore.agents._
-import uncore.coherence._
-import uncore.devices._
-import uncore.tilelink._
 import uncore.tilelink2._
-import uncore.util._
-import util._
 
-case object RoccMaxTaggedMemXacts extends Field[Int]
-case object RoccNMemChannels extends Field[Int]
 case object RoccNPTWPorts extends Field[Int]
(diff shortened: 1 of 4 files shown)


colinschmidt:
I've clearly borked something up in travis. Investigating

colinschmidt:
Should be good to review now. Sorry for the hiccup.

src/main/scala/tile/LazyRoCC.scala
```diff
@@ -316,14 +257,20 @@ class TranslatorExample(implicit p: Parameters) extends RoCC()(p) {
...
   io.mem.req.valid := Bool(false)
-  io.autl.acquire.valid := Bool(false)
-  io.autl.grant.ready := Bool(false)
 }
 
-class CharacterCountExample(implicit p: Parameters) extends RoCC()(p)
-    with HasTileLinkParameters {
+class  CharacterCountExample(implicit p: Parameters) extends LazyRoCC {
+  override lazy val module = new CharacterCountExampleModule(this)
+  override val atlNode = TLClientNode(TLClientParameters("CharacterCountRoCC"))
```
hcook: This override seems like a reasonable pattern to me.

PR #808, opened 2017-06-20
refactor ICache to be reusable by other frontends

colinschmidt:
specifically ones that would like to change the fetch width and number of
bytes in an instruction

This should keep the status quo for everyone but enable some new use-cases.

### DIFF
--- a/src/main/scala/rocket/Frontend.scala
+++ b/src/main/scala/rocket/Frontend.scala
@@ -45,9 +45,9 @@ class FrontendIO(implicit p: Parameters) extends CoreBundle()(p) {
   val perf = new FrontendPerfEvents().asInput
 }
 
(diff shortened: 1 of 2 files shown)


aswaterman:
I think the I$ doesn't need fetchWidth and coreInstBytes as separate parameters - can't it just have a single fetchBytes parameter, which is the product of the two?

colinschmidt:
I started that way but for a while I tried including the frontend in the refactor. But you're right, now that I've restricted the scope I can merge them. It doesn't seem like the require on coreInstBytes being a power of 2 is necessary in the Icache code. I'll submit a new PR soon.
Thanks for the CR.

aswaterman:
agreed.  i'll take a look again after the change

PR #822, opened 2017-06-26
tilelink2 ToAXI4: don't interlock R+W for non-FIFO masters

terpstra:
idCount <= 1 implies that no more than one transaction can be inflight,
so there is no need to interlock R+W. However, when stripBits > 0, it is
possible for a non-FIFO master to have > 1 idCount. In this case, we
also don't need to interlock R+W.

### DIFF
--- a/src/main/scala/uncore/tilelink2/ToAXI4.scala
+++ b/src/main/scala/uncore/tilelink2/ToAXI4.scala
@@ -73,14 +73,14 @@ class TLToAXI4(beatBytes: Int, combinational: Boolean = true, adapterName: Optio
       val sourceStall = Wire(Vec(edgeIn.client.endSourceId, Bool()))
       val sourceTable = Wire(Vec(edgeIn.client.endSourceId, out.aw.bits.id))
       val idStall = Wire(init = Vec.fill(edgeOut.master.endId) { Bool(false) })
-      var idCount = Array.fill(edgeOut.master.endId) { 0 }
+      var idCount = Array.fill(edgeOut.master.endId) { None:Option[Int] }
       val maps = (edgeIn.client.clients.sortWith(TLToAXI4.sortByType) zip edgeOut.master.masters) flatMap { case (c, m) =>
         for (i <- 0 until c.sourceId.size) {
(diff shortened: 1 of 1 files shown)


src/main/scala/uncore/tilelink2/ToAXI4.scala
```diff
@@ -199,17 +199,18 @@ class TLToAXI4(beatBytes: Int, combinational: Boolean = true, adapterName: Optio
       val a_sel = UIntToOH(arw.id, edgeOut.master.endId).toBools
       val d_sel = UIntToOH(Mux(r_wins, out.r.bits.id, out.b.bits.id), edgeOut.master.endId).toBools
       val d_last = Mux(r_wins, out.r.bits.last, Bool(true))
-      (a_sel zip d_sel zip idStall zip idCount) filter { case (_, n) => n > 1 } foreach { case (((as, ds), s), n) =>
```
hcook: ```(a_sel zip d_sel zip idStall zip idCount) } foreach { case (((as, ds), s), n)  if n.get > 1 =>```? ;)

src/main/scala/uncore/tilelink2/ToAXI4.scala
```diff
@@ -199,17 +199,18 @@ class TLToAXI4(beatBytes: Int, combinational: Boolean = true, adapterName: Optio
...
-        val count = RegInit(UInt(0, width = log2Ceil(n + 1)))
+      // If FIFO was requested, ensure that R+W ordering is preserved
+      (a_sel zip d_sel zip idStall zip idCount) filter { case (_, n) => n.map(_ > 1).getOrElse(false) } foreach { case (((as, ds), s), n) =>
```
hcook: `(a_sel zip d_sel zip idStall zip idCount) foreach { case (((as, ds), s), n) if n.map(_ > 1).getOrElse(false) =>` for added cuteness?

terpstra: Would result in a match error due to no 'case _' fallback.

PR #823, opened 2017-06-26
Added writeCombining to AXI4 port parameters

solomatnikov:


terpstra:
This does not seem right. This is supposed to indicate that a given slave will write combine, right? Then it belongs on the slave device, not the port. However, do we even want to expose this information?

solomatnikov:
This is per slave port attribute - WCB would be enabled for any device downstream.

terpstra:
But if you put a hypothetical AXI4 Xbar between a master and that adapter, you see that your adapter was in fact modifying the slave devices, not the port.

terpstra:
My more pressing concern, though, is why we're doing the write combining in AXI and not TL? And also, should the presence of an adapter even modify the slave parameters at all? Does the existence of a write combiner actually modify the slave in a way that masters need to know about? I'd wager the answer is: no. It should be transparent.

solomatnikov:
WCB is in TL, not in AXI.

This is just a parameter to let dongle know to instantiate WCB before converter for AXI slave port.

terpstra:
Right. We don't generally use diplomacy to parameterize things like that. Diplomatic parameters are supposed to describe what slaves/masters ARE so that you can talk to them correctly, not what you want to instantiate. This sort of configuration you are talking about should be made by the dongle traits case class. Talk to @hcook about how he would like this parameterized.

solomatnikov:
I did talked to @hcook and that's what he suggested

terpstra:
Patch will be applied to another repository.

PR #825, opened 2017-06-27
Debug + WFI Interactions

mwachs5:
Fixes some interactions between WFI instruction and Debug Mode


src/main/scala/rocket/CSR.scala
```diff
@@ -214,6 +212,7 @@ class CSRFile(perfEventSets: EventSets = new EventSets(Seq()))(implicit p: Param
   reset_dcsr.xdebugver := 1
   reset_dcsr.prv := PRV.M
   val reg_dcsr = Reg(init=reset_dcsr)
+  val reg_debugint = Reg(init=Bool(false), next=io.interrupts.debug)
```
aswaterman: For symmetry with all other interrupts I don't think this should be reset

mwachs5: I was just taking it from its previous functionality where reset_dcsr was 0. But I agree it shouldn't be reset.

mwachs5:
This last change has exposed a bug in FESVR.

PR #827, opened 2017-06-28
Dts improvements

terpstra:
This adds compatible and reg-names to most everything.

### DIFF
--- a/src/main/scala/uncore/devices/Rom.scala
+++ b/src/main/scala/uncore/devices/Rom.scala
@@ -13,7 +13,7 @@ import uncore.util._
 import config._
 
 class TLROM(val base: BigInt, val size: Int, contentsDelayed: => Seq[Byte], executable: Boolean = true, beatBytes: Int = 4,
-  resources: Seq[Resource] = new SimpleDevice("rom", Nil).reg)(implicit p: Parameters) extends LazyModule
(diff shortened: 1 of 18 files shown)


src/main/scala/uncore/devices/Rom.scala
```diff
@@ -13,7 +13,7 @@ import uncore.util._
 import config._
 
 class TLROM(val base: BigInt, val size: Int, contentsDelayed: => Seq[Byte], executable: Boolean = true, beatBytes: Int = 4,
-  resources: Seq[Resource] = new SimpleDevice("rom", Nil).reg)(implicit p: Parameters) extends LazyModule
+  resources: Seq[Resource] = new SimpleDevice("rom", Nil).reg("mem"))(implicit p: Parameters) extends LazyModule
```
richardxia: Do we not want to give this one a `compatible` string since it's just a basic memory?

src/main/scala/uncore/tilelink2/SRAM.scala
```diff
@@ -14,7 +14,7 @@ class TLRAM(address: AddressSet, executable: Boolean = true, beatBytes: Int = 4,
   val node = TLManagerNode(Seq(TLManagerPortParameters(
```
richardxia: Same question about providing a `compatible` string for this device.

terpstra: Both of those should indeed have a compatible field set. I've added it.

PR #828, opened 2017-06-28
diplomacy: a type of connect that always disables monitors

hcook:


src/main/scala/diplomacy/Nodes.scala
```diff
@@ -18,7 +18,8 @@ trait InwardNodeImp[DI, UI, EI, BI <: Data]
   def bundleI(ei: EI): BI
   def colour: String
   def reverse: Boolean = false
-  def connect(edges: () => Seq[EI], bundles: () => Seq[(BI, BI)])(implicit p: Parameters, sourceInfo: SourceInfo): (Option[MonitorBase], () => Unit) = {
+  def connect(edges: () =>  Seq[EI], bundles: () => Seq[(BI, BI)], enableMonitoring: Boolean)
```
colinschmidt: Not that it maters but you put an extra space before Seq[EI]

src/main/scala/rocket/Frontend.scala
```diff
@@ -59,7 +59,7 @@ class Frontend(val icacheParams: ICacheParams, hartid: Int)(implicit p: Paramete
   val masterNode = TLOutputNode()
   val slaveNode = TLInputNode()
 
-  icache.slaveNode.map { _ := slaveNode }
+  icache.slaveNode.map { _ connectButDontMonitor slaveNode }
```
colinschmidt: Why is this one no longer being monitored? The other place monitors are disabled has a nice comment.

src/main/scala/rocketchip/Periphery.scala
```diff
@@ -349,8 +349,7 @@ trait HasPeripheryErrorSlave extends HasSystemNetworks {
 
   // Override the default Parameters to exclude the TLMonitor between the Fragmenter and error slave.
```
colinschmidt: This comment could be updated now that its not an "override" of parameters.

"Most slaves...than others, therefore we exclude the TLMonitor."

hcook:
i wanted to make `:/!/=" but @terpstra said i couldnt :(

PR #829, opened 2017-06-29
Update README_TRAVIS.md

mwachs5:
I don't think there is any reason to delete the branch caches, especially since we now allow merges to master without merging/rebasing off all prior commits.


mwachs5:
See #831 , closing this PR in favor of that solution.

Issue #831, opened 2017-06-29
Make it so we don't need to do the travis hackery ever again


mwachs5:
branches:
  only:
    - "master"

Apparently we can add that to our travis.yml and then we dont need to keep switching the "build branch updates" toggle:

(from this thread: https://github.com/travis-ci/travis-ci/issues/1147)

terpstra:
You would still end up having to run travis twice for every PR. Once in push and once on merge. Given the limited build power, I'm not sure this is a good plan.

mwachs5:
You could always kill the merge job... Now that you are allowing merge commits which are true merges, I don't think that's really a bad idea.

terpstra:
If we had infinite build jobs, I would totally be in favour of this PR. I suppose it won't bother anyone with Travis admin privileges, because they can cancel it, as you suggest. I dunno. I abstain from reviewing this PR. Let's see what @aswaterman and @hcook think.

mwachs5:
Yeah. I think the other right fix is to find a way to make the riscv-tools bump less painful. Anyway I don't feel too strongly about it, just googling while waiting for runs to pass :)

PR #832, opened 2017-06-29
travis: add a branch whitelist with just 'master'.

mwachs5:
Travis allows us to build on branch updates and on PRs. Right now our process is to build always on PRs, but to manually only build 'master' when we know that riscv-tools has been bumped (because that rebuilds the master cache which holds riscv-tools). But this is an annoying and error-prone processes that requires extra admin permissions.

With this change, i think we can just leave "Build Branch Updates" to "ON" and get the same effect we are currently doing manually, because only on update to master branch will it do a build. PRs to master branch will get a build. PRs to other branches will I believe NOT get a build. We can add more to the whitelist if that's a problem.

Addresses #831

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,6 +40,18 @@ branches:
     - hwacha
     - boom
     - /^hurricane.*$/
+    
+# These branches are the only ones that
+# will build when "build branch updates"
+# is set in settings (branches which PR against
+# them are still built). With this set, 
+# the above blacklist is not useful.
+# Adding this allows us to keep "Build Branch Updates" 
+# set to 'ON'.
+
+branches:
+  only:
+  - master
 
 jobs:
   include:


terpstra:
This will still mean every successful PR builds twice. Once on the PR and once on the merge. We have pretty limited build resources. Is this the best use of them?

mwachs5:
Alright, but I think this travis change doesn't actually hurt anything. If our model is that you still have to go flick the switch to ON, this would then actually save resources so that other branches don't get built in the meantime.

mwachs5:
I wonder if we can only do the "build cache" steps for branch updates, and something else for PRs.

terpstra:
I agree that this doesn't hurt as long as we leave build merges off and can help. So I'll approve it on that basis. If there were some way to run only the first two steps of the build on branch builds, that would be amazing!

colinschmidt:
To me this change seems neutral in terms of build resources assuming everyone who bumps the tools follows the travis readme. And removing the manual switch flicking seems like a plus, so lgtm.

I think it would be tricky to have the tools/cache build only happen on branch updates because then you would potentially using the wrong tools during the test steps, or not running the tests after a merge.
I think the optimal thing would be able to move the tools cache from the PR to master on a merge, but I doubt thats possible, I'll browse a bit.

mwachs5:
This was merged with #842

PR #836, opened 2017-06-30
Excise the last instance of run-bmarks-test

ben-k:
I never quite understood why it used to be `run-bmarks-test` (as compared to `run-asm-tests`), but it looks like it's been changed everywhere else.


colinschmidt:
I believe this is one of those "lost to antiquity" things.

hcook:
that drove me mildly insane for literally years 😬

PR #837, opened 2017-06-30
plic: Recode to use OH knowledge

mwachs5:
This reduces the lines of code in the PLIC by 30%. Not yet tested.

### DIFF
--- a/src/main/scala/uncore/devices/Plic.scala
+++ b/src/main/scala/uncore/devices/Plic.scala
@@ -176,11 +176,11 @@ class TLPLIC(params: PLICParams)(implicit p: Parameters) extends LazyModule
       PLICConsts.enableBase(i) -> e.map(b => RegField(1, b))
     }
 
-    val claimer = Wire(init = Vec.fill(nHarts){Bool(false)})
-    val claiming = Wire(init = Vec.tabulate(nHarts){i => Mux(claimer(i), UIntToOH(maxDevs(i)), UInt(0))})
-    val claimedDevs = Wire(init = Vec(claiming.reduceLeft( _ | _ ).toBools))
+    val claimer = Wire(Vec(nHarts, Bool()))
+    val claiming = Vec.tabulate(nHarts){i => Mux(claimer(i), UIntToOH(maxDevs(i), nDevices+1), UInt(0))}
+    val claimedDevs = Vec(claiming.reduceLeft( _ | _ ).toBools)
 
-    for ((p, g), c) <- (pending zip gateways) zip claimedDevs) {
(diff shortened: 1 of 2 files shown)


src/main/scala/uncore/devices/Plic.scala
```diff
@@ -182,14 +176,28 @@ class TLPLIC(params: PLICParams)(implicit p: Parameters) extends LazyModule
       PLICConsts.enableBase(i) -> e.map(b => RegField(1, b))
     }
 
+    val claimer = Wire(init = Vec.fill(nHarts){Bool(false)})
+    val claiming = Wire(init = Vec.tabulate(nHarts){i => Mux(claimer(i), UIntToOH(maxDevs(i)), UInt(0, width=log2Up(nDevices+1)))})
+    val claimedDevs = Wire(init = Vec(claiming.reduceLeft( _ | _ ).toBools))
+
+    for ((pg, c) <- (pending zip gateways) zip claimedDevs) {
```
aswaterman: ```for (((p, g), c) <- (pending zip gateways) zip claimedDevs) {```

then you don't need the ```val p``` and ```val g```

src/main/scala/uncore/devices/Plic.scala
```diff
@@ -182,14 +176,28 @@ class TLPLIC(params: PLICParams)(implicit p: Parameters) extends LazyModule
...
+      val p = pg._1
+      val g = pg._2
+      g.ready := !p
+      g.complete := false
+      when(c) {
+        p := false
+      }.elsewhen (g.valid) {
+        p := true
+      }
```
aswaterman: Given that the goal is to reduce line count, it's worth seeing if this does better:

    when (c || g.valid) { p := !c }

mwachs5: I tried it out, overall LoC is not affected, they just get turned from if-else into assigns.

mwachs5: Actually I was wrong, it does help the LoC considerably.

mwachs5:
This passes tests.

src/main/scala/uncore/devices/Plic.scala
```diff
@@ -182,27 +176,60 @@ class TLPLIC(params: PLICParams)(implicit p: Parameters) extends LazyModule
       PLICConsts.enableBase(i) -> e.map(b => RegField(1, b))
     }
 
+    // When a hart reads a claim/complete register, then the
+    // device which is currently its highest priority is no longer pending.
+    // This code expolits the fact that, practically, only one claim/complete
```
colinschmidt: expolits -> exploits

PR #838, opened 2017-06-30
plic: Use same recoding technique on complete as well as claim

mwachs5:
I tried the same trick on the 'completers' and dropped another 10% in output LoC. The if-elses are still pretty ugly for the completers so I didn't push it directly to the 'recode' PR #837 

This is also yet to be tested.


mwachs5:
Passes tests

mwachs5:
I am just going to merge this and fix it in #837. Will squash all these down in the end.

PR #839, opened 2017-06-30
Bump firrtl

jackkoenig:
~Making sure this Firrtl branch doesn't break things~

This is now bumping Firrtl to master, includes const prop registers and better names but no cross module boundaries constant prop yet


mwachs5:
giving this a try now

PR #840, opened 2017-07-01
Make sure that DCache s2_xcpt data scratchpad case is assigned to after initial assignment.

richardxia:
This is the patch that Yunsup asked me to make to fix the issue I was seeing where access exceptions weren't being properly propagated.


richardxia:
Removed the "Do not merge" label, and will merge when this passes Travis.

PR #842, opened 2017-07-03
Bump FESVR for multi-core support

mwachs5:


mwachs5:
By the way @terpstra , @colinschmidt , I set the "build Branch Updates" switch to "ON" before making this PR. I think that means that it won't have to build again on master, but who knows.

PR #844, opened 2017-07-04
travis: do the two cache prep stages in parallel

mwachs5:
Originally we broke these up into two stages because running them in series in one job made it take > 2 hours. But I'm wondering if we can't just do this part in parallel.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -55,14 +55,14 @@ branches:
 
 jobs:
   include:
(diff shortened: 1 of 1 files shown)


terpstra:
This won't work. We had this before. The problem is that caches are grabbed at the start and saved at the end. If you do it how you've done it, it will only save one of the results.

PR #849, opened 2017-07-06
L1 memory system improvements

aswaterman:
- Handle TL errors in I$
- Fix D$ probe bug
- Add option for shared L2 TLB (disabled by default)

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -157,8 +157,8 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
   val accruedRefillError = Reg(Bool())
   val refillError = tl_out.d.bits.error || (refill_cnt > 0 && accruedRefillError)
   when (refill_done) {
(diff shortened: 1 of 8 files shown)


src/main/scala/rocket/ICache.scala
```diff
@@ -164,6 +174,7 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
   }
 
   val s1_tag_disparity = Wire(Vec(nWays, Bool()))
+  val s1_tlError = Wire(Vec(nWays, Bool()))
```
hcook: The casing of this variable name makes me want to cry. Personally I'm in favor of continuing the convention of `naming_wires_with_underscores` as opposed to `namingAbstractScalaObjectsWithCamelCase`. If you really want to switch everything to camel case then fine, but you really can't have both in one name XD

aswaterman: OK

aswaterman:
Yes @terpstra, I believe that was the problem.

PR #850, opened 2017-07-07
PLIC: (undefZero=true) Don't allow addresses to alias

mwachs5:
While the spec is unclear what happens when you access unused registers in the PLIC, for simplicity turn off register aliasing. If this becomes a performance/area issue we can revisit.


mwachs5:
True, but then I'd lose the blame record ...?

PR #854, opened 2017-07-10
regmapper: add byte-sized RegField helper function

terpstra:
Create byte-sized read-write RegFields out of a large UInt register.
It is updated when any of the bytes are written. Because the RegFields
are all byte-sized, this is also suitable when a register is larger
than the intended bus width of the device (atomic updates are impossible).

### DIFF
--- a/src/main/scala/regmapper/RegField.scala
+++ b/src/main/scala/regmapper/RegField.scala
@@ -77,6 +77,7 @@ case class RegField(width: Int, read: RegReadFn, write: RegWriteFn, name: String
 {
   require (width > 0)
   def pipelined = !read.combinational || !write.combinational
+  def readOnly = this.copy(write = ())
 }
 
 object RegField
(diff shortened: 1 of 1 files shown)


mwachs5:
Does this allow a processor to do byte access to something that is defined in the device as a multiple-byte UINT? If so I could use this in Debug Module to simplify the code significantly.

terpstra:
That is the intent. Yes. It breaks atomicity of the register updates, though, I guess you never had this?

mwachs5:
In debug module, there are registers which are only accessed by Debugger as 32-bit words, but can be accessed by processor as bytes. So for the former I basically just had 4x as many signals as I needed (I need to know when these are written so I have a 'notify' when the debugger writes them, so I had 4 signals instead of 1 which was just useless logic). 

From the processor side, it's OK if the accesses are not atomic.

terpstra:
Oh. I see. To get the 'notify' part, you would need to return the valids.reduce from the function as well. Feel free to embrace and extend. ;)

mwachs5:
Yeah I already have the notify signals generated on the should-be-a-Uint side, so I think this would be perfect for this.

PR #861, opened 2017-07-13
tile: disable all monitors inside the tile

hcook:
TLMonitors keep breaking deduplication when e.g. ITIMs are enabled. For now, disable all Monitors in all  tiles.


terpstra:
I think this is fine for using for a quick release, but should not be merged to master?

PR #869, opened 2017-07-17
Fix the TLFuzzer IO

zhemao:
If you don't mark "finished" specifically as output, it screws up inter-operation with code written in chisel3.


terpstra:
I'm surprised to hear this is a problem. This is by far not the only place in our codebase where we use the fact that the default direction is output.

PR #886, opened 2017-07-25
chiplink: adjust bus view to include the splitter

terpstra:


terpstra:
This should also fix your fadu problem, @hcook

PR #887, opened 2017-07-25
coreplex: retire RTCPeriod & introduce PeripheryBusParams.frequency

yunsup:
This PR is the first step to refactor devices that need to know the PeripheryBus frequency to setup the default divisor value (e.g., UART).

### DIFF
--- a/src/main/scala/coreplex/RTC.scala
+++ b/src/main/scala/coreplex/RTC.scala
@@ -3,21 +3,22 @@
 package freechips.rocketchip.coreplex
 
 import Chisel._
-import freechips.rocketchip.config.Field
-import freechips.rocketchip.diplomacy.LazyMultiIOModuleImp
+import freechips.rocketchip.diplomacy.{LazyMultiIOModuleImp, DTSTimebase}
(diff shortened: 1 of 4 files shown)


src/main/scala/coreplex/RTC.scala
```diff
@@ -3,21 +3,18 @@
...
-case object RTCPeriod extends Field[Option[Int]]
-
 trait HasRTCModuleImp extends LazyMultiIOModuleImp {
   val outer: HasPeripheryClint
-  private val internalPeriod: Option[Int] = outer.p(RTCPeriod)
-  require(internalPeriod.isDefined, "RTCPeriod is not defined")
+  private val pbusFreq = outer.p(PeripheryBusParams).frequency
+  private val rtcFreq = outer.p(DTSTimebase)
+  require(pbusFreq >= rtcFreq)
+  private val internalPeriod: BigInt = pbusFreq / rtcFreq
```
terpstra: I'd add a check here that the integer division is within 5% of the real division

yunsup: OK

PR #888, opened 2017-07-25
tilelink: fix Fragmenter source re-use bug

terpstra:
Consider the following waveform for two 4-beat bursts:
```
---A----A------------
-------D-----DDD-DDDD
Under TL rules, the second A can use the same source as the first A,
because the source is released for reuse on the first response beat.

However, if we fragment the requests, it looks like this:
---3210-3210---------
-------3-----210-3210
... now we've broken the rules because 210 are twice inflight.
```
To solve this, we alternate an a.source bit every time D completes a txn.


terpstra:
The bug affected also the NON-earlyAck case. This fix eliminates the per-source register overhead, fixing the bug in non-earlyack and also making it possible to use earlyAck in more places for free!

PR #889, opened 2017-07-25
Acquire before release; jump in frontend

aswaterman:
Enabling acquire-before-release will break for now, because of a limitation in TLBroadcast, to be fixed soon by @terpstra


src/main/scala/rocket/Events.scala
```diff
@@ -10,9 +10,13 @@ class EventSet(gate: (UInt, UInt) => Bool, events: Seq[(String, () => Bool)]) {
   def size = events.size
   def hits = events.map(_._2()).asUInt
   def check(mask: UInt) = gate(mask, hits)
+  def dump() {
+    for (((name, _), i) <- events.zipWithIndex)
+      when (check(1.U << i)) { printf(s"Event $name\n") }
+  }
```
hcook: cute

src/main/scala/rocket/HellaCache.scala
```diff
@@ -26,6 +26,7 @@ case class DCacheParams(
     nRPQ: Int = 16,
     nMMIOs: Int = 1,
     blockBytes: Int = 64,
+    acquireBeforeRelease: Boolean = false,
```
hcook: Shouldn't this default to true (if it's temporarily broken)?

PR #899, opened 2017-07-28
Stabilize tile wrappers for downstream tools

hcook:
The wrappers around the tiles needs to have consistent names, monitors, and boundary buffers to placate downstream toolflows.

### DIFF
--- a/src/main/scala/coreplex/RocketCoreplex.scala
+++ b/src/main/scala/coreplex/RocketCoreplex.scala
@@ -57,8 +57,6 @@ trait HasRocketTiles extends HasSystemBus
       }
       case RationalCrossing(direction) => {
         val wrapper = LazyModule(new RationalRocketTile(c, i)(pWithExtra))
-        val sink = LazyModule(new TLRationalCrossingSink(direction))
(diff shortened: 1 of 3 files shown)


src/main/scala/coreplex/RocketCoreplex.scala
```diff
@@ -57,13 +57,12 @@ trait HasRocketTiles extends HasSystemBus
...
       case RationalCrossing(direction) => {
         val wrapper = LazyModule(new RationalRocketTile(c, i)(pWithExtra))
-        val sink = LazyModule(new TLRationalCrossingSink(direction))
-        val source = LazyModule(new TLRationalCrossingSource)
         sbus.fromRationalTiles(direction) :=* wrapper.masterNode
         wrapper.slaveNode :*= pbus.toRationalSlaves
         wrapper
       }
     }
+    wrapper.suggestName("tile") // Try to stabilize this name for downstream tools
```
colinschmidt: Should you suggest ```tile_$i```? Or does suggestName just do that automatically?

hcook: automatically

PR #904, opened 2017-07-29
Fix D$ ready-valid signaling bug

aswaterman:


yunsup:
However, travis didn't pass, but it seems like a different problem.  @terpstra can you check this out?

terpstra:
I told you about this yesterday. I strengthened the corner case stressing ability of the TileLink fuzzer and it turns up an occasional issue in these unit tests. Just rerun them. The failure is random. I will debug it after we deal with all the other problems. The particular corner case involved will almost certainly not affect any of our current designs.

yunsup:
Yeah, just wanted to make sure this was the problem you told me.

PR #905, opened 2017-07-31
test to add buffering to tile

yunsup:


yunsup:
Don't merge this branch.  I'm going to make the buffering optional.  I'll follow up with a different PR.

hcook:
When making the buffering optional, consider whether this is something that should apply to the other types of tile crossings as well. I'm not sure whether it should be associated with `CoreplexCrossing` or `RocketTileParams`, but am leaning towards the latter because the former is used in more places

terpstra:
It does NOT apply to an async crossing.

I think it does apply to rational an synchronous connections.

yunsup:
@terpstra that is correct.  @hcook I wanted to put the option as part of the RocketTileParams, and only put the buffering in rational and synchronous rocket tiles.  I'm going to close this PR, and make another one soon.  Thanks everyone.

yunsup:
now #909

PR #906, opened 2017-07-31
Mitigate I$->D$->I$ critical path

aswaterman:
This seemingly irrelevant change shaves several gate delays off the I$ tl.a.valid path.


aswaterman:
Also pushed to test-fq2 branch.

aswaterman:
thanks @yunsup, we can do a design review after the tapeout

PR #912, opened 2017-08-01
tilelink: add mask rom

yunsup:


scripts/vlsi_rom_gen
```diff
@@ -0,0 +1,134 @@
+#!/usr/bin/env python3
```
richardxia: I think nothing in this file relies on Python 3-only features, so it might be better to make this Python 2-compatible to save future headaches with users of this repo.

If you do choose to make this Python 2-compatible, you'll want to add the following 2- and 3-compatible shim imports:

```python
from __future__ import division
from __future__ import print_function
from __future__ import unicode_literals
```

yunsup: Thanks, taking your comments.

PR #914, opened 2017-08-02
Fix some critical paths

aswaterman:


aswaterman:
Need to set DCacheParams.retimeWayMux to take advantage of the last commit

PR #918, opened 2017-08-03
Icache prefetch

aswaterman:


src/main/scala/rocket/TLB.scala
```diff
@@ -226,6 +229,7 @@ class TLB(lgMaxSize: Int, nEntries: Int)(implicit edge: TLEdgeOut, p: Parameters
   io.resp.ma.st := (ma_st_array & hits).orR
   io.resp.ma.inst := false // this is up to the pipeline to figure out
   io.resp.cacheable := (c_array & hits).orR
+  io.resp.prefetchable := (prefetchable_array & hits).orR && edge.manager.managers.forall(m => !m.supportsAcquireB || m.supportsHint)
```
hcook: why not only look at `m.supportsHint`?

aswaterman: I didn't want to need to store another bit in the TLB; I wanted to reuse the cacheability property, which is already stored in the TLB.

I can use cacheability as a proxy for prefetchability if all cacheable regions support hints (as is the case when there's an L2).

PR #925, opened 2017-08-06
diplomacy: provide default clock/reset for LazyRawModuleImp

yunsup:
so that we can instantiate LazyModules within LazyRawModules.


yunsup:
please make a merge commit for this

PR #926, opened 2017-08-06
bump riscv-tools

yunsup:


yunsup:
is there anything special I need to do for riscv-tools caching?

palmer-dabbelt:
I think you need to turn on the build for PRs and then clear the cache?

hcook:
https://github.com/freechipsproject/rocket-chip/blob/master/README_TRAVIS.md

yunsup:
thanks @hcook

aswaterman:
I'm looking into the failure

PR #939, opened 2017-08-08
tilelink: PMP controlled BusBlocker prevents bus accesses

terpstra:
Sometimes you just need to put IP in a cage.

### DIFF
--- a/src/main/scala/devices/tilelink/BusBlocker.scala
+++ b/src/main/scala/devices/tilelink/BusBlocker.scala
@@ -62,10 +62,9 @@ object DevicePMP
   def apply(addressBits: Int) = {
     val out = Wire(new DevicePMP(DevicePMPParams(addressBits)))
     out.l := UInt(0)
-    out.a := UInt(1) // TOR
+    out.a := UInt(0)
     out.r := UInt(0)
     out.w := UInt(0)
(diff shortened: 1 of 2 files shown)


src/main/scala/devices/tilelink/BusBlocker.scala
```diff
@@ -0,0 +1,108 @@
...
+      }))
+    Seq(
+      RegField(10),
+      field(params.addressBits-12, addr_hi),
+      RegField(56 - (params.addressBits-2)),
+      field(1, r),
+      field(1, w),
+      RegField(1), // x
+      field(1, a),
+      RegField(3), // a high + 2 reserved
```
aswaterman: RegField(2)

src/main/scala/devices/tilelink/BusBlocker.scala
```diff
@@ -0,0 +1,108 @@
...
+        when (wen && !locked) { reg := data }
+        Bool(true)
+      }))
+    Seq(
+      RegField(10),
+      field(params.addressBits-12, addr_hi),
+      RegField(56 - (params.addressBits-2)),
+      field(1, r),
+      field(1, w),
+      RegField(1), // x
```
aswaterman: RegField(1), // m

PR #942, opened 2017-08-08
Bus blocker lock

terpstra:
d'oh!


aswaterman:
So just to confirm the functionality:

- pmp(i).l locks pmp(i).everything
- pmp(i+1).l locks only pmp(i).address

terpstra:
Err. I had pmp(i+1).l locking everything of pmp(i). What is the intended behaviour?

terpstra:
Ok. I've pushed a version that only protects the PRIO address if l AND a are true.

terpstra:
Done.

PR #943, opened 2017-08-09
Fix IBuf bug

aswaterman:
Don't examine a packet's xcpt signal if it might be invalid.  In this case,
the correct fix is to not examine xcpt at all; the deleted code was vestigial.
(Note, the other use of xcpt(j+1) in this code is indeed safe.)


yunsup:
restarting travis build

PR #944, opened 2017-08-09
memgen: also randomize ren and rand register

yunsup:


aswaterman:
sorry for the headache @terpstra

aswaterman:
I'll just make the change, to speed this along

PR #951, opened 2017-08-11
Enable building maskrom and bootrom code from src

colinschmidt:
The bootrom code is now parameterized to be able to
build bootrom or maskrom style images.
You can mark periphery mask rom's as bootable and
the bootrom is now optional to allow the maskrom to
take its place.
Also a small verilator fix to the vlsi_rom_gen script


colinschmidt:
@hcook Would you take a look at this when you get the chance.

src/main/scala/devices/tilelink/BootROM.scala
```diff
@@ -61,25 +61,31 @@ class TLROM(val base: BigInt, val size: Int, contentsDelayed: => Seq[Byte], exec
...
-    val romdata = Files.readAllBytes(Paths.get(params.contentFileName))
-    val rom = ByteBuffer.wrap(romdata)
-    rom.array() ++ dtb.contents
-  }
-  def resetVector: BigInt = params.hang
 
-  val bootrom = LazyModule(new TLROM(params.address, params.size, contents, true, pbus.beatBytes))
+  def resetVector: Option[BigInt] = params.map(_.hang)
+
+   private lazy val contents = params.map{ params =>
```
hcook: `params.map { params =>`

src/main/scala/system/Configs.scala
```diff
@@ -31,10 +31,19 @@ class BaseConfig extends Config(new BaseCoreplexConfig().alter((site,here,up) =>
   case ExtIn  => SlavePortParams(beatBytes = 8, idBits = 8, sourceBits = 4)
   // Additional device Parameters
   case ErrorParams => ErrorParams(Seq(AddressSet(0x3000, 0xfff)))
-  case BootROMParams => BootROMParams(contentFileName = "./bootrom/bootrom.img")
+  case BootROMParams => None
+  case MaskROMParams => Seq.empty
```
hcook: prefer `Nil`?

colinschmidt:
rebased for @yunsup

kammoh:
Any reasons why this was never merged? Is there currently any way to generate a MaskROM as the BootROM?

PR #954, opened 2017-08-13
Add a +max-core-cycles PlusArg

mwachs5:
The current TestDriver uses a +max-cycles plus arg to cleanly exit simulations after a certain number of simulation cycles. This works fine if your Core Clock is the same as your TestDriver clock, but this is not always the case. This PR adds an assertion within the core so that the max-core-cycles can be specified, which is always the same as what shows up in the processor trace.

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -702,6 +702,13 @@ class Rocket(implicit p: Parameters) extends CoreModule()(p)
          wb_reg_cinst, wb_reg_cinst)
   }
 
+  val max_core_cycles = PlusArg("max-core-cycles",
+    default = 0,
+    docstring = "Maximum Core Clock cycles simulation may run before timeout. Ignored if 0 (Default).")
+  when (max_core_cycles > UInt(0)) {
+    assert (csr.io.time < max_core_cycles, "Maximum Core Cycles reached.")
+  }
+
   def checkExceptions(x: Seq[(Bool, UInt)]) =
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/RocketCore.scala
```diff
@@ -702,6 +702,13 @@ class Rocket(implicit p: Parameters) extends CoreModule()(p)
          wb_reg_cinst, wb_reg_cinst)
   }
 
+  val max_core_cycles = PlusArg("max-core-cycles",
+    default = 0,
+    docstring = "Maximum Core Clock cycles simulation may run before timeout. Ignored if 0 (Default).")
+  when (max_core_cycles > UInt(0)) {
+    assert (csr.io.time < max_core_cycles, "Maximum Core Cycles reached.")
+  }
```
aswaterman: If max_core_cyles is 0, then csr.io.time can never be less than max_core_cycles, so I believe you don't need the when, just the assert.

terpstra: Pretty sure megan has it right.

aswaterman: Yeah, it's correct as-is.  I was just saying that the when is unnecessary, since the when is only false when max_core_cycles is 0, in which case the assertion is always false, too.

aswaterman: oops

aswaterman: I got the polarity wrong, my bad

PR #955, opened 2017-08-14
Fix acquire before release

aswaterman:


src/main/scala/rocket/DCache.scala
```diff
@@ -90,8 +90,14 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   val q_depth = if (rational) (2 min maxUncachedInFlight-1) else 0
   val tl_out_a = Wire(tl_out.a)
   tl_out.a <> (if (q_depth == 0) tl_out_a else Queue(tl_out_a, q_depth, flow = true))
-  val tl_out_c = Wire(tl_out.c)
-  tl_out.c <> (if (cacheParams.acquireBeforeRelease) Queue(tl_out_c, cacheDataBeats, flow = true) else tl_out_c)
+  val (tl_out_c, release_queue_empty) =
+    if (cacheParams.acquireBeforeRelease) {
+      val q = Module(new Queue(tl_out.c.bits, cacheDataBeats, flow = true))
+      tl_out.c <> q.io.deq
+      (q.io.enq, q.io.count === 0)
```
terpstra: io.count latency is super bad. any reason you don't use !io.enq.valid ? I constantly have had to avoid count on critical paths in the L2.

aswaterman: Yes, I thought about that. It's a flow queue, so deq.valid is much worse.

It's only a few gate delays in this case: 2-bit comparators/subtractors are fine...

terpstra: If it is flow, you're right. Why is it flow tho? Release latency is irrelevant. In fact, it's beneficial in case A is slow.

PR #959, opened 2017-08-16
bump riscv-tools to get updated debug tools

mwachs5:


mwachs5:
Unfortunately this version of riscv-tools isn't yet passing its own regression

mwachs5:
Will make a new PR with newer versions of tools

PR #965, opened 2017-08-21
async_reset_reg: Squash X's the same as for synchronous reg

mwachs5:
This adds the same code emitted by FIRRTL for squashing Xs in registers, memories, etc to the AsyncResetReg black box, so that they have the same behavior. This is important for simulations which may not actually assert reset at time 0, regardless of the clocking.

### DIFF
--- a/vsrc/AsyncResetReg.v
+++ b/vsrc/AsyncResetReg.v
@@ -49,18 +49,18 @@ module AsyncResetReg (
    integer                       initvar;
    reg [31:0]                    _RAND;
    initial begin
- `ifndef verilator
+`ifndef verilator
       #0.002 begin end
- `endif
- `ifdef RANDOMIZE_REG_INIT
+`endif
+`ifdef RANDOMIZE_REG_INIT
       _RAND = {1{$random}};
       q = _RAND[0];
- `endif
+`endif
    end
 `endif //  `ifdef RANDOMIZE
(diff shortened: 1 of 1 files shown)


vsrc/AsyncResetReg.v
```diff
@@ -32,6 +45,20 @@ module AsyncResetReg (
                       input      clk,
                       input      rst);
    
+`ifdef RANDOMIZE
+   integer                       initvar;
+   reg [31:0]                    _RAND;
+   initial begin
+ `ifndef verilator
```
richardxia: Is this intentionally indented by one space? And I also just realized that the other, non-macro lines are indented by 3 spaces.

mwachs5: my editor did not like this at all. I will tidy up

mwachs5: Well my editor is apparently pretty happy with 3-space tabs in Verilog files, as thats what it already had in the file. I can unindent the `ifdefs though

richardxia: Well, what do you know? Apparently the default indentation for Verilog files in emacs is 3 spaces: https://github.com/veripool/verilog-mode/blob/c36a88640d2b245e6bf591086186afe03b90805f/verilog-mode.el#L600

PR #968, opened 2017-08-22
firrtl: disable DCE until markDUT works in rocket

terpstra:
@jackkoenig Firrtl is removing ports that the TestHarness does not use. I understand why we want to apply DCE to interior modules, but I don't know how to apply DUT-status to a Module in rocket-chip. I tried applying markDUT! and DONTtouch! by hand, but it didn't work. Surely there should be some way to just markDUT() for rocketchip?


terpstra:
This issue still has not been resolved. Reopening this work-around.

terpstra:
Rebased against current master.

terpstra:
Obsoleted by #1033

PR #971, opened 2017-08-24
Bump chisel3 and firrtl, update plugin versions

jackkoenig:
Don't merge this yet, but this PR should be used ultimately to bump chisel and firrtl

### DIFF
--- a/project/build.properties
+++ b/project/build.properties
@@ -1,1 +1,1 @@
-sbt.version=0.13.12
+sbt.version=0.13.16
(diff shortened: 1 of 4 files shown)


jackkoenig:
Alright it's about time we finally bump chisel and firrtl.

In the near future we should look in to cross compiling rocket-chip to Scala 2.12 as well which chisel and firrtl both do now as well.

jackkoenig:
Alright I finally think this is ready, will merge after Travis

PR #982, opened 2017-09-01
Add a FrontBus to which low-bandwidth bus masters attach

hcook:
"Front" means in front of the system port.

Also added a bunch of `suggestNames` for modules generated as attachment points to the SystemBus. We should probably revisit how this is handled, and do it more programmatically.

### DIFF
--- a/src/main/scala/coreplex/SystemBus.scala
+++ b/src/main/scala/coreplex/SystemBus.scala
@@ -37,8 +37,8 @@ class SystemBus(params: SystemBusParams)(implicit p: Parameters) extends TLBusWr
 
   def toSplitSlaves: TLOutwardNode = outwardSplitNode
 
-  def toPeripheryBus(nBuffers: Int): TLOutwardNode = {
-    val (in, out) = bufferChain(nBuffers, name = Some("pbus"))
+  def toPeripheryBus(addBuffers: Int): TLOutwardNode = {
+    val (in, out) = bufferChain(addBuffers, name = Some("pbus"))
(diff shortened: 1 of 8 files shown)


src/main/scala/coreplex/SystemBus.scala
```diff
@@ -20,61 +20,72 @@ case object SystemBusParams extends Field[SystemBusParams]
...
-  pbusBuffer0.node :*= pbus_fixer.node
-  pbusBuffer1.node :*= pbusBuffer0.node
-  pbusBuffer2.node :*= pbusBuffer1.node
-  pbusBuffer3.node :*= pbusBuffer2.node
-  val toPeripheryBus: TLOutwardNode = pbusBuffer3.node
+  def toPeripheryBus(nBuffers: Int): TLOutwardNode = {
+    val (in, out) = bufferChain(nBuffers, name = Some("PeripheryBus"))
+    in := pbus_fixer.node
+    out
+  }
```
terpstra: This should be nBuffers-1. There is already a buffer before the pbus_fixer.

terpstra:
I'm still pretty skeptical about these multiple buffer chains. Adding one buffer strategically here or there helps. I have yet to see adding more than one buffer improve timing in our designs...

mwachs5:
I agree, but still think TLBufferChain is a useful primitive to have

henrystyles:
Putting multiple 'AXI register slice' IPs back to back is pretty common for FPGA designs that I've seen.

PR #983, opened 2017-09-01
Remove global fields PAddrBits and ResetVectorBits

hcook:
- [x] Remove all instances of `PAddrBits`
- [x] Break the circular dependency on `HasCoreParameters` in  `ScratchpadSlavePort`
- [x] Ensure `ResetVectorBits` is always wide enough

### DIFF
--- a/src/main/scala/coreplex/Configs.scala
+++ b/src/main/scala/coreplex/Configs.scala
@@ -15,11 +15,10 @@ import freechips.rocketchip.util._
 
 class BaseCoreplexConfig extends Config ((site, here, up) => {
   // Tile parameters
-  case PAddrBits => 32
   case PgLevels => if (site(XLen) == 64) 3 /* Sv39 */ else 2 /* Sv32 */
   case ASIdBits => 0
   case XLen => 64 // Applies to all cores
-  case ResetVectorBits => site(PAddrBits)
+  case ResetVectorBits => 32 // TODO: site(SharedMemoryTLEdge).bundle.addressBits
   case MaxHartIdBits => log2Up(site(RocketTilesKey).size)
   case BuildCore => (p: Parameters) => new Rocket()(p)
   case RocketTilesKey =>  Nil // Will be added by partial configs found below
--- a/src/main/scala/coreplex/Configs.scala
+++ b/src/main/scala/coreplex/Configs.scala
@@ -18,7 +18,6 @@ class BaseCoreplexConfig extends Config ((site, here, up) => {
   case PgLevels => if (site(XLen) == 64) 3 /* Sv39 */ else 2 /* Sv32 */
   case ASIdBits => 0
   case XLen => 64 // Applies to all cores
-  case ResetVectorBits => 32 // TODO: site(SharedMemoryTLEdge).bundle.addressBits
   case MaxHartIdBits => log2Up(site(RocketTilesKey).size)
   case BuildCore => (p: Parameters) => new Rocket()(p)
   case RocketTilesKey =>  Nil // Will be added by partial configs found below
--- a/src/main/scala/rocket/ScratchpadSlavePort.scala
+++ b/src/main/scala/rocket/ScratchpadSlavePort.scala
@@ -98,24 +98,23 @@ class ScratchpadSlavePort(address: AddressSet, coreDataBytes: Int, usingAtomics:
 /** Mix-ins for constructing tiles that have optional scratchpads */
 trait CanHaveScratchpad extends HasHellaCache with HasICacheFrontend {
   val module: CanHaveScratchpadModule
-  val xLenBytes = p(XLen)/8
   val cacheBlockBytes = p(CacheBlockBytes)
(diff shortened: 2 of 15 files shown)


aswaterman:
praise allah

src/main/scala/coreplex/Configs.scala
```diff
@@ -15,11 +15,10 @@ import freechips.rocketchip.util._
 
 class BaseCoreplexConfig extends Config ((site, here, up) => {
   // Tile parameters
-  case PAddrBits => 32
   case PgLevels => if (site(XLen) == 64) 3 /* Sv39 */ else 2 /* Sv32 */
   case ASIdBits => 0
   case XLen => 64 // Applies to all cores
-  case ResetVectorBits => site(PAddrBits)
+  case ResetVectorBits => 32 // TODO: site(SharedMemoryTLEdge).bundle.addressBits
```
aswaterman: If this proves difficult to get rid of, it's probably OK to use ```xLen``` for this, then truncate it to ```paddrBits min xLen``` in the Frontend.

aswaterman: (I don't mean ```case ResetVectorBits => site(XLen)```, I mean delete the parameter altogether)

src/main/scala/rocket/ScratchpadSlavePort.scala
```diff
@@ -97,22 +96,25 @@ class ScratchpadSlavePort(address: AddressSet)(implicit p: Parameters) extends L
 }
 
 /** Mix-ins for constructing tiles that have optional scratchpads */
-trait CanHaveScratchpad extends HasHellaCache with HasICacheFrontend with HasCoreParameters {
+trait CanHaveScratchpad extends HasHellaCache with HasICacheFrontend {
   val module: CanHaveScratchpadModule
+  val xLenBytes = p(XLen)/8
```
aswaterman: Since XLen means "length of an x-register" maybe this could be ```xBytes```?  I figure it will show up a lot, so might as well pick a good name now.

src/main/scala/rocket/ScratchpadSlavePort.scala
```diff
@@ -97,22 +96,25 @@ class ScratchpadSlavePort(address: AddressSet)(implicit p: Parameters) extends L
...
+  val cacheBlockBytes = p(CacheBlockBytes)
+  val fetchBytes = tileParams.core.fetchBytes
 
   val slaveNode = TLInputNode() // Up to two uses for this input node:
 
   // 1) Frontend always exists, but may or may not have a scratchpad node
-  val fg = LazyModule(new TLFragmenter(fetchWidth*coreInstBytes, p(CacheBlockBytes), earlyAck=true))
-  val ww = LazyModule(new TLWidthWidget(xLen/8))
+  val fg = LazyModule(new TLFragmenter(fetchBytes, cacheBlockBytes, earlyAck=true)) // TODO: should these be moved outside the tile &
+  val ww = LazyModule(new TLWidthWidget(xLenBytes))                                 //       generated by a new PeripheryBus method?
```
aswaterman: If the tile remains a boundary for hierarchical physical design, then these should stay within the tile, to reduce the number of wires crossing that boundary.

src/main/scala/rocket/TLB.scala
```diff
@@ -13,7 +13,6 @@ import freechips.rocketchip.tile.{XLen, CoreModule, CoreBundle}
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.util._
 
-case object PAddrBits extends Field[Int]
```
aswaterman: Kind of funny this was declared inside the module that implements _virtual_ memory...

terpstra: lol

terpstra:
Thank god.

src/main/scala/rocket/ScratchpadSlavePort.scala
hcook: Depends on whether you expect the PeripheryBus to be wider or narrower than `xBytes`... Also as we discussed I think these should actually both go on some kind of intra-tile front bus

aswaterman: Yeah my misunderstanding is that there wasn't already an intra-tile bus.  That seems to me to be the right decision (in a separate PR).

src/main/scala/coreplex/Configs.scala
hcook: The problem isn't so much inside the tile as it is at the coreplex/system level. Some coreplexes pop this wire out as an external signal, but the width of that signal needs to be based on how much memory the tiles can actually see. I'm not even sure what it should be if e.g. different tiles can see different amounts of memory. It's possibly that having a single global reset vector is actually a misnomer

mwachs5: Agreed. I can think of lots of reasons you may want different reset vector for different cores.

PR #984, opened 2017-09-06
Jtagresettobool - add explicit toBool cast now required on reset.

ucbjrl:
There may be a better way to do this, but this is sufficient to get tests to pass with current chise3.


ucbjrl:
Thanks @mwachs5. @jackkoenig may have a better solution, so I'll leave this branch around until he gets a chance to comment on it.

PR #985, opened 2017-09-06
Add Parameters to diplomatic edges

terpstra:
Put some registers in to help timing


src/main/scala/devices/tilelink/Clint.scala
```diff
@@ -71,8 +71,8 @@ class CoreplexLocalInterrupter(params: ClintParams)(implicit p: Parameters) exte
     val ipi = Seq.fill(nTiles) { RegInit(UInt(0, width = 1)) }
 
     io.int.zipWithIndex.foreach { case (int, i) =>
-      int(0) := ipi(i)(0) // msip
-      int(1) := time.asUInt >= timecmp(i).asUInt // mtip
+      int(0) := ShiftRegister(ipi(i)(0), params.intStages) // msip
```
mwachs5: can you add a comment here that these are NOT for synchronization?

PR #989, opened 2017-09-09
Clean up various Configs/Fields/Params

### LINKED ISSUE #892: Error with "make verilog CONFIG=DefaultRV32Config"
Hi,

after updating to the newest rocket-chip I got this error while trying to build the verilog. In a clean checkout, I did the following to produce the error:
1. cd vsim
2. make verilog CONFIG=DefaultRV32Config

The (most sensible part of the) stack trace is:
Caused by: java.lang.IllegalArgumentException: requirement failed: Xbar data widths don't match: List(ExampleRocketSystem) has 8B vs List(plic, clint, dmInner, bootrom, error) has 4B

Thanks,

Kai Kristian

### LINKED ISSUE #962: Support for underscores in diplomacy.AddressSet
Scala does not natively support underscores in numeric literals, but it is fairly straightforward to add via an implicit class:
https://stackoverflow.com/questions/31275887/underscores-in-numeric-literals-in-scala

Since Diplomacy is dependent on AddressSets/BigInts to create the address map, would it be reasonable to add support for StringContext=>IntContext to the library?

hcook:
Diplomacy now provides the `x"..."` string context, which converts string literals representing hexadecimal numbers into `BigInt`s. You can insert arbitrary underscores, e.g., `x"8000_0000_0000"` instead of `BigInt(0x800000000000)`.

Config now provides optional default values for `Field` objects that are used as lookup keys. However, using the same class names for key objects as the case class name of the value they are holding results in `illegal cyclic reference`, so I'm trying to start walking back that anti-pattern.

Closes #962, closes #892

### DIFF
--- a/src/main/scala/config/Config.scala
+++ b/src/main/scala/config/Config.scala
@@ -5,7 +5,7 @@ package freechips.rocketchip.config
 abstract class Field[T] private (val default: Option[T])
 {
   def this() = this(None)
-  def this(x: T) = this(Some(x))
+  def this(default: T) = this(Some(default))
 }
 
 abstract class View {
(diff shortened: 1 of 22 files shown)


colinschmidt:
cool.

PR #990, opened 2017-09-12
Allow configuration to directly edit BootROM contents

zhemao:
Trying to get multiple cores to boot in TSI. We decided the easiest way was to have the chip generator patch the BootROM contents (loaded from bootrom.img) to insert the number of cores in the system.

This adds a function hook in BootROMParams to make this possible. The function gets passed the ByteBuffer holding the BootROM contents and can edit the buffer in-place.

We're working off an older version of RocketChip, but it should be trivial to rebase this change onto the latest master.

### DIFF
--- a/src/main/scala/devices/tilelink/BootROM.scala
+++ b/src/main/scala/devices/tilelink/BootROM.scala
@@ -14,7 +14,9 @@ case class BootROMParams(
   address: BigInt = 0x10000,
   size: Int = 0x10000,
   hang: BigInt = 0x10040,
-  contentFileName: String)
+  contentFileName: String,
+  patchFunction: ByteBuffer => Unit = (buffer: ByteBuffer) => ())
+
 case object BootROMParams extends Field[BootROMParams]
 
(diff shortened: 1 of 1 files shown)


aswaterman:
Why not have the boot software parse the device tree?  For anything nontrivial you're going to need to do that, anyway.

zhemao:
The code that's using this is in the first-stage bootloader. The functionality we're adding is pretty trivial. Core 0 is just iterating through the other cores and writing a 1 to their MSIP registers. I'd rather not add a DTB parser to the BootROM code just for that.

terpstra:
The way we've done this in our own codebase is to just iterate up to MAX_CORES. The fact that there is a register on the clint that toggles is proof-positive of the existence of a core with that hartid.

aswaterman:
Ah, good point, you can discover the number of harts by finding an MSIP that doesn't toggle.

zhemao:
So you write the MSIP, then read it back to check that it got set?

aswaterman:
Yeah

zhemao:
OK that seems to work.

PR #992, opened 2017-09-13
reset_catch: Allow Test Mode Overrides

mwachs5:
Generally, any generated resets need a test mode override to allow controllability of the resets during post-silicon-testing. Since all resets are (or should be) generated with the ResetCatchAndSync utility, add this functionality there. This is disabled by default so should have no effect until it is turned on with the Config parameter.

### DIFF
--- a/src/main/scala/util/PSDTestMode.scala
+++ b/src/main/scala/util/PSDTestMode.scala
@@ -1,0 +1,19 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import Chisel._
+import freechips.rocketchip.config._
+
+case object IncludePSDTest extends Field[Boolean](false)
+
+class PSDTestModeIO extends Bundle {
+  val test_mode       = Bool(INPUT)
+  val test_mode_reset = Bool(INPUT)
+  // TODO: Clocks?
+}
+
+trait CanHavePSDTestModeIO {
+  implicit val p: Parameters
+  val psd = p(IncludePSDTest).option(new PSDTestModeIO())
+}
--- a/src/main/scala/util/PSDTestMode.scala
+++ b/src/main/scala/util/PSDTestMode.scala
@@ -7,13 +7,13 @@ import freechips.rocketchip.config._
 
 case object IncludePSDTest extends Field[Boolean](false)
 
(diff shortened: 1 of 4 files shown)


src/main/scala/util/ResetCatchAndSync.scala
```diff
@@ -11,27 +11,40 @@ import Chisel._
...
   }
 
   def apply(clk: Clock, rst: Bool, sync: Int, name: String): Bool = apply(clk, rst, sync, Some(name))
   def apply(clk: Clock, rst: Bool, name: String): Bool = apply(clk, rst, name = Some(name))
 
+  def apply(clk: Clock, rst: Bool, sync: Int, name: String, psd_test_mode: Bool, psd_test_reset: Bool): Bool = apply(clk, rst, sync, Some(name),
+    psd_test_mode, psd_test_reset)
+  def apply(clk: Clock, rst: Bool, name: String, psd_test_mode: Bool, psd_test_reset: Bool): Bool = apply(clk, rst, name = Some(name),
+    psd_test_mode = psd_test_mode, psd_test_reset = psd_test_reset)
+
```
hcook: stylistically i'd break this out into
```
 def apply(...): Bool = apply(
  clk = clk,
  rst = rst,
  name = Some(name),
  psd_test_mode = psd_test_mode,
  psd_test_reset = psd_test_reset)
```

src/main/scala/util/PSDTestMode.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import Chisel._
+import freechips.rocketchip.config._
+
+case object IncludePSDTest extends Field[Boolean](false)
+
+class PSDTestModeIO extends Bundle {
+  val test_mode       = Bool(INPUT)
+  val test_mode_reset = Bool(INPUT)
+  // TODO: Clocks?
```
hcook: consider using `with Clocked` if that makes sense

mwachs5: ah, no the comment means that sometimes you may want to provide bypass clocks as well (e.g. for clock muxes). But often those are injected automatically by DFT tools. So it's more like "future work".

PR #994, opened 2017-09-16
Add L1 bus-error unit

aswaterman:
This reports correctable and uncorrectable ECC errors to a unit which sends interrupts to the PLIC.

Some TODOs/requests for feedback:
- Is the integration of this unit in CanHaveScratchpad and RocketTileWrapper done right?
- Should add a crossbar inside CanHaveScratchpad so the ITIM/DTIM/BEU only have one port to the outside.  But since they all have slightly different widths/capabilities, it's not abundantly clear how to do that.
- Need to handle/report TL errors for cached & uncached memory accesses.
- Need to write a software test for the above.

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -36,6 +36,11 @@ class ICacheReq(implicit p: Parameters) extends CoreBundle()(p) with HasL1ICache
   val addr = UInt(width = vaddrBits)
 }
 
+class ICacheErrors(implicit p: Parameters) extends CoreBundle()(p) with HasL1ICacheParameters {
+  val correctable = (cacheParams.tagECC.canDetect || cacheParams.dataECC.canDetect).option(Valid(UInt(width = paddrBits)))
+  val uncorrectable = (cacheParams.itimAddr.nonEmpty && cacheParams.dataECC.canDetect).option(Valid(UInt(width = paddrBits)))
+}
+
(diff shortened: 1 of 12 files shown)


src/main/scala/rocket/ICache.scala
```diff
@@ -48,12 +48,13 @@ class ICache(val icacheParams: ICacheParams, val hartid: Int)(implicit p: Parame
     name = s"Core ${hartid} ICache"))
 
   val size = icacheParams.nSets * icacheParams.nWays * icacheParams.blockBytes
+  val itimAddrSpaceSize = size - icacheParams.nSets / 2 * icacheParams.blockBytes
```
hcook: I support this reallocation of the space associated with the un-TIM-able way, but I'm not confident in this calculation's robustness. What should happen if we start scaling up icache way count for whatever reason? You can allocate N-1 ways, right?

hcook:
`CanHaveScratchpad` and `RocketTileWrapper` integrations look good to me. @terpstra's PR will simplify things further by obviating the need for output nodes.

> Should add a crossbar inside CanHaveScratchpad

I think the width and functionality mismatch can be handled with the right set of adapters, but you will also have to deal with the case where none of the three slave device options are included. Unfortunately we still can't handle xbars with 0 masters/slaves :/

aswaterman:
@hcook added the xbar, can you approve?

Merging is blocked on #1001

PR #998, opened 2017-09-18
Add cover property API that can be refined through Config PropertyLibrary

jchang0:
Also make variable to TLToAXI4 public

### DIFF
--- a/src/main/scala/util/Property.scala
+++ b/src/main/scala/util/Property.scala
@@ -1,0 +1,58 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util.property
+
+import Chisel._
+import chisel3.internal.sourceinfo.{SourceInfo, SourceLine}
+import freechips.rocketchip.config.{Field, Parameters}
+
+case object PropertyLibrary extends Field[BasePropertyLibrary](new DefaultPropertyLibrary)
+
+
+sealed abstract class PropertyType(name: String) {
+  override def toString: String = name
+}
+
+object PropertyType {
+  object Assert extends PropertyType("Assert")
+  object Assume extends PropertyType("Assume")
+  object Cover extends PropertyType("Cover")
+}
+
+trait BasePropertyParameters { 
+  val pType: PropertyType
+  val cond: Bool
+  val label: String
+  val message: String
+}
+
+case class CoverPropertyParameters(
+    cond: Bool,
+    label: String = "",
+    message: String = "") extends BasePropertyParameters {
+  val pType = PropertyType.Cover
+}
+
+abstract class BasePropertyLibrary {
+  def generateProperty(prop_param: BasePropertyParameters)(implicit sourceInfo: SourceInfo)
+}
+
+class DefaultPropertyLibrary extends BasePropertyLibrary {
+  def generateProperty(prop_param: BasePropertyParameters)(implicit sourceInfo: SourceInfo) {
+    // default is to do nothing
+    Unit
+  }
+}
+
+object cover {
+  def apply(cond: Bool)(implicit sourceInfo: SourceInfo, p: Parameters): Unit = {
+    p(PropertyLibrary).generateProperty(CoverPropertyParameters(cond))
+  }
+  def apply(cond: Bool, label: String)(implicit sourceInfo: SourceInfo, p: Parameters): Unit = {
+    p(PropertyLibrary).generateProperty(CoverPropertyParameters(cond, label))
+  }
+  def apply(cond: Bool, label: String, message: String)(implicit sourceInfo: SourceInfo, p: Parameters): Unit = {
+    p(PropertyLibrary).generateProperty(CoverPropertyParameters(cond, label, message))
+  }
+}
+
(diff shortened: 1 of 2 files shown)


src/main/scala/util/Property.scala
```diff
@@ -0,0 +1,45 @@
...
+  override def toString: String = name
+}
+
+object PropertyType {
+  object Assert extends PropertyType("Assert")
+  object Assume extends PropertyType("Assume")
+  object Cover extends PropertyType("Cover")
+}
+
+class BasePropertyParameters(val p_type: PropertyType, val cond: Bool, val label: String, val message: String) {}
```
hcook: i would turn this into a trait (can be extended by case classes):
```
trait BasePropertyParameters {
  val pType: PropertyType
  val cond: Bool
  val label: String
  val message: String
}

case class CoverPropertyParameters(
    cond: Bool,
    label: String = "",
    message: String = "") extends BasePropertyParameters {
  val pType =  PropertyType.Cover
}
```

src/main/scala/util/Property.scala
```diff
@@ -0,0 +1,45 @@
...
+
+class BasePropertyLibrary[PropParam <: BasePropertyParameters] {
+  def generateCover(prop_param: PropParam)(implicit sourceInfo: SourceInfo): Unit = {
+    // Currently empty for user to extend their own cover methods
+    Unit
+  }
+}
```
hcook: i would turn this into an abstract class:
```
abstract class BasePropertyLibrary[PP <: BasePropertyParameters] {
  def generateCover(pp: PP)(implicit sourceInfo: SourceInfo): Unit
}
```
and then maybe add a default concrete class that does something more specific:
```
class PrintCoverPropertyLibrary extends BasePropertyLibrary[CoverPropertyParameters] {
   def generateCover(pp: CoverPropertyParameters)(implicit sourceInfo: SourceInfo) { printf(p.message) }
}
```

src/main/scala/util/Property.scala
```diff
@@ -0,0 +1,45 @@
...
+  }
+  def apply(cond: Bool, label: String)(implicit sourceInfo: SourceInfo, p: Parameters): Unit = {
+    val prop_param = new BasePropertyParameters(PropertyType.Cover, cond, label, "")
+    p(PropertyLibrary).generateCover(prop_param)
+  }
+  def apply(cond: Bool, label: String, message: String)(implicit sourceInfo: SourceInfo, p: Parameters): Unit = {
+    val prop_param = new BasePropertyParameters(PropertyType.Cover, cond, label, message)
+    p(PropertyLibrary).generateCover(prop_param)
+  }
+}
```
hcook: If all `PropertyLibrary` child classes can agree that `generateCover` takes a `CoverPropertyParamers` then you can simplify this to the following
:
```
object cover {
  def apply(cond: Bool)(implicit sourceInfo: SourceInfo, p: Parameters) {
    p(PropertyLibrary).generateCover(CoverPropertyParamers(cond))
  }
  def apply(cond: Bool, label: String)(implicit sourceInfo: SourceInfo, p: Parameters) {
    p(PropertyLibrary).generateCover(CoverPropertyParamers(cond, label))
  }
  def apply(cond: Bool, label: String, message: String)(implicit sourceInfo: SourceInfo, p: Parameters) {
    p(PropertyLibrary).generateCover(CoverPropertyParamers(cond, label, message))
  }
}
```

src/main/scala/util/Property.scala
hcook: If you're returning Unit stylistically omit the `...): Unit = {...}` and just write `...) {...} `

hcook: Should `generateCover` not admit instances of `BasePropertyParameters` where `p_type != Cover`?

src/main/scala/coreplex/Configs.scala
```diff
@@ -27,6 +28,7 @@ class BaseCoreplexConfig extends Config ((site, here, up) => {
   case ErrorParams => ErrorParams(Seq(AddressSet(0x3000, 0xfff)))
   case BootROMParams => BootROMParams(contentFileName = "./bootrom/bootrom.img")
   case DebugModuleParams => DefaultDebugModuleParams(site(XLen))
+  case PropertyLibrary => (new BasePropertyLibrary[BasePropertyParameters])
```
hcook: I wouldn't add a default here until we come up with a specific default behavior applied to instances of `cover(...)` in rocket-chip. See below...

jchang0:
Made the requested changes.

jackkoenig:
This looks good for federation and rocket-chip, but I'm wondering how we should approach it in a more general library that is not beholden to rocket-chips parameterization system. We don't have to solve that yet but just something to think about.

jerryhethatday:
> This looks good for federation and rocket-chip, but I'm wondering how we should approach it in a more general library that is not beholden to rocket-chips parameterization system. We don't have to solve that yet but just something to think about.

Hey Jack, can you clarify in general what this `cover` object does? I saw many locs calling this `cover` functionality. But have no idea what this does. Thanks!

PR #1000, opened 2017-09-19
try to give seqmems clearer names for use with external tools

hcook:
allow the instance name to disambiguate without reference to the full hierarchical path


aswaterman:
@a0u is renaming these ok re: backend flow?

a0u:
@aswaterman They will impact the floorplanning scripts, but in this case updating those should be straightforward.

PR #1004, opened 2017-09-20
Bumpplugins - add sbt-coverage

ucbjrl:


jackkoenig:
Think I can merge this into #971 ? I'm bumping versions there

ucbjrl:
I'd do this merge (the plugins bump) before #971
<https://github.com/freechipsproject/rocket-chip/pull/971>. We'll need to
do another bump to the versions that will work with the version of chisel3
integrated by that PR (probably as part of that PR).

I'd initially bumped all the plugin versions, but that presented problems
with the version of chisel that rocket-chip is using, so I backed off.

jackkoenig:
#971 has the plugin bumps so I think it's fine as is, but we can go ahead and merge this first

ucbjrl:
I didn't realize #971 was so imminent. Your initial observation was correct. This could have been superseded by it.

PR #1012, opened 2017-09-21
Halt and Catch Fire

hcook:
Adds optional support for signaling unrecoverable errors [1] out of the tiles. For now, such a condition can only be triggered by uncorrectable ECC errors occurring in either the DCache or ICache. 

[1] https://en.wikipedia.org/wiki/Halt_and_Catch_Fire


hcook:
It makes the ports. Do we have a test that injects uncorrectable errors?

aswaterman:
The L1 doesn't have error injection but you could manually verify it by inspection by editing the D$ RTL to set io.errors.uncorrectable.valid := true

PR #1014, opened 2017-09-25
Fix omitted parameter

pbing:
Altera Synplify and Synopsis Design Compiler complained.

### DIFF
--- a/vsrc/plusarg_reader.v
+++ b/vsrc/plusarg_reader.v
@@ -2,7 +2,7 @@
 
 // No default parameter values are intended, nor does IEEE 1800-2012 require them (clause A.2.4 param_assignment),
 // but Incisive demands them. These default values should never be used.
-module plusarg_reader #(FORMAT="borked=%d", DEFAULT=0) (
+module plusarg_reader #(parameter FORMAT="borked=%d", DEFAULT=0) (
    output [31:0] out
 );
 


vsrc/plusarg_reader.v
```diff
@@ -2,7 +2,7 @@
 
 // No default parameter values are intended, nor does IEEE 1800-2012 require them (clause A.2.4 param_assignment),
 // but Incisive demands them. These default values should never be used.
-module plusarg_reader #(FORMAT="borked=%d", DEFAULT=0) (
+module plusarg_reader #(parameter FORMAT="borked=%d", DEFAULT=0) (
```
terpstra: Are we sure this is portable? We had 'string FORMAT' before and it had to be removed for vivado.

pbing:
For IEEE 1800-2012 (SystemVerilog) it is true that the 'parameter' keyword can be omitted (6.20.1 Parameter declaration syntax).

But this file has a .v suffix, so both tools assume a Verilog Syntax (IEEE 1364-2005):

A.1.3 Module parameters and ports
```
module_parameter_port_list ::= # ( parameter_declaration { , parameter_declaration } ) 
```

A.2.1.1 Module parameter declarations
```
parameter_declaration ::=
parameter [ signed ] [ range ] list_of_param_assignments | parameter parameter_type list_of_param_assignments
```

PR #1019, opened 2017-09-27
Move rocket output interrupt syncronizers

hcook:
From tile wrapper (fast clock) to coreplex (slow clock)


hcook:
@terpstra i'm making a release literally at this moment

PR #1021, opened 2017-09-29
FPU FMA FPGA retiming assist

henrystyles:
To assist retiming, manually move a max of 2 register stages of the FPU FMA pipeline into the FMA module

### DIFF
--- a/src/main/scala/tile/FPU.scala
+++ b/src/main/scala/tile/FPU.scala
@@ -579,42 +579,27 @@ class MulAddRecFNPipe(latency: Int, expWidth: Int, sigWidth: Int) extends Module
     val valid_stage0 = Wire(Bool())
     val roundingMode_stage0 = Wire(UInt(width=3))
     val detectTininess_stage0 = Wire(UInt(width=1))
-   
-    if(latency>0) {
-      mulAddRecFNToRaw_postMul.io.fromPreMul   := RegEnable(mulAddRecFNToRaw_preMul.io.toPostMul,io.validin)
-      mulAddRecFNToRaw_postMul.io.mulAddResult := RegEnable(mulAddResult,io.validin)
-      mulAddRecFNToRaw_postMul.io.roundingMode := RegEnable(io.roundingMode,io.validin)  
-      roundingMode_stage0                      := RegEnable(io.roundingMode,io.validin) 
-      detectTininess_stage0                    := RegEnable(io.detectTininess,io.validin)
-      valid_stage0                             := RegNext(io.validin,false.B)
-    } else {
-      mulAddRecFNToRaw_postMul.io.fromPreMul   := mulAddRecFNToRaw_preMul.io.toPostMul
-      mulAddRecFNToRaw_postMul.io.mulAddResult := mulAddResult
-      mulAddRecFNToRaw_postMul.io.roundingMode := io.roundingMode
-      roundingMode_stage0                      := io.roundingMode
-      detectTininess_stage0                    := io.detectTininess
-      valid_stage0                             := io.validin
-    }
- 
+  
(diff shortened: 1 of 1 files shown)


src/main/scala/tile/FPU.scala
```diff
@@ -542,7 +542,87 @@ class FPToFP(val latency: Int)(implicit p: Parameters) extends FPUModule()(p) {
...
+    val mulAddResult =
+        (mulAddRecFNToRaw_preMul.io.mulAddA *
+             mulAddRecFNToRaw_preMul.io.mulAddB) +&
+            mulAddRecFNToRaw_preMul.io.mulAddC
+
+    val valid_stage0 = Wire(Bool())
+    val roundingMode_stage0 = Wire(UInt(width=3))
+    val detectTininess_stage0 = Wire(UInt(width=1))
+   
+    if(latency>0) {
```
aswaterman: I think you can avoid the code repetition here by doing something like

    val postmul_regs = if (latency>0) 1 else 0
    mulAddRecFNToRaw_postMul.io.fromPreMul := Pipe(io.validin, mulAddRecFNToRaw_preMul.io.toPostMul, postmul_regs).bits
    ...
    valid_stage0                             := Pipe(io.validin, false.B, postmul_regs).valid

instead of the if/else.

henrystyles: I guess that would duplicate the enabling 'valid' register chain - but for timing that is probably no bad thing and synth is going to merge / dup it anyway.  I'll test out this change.

henrystyles: Actually forget that - it's not going to duplicate anything. I'll change this

aswaterman: Since it's only 1 pipeline stage, the 'valid' registers that Pipe creates don't actually get used, so Firrtl should eliminate them as dead code.  (And yeah, synthesis should do it anyway.)

colinschmidt:
Is this for ASIC QoR or FPGA QoR? @davidbiancolin was actively looking at the FPU retiming in FPGA tools just the other day(where it seems to be quite difficult) while I have had no issues with ASIC tools being able to retime well

colinschmidt:
Ah I see the branch is titled fpga_pipeline_fpu.

davidbiancolin:
Im curious to see what the longest paths through the DP FMA look like after manually pipelining the FPU.

aswaterman:
@henrystyles if you still have the static timing report, mind posting the critical path in a github gist?  I don't think that snippet would have anything proprietary in it.

henrystyles:
It meets timing at 100 Mhz Core / 25Mhz TileLink on VC707 ( Virtex-7 485T ).

The timing report is not really telling me a valid critical path because Vivado just gives up once it meets timing. That said, none of the top 10 worst paths in the report feature the FPU. The worst path listed is +0.035ns slack :  
U500VC707System/tile_3/rocket/frontend/icache/_T_9446_1_reg to
U500VC707System/tile_3/rocket/core/ex_reg_rs_msb_0_reg[55]/CE 

I'll try increasing the clock rate to see where it breaks.

aswaterman:
The I$ -> EX-stage pipeline register a critical path in ASIC implementation, so that sounds like good news.

aswaterman:
No regression on ASIC QoR.

hcook:
Does this change the list of Modules that need to be re-timed?

colinschmidt:
Looks like no to me. The retimeable registers are still inside the same module albeit at multiple levels of the hierarchy. Presumably Andrew would have mentioned that in his ASIC QoR regression comment.

aswaterman:
Yes, what Colin said.

davidbiancolin:
This is ready to be merged right?

Issue #733, opened 2017-05-08
Request to decouple PerfCounterIO's increment width from RetireWidth


ccelio:
Sorry for probably introducing the original hack, but I'd like to make a request to decouple the PerfCounterIO event's width from the retireWidth.

https://github.com/ucb-bar/rocket-chip/blob/master/src/main/scala/rocket/CSR.scala#L152

I would like a RocketCoreParams field that will let me choose what I want the event increment width to be.

My use-case is that a processor's issueWidth is likely to be wider than its retireWidth. Separating out different issue ports to separate events in not tenable, since software may want to save/restore counter state on a fine granularity. Even for coarse performance analysis on FPGAs, 29 hpm counters can run out fast.

Does this seem reasonable?

aswaterman:
Yes, but this interacts with how we're going to handle counters more systematically, so I'd like to have time to think about it more carefully than just adding another parameter to expand their width.

ccelio:
Sure thing, just want to make sure this is on the radar.

In the meantime, I'll continue with my fork.

PR #1023, opened 2017-10-02
Generalize CSR file to support simpler cores

aswaterman:


hcook:
the `HasRocketCoreParameters` to `HasCoreParameters` stuff lgtm

hcook:
Any further thoughts on https://github.com/freechipsproject/rocket-chip/issues/733?

aswaterman:
Not yet; it might just be that #733 demands another field in ```HasCoreParameters```.  Note, it doesn't demand another argument to ```RocketCoreParameters```, since there you can just ```def maxPerfCounterIncrement = retireWidth``` or whatever.

aswaterman:
With that in mind, now that cross-module width inference is reliable, maybe we don't need a parameter at all.  There are a few details to get right if we go that route (e.g., WideCounter is optimized for the case that the increment is 1 bit, and that savings would be lost).

aswaterman:
Actually, since CSRFile has the perfEvents, there's no need to rely on inference - perfEvents should specify each width somehow.

ccelio:
On the topic of CSRFiles and the removal of the custom CSRs, have you guys considered how to support "chicken bits"? 

For example, I have a memory disambiguator predictor that clears all of the entries every N cycles. I would like N to be set by a CSR, and where N==0 disables the predictor.

aswaterman:
I prefer not to add these as CSRs,
because that address space is more limited, not indexable, messy to
assemble for, etc. I've recently added a tile control port (an MMIO device)
that is currently only used to report ECC/bus errors. It could also be used
to hold nonstandard configuration registers.

PR #1025, opened 2017-10-02
Error grants

terpstra:
Grant messages should match the Acquire they respond to. Nevertheless, a Grant error always means you must not retain ownership of the block and retry the Acquire.


hcook:
Consistent transaction behavior/outcome as indicated via `param` is nice, but this makes it imperative for clients and monitors to pay attention to the `error` field and disregard all other indicated metadata changes, not just the `data` values.

terpstra:
Correct.

PR #1027, opened 2017-10-03
Make use of the new DontTouch annotation

### LINKED ISSUE #964: Produce a Module wrapper for a LazyModule
What I really want is to produce Verilog from a LazyModule without knowning its IO bundles.

For what I understand, the current GeneratorApp accepts Module, rather than LazyModule. Therefore, I need to produce a Module wrapper which contains the LazyModule as a DUT.

However, I do not want to set the IO bundle statically as the parameters may affect the IO bundle.
Here is what I did:

~~~scala
class CoreplexTop()(implicit p: Parameters) extends RawModule {

  val coreplex = LazyModule(new ExampleRocketSystem)
  val clk = IO(Input(Clock()))
  val rst = IO(Input(Bool()))
  val interrupts = IO(UInt(INPUT, width = coreplex.nExtInterrupts))
  val mem = IO(coreplex.mem_axi4.bundleOut)
  val mmio_master = IO(coreplex.mmio_axi4.bundleOut)
  val mmio_slave = IO(coreplex.l2FrontendAXI4Node.bundleIn)

  withClockAndReset(clk, rst) {
    val rocket = Module(coreplex.module)

    rocket.interrupts <> interrupts
    mem <> rocket.mem_axi4
    mmio_master <> rocket.mmio_axi4
    rocket.l2_frontend_bus_axi4 <> mmio_slave
  }
}
~~~

However, this does not work because the IO() is applied twice. I think the problem is the `bundleIn` and `bundleOut` are val produced by MixedNode. Then this IO() is applied twice.

Is there a proper way that I can get a IO bundle cloned from a module? Also, is there a way to inspect all IO bundles of a module generated by a `LazyMultiIOModuleImp`? I assume the IO in such a module is not named after `io`?

### LINKED ISSUE #966: AXI4 Slave Port
Hi all,

I would like to have a Slave AXI4 port on the rocket in order to access the SystemBus from an AXI4 master device. So I include the trait HasSlaveAXI4Port and HasSlaveAXI4PortBundle in the Top module but in the generated verilog file the slave port has only input pins and no output ones. How could I manage to have a full AXI4 slave port on the Rocket?

Thanks,
Andrea

hcook:
- Const prop cannot shove hartid into the tile
- Keep all dcache resp fields
- Keep all system-level top ports

Closes #964, closes #966

### DIFF
--- a/src/main/scala/coreplex/RocketCoreplex.scala
+++ b/src/main/scala/coreplex/RocketCoreplex.scala
@@ -3,6 +3,7 @@
 package freechips.rocketchip.coreplex
 
 import Chisel._
+import chisel3.experimental.dontTouch
 import freechips.rocketchip.config.{Field, Parameters}
 import freechips.rocketchip.devices.tilelink._
 import freechips.rocketchip.devices.debug.{HasPeripheryDebug, HasPeripheryDebugModuleImp}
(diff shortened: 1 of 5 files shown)


aswaterman:
The trait is nice

PR #1035, opened 2017-10-09
Fix paddrBits < xLen && paddrBits == vaddrBits case

aswaterman:
Require and/or force vaddrBits to be bigger than paddrBits so there's room to zero-extend a physical address by 1 bit, so that when the virtual address is sign-extended, the sign is zero.


mwachs5:
testing in progress

terpstra:
We have a design with addresses up to 0x40_0000_0000 populated. That's 38 bits used. Since it's one less than 39, we're ok, right?

aswaterman:
Yep.  I'm nearly certain we would've run afoul of this problem if we exceeded 38 bits.

PR #1036, opened 2017-10-10
Add some covers for L1 memory system

aswaterman:


aswaterman:
FWIW I copied the pattern from @terpstra

terpstra:
I don't really see the need to change the API for cover. It's hardly a big burden to expect developers to write a helper method that captures the information common to their design block. Alex should just do the same thing...

PR #1048, opened 2017-10-11
Correctly hook up the Local Interrupts into the Coreplex. 

mwachs5:
Also names some IntXBars for easier understanding of the emitted code / graphml.

Previously, `localIntNodes` were created any time someone called myModule.localIntNodes, which I don't think was the intention. Change the `def` to a `lazy val` so that they only get created once and the same copy is used every time.


aswaterman:
I probably should've made ```def localIntNodes``` to take a tile ID, so it would've been called once per tile, but the ```lazy

PR #1054, opened 2017-10-13
Add a check that MaxHartIdBits is enough for all hartids

mwachs5:
Before it was possible to have your hartid truncated and not realize it.

I am not sure this is the right place to put this, is it expected that all subclasses of `HartedTile` would have this same requirement? Or  is this specific to RocketTiles?


aswaterman:
This should hold for any HartedTile, so it seems like a good place to put this assertion.

mwachs5:
ah yes

PR #1059, opened 2017-10-18
Add atomics support to DTS JSON file.

richardxia:
cc @mikeyangsiv

This updates the DTS JSON file to include an `A` property if it supports all types of atomics.

### DIFF
--- a/src/main/scala/coreplex/BaseCoreplex.scala
+++ b/src/main/scala/coreplex/BaseCoreplex.scala
@@ -165,7 +165,7 @@ abstract class BaseCoreplexModule[+L <: BaseCoreplex](_outer: L) extends BareCor
   private val ranges = collect(Nil, outer.bindingTree).groupBy(_._2).toList.flatMap { case (key, seq) =>
     AddressRange.fromSets(key.address).map { r => (r, key.permissions, seq.map(_._1)) }
   }.sortBy(_._1)
-  private val json = ranges.map { case (range, ResourcePermissions(r, w, x, c), names) =>
+  private val json = ranges.map { case (range, ResourcePermissions(r, w, x, c, a), names) =>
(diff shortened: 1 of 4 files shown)


solomatnikov:
I think it makes sense to add LR/SC permission as well b/c some memories/devices support atomics but not LR/SC

richardxia:
I don't know the details of our TileLink implementation. Is there a simple expression for determining whether LR/SC is supported using the info available in `TLManagerParameters`?

solomatnikov:
AFAIK LR/SC are supported if atomics are supported and memory is cacheable but it's really a question for @aswaterman

hcook:
No. Those are handled internally to the L1 Dcache, i.e. before you even get to Diplomacy / Tilelink. I think they're supported for all cacheable address ranges?

solomatnikov:
So, cacheable + atomics is not sufficient for LR/SC. It must be cacheable in L1 DCache.

hcook:
From the perspective of software executing on a processor, what cacheable address ranges are not cacheable in the L1 DCache?

solomatnikov:
The core might have no L1 DCache but it can still access cacheable memory region, which can be cached by other cores that do have L1 DCache.

hcook:
There's a whole other can of worms related to heterogeneous processors that isn't even worth opening here. ATM, we only print out a single address map (from the perspective of the system bus crossbar), so you can't specialize it per tile anyways. I think that's out of scope for this PR.

terpstra:
LR/SC is implied by the cacheability bit

aswaterman:
Yes, it must be cached by the L1, not just cacheable. Specifically, a DTIM can't do LR/SC, even though it might have access to memory marked cacheable.

PR #1060, opened 2017-10-18
Fix address format

terpstra:
Make sure the address map includes the device name.


richardxia:
Apparently this comes from Java, where ["If there are more arguments than format specifiers, the extra arguments are ignored"](https://docs.oracle.com/javase/7/docs/api/java/util/Formatter.html#format%28java.lang.String,%20java.lang.Object...%29).

PR #1065, opened 2017-10-23
Better crossings deployed to bus and tile wrappers

hcook:
- Simplifying re-write of `RocketTileWrapper` to exploit features of `HasCrossingHelper`
- New methods on SystemBus and PeripheryBus, adapters are inside bus scope

(subsumes #1062)

@terpstra I bumped into a few issues, would like some assistance if you have any down time today. See below:


src/main/scala/coreplex/CrossingWrapper.scala
```diff
@@ -0,0 +1,96 @@
...
+{
+  this: LazyModule =>
+  val crossing: CoreplexClockCrossing
+
+  def cross(x: TLCrossableNode): TLOutwardNode = {
+    val out = x.node.parentsOut.exists(_ eq this) // is the crossing exiting the wrapper?
+    crossing match {
+      case SynchronousCrossing(params) => {
+        // !!! Why does star resolution fail for tile with no slave devices?
+        // this { TLBuffer(params)(x.node) }
```
hcook: When a tile has 0 slave devices (e.g. `DefaultConfig`), `resolveStar` fails here and I cannot figure out why. I though I was missing a `:*=` but I am pretty confident `SinkCardinality` is applied by `pbus.toTile` to all these adapter/crossing functions. The below line works; it's only when a buffer with :*= on both sides of it (I think) is inserted...

Any chance you could poke at it a little? It's driving me crazy.

src/main/scala/coreplex/RocketCoreplex.scala
```diff
@@ -95,67 +96,72 @@ trait HasRocketTiles extends HasTiles
...
+        case TileKey => tp
+        case BuildRoCC => tp.rocc
+        case SharedMemoryTLEdge => sharedMemoryTLEdge
+        case RocketCrossingKey => List(crossing)
+      })
+    ).suggestName(tp.name)
+
+    // Connect the master ports of the tile to the system bus
+    sbus.fromTile(
+      adapt = {x: TLOutwardNode => wrapper.cross(x) } andThen
```
hcook: @terpstra Sadly I needed to be explicit to the type checker here, otherwise it couldn't decide which version of `wrapper.cross(...)` applied. If you can come up with a concise syntax that's cool, otherwise I'd rather just write `wrapper.crossTL _` than go through the extra steps.

terpstra:
Oh! Shoot. It was not ready for deployment yet. Rocket Tile needs some crossings to always be of one sort and not influenced by the core:bus relationship. I had a plan on how to change this but was consumed by slides...

hcook:
@terpstra I think I figured out one way to handle unconditional crossings, check it out and see what you think. I got excited: Overall I am super duper happy with how this is looking so just kept going :)

terpstra:
I see. You made a new method. My plan was to split the 'cross' function into three variants which you can call explicitly and then have the 'cross' method dispatch to them based on the val. I also wanted to make them two layered traits, in case you want the crossRational/Sync/etc methods and don't want a common default trait-defined crossing.

hcook:
@terpstra There were already two methods, one taking `TLCrossableNode` and one taking `IntCrossableNode`. I just gave the `IntCrossableNode` variant an extra argument. As you can see in my comment on RocketCoreplex.scala:119 I think having a bunch of methods all named `cross` is a bit self-defeating anyways because it [unfortunately seems to] end up offloading the type specification on the API user. I like your idea of layering the trait-defined crossing on top of the raw  `cross` method(s); I guess I think that the raw form(s) should take the crossing type as an argument.

terpstra:
@hcook I think this is good to go now. Please give the final change-set a look!

terpstra:
FYI, I elected not to make the busses into implicits in this PR. I also did not change the adapter constructor objects to return a node. The later changed I intend to make in a new PR. I am having doubts that it is worth making the former change.

PR #1072, opened 2017-10-26
JTAG: Use new withClock way of overriding clocks

mwachs5:
the override clock way is deprecated

### DIFF
--- a/src/main/scala/jtag/Utils.scala
+++ b/src/main/scala/jtag/Utils.scala
@@ -15,7 +15,7 @@ class Tristate extends Bundle {
 
 /** Generates a register that updates on the falling edge of the input clock signal.
   */
-object NegativeEdgeLatch {
+object NegEdgeReg {
   def apply[T <: Data](clock: Clock, next: T, enable: Bool=true.B, name: Option[String] = None): T = {
     // TODO pass in initial value as well
     withClock((!clock.asUInt).asClock) {
--- a/src/main/scala/jtag/Utils.scala
+++ b/src/main/scala/jtag/Utils.scala
@@ -13,8 +13,7 @@ class Tristate extends Bundle {
   val driven = Bool()  // active high, pin is hi-Z when driven is low
 }
 
-class NegativeEdgeLatch[T <: Data](clock: Clock, dataType: T)
-    extends Module(override_clock=Some(clock)) {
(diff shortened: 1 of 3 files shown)


ducky64:
Does this compile? The NegativeEdgeLatch constructor has changed to not take block, but your `new NegativeEdgeLatch` is still being passed a clock and data type. And without the need to be at a module boundary to change clock domains, I'm not convinced a separate NegativeEdgeLatch class is needed, the NegativeEdgeLatch object can just create a new clock scope and instantiate the registers inside.

Also, this probably should be an upstream change in the chisel-jtag repo (along with some other general cleanup work there).

mwachs5:
Sorry no it doesn't compile, i got called into a meeting right after making this change, will fix it up.

mwachs5:
(I will squash all these)

src/main/scala/jtag/Utils.scala
```diff
@@ -12,40 +13,23 @@ class Tristate extends Bundle {
...
-  val reg = Reg(dataType)
-  when (io.enable) {
-    reg := io.next
-  }
-  io.output := reg
-}
-
 /** Generates a register that updates on the falling edge of the input clock signal.
   */
 object NegativeEdgeLatch {
```
aswaterman: This isn't germane to your pull request, but shouldn't this be called NegativeEdgeReg (or perhaps NegEdgeReg)?  It's not level-sensitive, so Latch confused me.

src/main/scala/jtag/Utils.scala
```diff
@@ -12,40 +13,23 @@ class Tristate extends Bundle {
...
+    }
   }
 }
 
 /** A module that counts transitions on the input clock line, used as a basic sanity check and
   * debug indicator clock-crossing designs.
   */
-class ClockedCounter(modClock: Clock, counts: BigInt, init: Option[BigInt])
-    extends Module(override_clock=Some(modClock)) {
+class ClockedCounter(counts: BigInt, init: Option[BigInt]) extends Module {
```
aswaterman: Now that we're using withClock, is this any different than using Counter(true.B, counts) from Chisel?

aswaterman: Oh, I guess the ability to set the reset value to something nonzero is a difference.

mwachs5:
I am holding off on merging this until it is actually tested, presumably with #1088

PR #1073, opened 2017-10-27
JTAG: Use sorted map for stability

mwachs5:
Otherwise the generated FIRRTL/Verilog is non deterministic


mwachs5:
I tried throwing a few 'reverse's in here and didn't get any different Verilog. Which makes me wonder if I am fixing the actual problem, or if something else was maintaining an order...

mwachs5:
Of course, the reason it wasn't diffing when I did a `reverse` is that the code I was generating didn't have a TAP in it. When I do it properly, using `reverse` does create the differing output as expected, and otherwise 2 consecutive runs generate the same.

PR #1076, opened 2017-10-27
coreplex: attach example external interrupts

### LINKED ISSUE #1071: Unconnected external interrupt port
Hello,

It seems that in the generated Verilog, the external interrupts port(s) are unconnected within the DUT. I've tried with the latest commit, as well as trying to tweak the TestHarness to make sure the signals weren't getting optimized out, but to no avail. I believe there should be something that connects the external interrupts to the PLIC, but I'm not sure how to determine if this connection exists. Any pointers?

Thanks

terpstra:
Fixes #1071


terpstra:
It's obviously wrong + fixed by looking at the generated verilog. Plus we'd had this exact same bug in another trait after the diplo refactor.

PR #1078, opened 2017-10-31
Error support

terpstra:
Implement the full TileLink specification for error handling.
- [x] modify TileLink native adapters (Fragmenter, WidthWidget)
- [x] modify TileLink <> foreign bus adapter
- [x] implement the rules in the Monitor


aswaterman:
Perhaps LFSR16 should only reset the LSB to 1, and not reset the MSBs?  This saves a tiny bit of hardware as well as giving the better simulation property.

terpstra:
Rebased to kill earlyAck assertion when !valid.

PR #1082, opened 2017-10-31
tilelink: fix error fragmentation from multibeat to multibeat

terpstra:
Unfortunately, dLast is not actually correct for AccessAckData.
dFragnum is 0 for all the subbeats in the multibeat=>multibeat case.


terpstra:
It's not a particularly bad bug, since we never had this rule about errors in the past.

PR #1088, opened 2017-11-03
[WIP] Add Debug Tests to Regression

mwachs5:
This adds a single test from riscv-tools/riscv-tests/debug to our rocket chip regressions, for both a 32 and 64-bit core. So far it is working but not always passing. To run (`DEBUG=1` is optional)

```
cd regression
make vsim-jtag-dtm-regression SUITE=JtagDtmSuite DEBUG=1
```
To use Verilator:

```
cd regression
make emulator-jtag-dtm-regression SUITE=JtagDtmSuite DEBUG=1
```

I still need to stabilize it and make it deterministic (as the JTAG VPI version is) such that time does not advance in the simulation until each OpenOCD packet is received.

Both Verilator and VCS sims are supported. In order to run with verilator, this PR replaces the JTAG VPI with a  SimJTAG which understands OpenOCD's "Remote Bitbang" protocol. This is similar to what is used in Spike.  I have left the JTAG VPI in place, but this repo no longer uses it.

@timsifive FYI.


mwachs5:
@richardxia can you help me out... this is failing because we need pexpect:

```

**Comment by richardxia** [comment]
If `pip` is installed on the Travis workers, then you just need to run `pip install -r $PATH_TO_RISCV_TESTS/tests/requirements.txt`. If it's not, then try doing `apt-get install python-pip` before running the `pip` command.

**Comment by timsifive** [comment]
In riscv-tools, pexpect is installed by adding it to the list of packages: https://github.com/riscv/riscv-tools/blob/master/.travis.yml#L25

**Comment by mwachs5** [comment]
Ok at least this is running now... but hard to debug without the logfiles. Trying some more local runs. I think it fails without the DEBUG=1 flag. @timsifive would be great if you were able to get this running on your side.

**Comment by timsifive** [comment]
To see the log files when there's a failure, pass `--print-failures` to gdbserver.py.

**Comment by mwachs5** [comment]
@timsifive that --print-failures flag seems to have done nothing... what am I missing? These pass locally for me:

https://travis-ci.org/freechipsproject/rocket-chip/jobs/303793884

**Comment by timsifive** [comment]
I fixed one bug in this logic in
https://github.com/riscv/riscv-tests/commit/669ac25d0db3888f2c41e3f50344d728be29a701
Hopefully that's the problem you were running into.

Tim

**Comment by mwachs5** [comment]
Looks like the failure is nondeterministic, the emulator just fails to connect sometimes. I can get it to fail sometimes locally but haven't figured out what is going on yet.

**Comment by mwachs5** [comment]
```

mwachs5:
It seems like Verilator is not properly async-resetting flops? This probably isn't an issue in most of our flops because they get SOME sort of clock during reset, and nothing matters until several clocks have passed in reset. But in the jtag we do assert reset and then de-assert it before any clocks, so we really do need Asynchronous reset in order to avoid assertions firing and killing these simulations (this doesn't matter unless you have TCK ticks that actually take the dmiReset to 0).

mwachs5:
Still something up with reset behavior, will try to reproduce locally...

```
Assertion failed: 'A' channel opcode changed within multibeat operation (connected at Debug.scala:406:16)
    at Monitor.scala:322 assert (a.bits.opcode === opcode, "'A' channel opcode changed within multibeat operation" + extra)
[14] %Error: freechips.rocketchip.system.WithJtagDTMSystem_DefaultConfig.v:132241: Assertion failed in TOP.TestHarness.ExampleRocketSystem.debug_1.dmOuter.dmiXbar.TLMonitor
```

mwachs5:
closing in favor of #1179

PR #1091, opened 2017-11-06
tilelink: AtomicAutomata should OR the Get error with the Put error

terpstra:
Errors should be detected also on the swallowed read!


terpstra:
So this is how one bypasses needing approval for your commits, eh? :)

aswaterman:
yep

PR #1097, opened 2017-11-09
More ITIM error-handling improvements

aswaterman:


src/main/scala/rocket/ICache.scala
```diff
@@ -313,7 +313,7 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
         }
 
         respValid := s2_slaveValid || (respValid && !tl.d.ready)
-        val respError = RegEnable(s2_scratchpad_hit && s2_data_decoded.uncorrectable, s2_slaveValid)
+        val respError = RegEnable(s2_scratchpad_hit && s2_data_decoded.uncorrectable && !(edge_in.get.hasData(s1_a) && s1_a.mask.andR), s2_slaveValid)
```
terpstra: Is s1_a the right stage in the pipeline for this?

aswaterman: yes it's ok

terpstra:
There's something wrong with this PR. It times out in the unit test regressions, despite being restarted.

aswaterman:
it's rebuilding the tools, which I assume is why it's timing out...

terpstra:
Ok. Looks like the tools were bumped wrong. Blew out the cache and restarting this PR for the purpose of correcting the cache.

aswaterman:
thanks @terpstra

PR #1107, opened 2017-11-12
micro-refactoring pmp.scala

pentin-as:


pentin-as:
Sorry. I did not read the specification. Now I understand what NAPOT, TOR, etc. mean.

PR #1115, opened 2017-11-18
tile: put a BasicBusBlocker inside RocketTile

hcook:
...instead of on the master side of the system bus.

Child `Tile` classes inheriting from `HasTileMasterPort` but not `HasScratchpadSlavePort` might need to add `masterNode := tileBus.node` by hand, sorry (cc/ @ccelio)

### DIFF
--- a/src/main/scala/coreplex/RocketCoreplex.scala
+++ b/src/main/scala/coreplex/RocketCoreplex.scala
@@ -16,23 +16,16 @@ import freechips.rocketchip.util._
 // TODO: how specific are these to RocketTiles?
 case class TileMasterPortParams(
     addBuffers: Int = 0,
-    blockerCtrlAddr: Option[BigInt] = None,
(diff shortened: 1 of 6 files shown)


ccelio:
Is it possible for a Tile to *not* have `CanHaveScratchpad`? I tried to remove it and directly plug in `with HasHellaCache with HasICacheFrontend ` into a `BoomTile extends HartedTile`, but the Tile expected a lot more plumbing from the ScratchpadSlavePort like `def findScratchpadFromICache` and such.

hcook:
The `TraceGenTile` in `groundtest` doesn't, but it can get away with it because it only has a dcache. `CanHaveScratchpad` is definitely horribly bloated... it's really `CanHaveABunchOfSlaveDevicesInTile`. I think we will refactor some of these traits once we support crossbars with optionally 0 slaves (coming soon).

PR #1123, opened 2017-11-21
DTS global

terpstra:
This expands the DTS visibility to the whole design, not just the innermost Scope.


src/main/scala/tile/RocketTile.scala
```diff
@@ -66,6 +66,9 @@ class RocketTile(val rocketParams: RocketTileParams)(implicit p: Parameters) ext
       val itim = if (frontend.icache.slaveNode.edges.in.isEmpty) Map() else Map(
         "sifive,itim"          -> ofRef(frontend.icache.device))
 
+      val decoherent = if (!rocketParams.core.useAtomicsOnlyForIO) Map() else Map(
```
richardxia: Nit: Name this `incoherent` to match the name of the DTS key (and because decoherent isn't a term used anywhere else at all).

PR #1129, opened 2017-11-29
Add cover properties for exceptions in the core.

richardxia:
I've added cover properties for the exceptions in RocketCore.scala. I created a `coverExceptions()` method which should basically get called everywhere that the existing `checkExceptions()` method is called. I manually read through each of the pipeline stages and added cover properties for all the exceptions that should be possible to receive at that stage (sans interrupts).

~~A have a couple of questions, probably mostly aimed at @aswaterman:~~

1. ~~I only conditionally emit the page fault exception cover properties if `rocketParams.useVM` is true. Is this the right param to be checking?~~ Also, is it necessary for me to explicitly exclude them in Scala, or is there some way to get Chisel or Firrtl to constant propagate them away later?

@jackkoenig says no to the constant prop question, at least for now.

2. ~~I'm currently missing all of the ECALL exceptions, since they don't seem to be referenced in RocketCore.scala. Instead, they're referenced in CSR.scala, but I'm not sure exactly how they eventually end up causing a real exception to the core. Can you point out to me where you would recommend adding the cover properties? Otherwise I'll spend some time tomorrow reading the code in more detail.~~

For the ECALL exceptions, I emit a cover property for the cartesian product of {mcause,scause} x {User, Supervisor, Machine, Hypervisor (the `Causes` constant still exists)}. If a {{ucause}} CSR were implemented we would add a cover property for that as well.

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -8,6 +8,7 @@ import Chisel.ImplicitConversions._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.tile._
 import freechips.rocketchip.util._
+import freechips.rocketchip.util.property._
 import scala.collection.mutable.LinkedHashMap
 import Instructions._
 
@@ -532,6 +533,9 @@ class CSRFile(perfEventSets: EventSets = new EventSets(Seq()))(implicit p: Param
     Causes.load_page_fault, Causes.store_page_fault, Causes.fetch_page_fault)
   val badaddr_value = Mux(write_badaddr, io.badaddr, 0.U)
 
+  val noCause :: mCause :: hCause :: sCause :: uCause :: Nil = Enum(5)
+  val xcause_dest = Wire(init = noCause)
+
   when (exception) {
     when (trapToDebug) {
       when (!reg_debug) {
@@ -544,6 +548,7 @@ class CSRFile(perfEventSets: EventSets = new EventSets(Seq()))(implicit p: Param
     }.elsewhen (delegate) {
       reg_sepc := formEPC(epc)
       reg_scause := cause
+      xcause_dest := sCause
       reg_sbadaddr := badaddr_value
       reg_mstatus.spie := reg_mstatus.sie
       reg_mstatus.spp := reg_mstatus.prv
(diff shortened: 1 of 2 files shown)


richardxia:
After chatting with @mikeyangsiv and @jackkoenig, I've decided that I'm going to take out the `if (rocketParams.useVM)` conditionals that I originally had. This is because eventually, we should have constant propagation take care of removing cover properties that either will always or never be hit, and because there's a risk of false negatives if there is drift between the cover properties code and the actual code that implements the exception logic. For now, it's safer to emit cover properties that may never be hit and force ourselves to manually wave them than to risk not emitting cover properties that really should exist.

I also figured out a way to add cover properties for the various ECALL exceptions, but I'd like someone to review them, since I did add an extra Wire that should get blown away during synthesis.

Removing WIP.

jchang0:
If the cover property will never be hit due to dead code, then it will produce a error (non-blocking error for the time being).  Because it's "proveable" that it will never be covered, and we expect it to be covered.  If there is a configuration that won't hit the cover property, it should be explicitly coded as such.

jchang0:
Adding extra wires for cover property is fine, if it's not used in the actual design, they will get pulled out into the _sim files during one of the firrtl passes.

richardxia:
> If the cover property will never be hit due to dead code, then it will produce a error (non-blocking error for the time being). Because it's "proveable" that it will never be covered, and we expect it to be covered. If there is a configuration that won't hit the cover property, it should be explicitly coded as such.

What does it mean to "produce an error"? And how can we write a property that is only expected to be covered if it is not impossible to cover it? The problem with explicitly coding the property so that it is omitted if it is not expected to be covered is that if the implementation changes, then the condition that the cover property uses may no longer be semantically correct anymore.

solomatnikov:
In some cases covers can be become assertions if they are never true

richardxia:
> In some cases covers can be become assertions if they are never true

That's a good point. It might be a good idea to write a utility method that allows for cover properties that are conditional on elaboration-time information that either emits covers or assertions depending on whether the property is coverable or should never be covered. I'm not going to do that now, but will keep it in mind.

PR #1133, opened 2017-11-30
Cleanups, and print out log names ASAP.

timsifive:
Factor out gdbserver common invocation into GDBSERVER (fixing
--print-failtures).
Add --print-log-names to that command so the logfiles can be inspected
while the simulation is still running.
`RISCV=... cmd` is more idiomatic than `export RISCV=... && cmd`


timsifive:
FYI, I can't merge this due to permissions/whatever.

PR #1134, opened 2017-11-30
Bump Chisel and Firrtl

jackkoenig:
This bumps Chisel and Firrtl to just past 3.0.0 and 1.0.0 respectively

Note the DontCares in the Jtag code


mwachs5:
Sorry I didn't review this. This does break something in the JTAG TAP. The irReg is blown away.

PR #1135, opened 2017-12-01
TileLink compliance: d_bits may not depend on d_ready

terpstra:
The debug module and all slaves using RegField.bytes were out-of-spec.
This degrades QoR and could potentially create a combinational loop at some design points.


mwachs5:
MY approval is based on the debug changes

PR #1138, opened 2017-12-01
Added coverage point to cover the case when ECC error happens during …

p12nGH:
…fence.i execution


hcook:
should we also add one for data errors?

p12nGH:
Added similar cover point for data

PR #1139, opened 2017-12-02
diplomacy: remove the :=? operator in favour of magic :*=*

terpstra:
The reason for the `:=?` operator was for when you have an adapter chain
whose direction of cardinality you could not know. We used explicit
directives to tell these compositions which way to go.

Unfortunately, that makes the API leaky. You think the chain of adapters
is just one adapter, but you have to use strange Cardinality scopes to
use it. That's just bad.

The new `:*=*` just automagically figures it out from the graph.


jerryhethatday:
how can I understand the notion of "direction of cardinality", can I understand it as the "the direction that determines the cardinality(number) of the connection？？？"

PR #1144, opened 2017-12-06
ICache: stores to the ITIM have effects (shrinking valid ITIM data)

terpstra:


hcook:
maybe put in a comment?

PR #1153, opened 2017-12-12
Bump scala and sbt-site plugin versions.

ucbjrl:
Strictly speaking, this isn't required until chisel3 and firrtl submodules are bumped to versions incorporating similar updates. However current emulation and regression tests pass with this change, so it could be merged now.


jackkoenig:
Subsumed by #1154

PR #1156, opened 2017-12-14
coreplex: make HasTiles more generic

hcook:
HasTiles now deals with only extremely general tile IOs.
Some RocketTiles specific behavior moved into RocketCoreplex.
BaseTile now has optional LocalInterruptNode.

### DIFF
--- a/src/main/scala/coreplex/BaseCoreplex.scala
+++ b/src/main/scala/coreplex/BaseCoreplex.scala
@@ -5,11 +5,8 @@ package freechips.rocketchip.coreplex
 import Chisel._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
-import freechips.rocketchip.interrupts._
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.devices.tilelink._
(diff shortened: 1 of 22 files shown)


hcook:
@ccelio this should make it easier to make a `RocketAndBoomAreFriendsCoreplex` that sets `val tiles = rocketTiles ++ boomTiles` if you decide you want to do that

ccelio:
Thanks for the nod. 

FYI, here's what I'm currently doing. Mostly a copy/paste job for now. I'm a few weeks behind:

https://github.com/ucb-bar/riscv-boom/blob/53b805546c474b36098abf02f432e1a68f728ac3/src/main/scala/system/BoomCoreplex.scala#L20

https://github.com/ucb-bar/riscv-boom/blob/53b805546c474b36098abf02f432e1a68f728ac3/src/main/scala/common/tile.scala

hcook:
@ccelio I might tweak it a little more this week, and if I have some spare cycles I'll try to make `RocketTileWrapper` into `AnyBaseTileWrapper` to cut down on duplication on your end. I don't think your coreplex will change much but you can at least mix HasRocketTiles and HasBoomTiles together.

PR #1159, opened 2017-12-18
diplomacy: LazyModule factory uses ValName

hcook:
future-proofing the LazyModule API


terpstra:
I'd skip the use of the name to suggestName. The goal is to enforce API
compliance without an effect on existing design names, right?

PR #1162, opened 2017-12-19
Revert "ICache: stores to the ITIM have effects (shrinking valid ITIM…

hcook:
… data) (#1144)"

This reverts commit a542ae687e6bd2e459939aee7fd98dddb399bed1.

Reverted commit apparently makes `rv64mi-ma_addr` fail when executed out of an itim. I think it triggers an access exception instead of the expected ma exception? It's holding up a bunch of stuff atm, so I'm reverting this now and we can solve it when ppl are back from vacation.


mwachs5:
rubber stamp approval

mwachs5:
should we (re)open an issue to track whatever this was originally fixing?

terpstra:
I highly suspected this PR might cause a problem. It's why I made sure it
was just one change by itself. Whether or not the PMA change is considered
more or less correct than the failing test should decide if we want to try
merging it again later.

PR #1165, opened 2017-12-21
Bump riscv-tools to try to figure out what's wrong with Travis caching

mwachs5:
#1088  and #597 are having trouble due to Travis being unable to cache riscv-tools install. Trying to isolate that part of the problem to see if it's a stale cache issue or something else.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@ git:
   submodules: false
 language: scala
 # run on new infrastructure
-sudo: false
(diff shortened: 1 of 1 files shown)


mwachs5:
Well, this failed before riscv-tools even finished building, and comparing the `regression/install` directory before and after made no difference. I am wondering if this is just an artifact of building ANY cache and something else has changed.

PR #1166, opened 2017-12-21
ECC: support poison during encode

terpstra:
This makes it possible to update an ECC-protected word while
retaining the fact that the value has an ECC error.


aswaterman:
Long merge latency ;)

PR #1170, opened 2017-12-30
Add cover points for read on registers

jchang0:
PR for review.  Do not merge.

Is RegField the best place to put the cover properties?  (I would like to cover read and write to each field)  It seems to cause massive change because Parameters needs to be passed in.

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -11,6 +11,7 @@ import freechips.rocketchip.tile.XLen
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.interrupts._
 import freechips.rocketchip.util._
+import freechips.rocketchip.config.Parameters
 
 /** Constant values used by both Debug Bus Response & Request
   */
@@ -250,7 +251,7 @@ class DebugCtrlBundle (nComponents: Int)(implicit val p: Parameters) extends Par
 
 // Local reg mapper function : Notify when written, but give the value as well.  
 object WNotifyWire {
-  def apply(n: Int, value: UInt, set: Bool) : RegField = {
+  def apply(n: Int, value: UInt, set: Bool)(implicit p: Parameters) : RegField = {
     RegField(n, UInt(0), RegWriteFn((valid, data) => {
       set := valid
       value := data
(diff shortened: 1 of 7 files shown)


src/main/scala/regmapper/RegField.scala
```diff
@@ -6,18 +6,30 @@ import Chisel._
...
+import chisel3.internal.sourceinfo.{SourceInfo, SourceLine}
+import freechips.rocketchip.config.Parameters
 
 case class RegReadFn private(combinational: Boolean, fn: (Bool, Bool) => (Bool, Bool, UInt))
 object RegReadFn
 {
+  def coverread(fn: (Bool, Bool) => (Bool, Bool, UInt))(implicit sourceInfo: SourceInfo, p: Parameters): (Bool, Bool) => (Bool, Bool, UInt) = {
+    case (ivalid: Bool, oready: Bool) => {
+      val (iready, ovalid, data) = fn(ivalid, oready)
+      cover(iready && ivalid, "regread_in", "RegField Read Request Initiate")
```
mwachs5: Why not regread_start/finish instead of in/out, to match the longer description?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -250,7 +251,7 @@ class DebugCtrlBundle (nComponents: Int)(implicit val p: Parameters) extends Par
 
 // Local reg mapper function : Notify when written, but give the value as well.  
 object WNotifyWire {
-  def apply(n: Int, value: UInt, set: Bool) : RegField = {
+  def apply(n: Int, value: UInt, set: Bool)(implicit p: Parameters) : RegField = {
     RegField(n, UInt(0), RegWriteFn((valid, data) => {
```
mwachs5: Why doesn’t this need the sourceInfo implicit?

terpstra:
@jackkoenig Could cover() be a first-class Chisel method, like assert()?

jchang0:
I think it's in the works, but I'm not sure what stage it is in.  But I agree that should be the right long term solution.

We can also avoid the use of parameters if the other users of rocket-chip don't mind a whole bunch of printf showing when things are covered when not compiling into systemverilog (but that's also a hack) and probably not acceptable.

jchang0:
new PR related to this enhancement
https://github.com/freechipsproject/rocket-chip/pull/1206

PR #1174, opened 2018-01-05
Claim that ErrorDevice is TRACKED

aswaterman:
The DCache needs all executable regions be tracked, or else it has to be flushed on FENCE.I at significant performance cost.

(https://github.com/freechipsproject/rocketchip/blob/master/src/main/scala/rocket/DCache.scala#L725)

However, we marked the ErrorDevice executable in 6a0150aad72f4e3a34f1f9d674a6c185f7e30fcc to facilitate I$ testing.  We either need to revert that change, or we need to mark the ErrorDevice as TRACKED, or we need to special-case the DCache check to ignore the ErrorDevice.

This option seems least undesirable.

### DIFF
--- a/src/main/scala/devices/tilelink/Error.scala
+++ b/src/main/scala/devices/tilelink/Error.scala
@@ -26,7 +26,7 @@ abstract class DevNullDevice(params: ErrorParams, beatBytes: Int = 4)
     Seq(TLManagerParameters(
       address            = params.address,
       resources          = device.reg("mem"),
-      regionType         = RegionType.UNCACHED,
+      regionType         = RegionType.UNCACHEABLE,
       executable         = true,
       supportsAcquireT   = xfer,
       supportsAcquireB   = xfer,
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -722,7 +722,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   metaArb.io.in(5).bits.data := metaArb.io.in(4).bits.data
 
   // Only flush D$ on FENCE.I if some cached executable regions are untracked.
-  val supports_flush = !edge.manager.managers.forall(m => !m.supportsAcquireT || !m.executable || m.regionType >= RegionType.TRACKED)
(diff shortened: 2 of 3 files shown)


terpstra:
So this PR is failing because the CacheCork requires that slaves behind it are not already tracked. That said, I think it's safe to mark the ErrorDevice TRACKED as you propose. It supports Acquire already, and since it always Grants with an error, it's not like you can get a block ownership discrepancy.

We could put the 'ErrorDevice special case' in the CacheCork. The ErrorDevice is somewhat exceptional in a lot of ways and there is code that looks for it and treats it specially in a few places already, so while not-so-pretty it might just be the most expedient approach. The ErrorDevice is the only device I can imagine which simultaneously supports Acquire, never Probes, and yet is multi-master coherent.

Another option might be to remove the Acquire-ability of the ErrorDevice. Presumably then the D$ need not be flushed since it is also obviously unable to acquire. I believe it needed to support Acquire so that it could serve as a landing pad for ChipLink traffic that went to an illegal address.

I notice also that the device currently has a bug where it does not support AcquirePerm.

hcook:
Another option we discussed is changing the region type to UNCACHEABLE while continuing to support Acquires of all types. While counterintuitive, I think this accurately represents the fact that it should be legal to Acquire data blocks from the device, while also being assured that the  cached values of error blocks will never change and so no flushes are needed.

terpstra:
Interesting. The current TL parameters require that supportsAcquire -> at least UNCACHED, but we could change that. I see your point that the ErrorSlave is in a way like a ROM. I'm not sure I agree that UNCACHEABLE + Acquire makes sense for ROMs, though. I would have said they are UNCACHED and you can put a CacheCork on them to add supportsAcquireB, but not supportsAcquireT.

aswaterman:
I'm still OK with either approach, but changing the ErrorDevice to support AcquireB but not AcquireT seems to solve the problem without additional code changes.

terpstra:
Ok, but then you break ChipLink, which redirects Acquire.toT to a bad address to there error slave.

aswaterman:
Oh, right.

aswaterman:
With that in mind, I think Henry's suggestion is least disruptive--make the device uncacheable but support acquire, anyway.  I'll relax the diplomacy assertion (but only in that direction, not the other way around).

hcook:
I'm approving the the UNCACHEABLE version. Maybe going forward we should consider renaming it that type to IMMUTABLE , or adding IMMUTABLE as another region type?

PR #1176, opened 2018-01-05
Fix TL MMIO port

### LINKED ISSUE #1175: Fail to get a TL MMIO port for ROCKET system top
In trait HasMasterTLMMIOPortModuleImp (rocket-chip/src/main/scala/coreplex/Ports.scala), the master TL port may use a wrong direction:
"(mmio_tl zip outer.mmio_tl.out) foreach { case (i, (o, _)) => i <> o }" should change to "(mmio_tl zip outer.mmio_tl.in) foreach { case (i, (o, _)) => i <> o }"

terpstra:
Fixes #1175

@hcook Maybe we should consider an alternative design pattern for these diplomacy-IO connections. This is like the fourth time I've seen this bug. Even a zipEq that requires the two lists to be equal length would be an improvement. Or maybe better still we can have some sort of API that packages more of the plumbing.

### DIFF
--- a/src/main/scala/coreplex/Ports.scala
+++ b/src/main/scala/coreplex/Ports.scala
@@ -195,7 +195,7 @@ trait HasMasterTLMMIOPortBundle {
 trait HasMasterTLMMIOPortModuleImp extends LazyModuleImp with HasMasterTLMMIOPortBundle {
   val outer: HasMasterTLMMIOPort
   val mmio_tl = IO(HeterogeneousBag.fromNode(outer.mmio_tl.in))
-  (mmio_tl zip outer.mmio_tl.out) foreach { case (i, (o, _)) => i <> o }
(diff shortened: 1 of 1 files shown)


hcook:
I agree they have to match but is it really correct to change `out` to `in` and not `in` to `out`? That's confusing. My intuition for what "in" and "out" mean in the context of NodeHandles might be backwards?

`val mmio_tl = IO(outer.mmio_tl.toOutwardBag))` makes sense to me as an API...?

terpstra:
Yeah. The direction is confusing, which is why it's so often wrong. The reasoning is like this:
- You declared a diplomatic slave node (the TL MMIO slave)
- Therefore, you must implement it's inward facing interface
- When you wire it outside, you must thus connect the tl_in

How does your proposed toOutwardBag connect the IO?

hcook:
Can't it make a `HBag` that's a wire, connect to it, and then return the wire to `IO(...)`?

PR #1181, opened 2018-01-08
Dual-port RAM replaced with single-port RAM for tag_array in HellaCache

pentin-as:
In accordance with https://github.com/freechipsproject/chisel3/issues/752


aswaterman:
Looks fine.  I take it you verified that the RAM is actually inferred as single-ported now?

pentin-as:
Yes. I checked. Now single-port RAM, it was a two-port.

PR #1183, opened 2018-01-08
more detailed RegField descriptions

mwachs5:
Previously, RegField took a `name` and `description` string. This isn't really sufficient for flows that may want to use the designer's knowledge, especially whether the register was reset or not, whether it is R/W, or what (R/W behavior may be abled to get extracted automatically sometimes, but not always).

This replaces the two strings with a more descriptive case class that can be populated by the RegField author and used by other flows.

### DIFF
--- a/src/main/scala/regmapper/RegField.scala
+++ b/src/main/scala/regmapper/RegField.scala
@@ -7,6 +7,46 @@ import chisel3.util.{ReadyValidIO}
 
 import freechips.rocketchip.util.{SimpleRegIO}
 
+
+// This information is not used internally by the regmap(...) function.
+// However, the author of a RegField may be the best person to provide this
+// information which is likely to be needed by downstream SW and Documentation
+// tools.
+
+object RegFieldAccessType extends scala.Enumeration {
+  type RegFieldAccessType = Value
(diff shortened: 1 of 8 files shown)


src/main/scala/regmapper/RegField.scala
```diff
@@ -7,6 +7,37 @@ import chisel3.util.{ReadyValidIO}
...
+object RegFieldAccessType extends scala.Enumeration {
+  type RegFieldAccessType = Value
+  val R, W, RW, RSPECIAL, WSPECIAL, RWSPECIAL, OTHER = Value
+}
+import RegFieldAccessType._
+
+
+object RegFieldResetType extends scala.Enumeration {
+  type RegFieldResetType = Value
+  val S, A, N = Value
```
ahuntington: Would it be appropriate to elaborate in a comment here what S, A and N stand for?

mwachs5: done

terpstra:
For this API to be useful it has to be as painless and automatic as
possible. We need to try really hard to infer as much as possible before we
push this out...

PR #1184, opened 2018-01-09
TLRegMapper: emit a JSON file describing the register fields

mwachs5:
This is a first stab at functionality we need to enable, but don't think this is really the right place to do it.

We should be able to take in register maps as JSON inputs and output them as JSON for a register map created by a Chisel coder. 

As a strawman this PR changes it so that all `TLRegMapper` devices output an XSVD format description (https://xpack.github.io/xsvd/) which is basically a simplified , JSON-based subset of IP-XACT.

Some obvious problems with this right now:

- A lot of info has to be manually specified by the user (as noted by @terpstra  on #1183)
- XSVD, IP-XACT, etc often deal with "Registers" or "RegisterArrays", which contain "Fields". Right now the `RegField.Map` offers no way to enter any sort of name/description or other properties at that level. Only `RegField` can have any user defined properties. This is less problematic going the other way (it would be easy to parse such an XML/JSON and create the described  `RegField.Map`)
-- By putting this inside the `regmap` function, the TLRegisterRouterNode doesn't have information about the device name so there is a lot of stuff like `deviceAt0x20000` as placeholders.
-- There is no way to capture the "repetition" information. Again this isn't really a problem going in the other direction.
-- ~~It would be better to completely omit the fields from JSON if the corresponding description does not exist. `~` operator supports `Option` so this should be used.~~

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,6 +15,7 @@ lazy val commonSettings = Seq(
   traceLevel   := 15,
   scalacOptions ++= Seq("-deprecation","-unchecked"),
   libraryDependencies ++= Seq("org.scala-lang" % "scala-reflect" % scalaVersion.value),
+  libraryDependencies ++= Seq("org.json4s" %% "json4s-jackson" % "3.5.0"),
   addCompilerPlugin("org.scalamacros" % "paradise" % "2.1.0" cross CrossVersion.full)
 )
 
@@ -44,3 +45,4 @@ val chipSettings = Seq(
     s"make -C $makeDir  -j $jobs $target".!
   }
 )
+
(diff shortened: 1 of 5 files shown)


hcook:
You might try putting it as separate function in `trait HasRegMap`, and then have e.g. `TLRegModule ` call that feeding it `router.device` &etc

mwachs5:
I am rebasing this since I have changed the RegFieldDesc in #1183.  I need to update this PR to match. We have decided to ignore the usual concept of "Registers" and instead to allow naming "RegFieldGroups". Downstream tools will be able to use this information as they see fit, rather than trying to force the Chisel to output the format directly.

PR #1186, opened 2018-01-09
Update TestDriver module to support FSDB

edcote:
Add support for FSDB waveform dumping using existing API.  Feature is enabled using +define+FSDB.

Change has not been fully regressed (i.e., please don't pull blindly).  Impact on existing knobs is minimal, should not affect existing functionality.  Automated Travis builds should be sufficient to assess.

Alternatively could using +define+VCS.  Chose to introduce new define because multiple simulators support FSDB dumping.

### DIFF
--- a/vsrc/TestDriver.v
+++ b/vsrc/TestDriver.v
@@ -51,7 +51,7 @@ module TestDriver;
 `ifdef VCS
       $vcdplusfile(vcdplusfile);
 `else
-      $fdisplay(stderr, "Error: +vcdplusfile is VCS-only; use +vcdfile instead");
+      $fdisplay(stderr, "Error: +vcdplusfile is VCS-only; use +vcdfile instead or recompile with VCS=1");
       $fatal;
 `endif
     end
@@ -63,7 +63,7 @@ module TestDriver;
       $fsdbDumpvars("+all");
       //$fsdbDumpSVA;
 `else
-      $fdisplay(stderr, "Error: +fsdbfile is FSDB-only; use +vcdfile or +vcdplus instead");
+      $fdisplay(stderr, "Error: +fsdbfile is FSDB-only; use +vcdfile/+vcdplus instead or recompile with FSDB=1");
       $fatal;
 `endif
     end


vsrc/TestDriver.v
```diff
@@ -55,12 +56,28 @@ module TestDriver;
...
     end
 
+    if ($value$plusargs("fsdbfile=%s", fsdbfile))
+    begin
+`ifdef FSDB
+      $fsdbDumpfile(fsdbfile);
+      $fsdbDumpvars("+all");
+      //$fsdbDumpSVA;
+`else
+      $fdisplay(stderr, "Error: +fsdbfile is FSDB-only; use +vcdfile or +vcdplus instead");
```
mwachs5: can you enhance this error message to say something like "Please recompile with FSDB=1"

vsrc/TestDriver.v
```diff
@@ -55,12 +56,28 @@ module TestDriver;
...
+      $fatal;
+`endif
+    end
+
     if ($value$plusargs("vcdfile=%s", vcdfile))
     begin
```
mwachs5: here you seem to allow concurrent vcd/vpd and fsdb dumping, but below code enforces one or the other. What is the intention? Is someone allowed to run a sim with `+fsdbfile=myFSDB.fsdb +vcdfile=myVCD.vcd`? It seems like if they did that they would not get an error but they would also not end up with a `myVCD.vcd`

edcote:
Priority to FSDB was design choice. I will update to allow concurrent
waveform dumping.

edcote:
I think my/existing code is correct.

The existing code does not enable concurrent dumping.  It enables concurrent population of vcdplusfile, fsdbfile, and/or vcdfile variables using their associated plusarg.

Enable, disable of dumping is done further down in the file using `VCDPLUSON and `VCDPLUSOFF.

Though the macro is named VCDPLUS, it's used for VCD or VPD dumping.  My implementaiton assumes it will do the same for fsdb.

The format of the dump is based on plusargs and the define.

.vpd -> +define+DEBUG +define+VCS +vcdplusfile=/path/to/vpd
.fsdb -> +define+DEBUG +define+VCS +FSDB +fsdbfile=/path/to/fsdb
^^ fsdb takes priority, conceivable that user has both defines set
.vcd -> +define+DEBUG +vcdfile=/path/to/vcd

vsrc/TestDriver.v
edcote: I didn't rearchitect the existing code.  If the user did +fsdbfile=myFSDB.fsdb +vcdfile=myVCD.vcd and forgot to define `FSDB, then no error and only vcd dumped.

vsrc/TestDriver.v
edcote: yes, done

PR #1201, opened 2018-01-19
Debug: reduce default program buffer size since impebreak is an option

mwachs5:
Turn down the program buffer words to 2 and use implicit ebreak.

@timsifive is 2 program buffer words enough ?

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -121,7 +121,7 @@ import DebugAbstractCommandType._
 
 case class DebugModuleParams (
   nDMIAddrSize  : Int = 7,
-  nProgramBufferWords: Int = 16,
+  nProgramBufferWords: Int = 2,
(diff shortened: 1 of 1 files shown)


timsifive:
>
> @timsifive <https://github.com/timsifive> is 2 program buffer words
> enough ?
>
2 is enough. In fact OpenOCD won't even try to use more than that (assuming
impebreak, otherwise it will use 3).

mwachs5:
Blergh. I forgot that `riscv-fesvr` is not as sophisticated, and *doesn't* utlize the `impebreak`. So none of the tests that use `riscv-fesvr` to load the program succeed.

I'll try turning it down to 4.

mwachs5:
This is blocked on FESVR being updated to use less progbuf words as OpenOCD was.

mwachs5:
This depends on https://github.com/riscv/riscv-fesvr/issues/48

mwachs5:
this has been done long ago

PR #1202, opened 2018-01-20
ResetCatchAndSync: work also in the context of a RawModule

terpstra:


timsliu:
Hi there,

I've been trying to add an asynchronous assert and synchronous de-assert to my design for the past few days. It looks like ResetCatchAndSync is doing just that. Where could I find instructions on how to use it?

Thanks!

Tim

mwachs5:
It is doing just that. There aren't any real instructions written down, though we're planning to make it easier to add docs to this repo in the near future. Best thing to do at this time is look by example of where it is used, search in this repo as well as https://github.com/sifive/freedom/search?q=resetcatchandsync&unscoped_q=resetcatchandsync. 

Bonus points if you submit a scaladoc PR with something that makes it clearer for the next person. Even if it's not right we'll help you edit it until it is. (https://docs.scala-lang.org/style/scaladoc.html)

timsliu:
Hey Megan,

Thanks for the reply! I've been taking a look through the examples. I've followed the first search result and tried something like:

```scala
case object BuildTop extends Field[Parameters => TopModule[Top]] ((p: Parameters) => Module(new TopModuleWrapper(p)))

class TestHarness(implicit val p: Parameters) extends Module {
  val dut = p(BuildTop)(p)
  dut.reset := ResetCatchAndSync(clock, reset.asBool)
}

This compiles fine, but when I open the generated source the resets don't look asynchronous. Is there some other step I'm missing?

And yes, when I finally get this working I'll definitely look into submitting a PR!

Thanks!

Tim

mwachs5:
what do you mean "the resets don't look asynchronous"? Which resets? This would only make async resets within the `ResetCatchAndSync` module.

To make an asynchronous reset in your design check out the info in this PR:
https://github.com/freechipsproject/www.chisel-lang.org/pull/74

timsliu:
Hi Megan,

Ohh, I think I misunderstood what ResetCatchAndSync does. I thought that it generated a reset that is asynchronously asserted and synchronously deasserted, which was then used by another module. But is it's actual purpose to take an asynchronous reset and make the deassertion synchronous (like a reset synchronizer)?

I've read over that PR, but will take a more careful second look.

Thanks!

Tim

mwachs5:
> "But is it's actual purpose to take an asynchronous reset and make the deassertion synchronous (like a reset synchronizer)?" 


yes

PR #1206, opened 2018-01-24
Add cover points to RegFields

jchang0:
With the new hack for covers, we now do not need to pipe parameters all over the place for covers... I did pass in the sourceInfo so that the cover labels will point to a sensible location of where the registers are instantiated from.


jchang0:
superseded by 

https://github.com/freechipsproject/rocket-chip/pull/1208

PR #1208, opened 2018-01-25
Add cover points to registers

jchang0:


jchang0:
It'll generate a unique string, so it'll elaborate fine.  But it makes it hard to debug and figure out which what it is trying to cover.  So we'll have to put in a name and description for all the field in the future (or at least the ones that are not covered)

PR #1212, opened 2018-01-27
Add cover properties to Core CSRs

jchang0:


src/main/scala/rocket/CSR.scala
```diff
@@ -609,6 +609,11 @@ class CSRFile(perfEventSets: EventSets = new EventSets(Seq()))(implicit p: Param
 
   io.rw.rdata := Mux1H(for ((k, v) <- read_mapping) yield decoded_addr(k) -> v)
 
+  // cover access to register
+  read_mapping.foreach( {case (k, v) => {
+    cover(io.rw.cmd.isOneOf(CSR.W, CSR.S, CSR.C, CSR.I, CSR.R) && io.rw.addr===k, "CSR_access_"+k.toString, "Cover Accessing Core CSR field")
```
aswaterman: You should remove ```CSR.I``` from this list, but otherwise it looks good to me.

PR #1219, opened 2018-01-31
Bump Firrtl

jackkoenig:


richardxia:
If it's not too much trouble, could you also bump Chisel for https://github.com/freechipsproject/chisel3/pull/766?

jackkoenig:
So this has an important bug fix which is why I'm doing it on it's own. Chisel is slightly more trouble but I'm trying to have it ready today or tomorrow

richardxia:
Okay, no problem, this sounds more important.

PR #1228, opened 2018-02-06
Teach MulDiv to do div-only

aswaterman:


hcook:
Is this configuration expressible via ISA string? Like is it still "M"? What happens when a mul instruction ends up in a binary accidentally?

aswaterman:
The intent isn't to remove instructions from the ISA (note this doesn't change the instruction decoder), it just allows processors to implement the MUL instructions with a different unit than the divider, e.g., a fully-pipelined multiplier.

PR #1230, opened 2018-02-07
Add BlackBox resources to rocket.

### LINKED ISSUE #1217: using blackbox in the rocket-chip repo
I'm trying to use BlackBox in rocket-chip generator.

First of all, I tried it using chisel3 template, and it works well.

However, I tried it inside rocket-chip generator using setResourcee("/alu/customALU.v") in the class that extends BlackBox with HasBlackBoxResource.(The location of file:  "~/rocket-chip/src/main/resources/alu/customALU.v")

But it continues to show FileNotFoundException:

mkdir -p /home/ubuntu-cho/rocket-chip/emulator/generated-src/
java -Xmx2G -Xss8M -XX:MaxPermSize=256M -cp /home/ubuntu-cho/rocket-chip/firrtl/utils/bin/firrtl.jar firrtl.Driver -i /home/ubuntu-cho/rocket-chip/emulator/generated-src/freechips.rocketchip.system.DefaultConfig.fir -o /home/ubuntu-cho/rocket-chip/emulator/generated-src/freechips.rocketchip.system.DefaultConfig.v -X verilog --infer-rw TestHarness --repl-seq-mem -c:TestHarness:-o:/home/ubuntu-cho/rocket-chip/emulator/generated-src/freechips.rocketchip.system.DefaultConfig.conf -faf /home/ubuntu-cho/rocket-chip/emulator/generated-src/freechips.rocketchip.system.DefaultConfig.anno -ffaaf
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256M; support was removed in 8.0
**Exception in thread "main" java.io.FileNotFoundException: Resource '/rocket/customALU.v'**
        at firrtl.transforms.BlackBoxSourceHelper$.copyResourceToFile(BlackBoxSourceHelper.scala:145)
        at firrtl.transforms.BlackBoxSourceHelper$$anonfun$1.apply(BlackBoxSourceHelper.scala:107)
        at firrtl.transforms.BlackBoxSourceHelper$$anonfun$1.apply(BlackBoxSourceHelper.scala:103)
        at scala.collection.mutable.ResizableArray$class.foreach(ResizableArray.scala:59)
        at scala.collection.mutable.ArrayBuffer.foreach(ArrayBuffer.scala:48)
        at firrtl.transforms.BlackBoxSourceHelper.execute(BlackBoxSourceHelper.scala:103)
        at firrtl.Transform$$anonfun$4.apply(Compiler.scala:242)
        at firrtl.Transform$$anonfun$4.apply(Compiler.scala:242)
        at firrtl.Utils$.time(Utils.scala:135)
        at firrtl.Transform.runTransform(Compiler.scala:242)
        at firrtl.Compiler$$anonfun$13$$anonfun$apply$10.apply(Compiler.scala:461)
        at firrtl.Compiler$$anonfun$13$$anonfun$apply$10.apply(Compiler.scala:461)
        at scala.collection.LinearSeqOptimized$class.foldLeft(LinearSeqOptimized.scala:124)
        at scala.collection.immutable.List.foldLeft(List.scala:84)
        at firrtl.Compiler$$anonfun$13.apply(Compiler.scala:461)
        at firrtl.Compiler$$anonfun$13.apply(Compiler.scala:461)
        at firrtl.Utils$.time(Utils.scala:135)
        at firrtl.Compiler$class.compile(Compiler.scala:460)
        at firrtl.VerilogCompiler.compile(LoweringCompilers.scala:138)
        at firrtl.Driver$$anonfun$execute$1.apply(Driver.scala:176)
        at firrtl.Driver$$anonfun$execute$1.apply(Driver.scala:140)
        at logger.Logger$$anonfun$makeScope$1.apply(Logger.scala:129)
        at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
        at logger.Logger$.makeScope(Logger.scala:127)
        at firrtl.Driver$.execute(Driver.scala:140)
        at firrtl.Driver$.execute(Driver.scala:232)
        at firrtl.Driver$.main(Driver.scala:248)
        at firrtl.Driver.main(Driver.scala)
Makefrag-verilator:16: recipe for target '/home/ubuntu-cho/rocket-chip/emulator/generated-src/freechips.rocketchip.system.DefaultConfig.v' failed
make: *** [/home/ubuntu-cho/rocket-chip/emulator/generated-src/freechips.rocketchip.system.DefaultConfig.v] Error 1

grebe:
This repo contains a lot of IP that may be generally useful for other projects, but using rocket-chip as a library (i.e., an sbt dependency) for other projects can be a little tricky. One problem is blackboxes (plusarg_reader.v in particular!). Rocket's build system knows to look in `vsrc/` for extra verilog sources and `csrc/` for extra C/C++ sources, but if you are using rocket as an sbt dependency you need to do some extra bookkeeping to make sure the right blackboxes are included in your own build system.

Chisel has a mechanism that is intended to help resolve the general problem of chisel libraries that have blackboxes- `HasBlackBoxResource`. You can include the blackboxes in the library's jar and then pass an annotation to firrtl telling it to emit the blackboxes along with the generated verilog. I would consider [the implementation of simulation-only floating point in the dsptools library](https://github.com/ucb-bar/dsptools/blob/53f6ce7f2b559f1a2bbbbcfa59bbb1d3cd50f73b/src/main/scala/dsptools/numbers/blackbox_compatibility/DspRealVerilatorBB.scala) a good example of why you'd want this and how to use it. #1217 discussed this somewhat as well (this closes #1217).

This PR adds the trait `HasBlackBoxResource` to blackboxes in rocket and adds the contents of the `vsrc/` and `csrc/` directories to `rocketchip.jar`. When running firrtl, the blackboxes will be put in `generated-src/`. However, I haven't modified the build system at all, so they won't be used- the originals in `vsrc/` and `csrc/` will be used instead. However, people using rocket as a library for other projects (like me) will be able to use components from rocket without worrying about maintaining separate copies `vsrc/` and `csrc/`.

I also made some small changes to the `build.sbt` file that make rocket suitable for publishing on something like sonatype. I think @ucbjrl may be interested in doing so.


seldridge:
This is actually a partial solution to the issue of making Firrtl aware of custom passes that may exist outside of Firrtl, e.g., in Rocket Chip or in a `ROCKETCHIP_ADDONS` (some discussion here on this generally: https://github.com/freechipsproject/firrtl/issues/678). I'd suggest going even further in what you expose to Firrtl's class path to include `$(base_dir)/chisel3/target/scala-2.11/*` as that's needed if Firrtl is doing inline Chisel compilation. I'll add a review quickly.

Note: there's still a problem of Firrtl needing to run custom transforms automatically (Chisel's Driver can do this, but Rocket Chip's use of Firrtl as a standalone binary does not). That's outside the scope of this, though...

Makefrag
```diff
@@ -19,8 +19,9 @@ JVM_MEMORY ?= 2G
 SBT ?= java -Xmx$(JVM_MEMORY) -Xss8M -XX:MaxPermSize=256M -jar $(base_dir)/sbt-launch.jar
 SHELL := /bin/bash
 
+ROCKET_CLASSES := $(base_dir)/target/scala-2.11/classes
```
seldridge: This was desperately needed... One optional suggestion: add the Chisel class path and Chisel jar to enable in-line compilation from Firrtl passes. Granted, that is technically outside the scope of this PR and could be added separately.

e.g.: 
```make
ROCKET_CLASSES := $(base_dir)/target/scala-2.11/classes:$(base_dir)/chisel3/target/scala-2.11/*
```

grebe: This makes total sense. At the *very* least it seems like a good idea to s/:=/?=.

grebe:
Good catch with SimJTAG, I think it was merged into master after when I started playing around with this. I also added the `jtag_vpi.tab` file- is that a reasonable thing to do? Not really sure what it's for.

@mwachs5, you are correct that Makefiles are still including .v/.cpp files directly as before. This is intentional. My thinking is that the least disruptive thing would be to not touch the build system in rocket at all. My goal with this PR is to make it possible to use rocket as an SBT dependency (i.e. publishable as a maven/ivy artifact) without separately managing these blackboxes.

If you'd prefer that I modify the makefiles to point to the files in generated-src, I'd be happy to do that. Is that likely to break anything you guys use outside of this repo?

Also, there are some places where I'm not sure it is appropriate to use `HasBlackBoxResource` or not. I'm not sure exactly how `BlackBoxedROM` is meant to work, but my guess is that there is some out-of-band process for generating the ROMs. You may want to do something like [this](https://github.com/ucb-art/rocket-dsp-utils/blob/diplomacyPort/src/main/scala/dsptools/SyncROM.scala#L58) and use `HasBlackBoxInline`. There's also `CloneModule` in diplomacy that uses blackbox. I'm not sure what it's doing exactly, but I don't think there's an external verilog getting generated or linked in anywhere.

mwachs5:
Right now we don't really use either of the mechanisms (the Makefile or the HasBlackBoxResources) outside of this repo. If you think HasBlackBoxResources is the way to go, I think we should just make this repo an example of how to do so.

The `.tab` file is sort of like a binding file for VPI. I plan to deprecate `JTAGVPI` soon, so don't worry too much about it.

grebe:
OK, I'm back! I'm looking for some feedback about if this is the right kind of approach.

Right now I'm reading the `black_box_verilog_files.f` file that lists all the blackboxes and doing some awk/sed to separate c source/headers/verilog files. Perhaps the firrtl passes should be improved to generate a .d file that make can directly include (@chick and @jackkoenig, what do you think of this?), but for now I think this is workable.

I've also refactored the Makefrags a bit to try to cut down on the copy-pasting.

grebe:
Anyone have thoughts on why travis isn't finishing (or starting)?

mwachs5:
Probably because there are conflicts which must be resolved first?

grebe:
OK, finally have tests passing. Now I feel better about asking for feedback.

1) I split out the firrtl/verilog/etc. targets into `/Makefrag.generate_sources` rather than individual makefrags inside `vsim` and `vsrc`. I'm now thinking it would be better to put the contents in the existing `/Makefrag` and cut down the number of files. What do you think?

2) `/Makefrag.generate_sources` makes a makefrag in `$(generated_dir)/$(long_name)` and has an `-include $(sim_sources_makefrag)` as a result. If care isn't taken, a target that doesn't depend on the generated source will still try to include the makefrag and hence generate the source anyways. I have a `$filter(..., $(MAKECMDGOALS))` that conditionally prevents the include for some targets (`clean` and `verilator` are the problematic targets I've run into). I think this is more or less the right way to do this, but I like dumb and simple Makefiles and this is more complicated than I like. If you've got other ideas, I'd be happy to hear them.

3) There are still some hardcoded files in `$(sim_csrcs)` - it seems that `emulator.cc` tries to link against them even if the rocket configuration doesn't include them. 

4) I restructured the `build.sbt` a bit for publishing. @ucbjrl, does it look good to you?

grebe:
@mwachs5 I'm curious what you think about 1) and 2)

vsim/Makefile
```diff
@@ -32,3 +32,13 @@ clean:
 	rm -rf $(junk) simv* csrc *.key DVE* *.h *.a *.daidir $(generated_dir)
 
 .PHONY: default all debug clean
+
+sim_vsrcs = $(base_dir)/vsrc/$(TB).v $(generated_verilog) $(verilog)
+
+sim_csrcs = $(generated_cppfiles)              \
+			$(base_dir)/csrc/remote_bitbang.cc \
+			$(base_dir)/csrc/SimDTM.cc \
+			$(base_dir)/csrc/SimJTAG.cc
```
jackkoenig: Are these tabs?

grebe: They look to be. I suppose they shouldn't be, right?

jackkoenig: I would do spaces and line up the `$` but it's not that important

jackkoenig: The deleted Makefrag seems to have mixed spaces and tabs for this

mwachs5:
Sorry, I'm not sure what feedback I can contribute to (1) and (2). @hcook probably understands this PR better

Makefrag
```diff
@@ -19,8 +19,9 @@ JVM_MEMORY ?= 2G
 SBT ?= java -Xmx$(JVM_MEMORY) -Xss8M -XX:MaxPermSize=256M -jar $(base_dir)/sbt-launch.jar
 SHELL := /bin/bash
 
+ROCKET_CLASSES ?= "$(base_dir)/target/scala-2.11/classes:$(base_dir)/chisel3/target/scala-2.11/*"
```
jackkoenig: I don't think it's necessary for this PR but I think we should consider moving to an `sbt-pack` approach: https://github.com/xerial/sbt-pack

edcote:
Here are some comments.  My engineering manager "hat" is on for these.

I observe there is a lot going on here and judge risk on this PR.

Changes for build.sbt for publishing on sonatype should be separate.  @jackkoenig yes to sbt-pack, or at least to consider a strategy as a separate discussion.

It seems you are trying to manage csrc/vsrc like scala resources, but you are keeping them in current directories.  There is some "re-invent the wheel" when compared to use of standard maven directory structure (src/main/resources)

jackkoenig:
I get errors with this in both vsim and emulator, anyone else having problems?

grebe:
@edcote I agree there's a lot going on. I think I'll split this into a `build.sbt` and BlackBox PR and a separate `Makefile` PR to follow on after.

The easiest way to do that is to keep the current directory structure and a `csrc/` and `vsrc/` to the resources path because that doesn't necessitate any build system changes. Early versions of this PR did precisely that. I think the following build system PR can reorganize the directory structure if you decide that's what you want to do.

Does this sound reasonable?

@jackkoenig I haven't run this in a while, but last time I checked vcs and verilator were working and it looks like travis didn't mind. What kind of errors are you seeing?

edcote:
@grebe My feedback is, without attachment, to take steps to lessen the coupling between the verilog emulator, C emulator, and the chisel/scala generation components.  The decision on how to package resources best left to the gatekeeper.  This is largely academic.  I am not pushing for the change on this.

hcook:
I too would appreciate some smaller PRs, particularly if you can get one that makes the `ROCKET_CLASSES` enhancement and `build.sbt` changes separable from the Makefrag ones.

I have zero attachment to the current location of `csrc` and `vsrc` if there is a more appropriate one.

Thanks for your work on this, obviously of interest to many.

hcook:
Closing in favor of #1302 , feel free to submit make or sbt changes separately.

PR #1232, opened 2018-02-08
Bump chisel and firrtl

jackkoenig:
This is not yet ready to be merge, just trying to organize and stage necessary changes for bumping Chisel.

The biggest issue blocking bumping firrtl is that https://github.com/freechipsproject/chisel3/pull/766 exposed a partial recompilation bug in sbt that is fixed but not yet in an sbt release. See https://github.com/sbt/zinc/pull/484

The main issue requiring changes is that `IO` now calls `cloneType` (as it probably should have in the past) which causes a few Bundles that didn't use to need cloneType now need it. I made some of these work with autoclonetype by making args vals or deleting unused ones.

The TLRegRouter stuff is also an open question since the anonymous class TLRegBundles [as used in RRTest0](https://github.com/freechipsproject/rocket-chip/blob/9f6d586e8c4f3a7a076009d343b0e4d25781a270/src/main/scala/tilelink/RegisterRouterTest.scala#L219) cannot currently but cloned automatically but need to be because of IO now calling cloneType. It seems that this pattern is due for refactoring anyway so I'm looking into how to resolve this in the least boilerplate-y way.

Blocking issues:
- [x] SBT partial compilation bug https://github.com/sbt/zinc/pull/484
    * Resolved by sbt 1.1.1 which was just released, looking into bumping
- [x] TLRegRouter pattern
    * Resolved via https://github.com/freechipsproject/chisel3/pull/771 (changed to deprecation, note: will error in future, but not yet)


colinschmidt:
Were there no other cloneTypes to delete?

jackkoenig:
I'm going to delete the cloneType on [GenericParameterizedBundle](https://github.com/freechipsproject/rocket-chip/blob/4c595d175c2c3d3dfa721a78a4a2a9fab009621b/src/main/scala/util/GenericParameterizedBundle.scala#L7), but other than that I don't see any. Of course since these are elaboration-time issues, I'm probably missing some. If anyone wants to try using this branch with their super-project to help me identify other necessary changes, I would appreciate it.

jackkoenig:
Alright, with https://github.com/freechipsproject/chisel3/pull/787, this PR bumping Chisel and Firrtl removes the partial recompilation problem. I'm leaving removing cloneType on GenericParameterizedBundle as a TODO since this PR will save people enormous amounts of time.

This PR reduces the time taken for `make verilog` in vsim after touching a random, unrelated file from 1m57s to 56s (still too long but better).

davidbiancolin:
Hurry up Travis.

PR #1235, opened 2018-02-11
Catch zero length queue in DCache

colinschmidt:
If you have maxUncachedInFlight set to 0 but do set bufferUncachedRequests you end up with a Queue with zero entries which then generates invalid firrtl.

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -94,6 +94,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   val tl_out_a = Wire(tl_out.a)
   tl_out.a <> outer.bufferUncachedRequests
                 .map(_ min maxUncachedInFlight-1)
+                .filter(_ > 0)
                 .map(Queue(tl_out_a, _, flow = true))
                 .getOrElse(tl_out_a)
 


aswaterman:
I don't think maxUncachedInFlight=0 should be a legal configuration; any load or store to uncacheable memory will deadlock the system.  ```require(maxUncachedInFlight > 0)``` instead?

aswaterman:
Ah, that makes sense.

Chisel should either fail loudly in the Queue constructor or special-case 0 (my preference). @jackkoenig thoughts?

If chisel should fail, we should merge this PR. We can also merge it now as a workaround if it’s important.

colinschmidt:
Its not that important, so we can just wait for the chisel decision to be made.
Thanks for the review

aswaterman:
@jackkoenig note that Queue.irrevocable should explicitly reject 0-entry queues, since the property that Queues turn Decoupled into Irrevocable doesn't hold in this case.  So maybe the special casing should be: change Queue.apply to return enq when n=0, and change Queue.irrevocable to require(n>0).

aswaterman:
Here is my proposal in code https://github.com/freechipsproject/chisel3/tree/zero-entry-queue

jackkoenig:
I think that proposal looks good. Wanna make it a PR?

aswaterman:
https://github.com/freechipsproject/chisel3/pull/780/

aswaterman:
I'm closing this because it will be resolved through Chisel, but note it's not actually fixed until Chisel is bumped.

PR #1238, opened 2018-02-15
Error: don't be an exception wrt. caching

terpstra:
Prior to this PR, the error device was allowed to be cached by
multiple actors despite never probing any of them. This is a
pretty unusual set of properties that has caused us trouble
several times now in the past.

Let's instead put the Error device into one of two very well
established categories: a straight-up MMIO device or a tracked
memory region.


terpstra:
This effectively walks back commit 000cde2f8aa9bee7773cb155c27476a76c50d1c3.
That solved the issue that fence.i flushes D$, but made the error device even more unusual.

aswaterman:
LGTM, but we should get @hcook's take

PR #1239, opened 2018-02-15
Reduce Debug Module "flags"

mwachs5:
This PR does a few things:

1) Reduces the number of zero, read-only `flag` registers in the Debug Module.
2) Removes the `[X:Y]` notation from register names generated with the `RegField.bytes` function, since it doesn't really make sense to have those characters in the name of a register. Instead, they are simply indexed by `_Y` where `Y` is the byte number
3) Prevents the generated JSON files from overwriting eachother if they are actually referring to different address spaces.


src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -111,7 +111,12 @@ case class TLRegisterNode(
       ("baseAddress" -> base) ~
       ("regfields" -> regDescs)
     ))
-    ElaborationArtefacts.add(s"${base}.regmap.json", pretty(render(json)))
+
+    var suffix = 0
+    while( ElaborationArtefacts.contains(s"${base}.${suffix}.regmap.json")){
+      suffix = suffix + 1
+    }
+    ElaborationArtefacts.add(s"${base}.${suffix}regmap.json", pretty(render(json)))
```
hcook: `s"${base}.${suffix}**.**regmap.json"` != `s"${base}.${suffix}regmap.json"` one has a period and one doesnt

src/main/scala/util/GeneratorUtils.scala
```diff
@@ -139,7 +139,12 @@ trait GeneratorApp extends App with HasGeneratorUtilities {
 
 object ElaborationArtefacts {
   var files: Seq[(String, () => String)] = Nil
-  def add(extension: String, contents: => String) {
+
+  def add(extension: String, contents: => String) = {
```
hcook: Are you intentionally making this no longer a `Unit` function with `=`?

src/main/scala/tilelink/RegisterRouter.scala
mwachs5: thanks!

src/main/scala/util/GeneratorUtils.scala
mwachs5: no. thanks.

PR #1246, opened 2018-02-22
Change build to use local .ivy2 and publish Chisel and FIRRTL

jackkoenig:
This is a workaround for an issue with incremental compilation in the
SBT multi-project build that would require recompiling Chisel (and thus
everything) whenever any .scala file was modified.

This is mostly a proposal to see what people think. This requires some changes to any projects using rocket-chip so that's kind of a pain, but I think it should be fairly straight forward. I intend to try this out with a corresponding branch on project-template before actually trying to merge.

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -16,24 +16,34 @@ CXX ?= g++
 CXXFLAGS := -O1
 JVM_MEMORY ?= 2G
 
-SBT ?= java -Xmx$(JVM_MEMORY) -Xss8M -XX:MaxPermSize=256M -jar $(base_dir)/sbt-launch.jar
+JAVA ?= java -Xmx$(JVM_MEMORY) -Xss8M -XX:MaxPermSize=256M
+IVY ?= -Dsbt.ivy.home=$(base_dir)/.ivy2/
+SBT ?= $(JAVA) $(IVY) -jar $(base_dir)/sbt-launch.jar
 SHELL := /bin/bash
 
-FIRRTL_JAR ?= $(base_dir)/firrtl/utils/bin/firrtl.jar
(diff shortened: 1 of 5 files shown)


jackkoenig:
I should add that I looked into (and continue looking into) how to fix the multi-project build but I haven't been able to figure it out. SBT reports that some of the macros defined in Chisel are "not on the classpath" (despite seemingly having no problem compiling things that depend on them), and thus invalidates everything. I've tried deleting anything that seems suspect in the Chisel build.sbt (I'm looking at you `compile-internal`), but to no avail. If anyone figures something out, please let me know. (Note that I did run these experiments on Chisel with https://github.com/freechipsproject/chisel3/pull/766, ie. newer than the version in this PR and on master).

Makefrag
```diff
@@ -16,24 +16,34 @@ CXX ?= g++
...
-FIRRTL ?= java -Xmx$(JVM_MEMORY) -Xss8M -XX:MaxPermSize=256M -cp $(FIRRTL_JAR) firrtl.Driver
+FIRRTL ?= $(JAVA) -cp $(FIRRTL_JAR) firrtl.Driver
+
+$(FIRRTL_STAMP): $(call lookup_scala_srcs, $(firrtl_dir))
+	cd $(firrtl_dir) && $(SBT) publishLocal
+	touch $@
 
-# Build firrtl.jar and put it where chisel3 can find it.
-$(FIRRTL_JAR): $(shell find $(base_dir)/firrtl/src/main/scala -iname "*.scala")
+$(FIRRTL_JAR): $(FIRRTL_STAMP)
```
codelec: your changes work well even without line 35 & 36

jackkoenig:
https://github.com/freechipsproject/chisel3/pull/787 Fixed the main problem this was addressing so closing in favor of #1232 to just keep the build system the same.

PR #1247, opened 2018-02-22
Implement misa.C proposal

aswaterman:
This removes undefined behavior when clearing the misa.C bit when the PC is not currently aligned.  The manner in which it is fixed is still up for debate, but better to remove the undefined behavior in the mean time.  See https://github.com/riscv/riscv-isa-manual/pull/139


aswaterman:
hehe

PR #1250, opened 2018-02-23
Add jtag_vpi.c to sources for vsim

seldridge:
I wasn't able to build a VCS version without this. @mwachs5: seem reasonable to include jtag_vpi.c here?

This should fix https://github.com/freechipsproject/rocket-chip/issues/1061. This is mentioned in https://github.com/ucb-bar/riscv-boom/issues/38.


terpstra:
I've killed your build. I'll restart it once master has an up-to-date cache. (We've been rebuilding everything pointlessly)

mwachs5:
(my goal is to eventually deprecate and remove all the JTAGVPI stuff in favor of the Remote Bitbang stuff).

seldridge:
Seems logical to me. Thanks for identifying a better fix.

PR #1251, opened 2018-02-23
Added functional covers

bipult:
Functional covers for TL Fifofixer.

### DIFF
--- a/src/main/scala/tilelink/FIFOFixer.scala
+++ b/src/main/scala/tilelink/FIFOFixer.scala
@@ -5,6 +5,8 @@ package freechips.rocketchip.tilelink
 import Chisel._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.util._
+import freechips.rocketchip.util.property._
 import scala.math.max
 
 class TLFIFOFixer(policy: TLFIFOFixer.Policy = TLFIFOFixer.all)(implicit p: Parameters) extends LazyModule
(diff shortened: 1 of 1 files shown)


src/main/scala/tilelink/FIFOFixer.scala
```diff
@@ -98,6 +100,31 @@ class TLFIFOFixer(policy: TLFIFOFixer.Policy = TLFIFOFixer.all)(implicit p: Para
...
+        SourceIdSet := UIntToOH(in.a.bits.source)
+      }
+      when (d_first && in.d.fire())  {
+        SourceIdClear := UIntToOH(in.d.bits.source)
+      }
+
+      SourceIdFIFOed := SourceIdFIFOed | SourceIdSet
+      val allIDs_FIFOed = SourceIdFIFOed===Fill(SourceIdFIFOed.getWidth, 1.U)
+
+      cover(allIDs_FIFOed, "COVER all sources", "Cover: FIFOFIXER covers all Source IDs")
```
jchang0: Why don't you just do the following instead of having a tracking variable
(0 to edgeIn.client.endSourceId).foreach { id =>
  cover(a_first && in.a.fire() && !a_notFIFO && in.a.bits.source===id.U, s"Cover: FIFOFIXER covers id ${id}")
}

src/main/scala/tilelink/FIFOFixer.scala
```diff
@@ -98,6 +100,31 @@ class TLFIFOFixer(policy: TLFIFOFixer.Policy = TLFIFOFixer.all)(implicit p: Para
...
+    //cover(flight.reduce(_ && _), "COVER full", "Cover: FIFO is full with all Source IDs")
+      cover(!(flight.reduce(_ || _)), "COVER empty", "Cover: FIFO is empty")
```
jchang0: I don't think this is very interesting because FIFO gets reset to empty...

src/main/scala/tilelink/FIFOFixer.scala
jerryhethatday: I have no idea what is the purpose of `cove`? Can you clarify this? Is this way of doing code coverage? Is there some info I can get about the cover stuff? Is it a framework? Thanks!

PR #1262, opened 2018-03-03
Add BusErrorUnit RegFieldDesc

aswaterman:


src/main/scala/rocket/BusErrorUnit.scala
```diff
@@ -68,17 +68,17 @@ class BusErrorUnit[T <: BusErrors](t: => T, params: BusErrorUnitParams)(implicit
     io.interrupt := (accrued.asUInt & local_interrupt.asUInt).orR
     int_out(0) := (accrued.asUInt & global_interrupt.asUInt).orR
 
-    def reg(r: UInt) = RegField.bytes(r, (r.getWidth + 7)/8)
-    def reg(v: Vec[Bool]) = v.map(r => RegField(1, r))
+    def reg(r: UInt, name: String, reset: Option[BigInt]) = RegFieldGroup(name, None, RegField.bytes(r, (r.getWidth + 7)/8, Some(RegFieldDesc(name, "", reset=reset))))
+    def reg(v: Vec[Bool], name: String, reset: Option[BigInt]) = RegFieldGroup(name, None, v.map(r => RegField(1, r, RegFieldDesc(name, "", reset=reset))))
```
mwachs5: this seems sort of funky... the reset should be per-field so it should be 1 bit each. I'll try this out and push a fix if it doesn't do what I think you meant

aswaterman: That makes sense.  I'm not sure what these new annotations are supposed to do or how to test if they're doing it right; I can test & fix it if you let me know where to look for the output.

mwachs5:
I made these way more verbose (and added a `DescribedRegister` utility that can be used anywhere. It ended up not being that useful for this device because `Vec[Bool]` doesn't play very well with it.

aswaterman:
LGTM, though I can't approve my own PR ;)

hcook:
I like `DescribedReg` a lot. Should it live in `regmapper` rather than `util`? Since users will probably end up importing `RegFieldDesc` also anyways?

PR #1270, opened 2018-03-06
Align RoCCIO with new cloneType

seldridge:
This aligns the RoCC interface with the new cloneType. Simple fix, just make `outer` a `val` like was done in #1232. I'm testing this now.

I lament the state of RoCC regression tests... :sob:

### DIFF
--- a/src/main/scala/tile/LazyRoCC.scala
+++ b/src/main/scala/tile/LazyRoCC.scala
@@ -52,8 +52,6 @@ class RoCCCoreIO(implicit p: Parameters) extends CoreBundle()(p) {
   val busy = Bool(OUTPUT)
   val interrupt = Bool(OUTPUT)
   val exception = Bool(INPUT)
-
-  override def cloneType = new RoCCCoreIO()(p).asInstanceOf[this.type]
 }
 
 /** Base classes for Diplomatic TL2 RoCC units **/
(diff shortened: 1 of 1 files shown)


seldridge:
This builds and passes the [accumulator hello world test](https://github.com/seldridge/rocket-rocc-examples/blob/master/bareMetal/accumulator.S).

src/main/scala/tile/LazyRoCC.scala
```diff
@@ -64,7 +62,7 @@ abstract class LazyRoCC(implicit p: Parameters) extends LazyModule {
   val tlNode: TLNode = TLIdentityNode()
 }
 
-class RoCCIO(outer: LazyRoCC)(implicit p: Parameters) extends RoCCCoreIO()(p) {
+class RoCCIO(val outer: LazyRoCC)(implicit p: Parameters) extends RoCCCoreIO()(p) {
```
colinschmidt: I think this can/should be private. @jackkoenig is that right?

I'm not sure whats even using the member. Perhaps its vestigial?

hcook: The new `cloneType` doesn't care about public/private, so it's a matter of policy on the part of the device maker. @mwachs5 was advocating for more of them to be `private` for consistency, but there was at least one case in sifive-blocks where it had to be public.

I think you are right this is vestigal but I'm merging it anyway. All of this code should be revisited in light of auto-punching of diplomatic bundles and LazyScopes...

hcook:
@seldridge thanks. is there a particular `Config` you'd like to see added to travis?

seldridge:
@hcook: I would be very happy with just adding a test build of `RoCCExampleConfig` that doesn't run any tests.

Running riscv-tests with a RoCC config isn't providing any significant coverage.

To get coverage, we'd have to run https://github.com/seldridge/rocket-rocc-examples/ or something like it (bare metal tests like riscv-tests that hit the available RoCCs). This, however, would overly complicate the regression build flow. It likely should happen at some point, but there's no rush. Just building `RoCCExampleConfig` is likely enough to catch 95% of the issues related to "RoCCExampleConfig doesn't work".

I'm open to discussion here and for providing improved support/testing of the existing example RoCCs.

PR #1271, opened 2018-03-06
Debug: use non-deprecated notequals

mwachs5:
Avoids this error: 
[deprecated] Debug.scala:941 (1 calls): $bang$eq is deprecated: "Use \'=/=\', which avoids potential precedence problems"


hcook:
@mwachs5 I think @jackkoenig handled this in https://github.com/freechipsproject/rocket-chip/pull/1261/, which I am inclined to merge now-ish. Do you expect to have to make any further rocket bumps that would get stuck behind it if we run into trouble with it in federation?

mwachs5:
@hcook I'm fine with you merging #1261 instead of this

PR #1279, opened 2018-03-09
More IP-XACT-like RegFieldDesc

mwachs5:
I didn't realize that IP-XACT supported finer grained explanations of `WriteType` (and less interestingly `ReadAction`). Add these, as well as `volatile`, and replace the overly vague `*SPECIAL` access Types.

Also update some descriptions in Bus Error Unit, Debug, and PLIC to use these and correct some other inaccuracies


src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -229,7 +231,8 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
     }
 
     def thresholdRegDesc(i: Int) = RegFieldDesc(s"threshold_$i", s"Interrupt & claim threshold for target $i. Maximum value is ${nPriorities}.",
-      reset=if (nPriorities > 0) None else Some(1), access=RegFieldAccessType.RWSPECIAL)
+      reset=if (nPriorities > 0) None else Some(1),
```
aswaterman: Threshold,  priority, and enable registers are _not_ reset.

mwachs5: right, but if there is only one priority value they are constant 1, right? That's what the code is trying to say.

aswaterman: oooh I misread the code. ya

src/main/scala/regmapper/RegField.scala
```diff
@@ -7,34 +7,54 @@ import chisel3.util.{ReadyValidIO}
...
 ){
 }
 
-// Our descriptions are in terms of RegFields only, which is somewhat unusual for
-// developers who are used to things being defined as bitfields within registers.
-// The "Group" allows a string & (optional) description to be added which describes the conceptual "Group"
-// the RegField belongs to. This can be used by downstream flows as they see fit to
-// present the information.
+object RegFieldDescReserved {
+  def apply(): RegFieldDesc = RegFieldDesc("reserved", "", access=RegFieldAccessType.R, reset=Some(0))
```
hcook: I would probably code this as a `def reserved: RegFieldDesc` method on an companion object named `RegFieldDesc`, especially if you think there might be more of them eventually

mwachs5: 👍

PR #1286, opened 2018-03-15
Initial commit of SBA Debug feature

rmac-sifive:
Initial pull request for System Bus Access debug feature

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,8 @@ jobs:
         - travis_wait 80 make emulator-ndebug -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
     - <<: *test
       script:
         - travis_wait 80 make emulator-ndebug -C regression SUITE=UnittestSuite JVM_MEMORY=3G
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,8 +73,12 @@ jobs:
         - travis_wait 80 make emulator-ndebug -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
-        - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
-        - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST=MemTest64 JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST=MemTest64 JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST=MemTest32 JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST=MemTest32 JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST=MemTest8 JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST=MemTest8 JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
     - <<: *test
       script:
         - travis_wait 80 make emulator-ndebug -C regression SUITE=UnittestSuite JVM_MEMORY=3G
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -34,7 +34,6 @@ object DMIConsts{
 }
 
 object DsbBusConsts {
-
   def sbAddrWidth = 12
   def sbIdWidth   = 10 
 
@@ -128,7 +127,7 @@ case class DebugModuleParams (
   nProgramBufferWords: Int = 16,
   nAbstractDataWords : Int = 4,
   nScratch : Int = 1,
-  hasBusMaster : Boolean = true,
+  hasBusMaster : Boolean = false,
   supportQuickAccess : Boolean = false,
   supportHartArray   : Boolean = false,
   hartIdToHartSel : (UInt) => UInt = (x:UInt) => x,
@@ -267,8 +266,7 @@ object RWNotify {
 class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyModule {
 
   // For Shorter Register Names
-  import uncore.devices._
-  import uncore.devices.DMI_RegAddrs._
+  import DMI_RegAddrs._
 
   val intnode = IntNexusNode(
     sourceFn       = { _ => IntSourcePortParameters(Seq(IntSourceParameters(1, Seq(Resource(device, "int"))))) },
@@ -418,9 +416,6 @@ class TLDebugModuleOuterAsync(device: Device)(implicit p: Parameters) extends La
 
 class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: Int)(implicit p: Parameters) extends LazyModule
 {
-
-  import uncore.devices._
-  import uncore.devices.DMI_RegAddrs._
 
   val cfg = p(DebugModuleParams)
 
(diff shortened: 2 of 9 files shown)


src/main/scala/devices/debug/Debug.scala
```diff
@@ -123,28 +130,14 @@ case class DebugModuleParams (
...
   nAbstractDataWords : Int = 4,
   nScratch : Int = 1,
-  //TODO: Use diplomacy to decide if you want this.
-  hasBusMaster : Boolean = false,
-  hasAccess128 : Boolean = false,
-  hasAccess64  : Boolean = false,
-  hasAccess32  : Boolean = false,
-  hasAccess16  : Boolean = false,
-  hasAccess8   : Boolean = false,
+  hasBusMaster : Boolean = true,
```
mwachs5: i think the default should be false

src/main/scala/devices/debug/Debug.scala
```diff
@@ -276,7 +269,8 @@ object RWNotify {
 class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyModule {
 
   // For Shorter Register Names
-  import DMI_RegAddrs._
+  import uncore.devices._
```
mwachs5: what does this first import get you?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
       }
     }
 
+    //--------------------------------------------------------------
+    // System Bus Access
+    //--------------------------------------------------------------
+    // --- SBCS Registers ---
+
+    sb2tlOpt match {
```
mwachs5: rather than the generic `match { case (Some)... case (None) }` pattern you could use: `sb2tlOpt.foreach { sb2tl => ...}`.  This is preferable I believe. Or if you want to return something you can use `sb2tlOpt.map{...}.getOrElse{...}` (see below)

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+       
+        /* Flags for setting error flag registers, needed because the number of SBDATA and
+         * SBADDRESS registers depends on system bus widths
+         */
+        val anyAddressWrEn = Wire(init = false.B)
+        val anyDataRdEn    = Wire(init = false.B)
+        val anyDataWrEn    = Wire(init = false.B)
+
+        // --- SBCS Status Register ---
+        val SBCSFieldsReg      = Reg(new SBCSFields())
```
mwachs5: do these registers show up with reasonable names in the verilog (questionable because they are inside a code block). If not, we may want to `suggestName` on some of the key signals, or return them from this code block to get assigned some names.

src/main/scala/devices/debug/Debug.scala
```diff
@@ -1115,6 +1371,9 @@ class ClockedDMIIO(implicit val p: Parameters) extends ParameterizedBundle()(p){
 
 class DMIToTL(implicit p: Parameters) extends LazyModule {
 
+  import uncore.devices._
+  import uncore.devices.DMI_RegAddrs._
+
```
hcook: did you actually need to add this import here?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+        when (~io.dmactive) {
+          SBADDRESS0FieldsReg := new SBADDRESS0Fields().fromBits(0.U(32.W))
+        }.otherwise {
+          SBADDRESS0FieldsReg := Mux(SBADDRESS0WrEn && !SBCSFieldsReg.sberror && !SBCSFieldsReg.sbbusy, new SBADDRESS0Fields().fromBits(SBADDRESS0WrData),
+                                 Mux(sb2tl.module.io.dataOutValid && SBCSFieldsReg.sbautoincrement,     new SBADDRESS0Fields().fromBits(autoIncrementedAddr(31,0)),SBADDRESS0FieldsReg))
+        }
+
+        SBADDRESS0RdData := SBADDRESS0FieldsReg
+
+        dmiNode.regmap(
```
mwachs5: I didn't think you could do this, as it would override/get overriden by the other dmiNode.regmap() calls. I could be wrong (and if I am my mind is blown because a lot of things I have worked on would have been way easier), but I *think* you need to do something like:

```
val  (sbaddr0fields, sbaddr1fields, ...)= sb2tlOpt.map { .... 
...
val addr0fields = Seq(RegField(...))
...
(addr0fields, addr1fields, ...)
}.getOrElse{(Nil, Nil ...)}
```

Then later in your regmap call:

```
dmiNode.regmap(
...
(DMI_SBADDRESS0 << 2)  -> sbaddr0fields,
(DMI_SBADDRESS1 << 2)  -> sbaddr1fields,
...
)
```

src/main/scala/devices/debug/Debug.scala
```diff
@@ -1159,3 +1418,89 @@ class DMIToTL(implicit p: Parameters) extends LazyModule {
...
+      val sizeIn       = UInt(INPUT, 128)
+      val dataIn       = UInt(INPUT, 128)
+      val dmactive     = Bool(INPUT)
+      val dataOutValid = Bool(OUTPUT)
+      val dataOut      = UInt(OUTPUT, 128)
+      val sbStateOut   = UInt(OUTPUT, log2Ceil(SystemBusAccessState.maxId))
+    })
+
+    import uncore.devices._
+    import uncore.devices.DMI_RegAddrs._
```
hcook: I'd consider adding `import SystemBusAccessState._` here to reduce clutter on the below lines.

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+        dmiNode.regmap(
+          (DMI_SBADDRESS0 << 2) -> Seq(RWNotify(32, SBADDRESS0RdData.asUInt(), SBADDRESS0WrData, SBADDRESS0RdEn, SBADDRESS0WrEn,
+            Some(RegFieldDesc("dmi_sbaddress0", "", reset=Some(0))))))
+
+        sb2tl.module.io.addrIn := SBADDRESS0RdData.asUInt()
+
+        anyAddressWrEn         := SBADDRESS0WrEn 
+
+        // Instantiate ADDR1-3 registers as needed depending on system bus address width
+        val hasSBAddr1 = (sb2tl.module.edge.bundle.addressBits >= 33)
```
mwachs5: this is all good, it may be cleaner to define the SBADDRESS1FieldsReg as an Option on hasSBAddr1. Then you can do all of this stuff on optional functions, including adding it to the RegMap (since you can't do what you are doing here to only sometimes call `dmiNode.regmap`)

For example, val SBADDRESS1FieldsReg = hasSBAddr1.option(Reg(...))
sbaddress1fields = SBADDRESS1FieldsReg.map {
  (all the logic) 
Seq(RegField...)
}.getOrElse(Nil)

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
       }
     }
 
+    //--------------------------------------------------------------
+    // System Bus Access
+    //--------------------------------------------------------------
+    // --- SBCS Registers ---
+
+    sb2tlOpt match {                   
+      case Some(sb2tl) => {
```
hcook: Now seeing the final verison, I'd swap the above two lines out out for `sb2tlOpt.foreach { sb2tl => ... ` (and drop the empty `None` case)

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+
+
+        // --- System Bus Data Registers ---
+        val SBDATA0FieldsReg = Reg(new SBDATA0Fields())
+        val SBDATA1FieldsReg = Reg(new SBDATA1Fields())
+        val SBDATA2FieldsReg = Reg(new SBDATA2Fields())
+        val SBDATA3FieldsReg = Reg(new SBDATA3Fields())
+        val SBDATA1RdData    = Wire(init = SBDATA1FieldsReg)
+        val SBDATA2RdData    = Wire(init = SBDATA2FieldsReg)
+        val SBDATA0RdData    = Wire(init = SBDATA0FieldsReg)
```
mwachs5: kinda weird these are out of order

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+        cover(SBCSFieldsReg.sbaccess === 3.U && !SBAlignmentError,    "SBCS Cover", "64-bit access")
+        cover(SBCSFieldsReg.sbaccess === 4.U && !SBAlignmentError,    "SBCS Cover", "128-bit access")
+        cover(SBCSFieldsReg.sbautoincrement && SBCSFieldsReg.sbbusy,  "SBCS Cover", "Access with autoincrement set")
+        cover(!SBCSFieldsReg.sbautoincrement && SBCSFieldsReg.sbbusy, "SBCS Cover", "Access without autoincrement set")
+
+        SBCSRdData             := SBCSFieldsReg
+        SBCSRdData.sbversion   := 1.U(1.W) // This code implements a version of the spec after January 1, 2018
+        SBCSRdData.sbbusy      := (sb2tl.module.io.sbStateOut =/= SystemBusAccessState.Idle.id.U)
+        SBCSRdData.sbasize     := sb2tl.module.edge.bundle.addressBits.U
+        SBCSRdData.sbaccess128 := (sb2tl.module.edge.bundle.dataBits == 128).B
```
mwachs5: can you use the already computed "has128" and such?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -1159,3 +1418,89 @@ class DMIToTL(implicit p: Parameters) extends LazyModule {
 
   }
 }
+
+class SBToTL(implicit p: Parameters) extends LazyModule {
+
+  val node = TLClientNode(Seq(TLClientPortParameters(Seq(TLClientParameters("sba")))))
```
mwachs5: what does this name get used for @hcook ? Depending on the answer, it might be better to call it "debug_sba" or similar.

src/main/scala/devices/debug/Debug.scala
```diff
@@ -1159,3 +1418,89 @@ class DMIToTL(implicit p: Parameters) extends LazyModule {
...
+    val src  = Wire(init = 0.U)
+    val (_,  gbits) = edge.Get(src, io.addrIn, io.sizeIn)
+    val (_, pfbits) = edge.Put(src, io.addrIn, io.sizeIn, io.dataIn)
+    val (_,  nbits) = edge.Put(src, io.addrIn, io.sizeIn, data=0.U, mask=0.U)
+    val sbState = Reg(init = 0.U)
+ 
+    io.sbStateOut := sbState
+    when       (sbState === SystemBusAccessState.SBReadRequest.id.U) { tl.a.bits := pfbits
+    }.elsewhen (sbState === SystemBusAccessState.SBWriteRequest.id.U){ tl.a.bits :=  gbits
+    }.otherwise                                                 {      tl.a.bits :=  nbits
```
hcook: the unaligned `{` D:

src/main/scala/devices/debug/Debug.scala
```diff
@@ -1159,3 +1418,89 @@ class DMIToTL(implicit p: Parameters) extends LazyModule {
...
+    val responseValid = tl.d.valid
+    val responseReady = tl.d.ready
+ 
+    when(~io.dmactive){
+      sbState := SystemBusAccessState.Idle.id.U
+    }.elsewhen (sbState === SystemBusAccessState.Idle.id.U){
+      sbState := Mux(io.rdEn, SystemBusAccessState.SBReadRequest.id.U,
+                 Mux(io.wrEn, SystemBusAccessState.SBWriteRequest.id.U,sbState))
+    }.elsewhen (sbState === SystemBusAccessState.SBReadRequest.id.U){
+      sbState := Mux(requestValid && requestReady, SystemBusAccessState.SBReadResponse.id.U, sbState)
```
mwachs5: this seems sort of circular logic. requestValid depends on tl.a.valid which is depending on SystmBusAccessState == SBReadRequest, so why include it in this check again? Basically as soon as you see the `ready` you can progress, right?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -1159,3 +1418,89 @@ class DMIToTL(implicit p: Parameters) extends LazyModule {
...
+      sbState := Mux(requestValid && requestReady, SystemBusAccessState.SBReadResponse.id.U, sbState) 
+    }.elsewhen (sbState === SystemBusAccessState.SBWriteRequest.id.U){
+      sbState := Mux(requestValid && requestReady, SystemBusAccessState.SBWriteResponse.id.U, sbState)
+    }.elsewhen (sbState === SystemBusAccessState.SBReadResponse.id.U){
+      sbState := Mux(responseValid && responseReady, SystemBusAccessState.Idle.id.U, sbState)
+    }.elsewhen (sbState === SystemBusAccessState.SBWriteResponse.id.U){
+      sbState := Mux(responseValid && responseReady, SystemBusAccessState.Idle.id.U, sbState)
+    }
+ 
+    io.dataOutValid := (sbState === SystemBusAccessState.SBReadResponse.id.U) && responseValid && responseReady
```
mwachs5: same note about the circular logic. `responseReady` is always true when `(sbState === SystemBusAccessState.SBReadResponse.id.U)`, right?

src/main/scala/devices/debug/dm_registers.scala
```diff
@@ -1,8 +1,8 @@
-package freechips.rocketchip.devices.debug
+package uncore.devices
```
mwachs5: you have to fix this. Ah, now I see why you were importing `uncore.devices` earlier. We should fix the `make chisel` target to to do the right thing in `riscv-debug-spec`

src/main/scala/devices/debug/dm_registers.scala
```diff
@@ -1,8 +1,8 @@
...
 
 import Chisel._
 
-// This file was auto-generated from the repository at https://github.com/riscv/riscv-debug-spec.git,
+// This file was auto-generated from the repository at https://github.com/sifive/riscv-debug-spec.git,
```
mwachs5: yeah we should fix this in the `make chisel` target as well.

src/main/scala/devices/debug/Debug.scala
mwachs5: This might also be a nice use case for the `DescribedReg` utility, if they aren't getting nice names.

src/main/scala/devices/debug/Debug.scala
hcook: I don't think it has impact on any verilog names. It appears when you connect the module to the rest of the diplomacy graph, i.e. `frontBus.fromPort { sb2tl.node }`. I think it's only worth being more specific if a single module has multiple nodes.

src/main/scala/devices/debug/Debug.scala
mwachs5: Am I correct @terpstra ? We can't just call `dmiNode.regmap()` several times, right?

src/main/scala/devices/debug/Debug.scala
rmac-sifive: I could've done "sbaccess128 := hasSBData2and3", but I still have to access dataBits for the others since those aren't needed above. I figured for code clarity this would be nicer

src/main/scala/devices/debug/Debug.scala
rmac-sifive: Yeah, the compiler complained that it couldn't find "DMI_RegAddrs.DMI_CONTROL", which is strange because this was working with the previous version of dm_registers.scala

src/main/scala/devices/debug/Debug.scala
rmac-sifive: Changing to "debug_sba" for clarity

src/main/scala/devices/debug/Debug.scala
mwachs5: yeah makes sense

src/main/scala/devices/debug/Debug.scala
rmac-sifive: The horror! Fixed :)

src/main/scala/devices/debug/Debug.scala
mwachs5: I think that is because you need to manually change the package name in the `dm_registers.scala`

src/main/scala/devices/debug/Debug.scala
rmac-sifive: Yeah, I see how it reads that way, and I could move to the Response state when I see "ready" from TileLink, I just wanted to be explicit that the request happens when valid && ready. There should be no circular dependency since sbState is a register

src/main/scala/devices/debug/Debug.scala
rmac-sifive: True, responseReady is always asserted. I want it to be explicit in the code that this is a valid/ready interface

src/main/scala/devices/debug/Debug.scala
rmac-sifive: Yeah, Wes confirmed I can't make multiple regmap() calls. I'll work on a change

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+        val SBCSFieldsReg      = Reg(new SBCSFields())
+        val SBCSFieldsRegReset = Wire(init = (new SBCSFields()).fromBits(0.U))
+        val SBCSRdData         = Wire(init = new SBCSFields().fromBits(0.U)) 
+        val SBCSWrDataVal      = Wire(init = 0.U(32.W))
+        val SBCSWrData         = Wire(init = new SBCSFields().fromBits(SBCSWrDataVal))
+        val SBCSRdEn, SBCSWrEn = Wire(init = false.B)
+
+
+        // --- System Bus Registers ---
+        val SBADDRESS0FieldsReg = Reg(new SBADDRESS0Fields())
```
mwachs5: If you ditch the "Fields" and just create `SBADDRESS` and `SBDATA` of type UInt(32.W), you can just treat these all as  Sequences of whatever they are, conditioned by their length. That would probably greatly reduce this code, especially the part below where you are doing the same thing for each one basically.

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +721,263 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+
+        val SBADDRESS0WrData, SBADDRESS1WrData, SBADDRESS2WrData, SBADDRESS3WrData = Wire(init = 0.U(32.W))
+        val SBADDRESS0RdEn,   SBADDRESS1RdEn,   SBADDRESS2RdEn,   SBADDRESS3RdEn   = Wire(init = false.B)  
+        val SBADDRESS0WrEn,   SBADDRESS1WrEn,   SBADDRESS2WrEn,   SBADDRESS3WrEn   = Wire(init = false.B)  
+
+        val autoIncrementedAddr = Wire(init = 0.U(128.W))
+        autoIncrementedAddr := sb2tl.module.io.addrIn + (1.U << SBCSFieldsReg.sbaccess)
+
+        // ADDR0 Register is required
+        when (~io.dmactive) {
```
mwachs5: RE my above comment, then this code would become something like 

```
SBADDRESSRegs.zipWithIndex.foreach { case (a, i) => 
  when (~io.dmactive) {
     a := 0.U
  }.otherwise {
     a:= Mux(SBADDRESSWrEn(i) && ...)
 }

...
}
```

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +716,184 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+          sbaddrfields(i) = Seq(RWNotify(32, SBADDRESSRdData(i).asUInt(), SBADDRESSWrData(i), SBADDRESSRdEn(i), SBADDRESSWrEn(i),
+                            Some(RegFieldDesc("dmi_sbaddr"+i, "", reset=Some(0)))))
+        } else {
+          sbaddrfields(i) =  Seq.empty[RegField]
+        }
+      }
+
+      if(hasAddr(3)) {
+        sb2tl.module.io.addrIn := Cat(SBADDRESSRdData(3),SBADDRESSRdData(2),SBADDRESSRdData(1),SBADDRESSRdData(0))
+        anyAddressWrEn         := SBADDRESSWrEn(3) || SBADDRESSWrEn(2) || SBADDRESSWrEn(1) || SBADDRESSWrEn(0)
```
mwachs5: You could simplify this a bit, by making `SBADDRESSWrEn = Seq.fill(4)(Wire (init = False.B))`, then `anyAddressWrEn := SBADDRESSWrEn.reduce(_ || _)` (no special casing needed).

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +716,184 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+          sbdatafields(i) = Seq(RWNotify(32, SBDATARdData(i).asUInt(), SBDATAWrData(i), SBDATARdEn(i), SBDATAWrEn(i),
+                            Some(RegFieldDesc("dmi_sbdata"+i, "", reset=Some(0)))))
+        } else {
+          sbdatafields(i) = Seq.empty[RegField]
+        }
+      }
+
+      if(hasData(3)) {
+        sb2tl.module.io.dataIn := Cat(SBDATARdData(3),SBDATARdData(2),SBDATARdData(1),SBDATARdData(0))
+        anyDataRdEn            := SBDATARdEn(3) || SBDATARdEn(2) || SBDATARdEn(1) || SBDATARdEn(0)
```
mwachs5: similar as above for how you could avoid writing this out 4 times

src/main/scala/devices/debug/Debug.scala
rmac-sifive: Thanks, I knew there'd be a way to do this better. Is there any similar trick for the address bus concatenation, or do I have to do special casing?

mwachs5: I think if you try to assign `sb2tl.moduile.io.addrIn` to something that is widier than it, it will just truncate the lower bits anyway. So you can try `sb2tl.module.io.addrIn := Cat(SBADDRESSRdData.reverse:_*`

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +716,158 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+          when (~io.dmactive) {
+            a := 0.U(32.W)
+          }.otherwise {
+            a := Mux(SBADDRESSWrEn(i) && !SBCSFieldsReg.sberror && !SBCSFieldsReg.sbbusy, SBADDRESSWrData(i),
+                 Mux(sb2tl.module.io.dataOutValid && SBCSFieldsReg.sbautoincrement,       autoIncrementedAddr(32*i+31,32*i), SBADDRESSFieldsReg(i)))
+          }
+
+          SBADDRESSRdData(i) := SBADDRESSFieldsReg(i)
+
+          sbaddrfields(i) = Seq(RWNotify(32, SBADDRESSRdData(i).asUInt(), SBADDRESSWrData(i), SBADDRESSRdEn(i), SBADDRESSWrEn(i),
```
mwachs5: wait... is this using a `var`? Let's avoid that...

src/main/scala/devices/debug/Debug.scala
```diff
@@ -722,6 +716,158 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
     }
 
+    //--------------------------------------------------------------
+    // System Bus Access
+    //--------------------------------------------------------------
+    // --- SBCS Registers ---
+    val (sbcsFields, sbAddrFields, sbDataFields):
+        (Seq[RegField], Array[Seq[RegField]], Array[Seq[RegField]]) = sb2tlOpt.map { sb2tl =>
+       
+      var sbaddrfields, sbdatafields: Array[Seq[RegField]] = new Array[Seq[RegField]](4)
```
mwachs5: do we really want to use a var here? Can we avoid that?

src/main/scala/devices/debug/Debug.scala
mwachs5: instead you can do: 
```
val sbaddrfields = SBADDRESSFieldsReg.zipWithIndex.map { case (a, i) => 
   if (hasAddr(i)) {
   ... 
  Seq(RWNotify...)
} else {
  Nil // (or Seq.empty, whatevz)
}
```

src/main/scala/devices/debug/SBA.scala
```diff
@@ -0,0 +1,320 @@
...
+    val tryRdEn = (SBADDRESSWrEn(0) && SBCSFieldsReg.sbreadonaddr) || (SBDATARdEn(0) && SBCSFieldsReg.sbreadondata)
+
+    val sbAccessError = (SBCSFieldsReg.sbaccess === 0.U) && (SBCSFieldsReg.sbaccess8   =/= 1.U) ||
+                        (SBCSFieldsReg.sbaccess === 1.U) && (SBCSFieldsReg.sbaccess16  =/= 1.U) ||
+                        (SBCSFieldsReg.sbaccess === 2.U) && (SBCSFieldsReg.sbaccess32  =/= 1.U) ||
+                        (SBCSFieldsReg.sbaccess === 3.U) && (SBCSFieldsReg.sbaccess64  =/= 1.U) ||
+                        (SBCSFieldsReg.sbaccess === 4.U) && (SBCSFieldsReg.sbaccess128 =/= 1.U)
+
+    val compareAddr = Wire(UInt(32.W)) // Need use written or latched address to detect error case depending on how transaction is initiated
+    compareAddr := Mux(SBADDRESSWrEn(0),SBADDRESSWrData(0),SBADDRESSFieldsReg(0))
```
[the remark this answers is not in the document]
rmac-sifive: Yes, according to the TileLink spec bad alignments aren't supported. Looks like there are also assertions in the TileLink code to catch this case

src/main/scala/devices/debug/SBA.scala
```diff
@@ -0,0 +1,320 @@
...
+
+object SystemBusAccessModule
+{
+  def apply(sb2tl: SBToTL, dmactive: Bool)(implicit p: Parameters): (Seq[RegField], Seq[Seq[RegField]], Seq[Seq[RegField]]) =
+  {
+    import DMI_RegAddrs._
+
+    val cfg = p(DebugModuleParams)
+
+    val anyAddressWrEn = Wire(init = false.B).suggestName("anyAddressWrEn")
```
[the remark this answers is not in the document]
rmac-sifive: I checked the waves after refactoring this into an object, and it didn't surface these signal names. If there's a better way to globally expose these names for debug, let me know

src/main/scala/devices/debug/SBA.scala
```diff
@@ -0,0 +1,320 @@
...
+    sb2tl.module.io.dmactive := dmactive
+
+    when (~dmactive) {
+      SBCSFieldsReg := SBCSFieldsRegReset
+    }.otherwise {
+      SBCSFieldsReg.sbbusyerror     := Mux(SBCSWrEn && SBCSWrData.sbbusyerror,                   false.B, // W1C
+                                       Mux(anyAddressWrEn && SBCSFieldsReg.sbbusy,               true.B, // Set if a write to SBADDRESS occurs while busy
+                                       Mux((anyDataRdEn || anyDataWrEn) && SBCSFieldsReg.sbbusy, true.B, SBCSFieldsReg.sbbusyerror))) // Set if any access to SBDATA occurs while busy
+      SBCSFieldsReg.sbreadonaddr    := Mux(SBCSWrEn, SBCSWrData.sbreadonaddr   , SBCSFieldsReg.sbreadonaddr)
+      SBCSFieldsReg.sbautoincrement := Mux(SBCSWrEn, SBCSWrData.sbautoincrement, SBCSFieldsReg.sbautoincrement)
```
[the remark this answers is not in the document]
rmac-sifive: Nothing in the spec prevents this explicitly. The only constraint on writing SBCS is: "To avoid race conditions, debuggers must not try to clear `sberror` until they read `sbbusy` as 0.'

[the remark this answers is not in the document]
mwachs5: Ok. I'll file an issue on the spec that clarifies more things the debugger should not try to do (shouldn't change what your implementation has to do).

src/main/scala/devices/debug/Debug.scala
```diff
@@ -725,6 +721,182 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
+                             (SBCSWrData.sbaccess === 2.U) && (SBCSFieldsReg.sbaccess32  =/= 1.U) ||
+                             (SBCSWrData.sbaccess === 3.U) && (SBCSFieldsReg.sbaccess64  =/= 1.U) ||
+                             (SBCSWrData.sbaccess === 4.U) && (SBCSFieldsReg.sbaccess128 =/= 1.U)  
+
+      when (~io.dmactive) {
+        SBCSFieldsReg := SBCSFieldsRegReset
+      }.otherwise {
+        SBCSFieldsReg.sbbusyerror     := Mux(SBCSWrEn && SBCSWrData.sbbusyerror,                   false.B, // W1C
+                                         Mux(anyAddressWrEn && SBCSFieldsReg.sbbusy,               true.B, // Set if a write to SBADDRESS occurs while busy
+                                         Mux((anyDataRdEn || anyDataWrEn) && SBCSFieldsReg.sbbusy, true.B, SBCSFieldsReg.sbbusyerror))) // Set if any access to SBDATA occurs while busy
```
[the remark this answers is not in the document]
rmac-sifive: Good catch. You're right, `SBCSFieldsReg.sbbusy` won't be asserted until the cycle after `sbStateOut =/= Idle`

src/main/scala/devices/debug/SBA.scala
```diff
@@ -177,17 +177,17 @@ object SystemBusAccessModule
...
       SBCSFieldsReg.sbaccess        := Mux(SBCSWrEn, SBCSWrData.sbaccess, SBCSFieldsReg.sbaccess)
       SBCSFieldsReg.sbversion       := 1.U(1.W) // This code implements a version of the spec after January 1, 2018
-      SBCSFieldsReg.sbbusy          := (sb2tl.module.io.sbStateOut =/= SystemBusAccessState.Idle.id.U)
-      SBCSFieldsReg.sbasize         := sb2tl.module.edge.bundle.addressBits.U
-      SBCSFieldsReg.sbaccess128     := (cfg.maxSupportedSBAccess == 128).B
-      SBCSFieldsReg.sbaccess64      := (cfg.maxSupportedSBAccess >=  64).B
-      SBCSFieldsReg.sbaccess32      := (cfg.maxSupportedSBAccess >=  32).B
-      SBCSFieldsReg.sbaccess16      := (cfg.maxSupportedSBAccess >=  16).B
-      SBCSFieldsReg.sbaccess8       := (cfg.maxSupportedSBAccess >=   8).B
+      SBCSFieldsReg.sbbusy          := sbBusy
```
mwachs5: I don't really have a problem with this change (not enough to block the PR), but do you ever need SBCSFieldsReg.sbbusy anymore? Couldnt you just use sbBusy everwhere?

If you don't use it Chisel will remove it anyway, but just curious.

src/main/scala/devices/debug/SBA.scala
```diff
@@ -173,8 +173,8 @@ object SystemBusAccessModule
       SBCSFieldsReg.sberror         := Mux(SBCSWrEn && SBCSWrData.sberror =/= 0.U, 0.U, // W1C
                                        Mux((sb2tl.module.io.wrEn && !sb2tl.module.io.wrLegal) || (sb2tl.module.io.rdEn && !sb2tl.module.io.rdLegal), 2.U, // Bad address accessed
                                        Mux((sb2tl.module.io.rdDone || sb2tl.module.io.wrDone) && sb2tl.module.io.respError, 3.U, // Response error from TL
-                                       Mux((tryWrEn || tryRdEn) && sbAccessError, 3.U, // Access size error
-                                       Mux((tryWrEn || tryRdEn) && sbAlignmentError, 3.U, SBCSFieldsReg.sberror))))) // Address alignment error
+                                       Mux((tryWrEn || tryRdEn) && sbAlignmentError, 3.U, // Access size error
```
mwachs5: the comment doesn't match the code, but fine with the code

src/main/scala/devices/debug/SBA.scala
```diff
@@ -0,0 +1,320 @@
...
+    SBCSWrEn.suggestName("SBCSWrEn")
+    SBCSRdEn.suggestName("SBCSRdEn")
+
+    val sbcsfields = Seq(RWNotify(32, SBCSRdData.asUInt(), SBCSWrDataVal, SBCSRdEn, SBCSWrEn,
+                        Some(RegFieldDesc("dmi_sbcs", "", reset=Some(0)))))
+
+    // --- System Bus Address Registers ---
+    // ADDR0 Register is required
+    // Instantiate ADDR1-3 registers as needed depending on system bus address width
+    val hasSBAddr1 = (sb2tl.module.edge.bundle.addressBits >= 33)
```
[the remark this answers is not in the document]
terpstra: Should be fine, but > 32 is more what you mean, right?

src/main/scala/devices/debug/SBA.scala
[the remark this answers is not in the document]
mwachs5: confiremd with @terpstra that you were right, Tile Link HW won't detect an error in this case, so you are correct ot check it before sending it

src/main/scala/devices/debug/SBA.scala
```diff
@@ -0,0 +1,320 @@
...
+    val rdTxValid = sbState === SBReadResponse.id.U && responseValid && responseReady
+    val txLast    = counter === ((1.U << io.sizeIn) - 1.U)
+    counter := Mux((wrTxValid || rdTxValid) && txLast, 0.U,
+               Mux((wrTxValid || rdTxValid)          , counter+1.U, counter))
+
+    for (i <- 0 until (cfg.maxSupportedSBAccess/8)) {
+      io.rdLoad(i) := rdTxValid && (counter === i.U)
+    }
+
+    // --- State Machine to interface with TileLink ---
```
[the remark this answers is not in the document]
terpstra: There is a problem: d.ready is low based on a state machine that expects a.fire => can lead to dead-lock. Otherwise, just a few stylistic things I'd change, but not worth getting into here.

src/main/scala/devices/debug/SBA.scala
```diff
@@ -233,33 +233,36 @@ class SBToTL(implicit p: Parameters) extends LazyModule {
  
     val (tl, edge) = node.out(0)
 
+    val sbState = Reg(init = 0.U)
```
mwachs5: minor nit... shouldn't this be one of your states? IDLE or whatever?

.travis.yml
```diff
@@ -73,6 +73,8 @@ jobs:
         - travis_wait 80 make emulator-ndebug -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST="MemTest64 MemTest32 MemTest8" JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
```
mwachs5: Nice. Let's keep an eye on the Travis runtime for this and make sure it's not excessive, otherwise more tests is better!

src/main/scala/devices/debug/SBA.scala
```diff
@@ -0,0 +1,320 @@
...
+    when (~dmactive) {
+      SBCSFieldsReg := SBCSFieldsRegReset
+    }.otherwise {
+      SBCSFieldsReg.sbbusyerror     := Mux(SBCSWrEn && SBCSWrData.sbbusyerror,     false.B, // W1C
+                                       Mux(anyAddressWrEn && sbBusy,               true.B, // Set if a write to SBADDRESS occurs while busy
+                                       Mux((anyDataRdEn || anyDataWrEn) && sbBusy, true.B, SBCSFieldsReg.sbbusyerror))) // Set if any access to SBDATA occurs while busy
+      SBCSFieldsReg.sbreadonaddr    := Mux(SBCSWrEn, SBCSWrData.sbreadonaddr   , SBCSFieldsReg.sbreadonaddr)
+      SBCSFieldsReg.sbautoincrement := Mux(SBCSWrEn, SBCSWrData.sbautoincrement, SBCSFieldsReg.sbautoincrement)
+      SBCSFieldsReg.sbreadondata    := Mux(SBCSWrEn, SBCSWrData.sbreadondata   , SBCSFieldsReg.sbreadondata)
+      SBCSFieldsReg.sberror         := Mux(SBCSWrEn && SBCSWrData.sberror =/= 0.U, 0.U, // W1C
```
mwachs5: The spec says 

```  When the debug module's system bus
            master causes a bus error, this field gets set. The bits in this
            field remain set until they are cleared by writing 1 to them.
```

@timsifive can confirm, but this was a question that was asked before, and I believe the intention is that *each* bit must be written that is set in order to clear it.

.travis.yml
```diff
@@ -73,8 +73,12 @@ jobs:
         - travis_wait 80 make emulator-ndebug -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
-        - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST="MemTest64 MemTest32 MemTest8" JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
-        - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST="MemTest64 MemTest32 MemTest8" JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
+        - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JTAG_DTM_TEST=MemTest64 JTAG_DTM_ENABLE_SBA=on JVM_MEMORY=3G
```
mwachs5: sigh. This is unfortunate that this didnd't work, since each test now needs to initialize OpenOCD and GDB. @timsifive didn't `gdbserver.py` use to take in a list of tests? Is that no longer supported?

timsifive:
gdbserver.py still takes a list of tests. It always initializes a new
gdb/OpenOCD for each test.

src/main/scala/devices/debug/Periphery.scala
```diff
@@ -30,6 +31,11 @@ class DebugIO(implicit val p: Parameters) extends ParameterizedBundle()(p) with
 trait HasPeripheryDebug { this: BaseSubsystem =>
   val debug = LazyModule(new TLDebugModule(pbus.beatBytes))
   pbus.toVariableWidthSlave(Some("debug")){ debug.node }
+
+
+  debug.dmInner.dmInner.sb2tlOpt.foreach { sb2tl  =>
+    fbus.fromPort(Some("debug_sb")){ TLWidthWidget(1) := sb2tl.node }
```
codelec: can i know the intention behind attaching debug_sb to fbus and not as a coherent master on sbus ?
this becomes an issue when using sba in fesvr/dtm.cc to read to_host which will most likely not be flushed out(by the core) from the cache to memory or is it that in that case you have to use progbuf to read the cache

rmac-sifive: Hey @codelec, see the debug spec at https://github.com/riscv/riscv-debug-spec. This decision was made to reduce complexity of the Debug unit such that it does not need to consider coherence

mwachs5: Fesvr uses the program buffer at this time

ccelio: Thank you for the clarification. I was hoping this would give us less invasive polling of the tohost.

mwachs5: Yes, we can probably do that in many cases. I don't believe that putting this on the FrontBus excludes it from being coherent, but @hcook may be able to answer that better than I.

PR #1296, opened 2018-03-20
Update Debug Module registers

mwachs5:
RISC-V Debug spec was recently updated to support more than 1024 harts. This change was made backwards compatible for systems with up to 32 harts, but some registers names changed and some fields were added/renamed.

Although the RISC-V debug spec now supports 2^20 harts, rocket-chip still only allows up to 1024 harts.

This PR also brings in the changes to the System Bus Access registers  version 1, and should probably be merged before #1286

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -571,7 +571,7 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
     val haltedSummary = Cat(haltedStatus.map(_.orR).reverse)
     val HALTSUM1RdData = (new HALTSUM1Fields()).fromBits(haltedSummary)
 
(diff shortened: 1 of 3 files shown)


src/main/scala/devices/debug/Debug.scala
```diff
@@ -569,7 +569,12 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
     }
 
     val haltedSummary = Cat(haltedStatus.map(_.orR).reverse)
-    val HALTSUMRdData = (new HALTSUMFields()).fromBits(haltedSummary)
+    val HALTSUM1RdData = (new HALTSUM1Fields()).fromBits(haltedSummary)
+
+    val selectedHaltedStatus = Mux((selectedHartReg >> 5) >= (nComponents.U >> 5), 0.U, haltedStatus(selectedHartReg >> 5))
```
rmac-sifive: Is Chisel smart enough to only generate a 5-bit comparator (instead of 10-bit) on this mux select (as opposed to if you'd coded the select as selectHartReg(10:5) >= nComponents(10:5))?

PR #1302, opened 2018-03-22
Add BlackBox resources to rocket, take 2

grebe:
This PR does a subset of #1230 (and also incorporates new feedback), namely:
1) Move csrc and vsrc into src/main/resources
2) Add `HasBlackBoxResource` to some black boxes
3) Adds `ROCKET_CLASSES` to Makefrag
4) Other small changes to the Makefrags to reflect these changes.

It **does not**:
1) Change how the build system works (beyond updating paths to verilog and cpp files)
2) Change build.sbt at all
3) Use the generated verilog or cpp files to build simulators (instead, it uses the originals in src/main/resources)

@hcook @edcote: If desired, I can split these two commits into separate PRs if you want even higher granularity for review.


hcook:
Planning to merge this shortly, thanks for your patience.

Closes #1217

PR #1304, opened 2018-03-22
TLRAM: add support for ECC

terpstra:
As described in the code, this SRAM uses either RMW with back-to-back SRAM read-write or a direct W if the operation is larger than the ECC granularity.


terpstra:
Also, when used with eccBytes=1 and code=Identity, the RTL degenerates into the same SRAM microarchitecture as the previous TLRAM implementation. There are a couple constants that firrtl fails to propagate, but those should go away after @jackkoenig bug fixes firrtl.

PR #1307, opened 2018-03-23
Testable field addition

srivatsa611y:
Adding testable parameter to the Debug register and including it in the Regfield description. 

Also added code to send that information to json files

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -344,7 +344,7 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
     // DMCONTROL is the only register, so it's at offset 0.
     dmiNode.regmap(
       0 -> Seq(RWNotify(32, DMCONTROLRdData.asUInt(),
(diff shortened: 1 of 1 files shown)


src/main/scala/devices/debug/Debug.scala
```diff
@@ -344,7 +344,7 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
     // DMCONTROL is the only register, so it's at offset 0.
     dmiNode.regmap(
       0 -> Seq(RWNotify(32, DMCONTROLRdData.asUInt(),
-        DMCONTROLWrDataVal, DMCONTROLRdEn, DMCONTROLWrEn, Some(RegFieldDesc("dmi_dmcontrol", "", reset=Some(0)))))
+        DMCONTROLWrDataVal, DMCONTROLRdEn, DMCONTROLWrEn, Some(RegFieldDesc("dmi_dmcontrol", "", reset=Some(0), testable=false))))
```
mwachs5: the commit comment is wrong for what you are actually changing here.

PR #1321, opened 2018-04-04
manual blackbox for synthesis (SimDTM)

edcote:
The DPI call definition, SV bit type, # delays, and DPI call in the file are not liked by the synthesis tool I am using (yosys).

The fix allows  the file to be read using yosys command: `read_verilog -lib SimDTM.v`

### DIFF
--- a/vsrc/SimDTM.v
+++ b/vsrc/SimDTM.v
@@ -1,6 +1,7 @@
 // See LICENSE.SiFive for license details.
 //VCS coverage exclude_file
 
+`ifndef SYNTHESIS
 import "DPI-C" function int debug_tick
 (
   output bit     debug_req_valid,
@@ -14,6 +15,7 @@ import "DPI-C" function int debug_tick
   input  int        debug_resp_bits_resp,
   input  int        debug_resp_bits_data
 );
+`endif //!SYNTHESIS
 
 module SimDTM(
   input clk,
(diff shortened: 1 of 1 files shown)


hcook:
I don't have an objection to this per se, but isn't the fact that you're trying and failing to synthesize a simulation only construct indicative of a problem in your flow? (E.g. incorrect identification of the DUT top module?)

@aswaterman and I were actually talking about doing what I think is the opposite of this: intentionally injecting unsynthesizable constructs into simulation-only blackboxes to ensure noisy failures should they accidentally be included in a synthesis run. What do you think about about that approach?

edcote:
No, the issue is not related to the incorrect setting of top module for synthesis (i.e. top module is _not_ set to TestHarness that contains the simulation models).

The synthesis tool's front end errors out during the initial verilog parse.

The workaround is to emit the generated Verilog per module (not single file) and not feed the test harness to Yosys.

I don't understand why you need to inject unsynthesizable constructs.  Just make sure your RTL is lint clean before attempting synthesis (i.e., verilator --lint-only or any of the commercial lint tools).

aswaterman:
Our internal flow does generate one file per module, then exclude simulation-only models from being parsed by the synthesis tool.  Perhaps that's the right thing to do here, too.  (I think it's just a Firrtl switch and a minor Makefile change.)

edcote:
I guess I could that.  I'll close.

PR #1322, opened 2018-04-04
manual blackbox for synthesis (SimJTAG)

edcote:
The DPI call definition, SV bit type, # delays, and DPI call are not liked by the synthesis tool I am using (yosys).

The fix allows the file to be read w/ yosys command: `read_verilog -lib SimJTAG.v`


edcote:
see https://github.com/freechipsproject/rocket-chip/pull/1321
closing

PR #1330, opened 2018-04-11
vsrc: AsyncResetReg adjustments

hcook:
Changing blackbox packaging strategy revealed bad behavior on certain simulators.

### DIFF
--- a/src/main/resources/vsrc/AsyncResetReg.v
+++ b/src/main/resources/vsrc/AsyncResetReg.v
@@ -52,7 +52,7 @@ module AsyncResetReg (
       integer                    initvar;
       reg [31:0]                 _RAND;
       _RAND = {1{$random}};
-`endif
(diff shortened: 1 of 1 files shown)


src/main/resources/vsrc/AsyncResetReg.v
```diff
@@ -39,43 +39,45 @@
...
-      q = 1'b0;
-`endif
+        q_reg = 1'b0;
       end 
 `ifdef RANDOMIZE
- `ifndef verilator
- `endif
  `ifdef RANDOMIZE_REG_INIT
       else begin
+  `ifdef verilator
```
mwachs5: Oops this should be ifndef, righ?

hcook: good catch, thx

PR #1332, opened 2018-04-11
Reg desc anno

depappas-sifive:
This code need to be cleaned up.

### DIFF
--- a/src/main/scala/regmapper/Annotation.scala
+++ b/src/main/scala/regmapper/Annotation.scala
@@ -10,6 +10,8 @@ import org.json4s.JsonAST.JValue
 import org.json4s.JsonDSL._
 import org.json4s.jackson.JsonMethods.{pretty, render}
 
+// Do not call this function
+// This function is called by DescribedRegChiselAnnotation
 case class RegFieldDescAnnotation(
     target: Named,
     desc: String) extends SingleTargetAnnotation[Named] {
@@ -24,9 +26,39 @@ object RegFieldDescAnnotation {
   }
 }
 case class DescribedRegChiselAnnotation(
-    target: InstanceId,
-    desc: RegFieldDesc) extends ChiselAnnotation with RunFirrtlTransform {
-  def toFirrtl = RegFieldDescAnnotation(target.toNamed, desc.toMap.toString)
+  target: InstanceId,
+  reg: RegFieldDesc) extends ChiselAnnotation with RunFirrtlTransform {
+  def toFirrtl: RegFieldDescAnnotation = {
+    val byte = 0 // TODO
+    val offset = 0 // TODO
+    val json = toJson(byte, offset).toString
+
+    RegFieldDescAnnotation(target.toNamed, json)
+  }
+
+  def toJson(byteOffset: Int, bitOffset: Int): JValue = {
+    (("byteOffset" -> s"0x${byteOffset.toHexString}") ~
+      ("bitOffset" -> bitOffset) ~
+      ("name" -> reg.name) ~
+      ("description" -> reg.desc) ~
+      ("resetValue" -> reg.reset) ~
+      ("group" -> reg.group) ~
+      ("accessType" -> reg.access.toString) ~
+      ("writeType" -> reg.wrType.getOrElse("").toString) ~
+      ("readAction" -> reg.rdAction.getOrElse("").toString) ~
+      ("volatile" -> reg.volatile) ~
+      ("enumerations" -> reg.enumerations.map {
+        case (key, (name, edesc)) => {
+          (("value" -> key) ~ ("name" -> name) ~ ("description" -> edesc))
+        }
+      })
+      )
+  }
+
+
+  // ("bitWidth"     -> reg.width) ~
+  // ("groupreg"    -> reg.map{_.groupreg}) ~
+
   def transformClass = classOf[DescribedRegDumpTransform]
 }
 
--- a/src/main/scala/regmapper/Annotation.scala
+++ b/src/main/scala/regmapper/Annotation.scala
@@ -121,7 +121,7 @@ object RegAnnotationUtil {
           }.getOrElse(anonName)) -> f.toJson(byte, offset)
           //RegFieldhHolder(named, f.toJson(byte, offset).toString)
           val json = f.toJson(byte, offset).toString
-          RegFieldDescAnnotation(named, json)
+          RegFieldDescAnnotation(named, json.toString)
         }
       }
     }
--- a/src/main/scala/regmapper/Annotation.scala
+++ b/src/main/scala/regmapper/Annotation.scala
@@ -25,6 +25,8 @@ object RegFieldDescAnnotation {
     }.toMap
   }
 }
+
+
 case class DescribedRegChiselAnnotation(
   target: InstanceId,
   json: String) extends ChiselAnnotation with RunFirrtlTransform {
@@ -157,7 +159,7 @@ object RegAnnotationUtil {
           //RegFieldhHolder(named, f.toJson(byte, offset).toString)
           val json = regField.toJson(byte, offset).toString
 
-          annotate(DescribedRegChiselAnnotation(named, json.toString()))
+          annotate(DescribedRegChiselAnnotation(named, json.toString))
         }
       }
     }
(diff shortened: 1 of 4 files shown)


mwachs5:
Also, can we have better commit messages in the final version than multiple commits that just say `annos`? I assume those will be squashed into something more meaningful?

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,17 +2,121 @@
...
+import firrtl.annotations._
+import freechips.rocketchip.regmapper.RegFieldAccessType.RegFieldAccessType
+import freechips.rocketchip.regmapper.RegFieldRdAction.RegFieldRdAction
+import freechips.rocketchip.regmapper.RegFieldWrType.RegFieldWrType
+import org.json4s.JsonAST.JValue
 import org.json4s.JsonDSL._
 import org.json4s.jackson.JsonMethods.{pretty, render}
 
+// Do not call this function
+// This function is called by DescribedRegChiselAnnotation
```
mwachs5: then can't it be placed inside that class and marked `private`?

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,17 +2,121 @@
...
+    val regDesc: freechips.rocketchip.regmapper.RegFieldDesc = regField.desc.get
+
+    //require("", regDesc.desc != None)
+
+    ( ("byteOffset"   -> s"0x${byteOffset.toHexString}") ~
+      ("bitOffset"    -> bitOffset) ~
+      ("bitWidth"     -> regField.width) ~
+      ("name"         -> regDesc.name) ~
+      ("description"  -> regDesc.desc) ~
+// TODO      ("resetValue"   -> regDesc.reset.getOrElse("0")) ~
```
mwachs5: why can't this handle optional parameters?

src/main/scala/regmapper/Annotation.scala
```diff
@@ -21,3 +125,64 @@ object RegMappingAnnotation {
...
+case class RegFieldDescSer(
+  name: String,
+  desc: String,
+  group: String,
+  groupDesc: String,
+  access: String,
+  wrType: String,
+  rdAction: String,
+  volatile: Boolean,
+  reset: BigInt
```
mwachs5: not all registers have reset values. This is a pretty critical thing to capture if it's not reset.

src/main/scala/regmapper/DescribedReg.scala
```diff
@@ -23,7 +24,14 @@ object DescribedReg {
     enumerations: Map[BigInt, (String, String)] = Map()): (T, RegFieldDesc) = {
     val rdesc = RegFieldDesc(name, desc, None, None,
       access, wrType, rdAction, volatile, reset.map{_.litValue}, enumerations)
+
     val reg = reset.map{i => RegInit(i)}.getOrElse(Reg(gen))
+    val byte = 0 // TODO
```
mwachs5: why is this in here? The register description doesn't know or need to know about byte and offset.

mwachs5: a `DescribedReg` may not even end up in a regmap call ever... I don't get what this is here for.

src/main/scala/regmapper/RegField.scala
```diff
@@ -141,9 +142,29 @@ case class RegField(width: Int, read: RegReadFn, write: RegWriteFn, desc: Option
 {
   require (width > 0, s"RegField width must be > 0, not $width")
 
-  def pipelined = !read.combinational || !write.combinational
+  def pipelined: Boolean = !read.combinational || !write.combinational
```
mwachs5: why are we randomly adding the type to some functions/vars and not others? This seems unrelated to the PR

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Ask Jack

src/main/scala/regmapper/RegField.scala
depappas-sifive: Rocket-chip is not run against the lint rules and does not pass lint. 

Public functions without a return type are a lint violation.

The return types were not added randomly. They were added to functions that were being modified. 

The Intellij IDE considers lint violations to be errors and this causes problems when using the IDE.

Henry suggested making a separate PR for rocket-chip lint fixes.

I can remove the Lint fixes and put them in a PR.

mwachs5: Sounds good... could address this ancient issue: https://github.com/freechipsproject/rocket-chip/issues/255

richardxia: I don't think we should make lint changes *just* based on what your editor considers a linting issue. If we don't unilaterally enforce lint at the CI level, then anyone not using the same editor as you may just continue to commit code that your editor considers to be a lint violation. If you want to make style changes, then please add a CI check to lint the code.

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Comment updated

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Optional params are added but I am not sure what I should put for the default param values in the Else.

Comments?

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Should I add a hasRest boolean?

src/main/scala/regmapper/Annotation.scala
mwachs5: I don't understand why you need anythng if it's optional. If it's not there, then there IS NO reset value. I don't know what you are doing with this so what the limitation is. If this is emitting JSON as it was before, then if the resetValue is omitted from the JSON, then the field is not reset.

src/main/scala/regmapper/Annotation.scala
mwachs5: Again, I don't know what you are doing and why the absence of a `resetValue` field in the emitted JSON is not allowed / sufficient. If there is no `resetValue`, that is the same as `hasReset = false`

src/main/scala/regmapper/RegField.scala
richardxia: There is no scalastyle-config.xml in rocket-chip.

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,196 @@
...
+object RegFieldDescAnnotation {
+  def nameMap(state: CircuitState): Map[String,Seq[String]] = {
+    state.annotations.collect {
+      case RegFieldDescAnnotation(mname, cname) => mname.toString -> cname
+    }.toMap
+  }
+}
+
+
+case class DescribedRegChiselAnnotation(
```
mwachs5: I think we chatted about this, but there is no need for a `DescribedReg` annotation. `DescribedReg` is just a convenience for the designer in coding up a `RegFieldDesc`. It is only once a `RegFieldDesc` is applied to a `RegField` that is passed into a call to `regmap()` that we care about it. At the point of the `DescribedReg` call it is meaningless.

depappas-sifive:
"There is no scalastyle-config.xml in rocket-chip."

Correct. But there is a  scalastyle-config.xml in federation which is used
by Intellij for both federation and rocket-chip.

depappas-sifive:
The Described Reg annotation was removed. I did not create it.

src/main/scala/regmapper/Annotation.scala
mwachs5: Not sure if you still have a pending change, but the DescribedReg annotation still exists in this version

hcook: I added it as an initial proposal for @depappas-sifive to work from, but I think we've decided to go the route of annotating the entire map at once inside the call to `regmap`.

I think the only thing we're giving up is the opportunity to annotate the specific `Reg` component with its description (rather than annotation the enclosing `Module` with the entire map). Annotating the `Reg` component in particular would make it much easier to extract e.g. the final signal name. Sadly, not all information is known inside of the call to `DescribedReg` (like the base address). A combination approach might be preferable but we can try that in a follow-up PR.

mwachs5: Yeah... but I expect there will be many more `RegFieldDesc` than there are `DescribedRegs`.

depappas-sifive: removed

depappas-sifive: Removed

PR #1333, opened 2018-04-11
Annotations: ShouldBeRetimed applies RetimeModuleAnnotation

hcook:
The trait `ShouldBeRetimed` can be used to apply `RetimeModuleAnnotations` to Modules that should be the target of [register retiming](https://en.wikipedia.org/wiki/Retiming) in downstream tools.

### DIFF
--- a/src/main/scala/util/Annotations.scala
+++ b/src/main/scala/util/Annotations.scala
@@ -1,0 +1,40 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import Chisel._
+import chisel3.experimental.{annotate, ChiselAnnotation, RawModule}
+import firrtl.annotations._
+
+// TODO: replace this with an implicit class when @chisel unprotects dontTouchPorts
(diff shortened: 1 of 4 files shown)


hcook:
@jackkoenig I was wondering whether there would be any advantage to using `ModuleName` as opposed to `Named`?

src/main/scala/util/Annotations.scala
```diff
@@ -0,0 +1,40 @@
...
+
+  def dontTouchPortsExcept(f: Data => Boolean): this.type = {
+    self.getModulePorts.filterNot(f).foreach(dontTouch(_))
+    self
+  }
+}
+
+case class RetimeModuleAnnotation(target: Named) extends SingleTargetAnnotation[Named] {
+  def duplicate(n: Named) = this.copy(n)
+}
```
jackkoenig: Using `ModuleName` here means any FIRRTL annotation can use that type information. It's not a huge deal but I think it also conveys intent better.

PR #1319, opened 2018-04-03
aggregate rocketchip subtasks

edcote:
My use case is:  I want to update my local repo to HEAD, compile, publish JARs in "less" steps.

Here are the commands I use:

1. Build FIRRTL and run `sbt publishLocal` (no change)

2. Build and publish other JARs in a single step:
~~~~
cd /path/to/rocketchip
# update to HEAD
git pull origin master
git submodule update --init --recursive
# overkill force re-publish (not exactly needed, but good practice under regressed environment)
rm -rf ~/.ivy2/local/berkeley
rm -rf ~/.ivy2/local/edu.berkeley.cs
sbt clean compile publishLocal
~~~~

more info : https://www.scala-sbt.org/0.13/docs/Multi-Project.html#Aggregation

Low urgency on this PR.  Taking the time to give back.

PR #1335, opened 2018-04-11
Update build.sbt for publishing.

grebe:
This should help people use rocket as an ivy dependency. Duplicates functionality from #1319.

@ucbjrl, would appreciate your feedback.


edcote:
My "manager" feedback is to ask why are you doing more than #1319?

Can you justify changing organization/version versions?  This could break someone else's API.  Imho, this would be at the discretion of the project's maintainers, not lowly contributors such as ourselves.

What value do you get from aggregate more than rocketchip level?

grebe:
This is to make it possible to publish on a repo like sonatype so you can use rocket-chip as an ivy dep. @ucbjrl publishes artifacts for edu.berkeley.cs. I'm setting the version to a snapshot, which is the standard practice for ivy stuff when it's not a release.

As far as I know, nobody who is a maintainer has any plans to use rocket as an ivy dep, so this change should be irrelevant to them, but I know several people who would like to use components from rocket in other projects (e.g. just to grab an async Fifi).

Maybe aggregating the subjects is unnecessary. Sometimes I'll mess around with a subproject and publish local the subproject rather than the whole thing to save time, so it's nice to aggregate them properly, but that's pretty niche.

build.sbt
```diff
@@ -20,11 +20,17 @@ lazy val commonSettings = Seq(
 )
 
 lazy val chisel = (project in file("chisel3")).settings(commonSettings)
-lazy val hardfloat  = project.dependsOn(chisel).settings(commonSettings)
-lazy val macros = (project in file("macros")).settings(commonSettings)
+lazy val hardfloat  = project.dependsOn(chisel).
+                              aggregate(chisel).
+                              settings(commonSettings)
```
hcook: can you format these with the same order and spacing as `rocketchip` below?

build.sbt
```diff
@@ -20,11 +20,17 @@ lazy val commonSettings = Seq(
...
+lazy val macros = Project(
+        id = "rocket-macros",
+        base = file("macros")).
+        settings(commonSettings)
```
hcook: Can you left align these two spaces?

edcote:
@hcook ack.

grebe:
Bump

grebe:
Closed by #1835

PR #1337, opened 2018-04-11
Add Debug resethaltreq feature

mwachs5:
This is an implementation of https://github.com/riscv/riscv-debug-spec/pull/255, ~~which is still just a proposal so this should not be merged yet. The point of this PR is to determine whether the proposal is practical to implement.~~

Another suggestion for how to implement the proposal is to condition the hart's reset vector on the `resethaltreq`, but since I am not sure what all is involved in entering debug mode (presumably more than just its reset vector), I would prefer to add this logic on the debug interrupt.

This does sketchy things crossing clock and reset domains without any synchronization, which is why I am not in love with the spec proposal. There is also the sketchy non-constant reset value in the `onResetHaltReq` registers which is key to the implementation.

This also assumes that debug module reset is equivalent to hart reset, but that assumption has been baked into the code for a while anyway.

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -244,6 +244,8 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
     val io = IO(new Bundle {
       val ctrl = (new DebugCtrlBundle(nComponents))
       val innerCtrl = new DecoupledIO(new DebugInternalBundle())
+      val innerOnResetHaltReq = Vec(nComponents, Bool()).asOutput
+      val innerDebugInt = Vec(nComponents, Bool()).asInput
     })
 
     //----DMCONTROL (The whole point of 'Outer' is to maintain this register on dmiClock (e.g. TCK) domain, so that it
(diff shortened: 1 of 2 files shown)


mwachs5:
This is part of the spec now, but probably shoudl update `dm_registers.scala` to match the current

src/main/scala/devices/debug/Debug.scala
```diff
@@ -368,10 +378,17 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
 
     for (component <- 0 until nComponents) {
       when (~dmactive) {
-        debugIntNxt(component) := false.B
+        debugIntNxt(component)        := false.B
+        onResetDebugIntNxt(component) := false.B
       }. otherwise {
         when (DMCONTROLWrEn && DMCONTROLWrData.hartsello === component.U) {
-          debugIntNxt(component) := DMCONTROLWrData.haltreq
+          debugIntNxt(component)        := DMCONTROLWrData.haltreq
```
hcook: nit: this spacing seems weird

src/main/scala/devices/debug/Debug.scala
```diff
@@ -475,6 +501,9 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
     val haltedBitRegs    = RegInit(Vec.fill(nComponents){false.B})
     val resumeReqRegs    = RegInit(Vec.fill(nComponents){false.B})
     val haveResetBitRegs = RegInit(Vec.fill(nComponents){true.B})
+    //!!! Non-constant reset value!
+    val onResetHaltRegs  = RegInit(io.innerOnResetHaltReq)
```
hcook: This gives me a bad feeling

mwachs5:
This was handled differently and better by @ernie-sifive

PR #1341, opened 2018-04-13
Adding --rbb-port=PORT parameter to the Emulator

noureddine-as:
This is added to allow defining a port for the JTAG Remote Bit Bang, instead of affecting one randomly.
remote_bitbang_t(port); instead of remote_bitbang_t(0);
If no port is selected default value is 0, then the port is chosen randomly.
Tested with OpenOCD + GDB

https://github.com/freechipsproject/rocket-chip/issues/1339

### DIFF
--- a/src/main/resources/csrc/emulator.cc
+++ b/src/main/resources/csrc/emulator.cc
@@ -69,8 +69,7 @@ EMULATOR OPTIONS\n\
   -m, --max-cycles=CYCLES  Kill the emulation after CYCLES\n\
        +max-cycles=CYCLES\n\
   -s, --seed=SEED          Use random number seed SEED\n\
-
-      --rbb-port=PORT      Use PORT for remote bit bang (with OpenOCD and GDB) \n\
+  -r, --rbb-port=PORT      Use PORT for remote bit bang (with OpenOCD and GDB) \n\
                            If not specified, a random port will be chosen\n\
                            automatically.\n\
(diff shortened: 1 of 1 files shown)


src/main/resources/csrc/emulator.cc
```diff
@@ -69,6 +69,10 @@ EMULATOR OPTIONS\n\
   -m, --max-cycles=CYCLES  Kill the emulation after CYCLES\n\
        +max-cycles=CYCLES\n\
   -s, --seed=SEED          Use random number seed SEED\n\
+
+      --rbb-port=PORT      Use PORT for remote bit bang (with OpenOCD and GDB) \n\
```
mwachs5: Can you add the “short” version to the help here?

src/main/resources/csrc/emulator.cc
```diff
@@ -110,6 +114,8 @@ int main(int argc, char** argv)
   uint64_t max_cycles = -1;
   int ret = 0;
   bool print_cycles = false;
+  // Port numbers are 16 bit unsigned integers. 
+  uint16_t port = 0;
```
mwachs5: Can you call this “rbb_port”

src/main/resources/csrc/emulator.cc
noureddine-as: Done.

src/main/resources/csrc/emulator.cc
noureddine-as: Done.

PR #1344, opened 2018-04-13
bump riscv-tools for debug related submodules

mwachs5:
Test PR for https://github.com/riscv/riscv-tools/pull/190, Don't merge until that passes and this on riscv-tools master.

Ok actually I also bumped `riscv-tests` according to https://github.com/riscv/riscv-tests/pull/123

And then I bumped riscv-openocd for https://github.com/riscv/riscv-openocd/pull/236


mwachs5:
Replaced by #1357

PR #1345, opened 2018-04-14
Subsystem: support optional ports via configs

hcook:
This PR adds support for specifying that individual ports are optional via parameters stored in the `Config` classes. This should allow their removal without modification of any `trait` extensions. I also:

- updated the SimAXIMem API to be a better example of using Edges
- cleaned up the various `Port` traits
- added helper methods for tying off various diplomatic bundles

(/cc @ppoocde  @mayoru)

### DIFF
--- a/src/main/scala/amba/ahb/Bundles.scala
+++ b/src/main/scala/amba/ahb/Bundles.scala
@@ -32,18 +32,18 @@ class AHBBundle(params: AHBBundleParameters) extends AHBBundleBase(params)
     hreadyout.dir match {
       case INPUT =>
         hreadyout := Bool(false)
-        hresp     := Bool(false)
+        hresp     := AHBParameters.RESP_OKAY
         hrdata    := UInt(0)
       case OUTPUT => 
         hmastlock := Bool(false)
-        htrans    := UInt(0)
+        htrans    := AHBParameters.TRANS_IDLE
         hsel      := Bool(false)
         hready    := Bool(false)
         hwrite    := Bool(false)
         haddr     := UInt(0)
         hsize     := UInt(0)
-        hburst    := UInt(0)
-        hprot     := UInt(0)
+        hburst    := AHBParameters.BURST_SINGLE
+        hprot     := AHBParameters.PROT_DEFAULT
         hwdata    := UInt(0)
       case _ =>
     }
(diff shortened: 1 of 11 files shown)


src/main/scala/amba/ahb/Bundles.scala
```diff
@@ -27,6 +27,27 @@ class AHBBundle(params: AHBBundleParameters) extends AHBBundleBase(params)
...
+        hwrite    := Bool(false)
+        haddr     := UInt(0)
+        hsize     := UInt(0)
+        hburst    := UInt(0)
+        hprot     := UInt(0)
+        hwdata    := UInt(0)
+      case _ =>
+    }
+  }
 }
```
terpstra: Could we use the named constants here?
htrans := TRANS_IDLE
hburst := BURST_SINGLE
hprot := PROT_DEFAULT
hresp := RESP_OKAY

src/main/scala/amba/axi4/Bundles.scala
```diff
@@ -70,6 +70,24 @@ class AXI4Bundle(params: AXI4BundleParameters) extends AXI4BundleBase(params)
...
+      case OUTPUT =>
+        ar.valid := Bool(false)
+        aw.valid := Bool(false)
+        w.valid  := Bool(false)
+        r.ready  := Bool(true)
+        b.ready  := Bool(true)
+      case _ =>
+    }
+  }
 }
```
terpstra: Here you've taken the opposite policy. You allowed requests to go through (ready=true) even though you never respond AR=>R, AW=>B. This seems the opposite of what you might want. At least AHB/APB actually make progress when you say 'ready true'; they are the zero device. Here, AXI is the request black-hold device.

src/main/scala/tilelink/Bundles.scala
```diff
@@ -191,6 +191,24 @@ class TLBundle(params: TLBundleParameters) extends TLBundleBase(params)
...
+        d.valid := Bool(false)
+      case OUTPUT =>
+        a.valid := Bool(false)
+        c.valid := Bool(false)
+        e.valid := Bool(false)
+        b.ready := Bool(true)
+        d.ready := Bool(true)
+      case _ =>
+    }
+  }
```
terpstra: Same as with the AXI ready.

src/main/scala/amba/ahb/Bundles.scala
hcook: I'll use the constants. I think the blocking-forever policy is the right one.

src/main/scala/amba/axi4/Bundles.scala
hcook: I think the blocking-forever policy is the right one. I'll make them consistent.

PR #1349, opened 2018-04-14
Updates to riscv-tools/Debug SBA error dfns

mwachs5:
Based on feedback from implementing SBA there were some edits/clarifications to the spec. Incorporate them here. 

Also the OpenOCD with `prefer_sba` command is now on master of riscv-tools, so this bumps riscv-tools to that version, which would be OK to merge.

### DIFF
--- a/src/main/scala/devices/debug/dm_registers.scala
+++ b/src/main/scala/devices/debug/dm_registers.scala
@@ -57,21 +57,20 @@ object DMI_RegAddrs {
   */
   def DMI_HARTINFO =  0x12
 
-  /* This register selects which of the 32-bit portion of the hart array mask register
-      is accessible in \Rhawindow.
-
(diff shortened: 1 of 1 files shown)


src/main/scala/devices/debug/dm_registers.scala
```diff
@@ -797,13 +805,19 @@ class SBCSFields extends Bundle {
...
             0: There was no bus error.
 
             1: There was a timeout.
 
             2: A bad address was accessed.
 
-            3: There was some other error (eg. alignment).
+            3: There was an alignment error.
+
+            4: An access of unsupported size was requested.
```
rmac-sifive: I'll need to make a change to my rocket-chip PR before submitting. Currently I don't differentiate between codes 3 and 4

PR #1351, opened 2018-04-15
Add simulation with Icarus Veirlog

jimmysitu:
Add simulation with Icarus Verilog, which is a free open source verilog simulator


ivlsrc/AsyncResetReg.v
```diff
@@ -0,0 +1,83 @@
+// See LICENSE.SiFive for license details.
+
```
mwachs5: why do you want separate VSRC for `ivlsim`? This looks the same as the other `vsrc` files

ivlsrc/ClockDivider3.v
```diff
@@ -0,0 +1,37 @@
+// See LICENSE.SiFive for license details.
+
```
mwachs5: In general I don't think you should copy these files, you should fix your makefile to get them from the right place.

ivlsim/Makefrag-iverilog
```diff
@@ -0,0 +1,174 @@
+#--------------------------------------------------------------------
+# Verilog Generation
+#--------------------------------------------------------------------
```
mwachs5: will this install ivlsim? Or can ivlsim be installed with a package manager? If you want to add this you should show it working within the Travis regression

mwachs5: (You can do this by editing the `.travis.yml` file in this repo)

seldridge:
(Not intending to be combative, I'm genuinely curious here... I looked at Icarus Verilog a long, long time ago, but have never used it with Rocket or any Chisel project.)

My understanding is that [Verilator is going to be 100x faster than Icarus Verilog in terms of simulation speed](https://www.veripool.org/wiki/verilator). Verilator currently produces a Rocket-Chip simulation executable that runs at ~10KHz. If that 100x is to be believed, then it may be a hard sell to include alternative, redundant simulation infrastructure that runs at 100Hz.

Have you done any benchmarking here? Is there an auxiliary benefit of Icarus Verilog over Verilator?

ivlsrc/ClockDivider3.v
jimmysitu: OK, I will fix it

ivlsrc/AsyncResetReg.v
jimmysitu: Hi, mwachs5
some files in vsrc is systemverilog, and using DPI, which iverilog not support yet. So I copy these files (together with cpp files in csrc) to ivlsrc, and rewrite the VPI instead. 
ivlsrc/AsyncResetReg.v is the same in vsrc, I will fix this in the Makefile

ivlsrc/SimDTM.cc
```diff
@@ -0,0 +1,113 @@
+// See LICENSE.SiFive for license details.
```
mwachs5: I see you fixed the duplication of VSRC files, but same goes for CSRC...you should not need to add copies of these files. If there are minor differences required you can use `#ifdef IVLSIM` or such to handle those.

ivlsim/Makefrag-iverilog
jimmysitu: Hi, mwachs5

I am newbie of Travis, I try the original .travis.yml on my travis
https://travis-ci.org/jimmysitu/rocket-chip/builds/367613403
It seems I cannot complete **make tools** in 50mins and Travis timeout.( job #1.1)
Then I check the travis log of the PR, this step done less than 3mins, I think that the travis cache helps.
My question is if there is any way to make my rocket-chip .travis.yml works without timeout?
Thanks.

jimmysitu:
@seldridge
Yes, I do the benchmark on a i5-8600K with 4.1GHz Tubro, 16G DDR4, 3200MHz
which is a 6core machine, the benchmark command is
```bash
time make -j6 run-asm-tests
```
The result is, Verilator: 42 mins, iverilog: 2242mins
Which means Verilator is 50x faster then iverilog.
No much surprise, but I think standard verilog simulation is need by rocket-chip.
Since not everyone has VCS installed, iverilog is a common choice
(Another choice is [oss-cvc](http://www.tachyon-da.com/) , a much faster open source simulator, but I give up since it fails with internal error while compiling rocket-chip's verilog code)

I think the benefit of using Icarus Verilog is you can use it as a 4-state simulator and cross check the result with verilator's.

ccelio:
37 hours for a 5-stage, in-order core is a total show stopper. If this code will never be regression tested, it will never be maintained, and it will break. 

Which is unfortunate, I'd love to see more demos of adding different backends. And if nothing else, it could serve as a performance torture tester for Icarus devs to bring the speed more inline with VCS. =(

jimmysitu:
Hi, All
I found it is impossible for iverilog to fininsh all regression test before travis timeout, so maybe it is not a good idea to merge this pull request to master.
I closed this PR and leave this porting in ivlsim in my fork.

ccelio:
Thanks for trying this out, sorry it didn't work for now. :(

PR #1358, opened 2018-04-18
Annotations: add some informative annotations

hcook:
These annotations capture interesting information about the design, mostly for documentation purposes.

### DIFF
--- a/src/main/scala/util/Annotations.scala
+++ b/src/main/scala/util/Annotations.scala
@@ -59,6 +59,15 @@ case class SlaveAddressMapChiselAnnotation(
     label = "slaves")
 }
 
+/** Record information about a top-level port of the design */
+case class TopLevelPortAnnotation(
+    target: ComponentName,
+    protocol: String,
+    tags: Seq[String],
+    mapping: Seq[AddressMapEntry]) extends SingleTargetAnnotation[ComponentName] {
+  def duplicate(n: ComponentName) = this.copy(n)
+}
+
 /** Helper object containing methods for applying annotations to targets */
 object annotated {
   def params[T <: Product](component: InstanceId, params: T): T = {
@@ -76,6 +85,14 @@ object annotated {
     mapping
   }
 
+  def port[T <: Data](
+      data: T,
+      protocol: String,
+      tags: Seq[String],
+      mapping: Seq[AddressMapEntry]): T = {
+    annotate(new ChiselAnnotation { def toFirrtl = TopLevelPortAnnotation(data.toNamed, protocol, tags, mapping) })
+    data
+  }
 }
 
 /** Mix this into a Module class or instance to mark its ports as untouchable */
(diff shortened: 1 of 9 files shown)


src/main/scala/util/Annotations.scala
```diff
@@ -3,12 +3,98 @@
...
+    target: ComponentName,
+    protocol: String,
+    tags: Seq[String],
+    mapping: Seq[AddressMapEntry]) extends SingleTargetAnnotation[ComponentName] {
+  def duplicate(n: ComponentName) = this.copy(n)
+}
+
+/** Helper object containing methods for applying annotations to targets */
+object annotated {
+  def params[T <: Product](component: InstanceId, params: T): T = {
```
jackkoenig: Can `Data` be annotated with `params` or just `Modules`? If it's only `Modules`, I would change this from `InstanceId` to `RawModule`

src/main/scala/util/Annotations.scala
```diff
@@ -3,12 +3,98 @@
...
 
-// TODO: replace this with an implicit class when @chisel unprotects dontTouchPorts
-trait DontTouch { self: RawModule =>
+/** Record a case class that was used to parameterize this target. */
+case class ParamsAnnotation(target: Named, paramsClassName: String, params: Map[String,Any]) extends SingleTargetAnnotation[Named] {
+  def duplicate(n: Named) = this.copy(n)
+}
+
+case class ParamsChiselAnnotation[T <: Product](target: InstanceId, params: T) extends ChiselAnnotation {
+  private val paramMap = params.getClass.getDeclaredFields.map(_.getName).zip(params.productIterator.to).toMap
```
jackkoenig: As far as I know, there isn't a better way to type bound case classes than by using `Product`, but note that lots of things implement `Product`, including `List`:
```scala
scala> val xs = List(1, 2, 3)
xs: List[Int] = List(1, 2, 3)

scala> xs.getClass.getDeclaredFields.map(_.getName).zip(xs.productIterator.to).toMap
res1: scala.collection.immutable.Map[String,Any] = Map(serialVersionUID -> 1, head -> List(2, 3))
```
The even more hilarious thing about the `List` above, is that 1 isn't actually the `serialVersionUID` of the `List`, this trick for case classes corresponds the names and values incorrectly for `List`.

hcook: Yeesh, I see, that is an annoying hole. Do you know any way to narrow down the scope further, e.g. I know cases classes also extend `Serializable`? Do Lists?

src/main/scala/util/Annotations.scala
hcook: It can also be applied to `Data` IMO. However, when I did so in practice (my example was adding the annotation to `HasBtbParameters` instead of `BTBModule`) I added a looooooooot of annotations. I'm not sure what the performance implications are of having 1000s of them, so I was more modest in practice so far.

jackkoenig: That's fair enough. I'm sure we'll hit some performance issues with toooooons of annotations so we should take a look at that before we have too much blow up.

src/main/scala/util/Annotations.scala
jackkoenig: Yeah Lists also extend `Serializable`, I'm not sure if there's much more we can do here.

PR #1361, opened 2018-04-19
Debug: work around param case class serialization limitation

hcook:
The functional parameters in `DebugModuleParams` render it unserializable as a Firrtl annotation. As a workaround I split them out into their own case class and key. I'm not sure if this is the right long term solution but it at least unblocks @rmac-sifive.

### DIFF
--- a/src/main/scala/devices/debug/SBA.scala
+++ b/src/main/scala/devices/debug/SBA.scala
@@ -208,7 +208,7 @@ object SystemBusAccessModule
     SBCSRdData.sbaccess16  := (cfg.maxSupportedSBAccess >=  16).B
     SBCSRdData.sbaccess8   := (cfg.maxSupportedSBAccess >=   8).B
     SBCSRdData.sbbusy      := sbBusy
-    SBCSRdData.sberror     := sbErrorReg.toBits
+    SBCSRdData.sberror     := sbErrorReg.asUInt
     
     cover(SBCSFieldsReg.sbbusyerror,    "SBCS Cover", "sberror set")
     cover(SBCSFieldsReg.sbbusy === 3.U, "SBCS Cover", "sbbusyerror alignment error")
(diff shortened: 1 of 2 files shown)


mwachs5:
I think this information is probably extractable some other way anyway (e.g. if something is outputting the `$mhartid` of the harts in the system). In general whatever is assigning HartIDs should be able to make this mapping, and it's just passed to the Debug Module because it needs to know it to properly interpret the execution of the Debug ROM code. There is currently no good way to to enforce this mapping anyway.

src/main/scala/devices/debug/SBA.scala
```diff
@@ -208,7 +208,7 @@ object SystemBusAccessModule
     SBCSRdData.sbaccess16  := (cfg.maxSupportedSBAccess >=  16).B
     SBCSRdData.sbaccess8   := (cfg.maxSupportedSBAccess >=   8).B
     SBCSRdData.sbbusy      := sbBusy
-    SBCSRdData.sberror     := sbErrorReg.toBits
+    SBCSRdData.sberror     := sbErrorReg.asUInt
```
rmac-sifive: Heh, thanks. Good to know this is deprecated

mwachs5:
It's because the serializer doesn't yet know how to turn "arbitrary function" into text it seems. Since these parameters are actually arbitrary functions, it doesn't know what to do with them.

hcook:
Nothing is fundamentally unserializable, but the question is "what can the JSON serializer that has been built in to Chisel/Firrtl handle without knowing anything a priori about the classes it is trying to serialize?"

The answer for now is: it can handle anything composed of built-in Scala types, where "composition" means: included as a member in a case class, or an element in a collection (nested case classes and collections are also allowed).

What is not allowed: user-defined/named Scala classes (that aren't case classes and don't inherit from `Serializable`), or anonymous classes created by the Scala compiler. Specifically the problem with the debug params was those two members that are functions from `UInt => UInt`; handling that member involves created an anonymous class that inherits from `Function1` or somesuch.

It a cool feature of Scala that you can have parameters that are functions, and it is definitely very useful in some cases, but it is hard to handle input from/output to non-Scala sources thats deals with such parameters in a meaningful way. While it's certainly awkward to have to manually segregate parameters-that-are-built-in-types from parameters-that-are-functions, it is also a fundamentally hard problem that I think must be explicitly skirted.

PR #1362, opened 2018-04-19
Misc. utility improvements and generalizations

aswaterman:
- Improve reusability of CSR unit and D$ for non-Rocket pipelines
- Several additions to Seq/UInt PML API

### DIFF
--- a/src/main/scala/util/package.scala
+++ b/src/main/scala/util/package.scala
@@ -141,6 +141,14 @@ package object util {
     def option[T](z: => T): Option[T] = if (x) Some(z) else None
   }
 
+  implicit class IntToAugmentedInt(val x: Int) extends AnyVal {
+    // exact log2
+    def log2: Int = {
+      require(isPow2(x))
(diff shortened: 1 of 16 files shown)


src/main/scala/util/package.scala
```diff
@@ -33,6 +33,29 @@ package object util {
...
+
+    def rotate(n: UInt): Seq[T] = {
+      require(isPow2(x.size))
+      val amt = n.padTo(log2Ceil(x.size))
+      (x /: (0 until log2Ceil(x.size)))((r, i) => (r.rotate(1 << i) zip r).map { case (s, a) => Mux(amt(i), s, a) })
+    }
+  }
+
+  // allow bitwise ops on Seq[Bool] just like UInt
+  implicit class SeqBoolBitwiseOps(val x: Seq[Bool]) extends AnyVal {
```
hcook: @mwachs5 I feel like you were just asking for this. You should get automagically if you've imported `rocketchip.util`

PR #1367, opened 2018-04-20
Make CLINT optional

solomatnikov:


src/main/scala/devices/tilelink/CLINT.scala
```diff
@@ -94,6 +94,9 @@ class CLINT(params: CLINTParams, beatBytes: Int)(implicit p: Parameters) extends
 
 /** Trait that will connect a CLINT to a subsystem */
 trait HasPeripheryCLINT { this: BaseSubsystem =>
```
mwachs5: should the trait be renamed to "CanHave..." ? @hcook mentioned the naming convention here

src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -124,7 +124,9 @@ trait HasRocketTiles extends HasTiles
 
     // 2. clint+plic conditionally crossing
     val periphIntNode = rocket.intInwardNode :=* rocket.crossIntIn
-    periphIntNode := clint.intnode                   // msip+mtip
+    p(CLINTKey).map { params =>
+      periphIntNode := clintOpt.get.intnode          // msip+mtip
```
mwachs5: doesn't this not work if you don't have a clint? I seem to recall that these interrupts are somewhat hard-coded in how they are hooked up. Shouldn't you tie `msip` and `mtip` to 0 if clintOpt is None?

mwachs5: See this comment: https://github.com/freechipsproject/rocket-chip/blob/20a88768560ca810c0dbe79f1a33604268d921c7/src/main/scala/tile/Interrupts.scala#L51

src/main/scala/devices/tilelink/CLINT.scala
hcook: Yes, that would be my proposal. @solomatnikov suggested something like `CouldGenerate...` which might be even clearer and help to distinguish the `LazyModule` vs `ModuleImp` distinction?

src/main/scala/subsystem/RocketSubsystem.scala
hcook: The option unpacking can also be streamlined:
`clintOpt.foreach { periphIntNode := _.intnode }`

I share @mwachs5's  concern, so maybe something like:
`periphIntNode := clintOpt.map(_.intnode).getOrElse(UInt(0))`

mwachs5: I don't think you can make an IntNode out of UInt(0). You need to make some sort of interrupt source that has 2 dummy MTIP and MSIP outputs.

src/main/scala/devices/tilelink/CLINT.scala
solomatnikov: I'll change to `CanHave...` to conform with existing pattern.

`CouldGenerate...` can be done in a separate PR.

src/main/scala/subsystem/RocketSubsystem.scala
mwachs5: Oh also I like @hcook 's `foreach` version of this better. Much more readable.

solomatnikov: Would `foreach` version work correctly if there is more than 1 CLINT (by mistake)?

mwachs5: the type of Clint is Option, not Seq, right?

solomatnikov: Yes

mwachs5: so then it does the right thing

PR #1368, opened 2018-04-20
PLIC optimizations

terpstra:
This PR attempts to reduce the PLIC line count and factor out common logic where possible.

### DIFF
--- a/src/main/scala/devices/tilelink/Plic.scala
+++ b/src/main/scala/devices/tilelink/Plic.scala
@@ -135,7 +135,7 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
     require(nHarts > 0 && nHarts <= PLICConsts.maxHarts, s"Must be: nHarts=$nHarts > 0 && nHarts <= PLICConsts.maxHarts=${PLICConsts.maxHarts}")
 
     // For now, use LevelGateways for all TL2 interrupts
-    val gateways = (false.B +: interrupts).map { case i =>
+    val gateways = interrupts.map { case i =>
       val gateway = Module(new LevelGateway)
       gateway.io.interrupt := i
       gateway.io.plic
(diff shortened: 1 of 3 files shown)


src/main/scala/util/MuxLiteral.scala
```diff
@@ -0,0 +1,58 @@
...
+    apply(index, default, first :: rest.toList)
+  def apply[T <: Data:ClassTag](index: UInt, default: T, cases: Seq[T]): T =
+    MuxTable(index, default, cases.zipWithIndex.map { case (v, i) => (BigInt(i), v) })
+}
+
+object MuxTable
+{
+  def apply[T <: Data:ClassTag](index: UInt, default: T, first: (BigInt, T), rest: (BigInt, T)*): T =
+    apply(index, default, first :: rest.toList)
+  def apply[T <: Data:ClassTag](index: UInt, default: T, cases: Seq[(BigInt, T)]): T = {
```
hcook: While `ClassTag` is kosher, chisel gets around using it by doing ```
class Vec[T <: Data] private[core] (gen: => T, val length: Int) { 
  ...
  private val self: Seq[T] = Vector.fill(length)(gen)
  ...
}
```

**Comment by mikeyangsiv** [comment]
FYI, I had to comment some stuff out from fadu in order to compile federation.jar

```[error] /scratch/mikeyang/federation.gamma02/src/main/scala/everywhere/coreip/fadu/Configs.scala:137:10: not found: value DebugModuleHartSelKey
[error]     case DebugModuleHartSelKey => DebugModuleHartSelFuncs(
[error]          ^
[error] /scratch/mikeyang/federation.gamma02/src/main/scala/everywhere/coreip/fadu/Configs.scala:137:35: not found: value DebugModuleHartSelFuncs
[error]     case DebugModuleHartSelKey => DebugModuleHartSelFuncs(
[error]                                   ^
[error] /scratch/mikeyang/federation.gamma02/src/main/scala/everywhere/coreip/fadu/Configs.scala:138:7: not found: value hartIdToHartSel
[error]       hartIdToHartSel = { x => x - 2.U },
[error]       ^
[error] /scratch/mikeyang/federation.gamma02/src/main/scala/everywhere/coreip/fadu/Configs.scala:139:7: not found: value hartSelToHartId
[error]       hartSelToHartId = { x => x + 2.U })
[error]       ^
[error] four errors found
```

terpstra:
So far, this PR reduces the PLIC in one of our designs from 10366906 bytes to 1479359 bytes, a 7-fold improvement and 118020 lines to 23707 lines, a 5-fold improvement. The output code now also has very readable enable logic on all the writable registers.

src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -135,71 +136,82 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
     require(nHarts > 0 && nHarts <= PLICConsts.maxHarts, s"Must be: nHarts=$nHarts > 0 && nHarts <= PLICConsts.maxHarts=${PLICConsts.maxHarts}")
 
     // For now, use LevelGateways for all TL2 interrupts
-    val gateways = Vec((false.B +: interrupts).map { case i =>
+    val gateways = interrupts.map { case i =>
```
mwachs5: how are you able to get rid of the 0th gateway?

src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -135,71 +136,82 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
...
     }
 
-    def priorityRegDesc(i: Int) = if (i > 0) {
-      RegFieldDesc(s"priority_$i", s"Acting priority of interrupt source $i",
-        reset=if (nPriorities > 0) None else Some(1),
-        wrType=Some(RegFieldWrType.MODIFY))
-    } else {
-      RegFieldDesc.reserved
+    def priorityRegDesc(i: Int) =
+      RegFieldDesc(
```
mwachs5: what happend to the 0th case?

src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -135,71 +136,82 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
...
+        name      = s"priority_$i",
+        desc      = s"Acting priority of interrupt source $i",
+        group     = Some("priority"),
+        groupDesc = Some("Acting priorities of each interrupt source."),
+        reset     = if (nPriorities > 0) None else Some(1))
```
mwachs5: not sure if this is handled below, but to be clear you could say accessType = R or RW accordingly

src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -135,71 +136,82 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
...
+
+    def pendingRegDesc(i: Int) =
+      RegFieldDesc(
+        name      = s"pending_$i",
+        desc      = s"Set to 1 if interrupt source $i is pending, regardless of its enable or priority setting.",
+        group     = Some("pending"),
+        groupDesc = Some("Pending Bit Array. 1 Bit for each interrupt source."),
+        volatile = true)
```
mwachs5: similar comment as above, you could say accessType=R but I think that is handled when you make the RegField

src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -135,71 +136,82 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
...
-      priority.zipWithIndex.map{case (p, i) => priorityRegField(p, i)}))
-    val pendingRegFields = Seq(PLICConsts.pendingBase  -> RegFieldGroup("pending", Some("Pending Bit Array. 1 Bit for each interrupt source."),
-      pending.zipWithIndex.map{case (b, i) => RegField.r(1, b, pendingRegDesc(i))}))
 
- 
+    def priorityRegField(x: UInt, i: Int) =
+      if (nPriorities > 0) {
+        RegField(prioBits, x, priorityRegDesc(i))
+      } else {
+        RegField.r(prioBits, x, priorityRegDesc(i))
```
mwachs5: oh yeah you did this here

src/main/scala/devices/tilelink/Plic.scala
terpstra: The 0th RegField is gone from the lists, replaced by explicit RegField(x)s

src/main/scala/devices/tilelink/Plic.scala
terpstra: The 0th gateway was always a no-op. The 0 interrupt never fires.

PR #1369, opened 2018-04-21
Annotate register field descriptions

depappas-sifive:
This PR adds remapper annotations to the regmap function. The RegFields are serialized into JSON and Firrtl annotations are created. A Firrtl transform then is used to write out JSON files which are equivalent to the current regmapper JSON files which are generated by ElborationArtefacts.


src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,196 @@
...
+      }.getOrElse("None"),
+      group = desc.map {
+        _.group.getOrElse("None")
+      }.getOrElse("None"),
+      groupDesc = desc.map {
+        _.groupDesc.getOrElse("None")
+      }.getOrElse("None"),
+      accessType = desc.map {
+        _.access.toString
+      }.getOrElse("r"), // TODO default?
```
mwachs5: It should be “unknown” or something. Please don’t just make things up that are wrong, this will only confuse downstream tools

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,196 @@
...
+      wrType = desc.map(_.wrType.toString).getOrElse("None"),
+      rdAction = desc.map(_.rdAction.toString).getOrElse("None"),
+      volatile = desc.map(_.volatile).getOrElse(false),
+      hasReset = desc.map { d => if (d.reset != None) true else false }.getOrElse(false), // TODO ugly
```
mwachs5: Isn’t if (x) true else false the same as “x”?

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,196 @@
...
+      resetValue = BigInt(0), // TODO desc.map{_.reset}.getOrElse(BigInt(0))
```
mwachs5: The TODO side of this looks fine

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,196 @@
...
+      enumerations = map
```
mwachs5: Not sure what the enumerations does

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,196 @@
...
+    width: Int,
+    byteOffset: Int,
+    bitOffset: Int,
+    regField: RegField): RegFieldSer = {
+
+    val anonRegFieldName = s"unnamedRegField${byteOffset.toHexString}_${bitOffset}"
+    val regFieldName = regField.desc.map {
+      _.name
+    }.getOrElse("")
+    val selectedRegFieldName = if (regFieldName.isEmpty /* selectedName.isEmpty */) anonRegFieldName else regFieldName
```
mwachs5: Seems like these 3 lines can be combined into one getOrElsd

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Replaced "r" with "UNKNOWN"

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Fixed.

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Fixed. But the _reset is an option so it needed a getOrElse

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Enumerations is a placeholder for future enumerations

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Fixed. Intellij formatting run amok...

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,186 @@
...
+    moduleName: String,
+    baseAddress: BigInt,
+    width: Int,
+    byteOffset: Int,
+    bitOffset: Int,
+    regField: RegField): RegFieldSer = {
+
+    val anonRegFieldName = s"unnamedRegField${byteOffset.toHexString}_${bitOffset}"
+    val regFieldName = regField.desc.map {_.name}.getOrElse("")
+    val selectedRegFieldName = if (regFieldName.isEmpty /* selectedName.isEmpty */) anonRegFieldName else regFieldName
```
hcook: If you implement my change above you can delete this and just assign `name = regFieldName` below.

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,186 @@
...
+    RegFieldSer(
+      moduleName, //selectedName,
+      regFieldDescSer
+     )
+  }
+
+
+  def anno(
+    rawModule: RawModule,
+    moduleName: String,
```
hcook: Why are you passing in moduleName when you assign it below?

src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -81,19 +85,52 @@ case class TLRegisterNode(
...
-    val json = RegMappingAnnotation.serialize(base, name, mapping:_*)
+    val json = GenRegDescJson.serialize(base, name, mapping:_*)
     var suffix = 0
-    while( ElaborationArtefacts.contains(s"${baseHex}.${suffix}.regmap.json")){
+    while( ElaborationArtefacts.contains(s"${baseHex}.${suffix}.regmap.json")) {
       suffix = suffix + 1
     }
     ElaborationArtefacts.add(s"${baseHex}.${suffix}.regmap.json", json)
+
+    /* GetMeMyModule is a hack to get the current module during elaboration.
```
hcook: I think this comment should go in the `GetMyModule` file itself.

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,186 @@
...
+    instanceCounter: Int,
```
hcook: Why not call `getInstanceCount(module.name, base)` inside this function instead of calling it outside and passing it in?

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,186 @@
...
+    baseAddress: BigInt,
+    mapping: RegField.Map*): Seq[RegField.Map] = {
+
+    val moduleName = rawModule.name
+    val baseHex = s"0x${baseAddress.toInt.toHexString}"
+    val displayName = s"${moduleName}.${baseHex}"
+
+    println(s"INFO: GenRegDescsAnno: annotating rawModule: ${moduleName}.${baseHex}")
```
hcook: I'm not convinced this needs to be output on the console.

src/main/scala/util/GeneratorUtils.scala
```diff
@@ -147,3 +153,24 @@ object ElaborationArtefacts {
     files.foldLeft(false)((t, s) => {s._1 == extension | t})
   }
 }
+
+
+object GenRegDescJson {
```
hcook: This doesn't really have anything to do with `Generators` i.e. Scala applications that execute Chisel programs.

I suggest moving it to `util/Annotations.scala`

hcook:
Please provide a short description of what this PR is doing, for public consumption.

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,188 @@
...
+  * Chisel Annotation
+  *
+  * This creates the firrtl annotation
+  *
+  * @param target
+  * @param regMappingSer
+  */
+case class RegMappingChiselAnnotation(
+  target: RawModule,
+  regMappingSer: RegistersSer) extends ChiselAnnotation with RunFirrtlTransform {
```
hcook: I think you can get rid of `RegistersSer` and just use its arguments directly in this Annotation class:
```
case class RegMappingChiselAnnotation(
   target: ModuleName,
   displayName: String,
  instanceCounter: Int,
  baseAddress: BigInt,
  regFields: Seq[RegFieldSer]) extends ...
```

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,188 @@
...
+  hasReset: Boolean = false,
+  enumerations: Map[BigInt, (String, String)] = Map()
+)
+
+case class RegFieldSer(
+  regFieldName: String,
+  desc: RegFieldDescSer
+)
+
+case class RegistersSer(
```
hcook: I don't think you need this extra level of case class, see below

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,188 @@
...
+    state.annotations.foreach {
+      case RegFieldDescMappingAnnotation(t, desc) => println(desc)
     }
+    state
+  }
+}
+
+object GenRegDescsAnno {
+  // Each class which has regmaps has instance counters
+  private var instanceCounters = scala.collection.mutable.Map[String, Int]()
```
hcook: I understand why you need this logic to match the behavior of `ElaborationAtrefacts`, but could you handle this on the `Transform` side instead? It only matters when you actually try to output some files, right?

src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -81,19 +85,52 @@ case class TLRegisterNode(
...
+      module.name,
+      GenRegDescsAnno.getInstanceCount(module.name, base),
+      base,
+      mapping:_*)
+
   }
 }
 
+
+
```
hcook: Why more whitespace

src/main/scala/util/GeneratorUtils.scala
```diff
@@ -4,13 +4,19 @@ package freechips.rocketchip.util
...
+import org.json4s.jackson.JsonMethods.{pretty, render}
 // TODO: better job of Makefrag generation for non-RocketChip testing platforms
 import freechips.rocketchip.system.{TestGeneration, DefaultTestSuites}
 import freechips.rocketchip.config._
 import freechips.rocketchip.diplomacy.LazyModule
 import java.io.{File, FileWriter}
 import firrtl.annotations.JsonProtocol
+import org.json4s.JsonAST.JValue
+import org.json4s.JsonDSL._
+import org.json4s.jackson.JsonMethods.{pretty, render}
```
hcook: duplicated import?

src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -81,19 +85,52 @@ case class TLRegisterNode(
...
+  }
+
+
+  def genRegDescsJson(mapping: RegField.Map*) : Unit = {
     // Dump out the register map for documentation purposes.
     val base = address.head.base
     val baseHex = s"0x${base.toInt.toHexString}"
     val name = s"deviceAt${baseHex}" //TODO: It would be better to name this other than "Device at ...."
-    val json = RegMappingAnnotation.serialize(base, name, mapping:_*)
+    val json = GenRegDescJson.serialize(base, name, mapping:_*)
```
hcook: Not really sure why you need both `GenRegDescJson` and `GenRegDescsAnno` objects

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,188 @@
...
+
+    val regFieldDescSer = RegFieldDescSer(
+      byteOffset = byteOffsetHex,
+      bitOffset = bitOffset,
+      bitWidth = width,
+      name = selectedRegFieldName,
+      desc = desc.map {_.desc}.getOrElse("None"),
+      group = desc.map {_.group.getOrElse("None")}.getOrElse("None"),
+      groupDesc = desc.map {_.groupDesc.getOrElse("None")}.getOrElse("None"),
+      accessType = desc.map {_.access.toString}.getOrElse("UNKNOWN"),
```
mwachs5: I know I suggested the "UNKNOWN" wording, but why not just make this "None" to be consistent with all the others?

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,188 @@
...
+      wrType = desc.map(_.wrType.toString).getOrElse("None"),
+      rdAction = desc.map(_.rdAction.toString).getOrElse("None"),
+      volatile = desc.map(_.volatile).getOrElse(false),
+      hasReset = desc.map { d => d.reset != None }.getOrElse(false),
```
mwachs5: everywhere else you use `_` instead of `d =>`, could clean this one up for consistency

src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -81,19 +85,52 @@ case class TLRegisterNode(
     bundleIn.c.ready := Bool(true)
     bundleIn.e.ready := Bool(true)
 
+    genRegDescsJson(mapping: _*)
```
mwachs5: I think the convention is no space here `:_*`

src/main/scala/regmapper/Annotation.scala
depappas-sifive: fixed

depappas-sifive: fixed

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,186 @@
...
+  baseAddress: BigInt,
+  regFields: Seq[RegFieldSer]
+)
+
+/**
+  * Firrtl annotation
+  *
+  * @param target
+  * @param regMappingSer
+  */
```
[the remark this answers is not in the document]
depappas-sifive: Added more text. 

For the uninitiated, noting that the ChiselAnno is converted to a FirrtlAnno is an important point.

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,186 @@
...
+case class RegFieldDescMappingAnnotation(
+  target: ModuleName,
+  regMappingSer: RegistersSer) extends SingleTargetAnnotation[ModuleName] {
```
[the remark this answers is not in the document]
depappas-sifive: That could be done however, I prefer to keep the class hierarchy as it is.

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Good point. This is a remnant of the changes that the annos went through while we figured out how to annotate all modules. 

Fixed.

src/main/scala/tilelink/RegisterRouter.scala
depappas-sifive: This comment serves as a reminder to change this code to use the new def in Chisel and remove the hack. Comment updated

src/main/scala/regmapper/Annotation.scala
depappas-sifive: fixed

src/main/scala/tilelink/RegisterRouter.scala
depappas-sifive: fixed

src/main/scala/util/GeneratorUtils.scala
depappas-sifive: Optimized imports

src/main/scala/tilelink/RegisterRouter.scala
depappas-sifive: GenRegDescJson is used by the old regmap file generator and by the new regmap annotation which is why I made a separate object.

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Good point. Using None.

src/main/scala/regmapper/Annotation.scala
depappas-sifive: Fixed

src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -119,7 +156,9 @@ class TLRegModule[P, B <: TLRegBundleBase](val params: P, bundleBuilder: => B, r
   val io = IO(bundleBuilder)
   val interrupts = if (router.intnode.out.isEmpty) Vec(0, Bool()) else router.intnode.out(0)._1
   val address = router.address
-  def regmap(mapping: RegField.Map*) = router.node.regmap(mapping:_*)
+  def regmap(mapping: RegField.Map*) : Unit = {
```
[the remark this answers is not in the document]
depappas-sifive: reverted

The lint rules say that public methods need a return type.

  def regmap(mapping: RegField.Map*) = router.node.regmap(mapping:_*)

should be

  def regmap(mapping: RegField.Map*): Unit = router.node.regmap(mapping:_*)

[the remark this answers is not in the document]
depappas-sifive: Intellij enforces the lint rules at the TOT in the scalastyle.xml file.

When there is a lint rule violation Intellij shows an error. When there is another error in the same code section which has a lint violation it is not obvious that there is a syntax error.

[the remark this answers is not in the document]
depappas-sifive: I would like to clean up the rocket-chip lint rule violations using Intellij at some point...

src/main/scala/tilelink/RegisterRouter.scala
depappas-sifive: fixed

src/main/scala/tilelink/RegisterRouter.scala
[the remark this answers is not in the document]
mwachs5: Yep, wholeheartedly agree

src/main/scala/regmapper/Annotation.scala
depappas-sifive: The code annotates the ChiselAnnotation which is converted into a FirrtlAnnotation.

  annotate(RegMappingChiselAnnotation(rawModule, registersSer))

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,186 @@
...
-      seq.map(_.width).scanLeft(0)(_ + _).zip(seq).map { case (bit, f) =>
-        val anonName = s"unnamedRegField${byte.toHexString}_${bit}"
-        (f.desc.map{ _.name}.getOrElse(anonName)) -> f.toJson(byte, bit)
-      }
+import chisel3.core.annotate
+import chisel3.experimental.{ChiselAnnotation, RawModule, RunFirrtlTransform}
+import firrtl.annotations._
+import firrtl.{CircuitForm, CircuitState, LowForm, Transform}
+
+case class RegFieldDescAnnotation(
```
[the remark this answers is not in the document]
depappas-sifive: fixed

src/main/scala/regmapper/RegField.scala
```diff
@@ -48,6 +48,7 @@ case class RegFieldDesc (
   reset: Option[BigInt] = None,
   enumerations: Map[BigInt, (String, String)] = Map()
 ){
+  def toMap = this.getClass.getDeclaredFields.map(_.getName).zip(this.productIterator.to).toMap
```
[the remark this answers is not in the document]
depappas-sifive: removed

src/main/scala/regmapper/Annotation.scala
depappas-sifive: I am going to leave this for now.

src/main/scala/util/Annotations.scala
```diff
@@ -39,3 +42,109 @@ case class RetimeModuleAnnotation(target: ModuleName) extends SingleTargetAnnota
...
+    moduleName: String,
+    baseAddress: BigInt,
+    width: Int,
+    byteOffset: Int,
+    bitOffset: Int,
+    regField: RegField): RegFieldSer = {
+
+    val anonRegFieldName = s"unnamedRegField${byteOffset.toHexString}_${bitOffset}"
+    val regFieldName = regField.desc.map {_.name}.getOrElse("")
+    val selectedRegFieldName = if (regFieldName.isEmpty /* selectedName.isEmpty */) anonRegFieldName else regFieldName
```
hcook: This can just be:
`val selectedRegFieldName = regField.desc.map(_.name).getOrElse(anonRegFieldName)`

src/main/scala/util/GeneratorUtils.scala
```diff
@@ -3,14 +3,16 @@
...
-import freechips.rocketchip.system.{TestGeneration, DefaultTestSuites}
-import freechips.rocketchip.config._
-import freechips.rocketchip.diplomacy.LazyModule
 import java.io.{File, FileWriter}
+
 import firrtl.annotations.JsonProtocol
+import freechips.rocketchip.config._
+import freechips.rocketchip.system.{DefaultTestSuites, TestGeneration}
+
+
```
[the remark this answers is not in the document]
depappas-sifive: Removed a couple of newlines

hcook:
i fixed a merge conflict with master on util/Annotations.scala imports

src/main/scala/regmapper/Annotation.scala
depappas-sifive: done

src/main/scala/regmapper/Annotation.scala
```diff
@@ -2,22 +2,186 @@
...
+  */
+case class RegMappingChiselAnnotation(
+  target: RawModule,
+  regMappingSer: RegistersSer) extends ChiselAnnotation with RunFirrtlTransform {
+  def toFirrtl: RegFieldDescMappingAnnotation = RegFieldDescMappingAnnotation(target.toNamed, regMappingSer)
+
+  def transformClass: Class[RegMappingDumpTransform] = classOf[RegMappingDumpTransform]
+}
+
+class RegMappingDumpTransform extends Transform {
```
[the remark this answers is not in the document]
depappas-sifive: deleted

PR #1375, opened 2018-04-24
Bump riscv-tools

timsifive:


timsifive:
Not quite a master of riscv-tools. It's pointing at https://github.com/riscv/riscv-tools/pull/196 before master was merged in. (There was a change that went into riscv-tools while my bump change was building.)

mwachs5:
I don't want to merge a PR to rocket-chip that isn't pointing to *some* master of riscv-tools. Can you re-bump to some version that includes the changes you want? I'll handle approving and merging once it goes through the regression again

timsifive:
I'll start my script again. I think it's close to me being able to just run it and have all the right git stuff happen.

PR #1376, opened 2018-04-25
Fix wrong include path in emulator/Makefile

jimmysitu:
Fix wrong include path in emulator/Makefile, as the csrc dir is move to src/main/resources/csrc

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -35,8 +35,8 @@ $(FIRRTL_JAR): $(shell find $(base_dir)/firrtl/src/main/scala -iname "*.scala")
 
 src_path := src/main/scala
 resources := $(base_dir)/src/main/resources
-csrc := $(resources)/csrc/
-vsrc := $(resources)/vsrc/
+csrc := $(resources)/csrc
+vsrc := $(resources)/vsrc
 default_submodules := . hardfloat chisel3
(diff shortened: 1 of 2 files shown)


hcook:
@grebe ?

grebe:
Hmm, I guess I missed that. It still built the emulator on my system (and travis, I guess), so perhaps the whole `-I` is not needed.

@jimmysitu, did you find this by running into a problem building the emulator or by reading through the source? If you didn't have any problem with the emulator, I'd suggest entirely removing the `-I`.

jimmysitu:
@grebe , I find this just by reading the Makefile, and it still build the emulator. I tried to remove the entriely -I, it works!

Makefrag
```diff
@@ -35,8 +35,8 @@ $(FIRRTL_JAR): $(shell find $(base_dir)/firrtl/src/main/scala -iname "*.scala")
 
 src_path := src/main/scala
 resources := $(base_dir)/src/main/resources
-csrc := $(resources)/csrc/
-vsrc := $(resources)/vsrc/
+csrc := $(resources)/csrc
```
grebe: My personal preference is to leave trailing slashes so even slashes are forgotten later, you're still safe. It's not wrong to remove them, though. Is there a good reason to remove them?

jimmysitu: I check most variables in Makefiles of the project, which have no tailing slashes. So I think it is better all variable follow the same rule.

PR #1383, opened 2018-04-28
Elborated config from annos rocket

depappas-sifive:
Fixing the elaborated json file output and adding an interrupt port placeholder

### DIFF
--- a/src/main/scala/util/Annotations.scala
+++ b/src/main/scala/util/Annotations.scala
@@ -15,13 +15,11 @@ import org.json4s.JsonDSL._
 import org.json4s.jackson.JsonMethods.{pretty, render}
 
 /** Record a set of interrupts. */
-case class InterruptsPortAnnotation(target: Named, nInterrupts: Seq[Int])
-  extends SingleTargetAnnotation[Named] {
+case class InterruptsPortAnnotation(target: Named, nInterrupts: Seq[Int]) extends SingleTargetAnnotation[Named] {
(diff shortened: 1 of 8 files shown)


src/main/scala/util/Annotations.scala
```diff
@@ -14,6 +14,10 @@ import freechips.rocketchip.tilelink.TLToAXI4IdMapEntry
 import org.json4s.JsonDSL._
 import org.json4s.jackson.JsonMethods.{pretty, render}
 
+/** Record a set of interrupts. */
+case class InterruptsPortAnnotation(target: Named, nInterrupts: Seq[Int]) extends SingleTargetAnnotation[Named] {
```
hcook: This should use `ComponentName`

src/main/scala/util/Annotations.scala
```diff
@@ -66,11 +70,27 @@ case class TopLevelPortAnnotation(
     protocol: String,
     tags: Seq[String],
     mapping: Seq[AddressMapEntry]) extends SingleTargetAnnotation[ComponentName] {
-  def duplicate(n: ComponentName) = this.copy(n)
+  def duplicate(n: ComponentName): TopLevelPortAnnotation = this.copy(n)
+}
+
+/** Record the resetVector. */
+case class ResetVectorAnnotation(target: Named, resetVec: BigInt) extends SingleTargetAnnotation[Named] {
```
hcook: This should use `ComponentName`

PR #1394, opened 2018-04-30
SEC: switch to systematic encoding

terpstra:
Also improve verilog output by using xorR.

### DIFF
--- a/src/main/scala/util/ECC.scala
+++ b/src/main/scala/util/ECC.scala
@@ -183,37 +183,6 @@ object ErrGen
   def apply(x: UInt, f: Int): UInt = x ^ apply(x.getWidth, f)
 }
 
-class SECDEDTest extends Module
-{
-  val code = new SECDEDCode
-  val k = 4
-  val n = code.width(k)
-
-  val io = new Bundle {
-    val original = Bits(OUTPUT, k)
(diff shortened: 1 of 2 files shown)


src/main/scala/util/ECC.scala
```diff
@@ -157,9 +157,9 @@ class SECDEDCode extends Code
   def width(k: Int) = sec.width(k)+1
   def encode(x: UInt, poison: Bool = Bool(false)) = {
     // toggling two bits ensures the error is uncorrectable
-    val toggle_lo = poison.asUInt // a non-data bit in SEC
-    val toggle_hi = toggle_lo << sec.width(x.getWidth) // the parity bit
-    par.encode(sec.encode(x)) ^ toggle_lo ^ toggle_hi
+    val toggle_lo = Cat(poison.asUInt, poison.asUInt)
```
aswaterman: Mind adding a comment to explain the rationale?

terpstra: The comment 'toggling two bits = ECC failure' is already there. The change was needed because when you poison, you want to toggle parity bits only, so that uncorrected is the same as corrected.

terpstra: The systematic SEC change moved the parity bits.

PR #1395, opened 2018-05-01
util: add explicitly reproducible Random

hcook:


src/main/scala/util/SeededRandom.scala
```diff
@@ -0,0 +1,12 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+/** Using e.g. Random.nextInt makes your hardware generation non-repoducible,
+  * which is almost certainly undesirable. Use LCG to increment random numbers in HW,
+  * or use this SeededRandom.fromSeed to make reproducible Scala PRNGs.
+  */
+object SeededRandom {
+  val fromCurrentTime = new scala.util.Random(System.currentTimeMillis())
```
aswaterman: Why provide this one if the goal is determinism?

PR #1415, opened 2018-05-10
Documenting emulator debugging process with GDB

noureddine-as:
This documents the process of debugging RISC-V programs running on the Emulator using the GNU debugger and OpenOCD in a similar manner as https://github.com/riscv/riscv-isa-sim#debugging-with-gdb .

https://github.com/freechipsproject/rocket-chip/issues/1339

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,11 +450,11 @@ Then you can build as usual with CONFIG=MyConfig.
 
 ## <a name="debug"></a> Debugging with GDB
 
-### 1) Generating the RBB emulator
+### 1) Generating the Remote Bit-Bang (RBB) Emulator
 
-The objective of this section is to use GNU debugger to debug programs RISC-V programs running on the emulator in the same fashion as in [Spike](https://github.com/riscv/riscv-isa-sim#debugging-with-gdb).
+The objective of this section is to use GNU debugger to debug RISC-V programs running on the emulator in the same fashion as in [Spike](https://github.com/riscv/riscv-isa-sim#debugging-with-gdb).
 
-For that we need to build a Remote Bit Bang enabled emulator, this could be any of the listed configurations in `rocket-chip/src/main/scala/system/Configs.scala`. Basically all we need is to extend our configuration with JtagDTMSystem that will deal with program loading and interface with the debugger. In the following example we added this extention to the DefaultConfig one.
+For that we need to add a Remote Bit-Bang client to the emulator. We can do so by extending our Config with JtagDTMSystem, which will add a DebugTransportModuleJTAG to the DUT and connect a SimJTAG module in the Test Harness. This will allow OpenOCD to interface with the emulator, and GDB can interface with OpenOCD. In the following example we added this Config extension to the DefaultConfig:
 
     class DefaultConfigRBB extends Config(
     new WithJtagDTMSystem ++ new WithNBigCores(1) ++ new BaseConfig)
@@ -462,12 +462,18 @@ For that we need to build a Remote Bit Bang enabled emulator, this could be any 
     class QuadCoreConfigRBB extends Config(
     new WithJtagDTMSystem ++ new WithNBigCores(4) ++ new BaseConfig)
 
-To build the configuration we use the command:
+To build the emulator with `DefaultConfigRBB` configuration we use the command:
 
     rocket-chip$ cd emulator
     emulator$ CONFIG=DefaultConfigRBB make
 
-By default the emulator is generated under the name `emulator-freechips.rocketchip.system-DefaultConfigRBB`.
+We can also build a debug version capable of generating VCD waveforms using the command:
+
+    emulator$ CONFIG=DefaultConfigRBB make debug
+
+By default the emulator is generated under the name `emulator-freechips.rocketchip.system-DefaultConfigRBB` in the first case and `emulator-freechips.rocketchip.system-DefaultConfigRBB-debug` in the second.
+
+Please note that generated VCD waveforms can be very voluminous depending on the size of the .elf file (i.e. code size + debugging symbols). If you need only to execute your program and get output, you may need to strip the .elf file. Make sure you keep `tohost` and `fromhost` because these are necessary in the HTIF interface.
 
 ### 2) Compiling the executable
 
@@ -512,9 +518,13 @@ We can then launch the emulator with
 	Listening on port 9823
 	Attempting to accept client socket
 
+You can also use the `emulator-freechips.rocketchip.system-DefaultConfigRBB-debug` version instead if you would like to generate VCD waveforms.
+
+Please note that if the argument `--rbb-port` is not passed, a default free TCP port on your computer will be chosen randomly. 
+
 ### 4) Launch OpenOCD
 
-We first need to ensure the OpenOCD has been generated. Generally it is in `$(RISCV)/bin/openocd`. We then define a configuration file that is going to define the RBB port we will use, which is in our case `9823`.
+You will need a RISC-V Enabled OpenOCD binary. This is installed with riscv-tools in `$(RISCV)/bin/openocd`, or can be compiled manually from riscv-openocd. OpenOCD requires a configuration file, in which we define the RBB port we will use, which is in our case `9823`.
 
     $ cat cemulator.cfg 
     interface remote_bitbang
@@ -534,7 +544,7 @@ We first need to ensure the OpenOCD has been generated. Generally it is in `$(RI
 
 Then we launch OpenOCD in another terminal using the command
     
-    $ openocd -f ./cemulator.cfg
+    $(RISCV)/bin/openocd -f ./cemulator.cfg
     Open On-Chip Debugger 0.10.0+dev-00112-g3c1c6e0 (2018-04-12-10:40)
     Licensed under GNU GPL v2
     For bug reports, read
@@ -555,10 +565,12 @@ Then we launch OpenOCD in another terminal using the command
     Info : Listening on port 3333 for gdb connections
     Info : Listening on port 6666 for tcl connections
     Info : Listening on port 4444 for telnet connections
+
+A `-d` flag can be added to the command to show further debug information.
     
 ### 5) Launch GDB
 
-Launch GDB in another terminal using
+In another terminal launch GDB and point to the elf file you would like to load then run it with the debugger (in this example, `helloworld`):
 
     $ riscv64-unknown-elf-gdb helloworld
     GNU gdb (GDB) 8.0.50.20170724-git
@@ -578,9 +590,9 @@ Launch GDB in another terminal using
     Reading symbols from ./proj1.out...done.
     (gdb)
 
-Compared to Spike, the C Emulator is very slow, so several problems are encountered due to timeouts between issuing commands and response from the emulator. Thats why we should increase the `remotetimeout`.
+Compared to Spike, the C Emulator is very slow, so several problems may be encountered due to timeouts between issuing commands and response from the emulator. To solve this problem, we increase the timeout with the GDB `set remotetimeout` command.
 
-After that we load our program by performing a `load` command.
+After that we load our program by performing a `load` command. This automatically sets the `$PC` to the `_start` symbol in our .elf file.
 
  	(gdb) set remotetimeout 2000
  	(gdb) target remote localhost:3333
@@ -600,10 +612,6 @@ After that we load our program by performing a `load` command.
 
 Now we can proceed as with Spike, debugging works in a similar way:
 
-	(gdb) target remote localhost:3333
-	Remote debugging using localhost:3333
-	0x000000008000104c in main (argc=0, argv=0x0) at src/main.c:16
-	16	    while (wait)
 	(gdb) print wait
 	$1 = 1
 	(gdb) print wait=0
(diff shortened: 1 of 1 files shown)


mwachs5:
Thanks! I took a quick glance over and it looks good, I'll take a closer look shortly.

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
 
 Then you can build as usual with CONFIG=MyConfig.
 
+## <a name="debug"></a> Debugging with GDB
+
+### 1) Generating the RBB emulator
```
mwachs5: Can you update this to:

Generating the Remote Bit-Bang (RBB) Emulator

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+
+The objective of this section is to use GNU debugger to debug programs RISC-V programs running on the emulator in the same fashion as in [Spike](https://github.com/riscv/riscv-isa-sim#debugging-with-gdb).
```
mwachs5: programs RISC-V programs (typo)

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+
+For that we need to build a Remote Bit Bang enabled emulator, this could be any of the listed configurations in `rocket-chip/src/main/scala/system/Configs.scala`. Basically all we need is to extend our configuration with JtagDTMSystem that will deal with program loading and interface with the debugger. In the following example we added this extention to the DefaultConfig one.
```
mwachs5: edited version:

For that we need to add a Remote Bit-Bang client to the emulator. We can do so by extending our `Config` with `JtagDTMSystem`, which will add a `DebugTransportModuleJTAG` to the DUT and connect a `SimJTAG` module in the Test Harness. This will allow OpenOCD to interface with the emulator, and GDB can interface with OpenOCD. In the following example we added this `Config` extension to the `DefaultConfig`:

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+        ;
+}
+```
+Do not forget to compile with `-g -Og` flags to provide debugging support just as explained [here](https://github.com/riscv/riscv-isa-sim#debugging-with-gdb).
+
+### 3) Launch the emulator
+
+We can then launch the emulator with
+
+    ./emulator-freechips.rocketchip.system-DefaultConfigRBB +jtag_rbb_enable=1 --rbb-port=9823 helloworld
```
mwachs5: There is no real point to the `helloworld` argument to the emulator, because you `load` it later through the debugger. What happens if you leave it off?

mwachs5: You should note that if you leave off the `--rbb-port` argument, a random port will be assigned.

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+	This emulator compiled with JTAG Remote Bitbang client. To enable, use +jtag_rbb_enable=1.
+	Listening on port 9823
+	Attempting to accept client socket
+
+### 4) Launch OpenOCD
+
+We first need to ensure the OpenOCD has been generated. Generally it is in `$(RISCV)/bin/openocd`. We then define a configuration file that is going to define the RBB port we will use, which is in our case `9823`.
```
mwachs5: Edited version:

You will need RISC-V Enabled OpenOCD binary. This is installed with `riscv-tools` in `$(RISCV)/bin/openocd`, or can be compiled manually from [riscv-openocd](https://github.com/riscv/riscv-openocd). OpenOCD requires a configuration file, in which we define the RBB port we will use, which is in our case `9823`.

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+    target create $_TARGETNAME riscv -chain-position $_TARGETNAME
+
+    gdb_report_data_abort enable
+
+    init
+    halt
+
+Then we launch OpenOCD in another terminal using the command
+    
+    $ openocd -f ./cemulator.cfg
```
mwachs5: please use the full path `$(RISCV)/bin/openocd`. This is a common mistake.

mwachs5: You can also suggest adding the `-d` flag for debug info

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+    Info : [0] Found 1 triggers
+    Info : Examined RISC-V core; found 1 harts
+    Info :  hart 0: XLEN=64, 1 triggers
+    Info : Listening on port 3333 for gdb connections
+    Info : Listening on port 6666 for tcl connections
+    Info : Listening on port 4444 for telnet connections
+    
+### 5) Launch GDB
+
+Launch GDB in another terminal using
```
mwachs5: Launch GDB in another terminal and point to the `elf` file you would like to load & run with the debugger (in this example, `helloworld`):

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+    For bug reporting instructions, please see:
+    <http://www.gnu.org/software/gdb/bugs/>.
+    Find the GDB manual and other documentation resources online at:
+    <http://www.gnu.org/software/gdb/documentation/>.
+    For help, type "help".
+    Type "apropos word" to search for commands related to "word"...
+    Reading symbols from ./proj1.out...done.
+    (gdb)
+
+Compared to Spike, the C Emulator is very slow, so several problems are encountered due to timeouts between issuing commands and response from the emulator. Thats why we should increase the `remotetimeout`.
```
mwachs5: Edited version:

Compared to Spike, the C Emulator is very slow, so several problems may be encountered due to timeouts between issuing commands and response from the emulator. To solve this problem, we increase the timeout with the GDB `remotetimeout` command.

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+
+After that we load our program by performing a `load` command.
```
mwachs5: You may want to note "This automatically sets the `$pc` to the `_start` symbol in our `.elf` file.

README.md
mwachs5: Also you could provide a link to the GDB documentation of this command.

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+ 	Loading section .rodata.str1.8, size 0x20 lma 0x80001b30
+ 	Loading section .data, size 0x22 lma 0x80001b50
+ 	Loading section .sdata, size 0x4 lma 0x80001b74
+ 	Start address 0x80000000, load size 3646
+ 	Transfer rate: 40 bytes/sec, 520 bytes/write.
+  	(gdb) 
+
+Now we can proceed as with Spike, debugging works in a similar way:
+
+	(gdb) target remote localhost:3333
```
mwachs5: you don't need to target remote localhost again, you did that step above.

mwachs5: Actually I'm not sure how you got from the previous step to this one. You're missing a "c" or "set breakpoint" or something.

README.md
```diff
@@ -447,6 +448,181 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+	^C
+	Program received signal SIGINT, Interrupt.
+	main (argc=0, argv=<optimized out>) at src/main.c:33
+	33	    while (!wait)
+	(gdb) print wait
+	$4 = 0
+	(gdb) print text
+	$5 = "Instruction sets want to be free!"
+	(gdb)
+
```
mwachs5: I think you should add a note either here or at the beginning that will note what to do to make it waveform-enabled (building the `-debug` emulator target)

noureddine-as:
Thanks for your reviews @mwachs5, I'll get back very soon to perform modifications.

README.md
noureddine-as: Actually without the program argument, the emulator simply shows the help message.

noureddine-as: Default RBB if no -rbb-port note added.

README.md
noureddine-as: Done

README.md
noureddine-as: Done

README.md
noureddine-as: Done

README.md
noureddine-as: Done

README.md
noureddine-as: Done.

README.md
noureddine-as: Done.

README.md
noureddine-as: Done + useful links added.

README.md
noureddine-as: Added.

README.md
noureddine-as: Yeah, I actually just copied that again, sorry. Adjusted now.

README.md
mwachs5: > Actually without the program argument, the emulator simply shows the help message.

Good to know. Maybe we should add a note here that it's not going to load your program, or that that argument doesn't matter, or use the term "dummy" ( I believe the latter is what the Debug Regressions actually do). Because it doesn't matter what you put there, since FESVR isn't attached it's not going to load that program.

README.md
noureddine-as: Added in 1).
+ Pointed out the fact that VCD waveforms can be voluminous depending on program size.
+ I'll add a section for executable stripping may be? I guess it affects the execution time, do you agree @mwachs5 ?.

mwachs5: Well, the program size and duration. I don't think it's really that relevant, people already know that VCD waveforms can be big and scale with runtime, so I woudl probably remove both those notes.

README.md
noureddine-as: Okay. I prefer to let the `helloworld` there so that new people don't get confused when they don't use RBB or something, but notified the fact that it is loaded from GDB (in contrast with Spike) and that's it's a dummy arg. Thanks for clarification.

README.md
noureddine-as: You're right experienced people could already know this. But when it comes to beginners discovering all these, it is not the case. An example is a friend who tried to boot BBL (that has now several MiB of size) and you think it just doesn't work, but you find out that it just takes hours to run. In these cases gaining more space means gaining hours of waiting.
Anyway, I didn't test it personally. Is it the case normally?

mwachs5: I mean, the larger program -> longer runtime is sort of orthogonal to longer runtime -> larger VCD file. So if you want to point out that it's going to be very slow to load a large program this way, I would separate it from the note that the generated waveforms with the `-debug` version of emulator may be large.

noureddine-as: I understand your point of view and I agree. 
As you said VCD size is trivial. Though, I pointed out the relation {larger program -> longer runtime} and how to reduce it as well as how to test the program on the emulator event after beginning debugging.
Again, I know these are basic ideas, but documenting can save new people some time to understand what's going on.
Thanks a lot.

noureddine-as: I guess it's okay now. Any other recommendations @mwachs5 ?

README.md
```diff
@@ -447,6 +448,217 @@ you can create your own Configuration(s) and compose them with Config's ++ opera
...
+First we can test if your program executes well in the simple version of emulator before moving to debugging in step 3 :
+
+	$ ./emulator-freechips.rocketchip.system-DefaultConfig helloworld 
+	
+Additional verbose information (clock cycle, pc, instruction being executed) can be printed using the following command:
+
+	$ ./emulator-freechips.rocketchip.system-DefaultConfig +verbose helloworld 2>&1 | spike-dasm 
+
+Please note that execution time of the program on the emulator depends on executable size. Stripping the .elf executable will unsurprisingly make it run faster. For this you can use `$RISCV/bin/riscv64-unknown-elf-strip` tool to reduce the size. This is good for accelerating execution but not for debugging. Keep in mind that the HTIF communication interface between our system and the emulator relies on `tohost` and `fromhost` symbols to communicate. This is why you may get the following error when you try to run a totally stripped executable on the emulator:
```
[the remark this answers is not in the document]
mwachs5: Also, minor nit, it's not really the execution time of the program, but the time it takes the emulator to **load** your program that depends on the size.  Minor rewording would be:

Please note that the time it takes the emulator to load your program depends on executable size. Stripping the .elf executable will unsurprisingly make it run faster. For this you can use `$RISCV/bin/riscv64-unknown-elf-strip` tool to reduce the size. This is good for accelerating your simulation but not for debugging. Keep in mind that the HTIF communication interface between our system and the emulator relies on `tohost` and `fromhost` symbols to communicate. This is why you may get the following error when you try to run a totally stripped executable on the emulator:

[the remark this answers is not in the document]
mwachs5: Actually, I re-read your justification for the two notes above, so if you want to keep the other note about VCD waveform size that's fine.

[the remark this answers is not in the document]
noureddine-as: Thanks Megan. I rearranged the idea in the 'Compiling and executing' part. I also deleted redundancies. Added also an example of using the -debug version.
I guess the text is now more coherent.
Thank you for the reviews. Glad to participate :)

noureddine-as:
Thank you Megan!

Best reagards.

mwachs5:
It was some sort of Travis timeout, perhaps the server was overloaded. I'll just restart the regression.

PR #1420, opened 2018-05-12
debug: assert on unexpected exceptions

mwachs5:
Harts should only be writing to the EXCEPTION location if doing work under the direction of the debugger. Otherwise this indicates some other sort of debug mode exception, which is unexpected since Debug ROM code is super benign.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -66,15 +66,22 @@ jobs:
       script:
         - travis_wait 120 make verilator -C regression SUITE=none
       before_install:
-        - export CXX=g++-4.8 CC=gcc-4.8    
+        - export CXX=g++-4.8 CC=gcc-4.8
     - &test
       stage: Test
+      script:
+        - travis_wait 80 make emulator-ndebug -C regression SUITE=UnittestSuite JVM_MEMORY=3G
+        - travis_wait 80 make emulator-regression-tests -C regression SUITE=UnittestSuite JVM_MEMORY=3G
+    - <<: *test
       script:
         - travis_wait 80 make emulator-ndebug -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-32 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
         - travis_wait 80 make emulator-jtag-dtm-tests-64 -C regression SUITE=JtagDtmSuite JVM_MEMORY=3G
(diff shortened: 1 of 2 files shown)


solomatnikov:
What is the other assertion for the exception?

https://github.com/freechipsproject/rocket-chip/blob/3075e10a0715ee5e96646b1eb45bb4d71a6da6e3/src/main/scala/devices/debug/Debug.scala#L1029

mwachs5:
Probably I should refactor these so there is not just one long-running test

.travis.yml
```diff
@@ -69,24 +69,25 @@ jobs:
         - export CXX=g++-4.8 CC=gcc-4.8    
     - &test
       stage: Test
+      script:
+        - travis_wait 80 make emulator-ndebug -C regression SUITE=UnittestSuite JVM_MEMORY=3G
+        - travis_wait 80 make emulator-regression-tests -C regression SUITE=UnittestSuite JVM_MEMORY=3G
```
richardxia: I have not tested this out, but I believe the way that YAML works, this will not do what you want it to do. I believe every place where you use `<<: *test`, when those define their own `script:` key, it will completely override these two lines rather than concatenate to it. I think the following Python pseudocode is equivalent to what you've done:

```python
test = {
  'stage': 'Test',
  'script': [ 'blah', 'blah2']
}

other_test_item = copy(test)
other_test_item['script'] = ['blah3', 'blah4', 'blah5]
```

You should perhaps test parsing this YAML in some language and seeing if its contents match what you want it to be.

mwachs5: Completely overriding the lines is the desired behavior... each "test/script" chunk is supposed to operate in parallel to the others.

Why the first one doesn't need a <<: *test is beyond my understanding of the Travis required syntax, I am just trying to move things around in the file.

mwachs5:
It appears that fixing my whitespace errors has allowed Travis to recognize this

.travis.yml
richardxia: > Completely overriding the lines is the desired behavior... each "test/script" chunk is supposed to operate in parallel to the others (including the first)

Oh, then the use of `&` and `<<` was confusing to me, since that's essentially YAML syntax for doing something similar to class inheritance.

> Why the first one doesn't need a <<: *test is beyond my understanding of the Travis required syntax, I am just trying to move things around in the file.

This is YAML syntax, not Travis, FYI. The first one can't use `<<: *test` because it *is* the "parent class". `&test` means "define a label 'test' on an object (a.k.a. a hashmap)" and `<<: *test` means "copy the object with the label 'test' into the current object."

PR #1423, opened 2018-05-14
modifying the regmapper annotations

depappas-sifive:
Removed an intermediate level in the case class hierachy
Removed regFiled name
Added deviceName 

This does not affect the old JSON regmap files.

There were merge problems with the old PR so I made a new PR.

### DIFF
--- a/src/main/scala/util/Annotations.scala
+++ b/src/main/scala/util/Annotations.scala
@@ -142,12 +142,17 @@ object GenRegDescsAnno {
     width: Int,
     byteOffset: Int,
     bitOffset: Int,
-    regField: RegField): RegFieldSer = {
+    regField: RegField): RegFieldDescSer = {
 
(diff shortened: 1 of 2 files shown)


src/main/scala/util/Annotations.scala
```diff
@@ -142,12 +142,17 @@ object GenRegDescsAnno {
...
     bitOffset: Int,
-    regField: RegField): RegFieldSer = {
+    regField: RegField): RegFieldDescSer = {
 
     val anonRegFieldName = s"unnamedRegField${byteOffset.toHexString}_${bitOffset}"
     val selectedRegFieldName = regField.desc.map(_.name).getOrElse(anonRegFieldName)
 
-    val map = Map[BigInt, (String, String)]() // TODO
+// TODO: enumerations will be handled in upcoming PR
+val map = Map[BigInt, (String, String)]() // TODO
```
mwachs5: Nit: wonky indentation

PR #1428, opened 2018-05-16
NegEdgeReg: use a real verilog negative edge register

mwachs5:
Certain toolflows (esp for FPGA) might not properly infer negative-edged flops when they are described in the Chisel way. Be explicit about it.

Note that this means these registers will NOT get `firrtl` initialization code (no `RANDOMIZE_REG_INIT` stuff)

### DIFF
--- a/src/main/scala/jtag/Utils.scala
+++ b/src/main/scala/jtag/Utils.scala
@@ -5,6 +5,7 @@ package freechips.rocketchip.jtag
 import Chisel._
 import chisel3.core.{Input, Output}
 import chisel3.experimental.withClock
+import chisel3.util.HasBlackBoxInline
 
 /** Bundle representing a tristate pin.
   */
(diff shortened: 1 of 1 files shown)


mwachs5:
@hcook when I tried doing this locally in this repo, I didn't actually see the `NegEdgeReg_1.v` module definition anywhere in `generated_src`.... am i missing some step? Or just looking in the wrong place?

hcook:
I don't know where else it would be

mwachs5:
@terpstra assigning to you to merge this if you think it's important/correct.

mwachs5:
This seems unnecessary if we have lived this long without merging this PR. closing and can always decide to make this change later.

sequencer:
I have some PR for clock edge:
https://github.com/freechipsproject/firrtl/pull/1125
However currently stack at `Mem` related issues
I'll work on this after https://github.com/freechipsproject/firrtl/pull/1210

PR #1430, opened 2018-05-16
Debug exception assert 2

mwachs5:
This is the same as #1420 , but for some reason Travis won't start that one. Opening this to attempt to jump-start it.


solomatnikov:
Does it make sense to add cover for debug exception for the case when it should happen?

mwachs5:
The more covers the better, but not sure exactly what you mean. You mean, the case when we are expecting to get an exception write?

solomatnikov:
I am assuming there are cases when debug exceptions are needed, right? Otherwise, why are they in the design. So, covers for those cases.

mwachs5:
This is not working any better than the previous one in terms of Travis not ignoring it so just going to close this one.

PR #1431, opened 2018-05-17
Upgrade to Scala 2.12.4

jackkoenig:
Unfortunately, it is not sufficient to just change the scalaVersion. SBT doesn't really support setting the version of subprojects in multi-project builds (see https://github.com/sbt/sbt/issues/1448), so we need to override the Scala version with `++2.12.4`.

~This currently points to a branch of hardfloat~

I'm PRing this to get Travis to run

EDIT: This no longer points to a branch of hardfloat, it turns out that you **can** add crossBuildVersions to a subproject so it is not necessary to change hardfloat at all.

Tagging some stakeholders to take a look/try it out. This obviously has implications for any project built on top of rocket-chip


jackkoenig:
This is ready to go. Projects built on top of rocket-chip should add `++2.12.4` or `++2.11.12` to their SBT invocation, otherwise there should be no problem.

PR #1433, opened 2018-05-17
Bump chisel3 and firrtl

jackkoenig:
Includes a new API to get current module and removes a broken Mem API


colinschmidt:
Shouldn't this remove the hacked in version of the module api?

jackkoenig:
Yes, I will add it

jackkoenig:
This ready to go, summary of changes:

Chisel
* Fix performance bug in VecInit
* Add Module.currentModule to get currently elaborating Module
  * Remove GetMeMyModule hack
* Make Mem constructions private (they didn't work, use object Mem.apply)
* Make UIntToOH work with output widths larger than 2^(input width)

FIRRTL
* Fix performance bug in Namespace
* Use absolute paths in BlackBox
* Support pad of less than argument width
* Better Verilog emission of some add/subs
* Fix some zero-width bugs
* Fix bug using temporary names in ports
* Deprecate some old WiringUtils methods
* Add TopWiringTransform

PR #1434, opened 2018-05-18
Elaborated config from annos3 rocket

depappas-sifive:
New elaborated configs. Add annotations for:
- Reset vectors
- Interrupts
- Registers
- Some Params

### DIFF
--- a/src/main/scala/util/Annotations.scala
+++ b/src/main/scala/util/Annotations.scala
@@ -14,6 +14,10 @@ import freechips.rocketchip.tilelink.TLToAXI4IdMapEntry
 import org.json4s.JsonDSL._
 import org.json4s.jackson.JsonMethods.{pretty, render}
 
+/** Record a set of interrupts. */
+case class InterruptsPortAnnotation(target: Named, name: String, nInterrupts: Seq[Int]) extends SingleTargetAnnotation[Named] {
+  def duplicate(n: Named) = this.copy(n)
+}
 /** Record a case class that was used to parameterize this target. */
 case class ParamsAnnotation(target: Named, paramsClassName: String, params: Map[String,Any]) extends SingleTargetAnnotation[Named] {
   def duplicate(n: Named) = this.copy(n)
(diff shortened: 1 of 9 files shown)


src/main/scala/util/Annotations.scala
```diff
@@ -14,6 +14,10 @@ import freechips.rocketchip.tilelink.TLToAXI4IdMapEntry
 import org.json4s.JsonDSL._
 import org.json4s.jackson.JsonMethods.{pretty, render}
 
+/** Record a set of interrupts. */
+case class InterruptsPortAnnotation(target: Named, name: String, nInterrupts: Seq[Int]) extends SingleTargetAnnotation[Named] {
```
hcook: Is this really `nInterrupts`, or is it storing the actual interrupt indexes?

src/main/scala/util/Annotations.scala
```diff
@@ -135,6 +156,10 @@ case class RegFieldDescMappingAnnotation(
 
 object GenRegDescsAnno {
 
+  var GLOBAL_EXTERNAL_INTERRUPTS = "global-external-interrupts"
+  var LOCAL_EXTERNAL_INTERRUPTS = "local-external-interrupts"
+  var LOCAL_INTERRUPTS_STARTING_NUMBER = 16 /* TODO the ISA specfication reserves the first 12 interrupts but
```
hcook: 1. Why are you adding these constants to the `GenRegDecsAnno` object, when they are not used in RegDesc annotations? 
2. Why are they `var`s if they are constants?
3. I found where the 16 is coming from: https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/tile/Interrupts.scala#L59
4. I think you should add a `def localInterruptOffset: Int = 16` function to `HasExternalInterrupts` and use it to fill in the value there and also in your annotation

src/main/scala/util/Annotations.scala
```diff
@@ -66,11 +70,28 @@ case class TopLevelPortAnnotation(
...
+object Annotated {
+  def interrupts(component: InstanceId, name: String, interrupts: Seq[Int]): Unit = {
+    annotate(new ChiselAnnotation {def toFirrtl: Annotation = InterruptsPortAnnotation(
+      component.toNamed,
+      name,
+      interrupts
+    )})
+  }
+
+  def resetVector(component: InstanceId, resetVec: BigInt): Unit = {
```
hcook: What do you think about having this function do the assignment as well:
` component := resetVec.U`
Seem like it will simplify using it downstream?

src/main/scala/util/Annotations.scala
depappas-sifive: 1,2 fixed

3. Andrew said to dot he following:

Add the following to Core.scala
HasCoreParameters

def firstLocalInterrupt = new MIP().getWidth - coreParams.nLocalInterrupts

Add to Interrupts.scala

import freechips.rocketchip.tile._

replace

GenRegDescsAnno.LOCAL_INTERRUPTS_STARTING_NUMBER

with

HasCoreParameters.firstLocalInterrupt

But HasCoreParameters.firstLocalInterrupt is not available in Annotations.scala...

src/main/scala/util/Annotations.scala
depappas-sifive: Lets talk about this in person.

PR #1443, opened 2018-05-23
Guard against integer overflow on the DiplomaticSRAM size.

ccelio:
A "size=(1<<31)" returns a negative integer size (and errors out in FIRRTL),
while size=(1<<32) wraps around back to 1 (instead of getting a 4B entry RAM).

Since comically large SRAMs aren't desirable anyways (and difficult to
simulate), let us place assertion guards on the caller side of the RAM creation.

### DIFF
--- a/src/main/scala/amba/ahb/SRAM.scala
+++ b/src/main/scala/amba/ahb/SRAM.scala
@@ -29,6 +29,7 @@ class AHBRAM(
 
   lazy val module = new LazyModuleImp(this) {
     val (in, _) = node.in(0)
+    require (mask.filter(b=>b).size < 31)
     val mem = makeSinglePortedByteWriteSeqMem(1 << mask.filter(b=>b).size)
 
     // The mask and address during the address phase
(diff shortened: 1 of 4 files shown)


ccelio:
Just a thought. I'm sure others may have differing opinions on how to protect against user error on mis-sizing SRAMs (like the simulated external memory SimAXIMem).

aswaterman:
True.  I'll leave it up to @terpstra but I think the following approach is more future-proof:

- change `makeSinglePortedByteWriteSeqMem` to accept `BigInt` instead of `Int`
- inside `makeSinglePortedByteWriteSeqMem`, add a `require(size.toInt == size)`
- change all uses of `makeSinglePortedByteWriteSeqMem` to pass `BigInt(1) << whatever` instead of `1 << whatever`

terpstra:
I'd rather that we make 'absurdly large' SRAMs work, along the lines @aswaterman suggests. For simulation-only, they could be quite useful. I presume this requires a chisel change, @jackkoenig ?

PR #1450, opened 2018-05-26
PhysicalFilter: unconditionally add an address bit

terpstra:
It is convenient in software to be able to describe the end of memory as a
power-of-two.  However, previously, if the largest physical address was
0xc0, the largest address you could program into the PhysicalFilter was
0xff.  While it did the useful thing for power-of-two largest addresses;
e.g., largest adress of 0xff made 0x100 programmable, not all designs end
up with power-of-two largest address. For example, a banked L2.

This PR changes the physical bits to always include one more bit than is
strictly necessary.


terpstra:
Closing in favor of #1451 which also resolves the banking issue, but by considering the maximum address across all ports.

PR #1459, opened 2018-05-30
Use only bus-error unit to report TL errors

aswaterman:
Having both synchronous and asynchronous reporting mechanisms is confusing,
and the synchronous reporting isn't possible for some reasonable design
points, e.g., in-order processors with nonblocking caches.

Make sure to instantiate a bus-error unit if any TL slaves can error!


aswaterman:
the DON'T MERGE is so @terpstra can do some additional testing with this approach

PR #1463, opened 2018-05-31
Factor out some generic code in HasRocketTiles

aswaterman:
HasRocketTile's clock-domain crossing code for TileLink and interrupts is pretty generic.  This PR moves it into helper methods on HasTile, so other tiles can leverage them without duplicating code.

### DIFF
--- a/src/main/scala/subsystem/HasTiles.scala
+++ b/src/main/scala/subsystem/HasTiles.scala
@@ -9,7 +9,7 @@ import freechips.rocketchip.devices.debug.TLDebugModule
 import freechips.rocketchip.devices.tilelink.{BasicBusBlocker, BasicBusBlockerParams, CLINT, TLPLIC}
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.interrupts._
-import freechips.rocketchip.tile.{BaseTile, TileParams, SharedMemoryTLEdge, HasExternallyDrivenTileConstants}
+import freechips.rocketchip.tile.{BaseTile, TileKey, TileParams, SharedMemoryTLEdge, HasExternallyDrivenTileConstants}
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.util._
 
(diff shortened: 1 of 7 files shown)


src/main/scala/subsystem/HasTiles.scala
```diff
@@ -22,6 +25,96 @@ trait HasTiles { this: BaseSubsystem =>
...
   def sharedMemoryTLEdge = sbus.busView
+
+  protected def augmentedTileParameters(tp: TileParams): Parameters = p.alterPartial {
+    // For legacy reasons, it is convenient to store some state
+    // in the global Parameters about the specific tile being built now
+    case TileKey => tp
+    case SharedMemoryTLEdge => sharedMemoryTLEdge
+  }
+
+  protected def connectMasterPortsToSBus(tile: BaseTile, crossing: RocketCrossingParams) {
```
hcook: Should `RocketCrossingParams` become `TileCrossingParams`?

src/main/scala/tile/Interrupts.scala
```diff
@@ -50,7 +50,7 @@ trait HasExternalInterrupts { this: BaseTile =>
 
   // TODO: the order of the following two functions must match, and
   //         also match the order which things are connected to the
-  //         per-tile crossbar in subsystem.HasRocketTiles
+  //         per-tile crossbar in subsystem.HasTiles.connectInterrupts
```
hcook: Will this code actually still work right if optional Debugs, CLINTs or PLICs are not present?

src/main/scala/subsystem/HasTiles.scala
aswaterman: Yes, I think so. I didn’t want to force that change as part of this refactor.

src/main/scala/tile/Interrupts.scala
aswaterman: Only for the ones that are currently optional. This is a stab at the API; I didn’t fix the fundamental problems.

PR #1464, opened 2018-06-01
LazyRawModuleImp: support a default clock for all children

terpstra:
This PR makes LazyRawModuleImp far more useful, as you can feed in a clock from one child to drive all the others. This is important when stitching together top-level modules with PLL children, for example.


colinschmidt:
This feature is very helpful!

PR #1472, opened 2018-06-05
tilelink: remove BCE channels completely if unnecessary

terpstra:
@hcook Let me know if this works to resolve your markDUT issue with TL-UH

### DIFF
--- a/src/main/scala/tilelink/Bundles.scala
+++ b/src/main/scala/tilelink/Bundles.scala
@@ -6,6 +6,7 @@ import Chisel._
 import chisel3.util.{ReadyValidIO}
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.util._
+import scala.collection.immutable.ListMap
 
 abstract class TLBundleBase(params: TLBundleParameters) extends GenericParameterizedBundle(params)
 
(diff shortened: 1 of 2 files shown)


hcook:
This resolved my issue!

terpstra:
I mostly don't like it, because it's cryptic code in a place where many people will end up looking eventually. However, if it works in all of the dependent repositories, then I guess it's ok to merge.

hcook:
I hear you.

Maybe we need some technology to materialize a particular view of an edge using a different Bundle/Record template? For example, I've noticed people creating "flattened" versions of the AXI4 bundle that does not use `Decoupled` (to elide the `.bits.` wrapper) and uses the ALL_CAPS lettering scheme from that spec. The flavor of TLBundle specific to TL-U[L/H] feels like another version of the same desire to customize a ports signal names.

I'm not sure whether it would be better expressed as a user-defined override supplied on a particular edge (`TLBundleParams => Data`), or whether each protocol type could have a built-in bundle type, e.g. `TLPortBundle` that has a particular format or additional logic to prune options signals even when a `DontTouch` annotation is going to be applied.  Any thoughts?

terpstra:
I don't think variable edge types is a good idea. You count on diplomacy to do the right thing with your bundles because it has very strict type and constructor argument discipline.

With respect to the AXI4 signal naming, I don't think we can ever match the names used in various IP blocks. They often call stuff MYSTUFFARVALID. So you need to have no _s in the name at all. That's going to be hard to match with nested Bundles. I agree we need a better way of creating blackbox wrappers, though.

hcook:
I used this again and I think I want to merge it.

The only other option I can come up with is to add a `makeIO` method on `TL[Manager|Client]Node`s that makes a special version of the bundle with the optional fields made optional and removed by the params. But it seems a little unnecessary to define a new top-level bundle type just for that?

hcook:
#YOLO

PR #1473, opened 2018-06-06
diplomacy: add BundleBridge for cross-module bundles

terpstra:
This can be quite useful for plumbing IO from deep inside the hierarchy to
the top of the design.

Example:
```
// Wrap up SomeDevice's module.io Bundle into an ioNode
val (ioNode, device) = BundleBridge(new SomeDevice)
// Somewhere else in the design in LazyModule scope:
val sink = ioNode.sink // creates a sink node of matching type
 // In LazyModuleImp scope:
val io = sink.io // io is a Wire () connected to device.module.io
```

### DIFF
--- a/src/main/scala/diplomacy/BundleBridge.scala
+++ b/src/main/scala/diplomacy/BundleBridge.scala
@@ -1,0 +1,20 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.diplomacy
+
+import chisel3._
+import chisel3.internal.sourceinfo.SourceInfo
+import freechips.rocketchip.config.{Parameters,Field}
+
+case class BundleBridgeParams[T <: Data](gen: () => T)
+case class BundleBridgeNull()
+
+class BundleBridgeImp[T <: Data]() extends SimpleNodeImp[BundleBridgeParams[T], BundleBridgeNull, BundleBridgeParams[T], T]
+{
+  def edge(pd: BundleBridgeParams[T], pu: BundleBridgeNull, p: Parameters, sourceInfo: SourceInfo) = pd
+  def bundle(e: BundleBridgeParams[T]) = e.gen()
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomacy/BundleBridge.scala
```diff
@@ -0,0 +1,57 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.diplomacy
+
+import chisel3._
+import chisel3.internal.sourceinfo.SourceInfo
+import freechips.rocketchip.config.{Parameters,Field}
+
+case class BundleBridgeParams[T <: Data](gen: () => T)
+case class BundleBridgeNull()
```
hcook: could this be a `case object`?

terpstra: No other parameters classes are objects, so even if it's possible, I think it would be non-conforming.

edcote:
I agree w/ @hcook.  This is very cool.

src/main/scala/diplomacy/BundleBridge.scala
```diff
@@ -0,0 +1,57 @@
...
+ *   // In LazyModuleImp scope:
+ *   val io = sink.io // io is a Wire () connected to device.module.io
+ */
+object BundleBridge
+{
+  def apply[D <: Data, T <: LazyModule](lm: => T { val module: { val io: D }})(implicit p: Parameters) = {
+    val bridge = LazyModule(new BundleBridge(lm))
+    (bridge.ioNode, bridge.child)
+  }
+}
```
jerryhethatday, line 57: I know this is out-of-date, but I wonder why you guys only defined io inside the BundleBridgeSink. When plumbing IO from deep inside the hierachy and use it as a sourceNode, how can I access its IO, just use ioNode.out(0)._1? Is this correct? So, from this viewpoint, the io defined inside the SinkNode is just a helper method?

PR #1477, opened 2018-06-08
Add InModuleBody API

terpstra:
This replaces the old instantiate() API.

This new API is strictly more powerful as you can use it from
helper functions. The instatiate() API required you to put your
extra glue code in one place and it had to be in the LazyModule.

Example:
```
def helper() {
  val child = LazyModule(new Device)
  InModuleBody { child.module.io.bypass := false.B }
}
```


colinschmidt:
Sounds super helpful!

hcook:
Ahh this is going to be nice. You still cannot create `IO(...)`'s in `InModuleBody` scope correct?

terpstra:
Actually... you totally can! You just need to proxy the IO member method inside a public method.

PR #1483, opened 2018-06-11
AsyncResetReg: support reset-value of '1'

terpstra:
While the inversion trick works most of the time, if you are putting a
register in an FPGA IO buffer, you can't have any logic between the register
and the output pin.

There was once a problem where LEC failed due to an async-set-reg, tho.
So I am not sure what the way forward should be.

### DIFF
--- a/src/main/resources/vsrc/AsyncResetReg.v
+++ b/src/main/resources/vsrc/AsyncResetReg.v
@@ -37,13 +37,14 @@
 `define RANDOMIZE
 `endif
 
-module AsyncResetReg (
-                      input      d,
-                      output reg q,
-                      input      en,
+module AsyncResetReg (d, q, en, clk, rst);
+parameter RESET_VALUE = 0;
 
-                      input      clk,
-                      input      rst);
+input  wire d;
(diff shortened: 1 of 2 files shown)


aswaterman:
IIRC this is what caused the LEC problem, and is why we were sticking to async-reset only, and avoiding async-set altogether.  Not sure how best to resolve this tension, aside from making the implementation of AsyncResetReg backend-specific.

terpstra:
If I recall correctly, when we had the LEC failure it was just for one of the registers. Others worked. So I wonder if that was a one-off or if it is a real problem.

mwachs5:
FWIW the original code was slightly different that was failing LEC. It
created two versions of the black box (AsyncResetReg vs AsyncSetReg). It
seems like both things are totally valid verilog so if it continues to fail
LEC we should probably figure out why it's confusing LEC (perhaps it was
synthesizing to the wrong type of flop, or the library didn't have an Async
Set flop, which is good to know).

In short LGTM.

terpstra:
@mwachs5 could you click the approve button, then? :)

PR #1489, opened 2018-06-12
Use DescribedSRAM to annotate SRAM instances

hcook:
Make it easier to generate nice descriptions of these important structures in downstream tools.


hcook:
Oh, I can't review it if I make the PR. Will have @depappas-sifive open the PR.

PR #1490, opened 2018-06-13
Sram annos1

depappas-sifive:
Adding DescribedSRAM


colinschmidt:
Does this integrate with/replace ReplSeqMem or is this a sideband interface consumed by some internal tooling?

PR #1494, opened 2018-06-14
BundleBridge: improve module and instance naming

terpstra:
Having a lot of children all called 'bridge' was not ideal.


hcook:
Haha makes sense

PR #1500, opened 2018-06-18
L1D: update how corrupt responses are handled

terpstra:
This PR makes the following changes:

1. Upon receipt of a corrupt block, inform the BEU, potentially taking an interrupt to handle it
2. When evicting a corrupt block, poison only those beats actually corrupt
3. Maintain wcorrupt in AXI4 test RAMs for unit testing

### DIFF
--- a/src/main/scala/amba/axi4/SRAM.scala
+++ b/src/main/scala/amba/axi4/SRAM.scala
@@ -12,7 +12,8 @@ class AXI4RAM(
     executable: Boolean = true,
     beatBytes: Int = 4,
     devName: Option[String] = None,
-    errors: Seq[AddressSet] = Nil)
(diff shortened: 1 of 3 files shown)


terpstra:
@aswaterman so ... do we want precise traps on 'denied' or not? My test cases now pass with both behaviors. However, I will note that it is a major PITA to work with the BEU-interrupt. The uncertainty of when it arrives is really not ideal. If we could, I'd rather go the other direction and make 'corrupt' precise as well (for memory). For MMIO, I think we both agree imprecise is just a fact of life.

terpstra:
The decision was to merge #1459 since the previous behaviour was buggy, anyway. That PR's commit has been removed from this branch which has been rebased.

PR #1504, opened 2018-06-18
add dts address anno

depappas-sifive:
adding dts address annos


src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -3,6 +3,7 @@
 package freechips.rocketchip.subsystem
 
 import Chisel._
+import chisel3.internal.InstanceId
```
jackkoenig: Is this import used?

depappas-sifive: fixed

PR #1509, opened 2018-06-19
adding width to port anno

depappas-sifive:
adding width to port anno


hcook:
Usually I'd want you to calculate "width" from `data: T`, but in this case you actually mean "width of the data bus of a potentially much wider port interface". I think it has to be up to the annotator to decide what "width" means and supply the argument accordingly, so this implementation is general enough.

PR #1511, opened 2018-06-20
Allow PLIC to be optional

hcook:
Allow for removal of the PLIC when the number of external or device interrupts is 0. All combinations are not yet gracefully handled:
- Some(PLIC) && IBus.intnode.sources.size > 0  => OK
- Some(PLIC) && IBus.intnode.sources.size == 0 => zero size vec error in PLIC
- None && IBus.intnode.sources.size > 0 => diplomacy error, no possible output for IBus.int_xbar
- None && IBus.intnode.sources.size == 0 => OK now

### DIFF
--- a/src/main/scala/devices/tilelink/Plic.scala
+++ b/src/main/scala/devices/tilelink/Plic.scala
@@ -60,7 +60,7 @@ case class PLICParams(baseAddress: BigInt = 0xC000000, maxPriorities: Int = 7, i
   def address = AddressSet(baseAddress, PLICConsts.size-1)
 }
 
-case object PLICKey extends Field(PLICParams())
+case object PLICKey extends Field[Option[PLICParams]](None)
 
 /** Platform-Level Interrupt Controller */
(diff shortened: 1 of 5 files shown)


aswaterman:
My understanding is that `HasExternalInterrupts.csrIntMap` and `decodeCoreInterrupts` have to match `HasTiles.connectInterrupts` exactly: that is, if an interrupt source is conditionalized, it must also be conditionalized in the other.  Note both currently check `useVM` to determine whether to wire up `seip`.

One option is to inform `decodeCoreInterrupts` that there's no PLIC so it can avoid wiring up `meip` and `seip`.  Is another option to change `connectInterrupts` to say "if PLIC, wire up PLIC, else, wire up a null interrupt source"?  In other words, make it so the PLIC interrupts are always presented to the tile; they just happen to be hardwired to 0.

solomatnikov:
I agree with @aswaterman hardwiring meip to 0 should be simpler b/c `decodeCoreInterrupts` does not need be modified.

hcook:
@aswaterman @solomatnikov Can one of you either approve this or take over responsibility for merging it?

PR #1513, opened 2018-06-21
chisel3: bump for the new dynamic port creation API

terpstra:
Now we can use helper methods/classes in Chisel to create IOs for large Modules!


jackkoenig:
I should note that this bump also includes a change that will reduce the suffixes on virtually every `_T_#` temporary. This is obviously a good thing but just want people to be aware.

PR #1516, opened 2018-06-25
RoCC Build Fixes

### LINKED ISSUE #1515: RoccExampleConfig throwing java.lang.IllegalAccessError
On master with `RoccExampleConfig` I'm getting a `java.lang.IllegalAccessError` tied to [line 91 of LazyRoCC.scala](https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/tile/LazyRoCC.scala#L91) which doesn't seem to make sense. This seems to be trying to access some reflection-related member in `RocketTileModuleImp` from `LazyRoCC` and failing, but I'm having some trouble understanding exactly what is going on here.

```
[error] Caused by: java.lang.IllegalAccessError: tried to access field freechips.rocketchip.tile.RocketTileModuleImp.reflPoly$Cache4 from class freechips.rocketchip.tile.HasLazyRoCCModule$class
[error] 	at freechips.rocketchip.tile.HasLazyRoCCModule$class.reflMethod$Method4(LazyRoCC.scala:91)
```

To reproduce this:

```bash
make CONFIG=RoccExampleConfig
```

Somewhat strangely, some initial triage of this where I commented out line 91, built, uncommented, and then built resulted in this switching to a `NullPointerException` tied to [line 105](https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/tile/LazyRoCC.scala#L105). (A stateful error? :fearful:)

h/t @aroelke

seldridge:
This fixes a problem where RoCC-enabled tiles would fail to build. This was due to bad inheritance where the inheriting class defined the FPU before it was used. The connections are now made more explicit by moving some code from the RoCC into the Tile.

- Creates a new `HasFpuOpt` trait that instantiates the FPU
- Mixes in the `HasFpuOpt` trait _before_ the `HasLazyRoCCModule` trait to make the `fpuOpt` member available for use inside the RoCC implementation
- Removes a kludge where the RoCC would first connect to a wire and then the core would connect to the same wire (assumedly intended to facilitate problems with trait ordering) in favor of making the connections explicit once the LHS and RHS exist
- Removes the type variable on `HasLazyRoCCModule` as this is no longer needed
- Changes the type of RoCC arbiters to `Option[T]` so that these now exist (possibly as `None`) inside of the Rocket tile and connections can be made there
- Enables building the `RoccExampleConfig` in Travis to hopefully prevent RoCC regressions going forward (ping: @colinschmidt)

Note: I was unable to root out what was exactly going on with the connections and the reflection error. Avoiding the wire kludge seemed to do it at the expense of having to migrate some logic from the RoCC implementation trait into the Rocket tile implementation class.

Fixes #1515

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,13 +40,13 @@ branches:
     - hwacha
     - boom
     - /^hurricane.*$/
-    
+
 # These branches are the only ones that
 # will build when "build branch updates"
 # is set in settings (branches which PR against
-# them are still built). With this set, 
+# them are still built). With this set,
 # the above blacklist is not useful.
-# Adding this allows us to keep "Build Branch Updates" 
+# Adding this allows us to keep "Build Branch Updates"
 # set to 'ON'.
 
 branches:
@@ -100,3 +100,6 @@ jobs:
       script:
         - travis_wait 80 make emulator-ndebug -C regression SUITE=RocketSuiteC JVM_MEMORY=3G
         - travis_wait 80 make emulator-regression-tests -C regression SUITE=RocketSuiteC JVM_MEMORY=3G
+    - <<: *test
+      script:
+        - make emulator-ndebug -C regression SUITE=Miscellaneous JVM_MEMORY=3G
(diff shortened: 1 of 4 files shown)


colinschmidt:
Thanks for fixing this!

PR #1517, opened 2018-06-25
adding addresses to port annos

depappas-sifive:
adding addresses to port annos


hcook:
Ports can have multiple AddressSets associated with them.

PR #1526, opened 2018-06-30
Add the GlobalConstantsAnnotation

depappas-sifive:


depappas-sifive:
@jackkoenig  "are there going to be lots of these?"

No idea

PR #1537, opened 2018-07-13
Clock Crossings: make API more flexible with CrossingHelper

hcook:
I added a `CrossingHelper` class that makes it easier for multiple crossing types to be used within a single wrapper. The helper also uses `ValName` to capture a nice name for the signals created by calls to its methods.

The change is backwards compatible with the extant `HasCrossing` trait, but I also started to retire use of that trait in cases where I needed better interface signal names (tiles) or know there will soon be multiple crossing types (bus wrappers). Using individual helpers instead of `HasCrossing` also allows client code to narrow the interface of `cross` methods to ones they actually support.

I think more helper objects for concisely creating named crossings and identity nodes could be useful, but will leave that to future work.

### DIFF
--- a/src/main/scala/subsystem/CrossingWrapper.scala
+++ b/src/main/scala/subsystem/CrossingWrapper.scala
@@ -42,27 +42,28 @@ class CrossingHelper(parent: LazyModule with LazyScope)(implicit valName: ValNam
   def crossTLSyncInOut(out: Boolean)(params: BufferParams = BufferParams.default)(implicit p: Parameters): TLNode = {
     val sync_xing = parent { LazyModule(new TLBuffer(params)).node }
     crossingCheck(out, sync_xing, sync_xing)
-    sync_xing
+    if (out) parent { TLIdentityNode()(valName) :*=* sync_xing }
+      else   parent { sync_xing :*=* TLIdentityNode()(valName) }
   }
 
   def crossTLAsyncInOut(out: Boolean)(depth: Int = 8, sync: Int = 3)(implicit p: Parameters): TLNode = {
     lazy val async_xing_source = LazyModule(new TLAsyncCrossingSource(sync))
     lazy val async_xing_sink = LazyModule(new TLAsyncCrossingSink(depth, sync))
-    val source = if (out) parent { async_xing_source } else async_xing_source
-    val sink = if (out) async_xing_sink else parent { async_xing_sink }
-    sink.node :*=* source.node
-    crossingCheck(out, source.node, sink.node)
-    NodeHandle(source.node, sink.node)
+    val source = if (out) parent { TLAsyncIdentityNode()(valName) :*=* async_xing_source.node } else async_xing_source.node
+    val sink = if (out) async_xing_sink.node else parent { async_xing_sink.node :*=* TLAsyncIdentityNode()(valName) }
+    crossingCheck(out, async_xing_source.node, async_xing_sink.node)
+    sink :*=* source
+    NodeHandle(source, sink)
   }
 
   def crossTLRationalInOut(out: Boolean)(direction: RationalDirection)(implicit p: Parameters): TLNode = {
     lazy val rational_xing_source = LazyModule(new TLRationalCrossingSource)
     lazy val rational_xing_sink = LazyModule(new TLRationalCrossingSink(if (out) direction else direction.flip))
-    val source = if (out) parent { rational_xing_source } else rational_xing_source
-    val sink = if (out) rational_xing_sink else parent { rational_xing_sink }
-    sink.node :*=* source.node
-    crossingCheck(out, source.node, sink.node)
-    NodeHandle(source.node, sink.node)
+    val source = if (out) parent { TLRationalIdentityNode()(valName) :*=* rational_xing_source.node } else rational_xing_source.node
+    val sink = if (out) rational_xing_sink.node else parent { rational_xing_sink.node :*=* TLRationalIdentityNode()(valName) }
+    crossingCheck(out, rational_xing_source.node, rational_xing_sink.node)
+    sink :*=* source
+    NodeHandle(source, sink)
   }
 
   def crossTLSyncIn (params: BufferParams = BufferParams.default)(implicit p: Parameters): TLNode = crossTLSyncInOut(false)(params)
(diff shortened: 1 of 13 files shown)


src/main/scala/subsystem/CrossingWrapper.scala
```diff
@@ -22,47 +22,48 @@ case class SynchronousCrossing(params: BufferParams = BufferParams.default) exte
...
-    sync_xing
+    lazy val sync_xing = LazyModule(new TLBuffer(params))
+    crossingCheck(out, sync_xing.node, sync_xing.node)
+    if (!out) parent { TLNameNode(valName) :*=* sync_xing.node }
+    else      parent { sync_xing.node :*=* TLNameNode(valName) }
   }
 
   def crossTLAsyncInOut(out: Boolean)(depth: Int = 8, sync: Int = 3)(implicit p: Parameters): TLNode = {
     lazy val async_xing_source = LazyModule(new TLAsyncCrossingSource(sync))
     lazy val async_xing_sink = LazyModule(new TLAsyncCrossingSink(depth, sync))
```
terpstra: Do these LazyModules end up taking their name from the CrossingHelper's implicit ValName? Do we want this behaviour?

src/main/scala/subsystem/CrossingWrapper.scala
```diff
@@ -158,36 +160,39 @@ trait HasCrossingMethods extends LazyModule with LazyScope
...
-  val crossing: SubsystemClockCrossing
 
-  def crossTLIn   (implicit p: Parameters): TLNode  = crossTLIn   (crossing)
-  def crossTLOut  (implicit p: Parameters): TLNode  = crossTLOut  (crossing)
-  def crossAXI4In (implicit p: Parameters): AXI4Node= crossAXI4In (crossing)
-  def crossAXI4Out(implicit p: Parameters): AXI4Node= crossAXI4Out(crossing)
-  def crossIntIn  (implicit p: Parameters): IntNode = crossIntIn  (crossing)
-  def crossIntOut (implicit p: Parameters): IntNode = crossIntOut (crossing)
+  def crossing: SubsystemClockCrossing
+  protected lazy val xing = new CrossingHelper(this, crossing)
```
terpstra: This should not be lazy. The class has no effects, so there is no reason to be lazy. Making it lazy will make it called 'xing_lzy' in signal names. We can skip the _lzy I think.

src/main/scala/subsystem/FrontBus.scala
```diff
@@ -18,14 +18,15 @@ case object FrontBusKey extends Field[FrontBusParams]
 
 class FrontBus(params: FrontBusParams)
               (implicit p: Parameters) extends TLBusWrapper(params, "front_bus")
-    with HasTLXbarPhy
-    with HasCrossing {
-  val crossing = params.sbusCrossing
+    with HasTLXbarPhy {
+
+  protected val sbus_xing = new CrossingHelper(this, params.sbusCrossing)
+  def crossTLOut(implicit p: Parameters): TLNode  = sbus_xing.crossTLOut
```
terpstra: I am not sure I approve of these forwarding functions. I think it might be more clear to have the caller explicitly refer to the crossing by name. eg: fbus.sbus_xing.crossTLOut(...)

src/main/scala/subsystem/FrontBus.scala
```diff
@@ -18,14 +18,15 @@ case object FrontBusKey extends Field[FrontBusParams]
...
 
   def fromPort[D,U,E,B <: Data]
       (name: Option[String] = None, buffer: BufferParams = BufferParams.none)
       (gen: => NodeHandle[D,U,E,B,TLClientPortParameters,TLManagerPortParameters,TLEdgeOut,TLBundle] =
-        TLIdentity.gen): InwardNodeHandle[D,U,E,B] = {
+        TLNameNode(name)): InwardNodeHandle[D,U,E,B] = {
     from("port" named name) { fixFrom(TLFIFOFixer.all, buffer) :=* gen }
```
terpstra: Nice. 👍

src/main/scala/subsystem/PeripheryBus.scala
```diff
@@ -21,14 +21,15 @@ case object PeripheryBusKey extends Field[PeripheryBusParams]
 
 class PeripheryBus(params: PeripheryBusParams)
                   (implicit p: Parameters) extends TLBusWrapper(params, "periphery_bus")
-    with HasTLXbarPhy
-    with HasCrossing {
-  val crossing = params.sbusCrossingType
+    with HasTLXbarPhy {
+
+  protected val sbus_xing = new CrossingHelper(this, params.sbusCrossingType)
+  def crossTLIn(implicit p: Parameters): TLNode  = sbus_xing.crossTLIn
```
terpstra: Again, I think it might be better to make the crossing public with a good name. I understand you need the forwarding functions for backwards compatibility, but maybe we could just leave them in as deprecated?

src/main/scala/tilelink/Nodes.scala
```diff
@@ -43,6 +43,13 @@ case class TLAdapterNode(
 
 case class TLIdentityNode()(implicit valName: ValName) extends IdentityNode(TLImp)()
 
+object TLNameNode {
+  def apply(name: ValName) = TLIdentityNode()(name)
+  def apply(name: String): TLIdentityNode = apply(Some(name))
+  def apply(name: Option[String]) =
+    TLIdentityNode()(ValName(name.getOrElse("with_no_name")))
+}
+
```
terpstra: This is pretty handy, but I wonder if it should exist for Interrupts and AXI as well?

src/main/scala/subsystem/CrossingWrapper.scala
hcook: It's for backwards compatibility with things following the `Island` pattern that declare a hard-coded `val crossing` in their constructor body rather than taking it as a constructor parameter.

src/main/scala/subsystem/CrossingWrapper.scala
hcook: I want the CrossingHelper to determine the name some how. Taking it from the val name was your suggestion originally iirc, haha...  But if we are going to expose the val name of the CrossingHelper as part of e.g. the bus wrapper's API, maybe it would be better to decouple that from the signal name.

src/main/scala/subsystem/PeripheryBus.scala
hcook: I'm not especially worried about breaking this BusWrapper api compatibility at this juncture because it only affects the two points where they connect to each other.

src/main/scala/tilelink/Nodes.scala
hcook: Yeah I wanted your approval of it first. Will do

src/main/scala/subsystem/FrontBus.scala
hcook: I think I agree

rmac-sifive:
Repeating @terpstra's concerns: will this affect existing backend flows? Since we're using the legacy Aloe flow to prove out the pbus clocking changes, and I'm already hacking the generated Almond verilog quite a bit before giving it to Han, so it's preferable that this PR doesn't introduce the need for additional hacks

PR #1545, opened 2018-07-19
BundleBridge and CrossingWrapper API cleanups (pt 2)

hcook:
Follow on to #1544

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -138,9 +138,9 @@ abstract class BaseTile(tileParams: TileParams, val crossing: SubsystemClockCros
   def intInwardNode: IntInwardNode
   def intOutwardNode: IntOutwardNode
 
-  protected val tlMasterXing = new CrossingHelper(this, crossing, "tl_master_xing")
-  protected val tlSlaveXing = new CrossingHelper(this, crossing, "tl_slave_xing")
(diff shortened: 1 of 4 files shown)


src/main/scala/tile/BaseTile.scala
```diff
@@ -138,13 +138,13 @@ abstract class BaseTile(tileParams: TileParams, val crossing: SubsystemClockCros
...
-  protected val tlMasterXing = new CrossingHelper(this, crossing, "tl_master_xing")
-  protected val tlSlaveXing = new CrossingHelper(this, crossing, "tl_slave_xing")
-  protected val intXing = new CrossingHelper(this, crossing, "int_xing")
-  def crossTLOut (implicit p: Parameters): TLNode  = tlMasterXing.crossTLOut
-  def crossTLIn  (implicit p: Parameters): TLNode  = tlSlaveXing.crossTLIn
-  def crossIntIn (implicit p: Parameters): IntNode = intXing.crossIntIn
-  def crossIntOut(implicit p: Parameters): IntNode = intXing.crossIntOut
+  protected val tlMasterXingNode = new CrossingHelper(this, crossing)
+  protected val tlSlaveXingNode = new CrossingHelper(this, crossing)
+  protected val intXingNode = new CrossingHelper(this, crossing)
```
terpstra: Wut? Why are these called nodes? They are not. You had _xing in the name before, so leave it called Xing at the end to get the same effect.

PR #1546, opened 2018-07-19
Revamped Register Router

hcook:
`RegisterRouters` make use of several new APIs that give them advantages over generic `TLRegisterRouters`. They are also less TL-specific.
- They take a `ClockCrossingType` as an method argument to `CrossingHelper` to supply two standard crossing methods
- They take a `Data` as an argument and use it with `BundleBridgeSource` to supply a method that can be used to create a node that then can punch an IO out of an arbitrary location.

This is intended for use with some matching changes to `sifive-blocks`.


terpstra:
Interestingly, the name 'RegisterRouter' is actually NOT used anywhere yet. There is TLRegisterRouter/etc, but no RegisterRouter. So we could totally give this the 'final' name now and not have to rename it later. Since you are just mixing a TLRegMap trait with the RegisterRouter, there is no namespace collision to worry about at the TL/etc level, either.

hcook:
Ok! Will do!

src/main/scala/regmapper/RegisterRouter.scala
```diff
@@ -0,0 +1,45 @@
...
+  compat: Seq[String],
+  base: BigInt,
+  size: BigInt = 4096,
+  concurrency: Int = 0,
+  beatBytes: Int = 4,
+  undefZero: Boolean = true,
+  executable: Boolean = false)
+
+abstract class RegisterRouter[T <: Data](val devParams: RegisterRouterParams, portBundle: => T)
+                                        (implicit p: Parameters)
```
terpstra: Why do you proxy all the devParams? Can the traits not reach devParams.undefZero/etc from the constructor val for some reason?

PR #1548, opened 2018-07-19
Refactor TLBusWrapper and CrossingHelper

hcook:
I refactored both the TLBusWrapper and CrossingHelper APIs. The former is slightly more flexible and has better code reuse. The latter is a complete re-write No intended functional RTL changes other than signal names.

TLBusWrapper:
- Unprotect or removed the base class's adapter methods
- Moved the common master and slave attachment functions into traits `CanAttachTLSlaves` and `CanAttachTLMasters`
- Changed how the existing buses clock cross to each other based on the following changes:

CrossingHelper:
- `HasCrossing` is deprecated, `CrossingWrapper` uses the replacement trait `CrossesToOnlyOneClockDomain`
- Renames `subsystem.SubsystemClockCrossing` to `diplomacy.ClockCrossingType`
- Puts `HasClockDomainCrossing` base trait in diplomacy
- Defines several `[Protocol][Out|In]wardCrossingHelper` case classes for each protocol type and `[Protocol]ClockDomainCrossing` implicit class conversions in each package to add crossing methods to instances of `HasClockDomainCrossing` that need to be crossed
- Adds `ClockCrossingType` argument to individual calls to `CrossingHelper` methods


src/main/scala/amba/axi4/CrossingHelper.scala
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  def crossAXI4Out(implicit p: Parameters): AXI4Node = crossingType match {
+    case x: SynchronousCrossing  => crossAXI4SyncOut(x.params)
+    case x: AsynchronousCrossing => crossAXI4AsyncOut(x.depth, x.sync)
+    case x: RationalCrossing     => throw new IllegalArgumentException("AXI4 Rational crossing unimplemented")
+  }
+}
+
+class AXI4CrossingWrapper(val crossing: ClockCrossingType)(implicit p: Parameters) extends SimpleLazyModule with LazyScope {
```
terpstra: If you split the crossing helpers by type like you've done, then I think this API (the wrapper) has to die. You typically need to cross more than one bus type and this wrapper can only cross AXI. So it's mostly useless.

terpstra: The way these wrappers work is to drop a half of the crossing in the island and place the other half in the caller. We need to IMO reconsider a few things: is the island a new module layer just doing the crossing? How do we relate multiple clocks each with multiple busses.

terpstra: On the one hand, you really want only one helper per clock domain in the island. On the other hand, you want to change the crossing type based on the relationship between that domain and the clock where the other half is dropped.

hcook: We decided to:
- Put the deprecation only on the methods in `HasCrossing`
- de-deprecate `CrossingWrapper` and delete the protocol-specific Wrappers
- Move the `crossingType` arg from the constructor to the function call
- Remove the prefixes on the protocol-specific function name

hcook:
We decided to:
- Put the deprecation only on the methods in HasCrossing
- De-deprecate CrossingWrapper and delete the protocol-specific Wrappers
- Move the crossingType arg from the constructor to the function call
- Remove the prefixes on the protocol-specific function name

PR #1550, opened 2018-07-25
Several QoR and minor generator functionality improvements for Rocket-related utilities

aswaterman:
In trying to reuse some of these units for another generator I'm working on,  I noticed some deficiencies in QoR and also wanted a few more knobs to turn.

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -700,11 +700,13 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
     amoalu.io.cmd := (if (usingAtomicsInCache) pstore1_cmd else M_XWR)
     amoalu.io.lhs := s2_data_word
     amoalu.io.rhs := pstore1_data
-    pstore1_storegen_data := amoalu.io.out
+    pstore1_storegen_data := (if (!usingDataScratchpad) amoalu.io.out else {
+      val mask = FillInterleaved(8, Mux(s2_correct, 0.U, pstore1_mask))
+      amoalu.io.out_unmasked & mask | s2_data_word_corrected & ~mask
+    })
   } else if (!usingAtomics) {
     assert(!(s1_valid_masked && s1_read && s1_write), "unsupported D$ operation")
   }
-  when (s2_correct) { pstore1_storegen_data := s2_data_word_corrected }
 
   // flushes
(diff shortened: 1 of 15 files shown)


src/main/scala/rocket/DCache.scala
```diff
@@ -449,19 +459,24 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
...
+    val uncachedGrantOpcodes = uncachedGrantOpcodesWithData ++ uncachedGrantOpcodesSansData
+    val whole_opc = tl_out.d.bits.opcode
     if (usingDataScratchpad) {
-      assert(!(tl_out.d.valid && res))
-      false.B
+      assert(!tl_out.d.valid || whole_opc.isOneOf(uncachedGrantOpcodes))
+      // the only valid TL-D messages are uncached, so we can do some pruning
+      val opc = whole_opc(uncachedGrantOpcodes.map(_.getWidth).max - 1, 0)
+      val data = DecodeLogic(opc, uncachedGrantOpcodesWithData, uncachedGrantOpcodesSansData)
+      (opc, true.B, data)
```
hcook: Seems like all of this should get sucked into some kind of helper method on `TLMessages`?

aswaterman: Per in-person discussion, postponing until TL message-ID mapping is revisited and associated code is refactored.

src/main/scala/rocket/TLB.scala
```diff
@@ -79,31 +82,92 @@ class TLB(instruction: Boolean, lgMaxSize: Int, nEntries: Int)(implicit edge: TL
...
   }
 
-  val totalEntries = nEntries + 1
-  val normalEntries = nEntries
-  val specialEntry = nEntries
-  val aeEntry = specialEntry - (1 << log2Floor(nEntries))
-  val valid = Reg(init = UInt(0, totalEntries))
-  val reg_entries = Reg(Vec(totalEntries, UInt(width = new Entry().getWidth)))
-  val entries = reg_entries.map(_.asTypeOf(new Entry))
+  class Entry(nSectors: Int, superpageOnly: Boolean) extends Bundle {
```
hcook: FYI if you turn these into `val`s you shouldn't need to define your own `cloneType` anymore

aswaterman: awesome

PR #1551, opened 2018-07-25
TLToAPB: cut paths from A to D

terpstra:
One small step for flops, but one large step for timing!

### DIFF
--- a/src/main/scala/tilelink/ToAPB.scala
+++ b/src/main/scala/tilelink/ToAPB.scala
@@ -63,6 +63,11 @@ class TLToAPB(val aFlow: Boolean = true)(implicit p: Parameters) extends LazyMod
       val a_sel    = a.valid && RegNext(!in.d.valid || in.d.ready)
       val a_write  = edgeIn.hasData(a.bits)
 
+      val enable_d = a_sel && !a_enable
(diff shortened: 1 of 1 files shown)


src/main/scala/tilelink/ToAPB.scala
```diff
@@ -63,6 +63,10 @@ class TLToAPB(val aFlow: Boolean = true)(implicit p: Parameters) extends LazyMod
       val a_sel    = a.valid && RegNext(!in.d.valid || in.d.ready)
       val a_write  = edgeIn.hasData(a.bits)
 
+      val d_write  = RegNext(a_write)
+      val d_source = RegNext(a.bits.source)
+      val d_size   = RegNext(a.bits.size)
```
aswaterman: Is the reason you aren't clock-gating these that they are tiny in practice?

terpstra: Also, the only criteria that would be safe is '!a_enable', which is true nearly all the time.

terpstra: Nevermind. I found a better condition.

aswaterman:
cool

PR #1552, opened 2018-07-25
Ignore SFENCE rs1 argument when VM is disabled

aswaterman:
Using SFENCE with rs1 != x0 relies on reusing the TLB hit path, but it's
not possible to hit when VM is disabled, so the logic did the wrong thing.
The simple fix is to flush the whole TLB when using SFENCE when VM is
disabled.  This shouldn't be an important performance case, because SFENCE
is nearly only used when VM is enabled (for obvious reasons).


aswaterman:
@hcook can you approve this?  It's not yet clear whether the old behavior will be valid under the ISA, but this commit is conservative so it'll be valid regardless.

PR #1558, opened 2018-07-30
Adding Seq for opcode and param translation in TLMessages

srivatsa611y:
We are adding methods to decode the current Opcode and Param information on the tilelink bundle.

The way it is done might be messy and hoping for the reviewers for better ways.

### DIFF
--- a/src/main/scala/tilelink/Bundles.scala
+++ b/src/main/scala/tilelink/Bundles.scala
@@ -141,14 +141,14 @@ final class TLBundleA(params: TLBundleParameters)
   def opcodePrint()
   {
     switch (opcode) {
-      is(0.U) {printf("OP: PutFullData    ")} 
-      is(1.U) {printf("OP: PutPartialData ")}
-      is(2.U) {printf("OP: ArithmeticData ")} 
-      is(3.U) {printf("OP: LogicalData    ")} 
-      is(4.U) {printf("OP: Get            ")}
-      is(5.U) {printf("OP: Hint           ")}
-      is(6.U) {printf("OP: AcquireBlock   ")}
-      is(7.U) {printf("OP: AcquirePerm    ")}
+      is(0.U) {printf("OP: PutFullData      ")} 
+      is(1.U) {printf("OP: PutPartialData   ")}
+      is(2.U) {printf("OP: ArithmeticData   ")} 
+      is(3.U) {printf("OP: LogicalData      ")} 
+      is(4.U) {printf("OP: Get              ")}
+      is(5.U) {printf("OP: Hint             ")}
+      is(6.U) {printf("OP: AcquireBlock     ")}
+      is(7.U) {printf("OP: AcquirePerm      ")}
     }
   }
   
@@ -157,11 +157,11 @@ final class TLBundleA(params: TLBundleParameters)
     when(opcode === 2.U) 
     {
       switch (param) {
-        is(0.U) {printf(" P: MIN           ")}
-        is(1.U) {printf(" P: MAX           ")}
-        is(2.U) {printf(" P: MINU          ")}
-        is(3.U) {printf(" P: MAXU          ")}
-        is(4.U) {printf(" P: ADD           ")}
+        is(0.U) {printf(" P: MIN            ")}
+        is(1.U) {printf(" P: MAX            ")}
+        is(2.U) {printf(" P: MINU           ")}
+        is(3.U) {printf(" P: MAXU           ")}
+        is(4.U) {printf(" P: ADD            ")}
       }
     }
     .elsewhen(opcode === 3.U) 
--- a/src/main/scala/tilelink/Bundles.scala
+++ b/src/main/scala/tilelink/Bundles.scala
@@ -141,14 +141,14 @@ final class TLBundleA(params: TLBundleParameters)
   def opcodePrint()
   {
     switch (opcode) {
-      is(0.U) {printf("OP: PutFullData")} 
-      is(1.U) {printf("OP: PutPartialData")}
-      is(2.U) {printf("OP: ArithmeticData")} 
-      is(3.U) {printf("OP: LogicalData")} 
-      is(4.U) {printf("OP: Get")}
-      is(5.U) {printf("OP: Hint")}
-      is(6.U) {printf("OP: AcquireBlock")}
-      is(7.U) {printf("OP: AcquirePerm")}
+      is(0.U) {printf("OP: PutFullData    ")} 
+      is(1.U) {printf("OP: PutPartialData ")}
+      is(2.U) {printf("OP: ArithmeticData ")} 
+      is(3.U) {printf("OP: LogicalData    ")} 
+      is(4.U) {printf("OP: Get            ")}
+      is(5.U) {printf("OP: Hint           ")}
+      is(6.U) {printf("OP: AcquireBlock   ")}
+      is(7.U) {printf("OP: AcquirePerm    ")}
     }
   }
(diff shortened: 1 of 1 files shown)


src/main/scala/tilelink/Bundles.scala
```diff
@@ -134,8 +137,59 @@ final class TLBundleA(params: TLBundleParameters)
...
+      is(0.U) {printf("OP: PutFullData      ")} 
+      is(1.U) {printf("OP: PutPartialData   ")}
+      is(2.U) {printf("OP: ArithmeticData   ")} 
+      is(3.U) {printf("OP: LogicalData      ")} 
+      is(4.U) {printf("OP: Get              ")}
+      is(5.U) {printf("OP: Hint             ")}
+      is(6.U) {printf("OP: AcquireBlock     ")}
+      is(7.U) {printf("OP: AcquirePerm      ")}
+    }
+  }
```
seldridge: A functional style can help with the clutter. One equivalent approach would be:
```scala
  def opcodePrint(): Unit =
    Seq("PutFullData", "PutPartialData", "ArithmeticData", "LogicalData", "Get", "Hint", "AcquireBlock", "AcquirePerm")
      .zipWithIndex
      .map{ case (str, idx) => when (opcode === idx.U) { printf(s"OP: $str\n") } }
```

srivatsa611y: Can this be done in runtime, because thats where we will be using it.

seldridge: Both methods will statically emit circuits with `printf`s that trip on the conditions specified. So, they will generate hardware `printf`s that are usable at runtime.

Note: there is a subtle difference here in that `switch`/`is` produces a `when`/`else` nesting while the explicit `when` blocks of the functional style are all flat. Functionally, there's no difference, though.

The original is doing something like:
```
    node _T_33 = eq(UInt<1>("h00"), UInt<1>("h01")) 
    when _T_33 : 
      node _T_34 = bits(reset, 0, 0) 
      node _T_35 = eq(_T_34, UInt<1>("h00")) 
      when _T_35 : 
        printf(clock, UInt<1>(1), "OP: PutFullData      ") 
    else : 
      node _T_36 = eq(UInt<1>("h01"), UInt<1>("h01")) 
      when _T_36 : 
        node _T_37 = bits(reset, 0, 0) 
        node _T_38 = eq(_T_37, UInt<1>("h00")) 
        when _T_38 : 
          printf(clock, UInt<1>(1), "OP: PutPartialData   ") 
```

The `map` is generating:
```
    node _T_9 = eq(UInt<1>("h01"), UInt<1>("h00")) 
    when _T_9 : 
      node _T_10 = bits(reset, 0, 0) 
      node _T_11 = eq(_T_10, UInt<1>("h00")) 
      when _T_11 : 
        printf(clock, UInt<1>(1), "OP: PutFullData\n") 
    node _T_12 = eq(UInt<1>("h01"), UInt<1>("h01")) 
    when _T_12 : 
      node _T_13 = bits(reset, 0, 0) 
      node _T_14 = eq(_T_13, UInt<1>("h00")) 
      when _T_14 : 
        printf(clock, UInt<1>(1), "OP: PutPartialData\n") 
```

hcook:
Why are there two distinct print functions instead of one a single one that prints something about every field in the channel?

hcook:
You should use the defs in the `TLMessages` object, rather than hardcoding the values in two places: https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/tilelink/Bundles.scala#L16

We might change the encoding in the future!

srivatsa611y:
yea, we wanted your reviews. the print is all being moved out of rocketchip. We are planning to have maps as suggested by @seldridge

srivatsa611y:
Changes done as requested.

srivatsa611y:
@hcook  @terpstra  Can you guys please review the pr. I have done the changes as requested.

src/main/scala/tilelink/Bundles.scala
```diff
@@ -44,6 +44,41 @@ object TLMessages
...
   def bcResponse = Vec(AccessAck, AccessAck, AccessAckData, AccessAckData, AccessAckData, HintAck, ProbeAck, ProbeAck)
+  
+  def a = Seq( ("PutFullData",TLPermissions.PermMsgReserved),
+               ("PutPartialData",TLPermissions.PermMsgReserved),
+               ("ArithmeticData",TLAtomics.ArithMsg),
+               ("LogicalData",TLAtomics.LogicMsg),
+               ("Get",TLPermissions.PermMsgReserved),
+               ("Hint",TLHints.HintsMsg),
+               ("AcquireBlock",TLPermissions.PermMsgGrow),
+               ("AcquirePerm",TLPermissions.PermMsgReserved))
```
terpstra: Oh. This should be PermMsgGrow.

src/main/scala/tilelink/Bundles.scala
```diff
@@ -44,6 +44,41 @@ object TLMessages
...
+
+  def b = Seq( ("PutFullData",TLPermissions.PermMsgReserved),
+               ("PutPartialData",TLPermissions.PermMsgReserved),
+               ("ArithmeticData",TLPermissions.PermMsgReserved),
+               ("LogicalData",TLPermissions.PermMsgReserved),
```
terpstra: This should be LogicMsg

src/main/scala/tilelink/Bundles.scala
```diff
@@ -44,6 +44,41 @@ object TLMessages
...
+               ("ArithmeticData",TLAtomics.ArithMsg),
+               ("LogicalData",TLAtomics.LogicMsg),
+               ("Get",TLPermissions.PermMsgReserved),
+               ("Hint",TLHints.HintsMsg),
+               ("AcquireBlock",TLPermissions.PermMsgGrow),
+               ("AcquirePerm",TLPermissions.PermMsgReserved))
+
+  def b = Seq( ("PutFullData",TLPermissions.PermMsgReserved),
+               ("PutPartialData",TLPermissions.PermMsgReserved),
+               ("ArithmeticData",TLPermissions.PermMsgReserved),
```
terpstra: This should be ArithMsg

src/main/scala/tilelink/Bundles.scala
```diff
@@ -44,6 +44,41 @@ object TLMessages
...
+               ("LogicalData",TLPermissions.PermMsgReserved),
+               ("Get",TLPermissions.PermMsgReserved),
+               ("Hint",TLPermissions.PermMsgCap),
```
terpstra: Should be HintsMsg

src/main/scala/tilelink/Bundles.scala
```diff
@@ -44,6 +44,41 @@ object TLMessages
...
+               ("Probe",TLPermissions.PermMsgReserved))
```
terpstra: Should be PermsMsg. something. can't review now.

src/main/scala/tilelink/Bundles.scala
srivatsa611y: Well they are for the reserved fields. So i put them all in 1 object and use the same seq in other places. Do you want me to change is accordingly?

srivatsa611y: OOh sorry my bad. I'll cross verify it

srivatsa611y: The AcquirePerm Is not there in the TileLink Spec. Draft 1.7

PR #1559, opened 2018-07-31
subsystem: make sbus arbitration configurable

hcook:
Note: using this is kind of dangerous; make sure you aren't going to starve anyone!


hcook:
Correct

PR #1560, opened 2018-08-01
Add a clock divider with explicit input clock

rmac-sifive:
Adds a clock divider module which allows for an explicit clock input, and rewrites `Pow2ClockDivider` to instantiate this module for legacy compatibility

### DIFF
--- a/src/main/scala/util/ClockDivider.scala
+++ b/src/main/scala/util/ClockDivider.scala
@@ -31,11 +31,12 @@ class ClockDivider3 extends BlackBox with HasBlackBoxResource {
   setResource("/vsrc/ClockDivider3.v")
 }
 
-/** Divide the clock by power of 2 times.
- *  @param pow2 divides the clock 2 ^ pow2 times */
-class Pow2ClockDivider(pow2: Int) extends Module {
+/** Divide an input clock by power of 2 times.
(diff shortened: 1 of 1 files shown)


aswaterman:
@hcook should make the call, but another option is making a companion object, and for the explicit version, use withClock as the means to pass it in:

    object Pow2ClockDivider {
      // legacy version
      def apply(pow2: Int): Clock = Module(new Pow2ClockDivider(pow2)).clock_out

      // new version
      def apply(clock_in: Clock, pow2: Int): Clock = withClock(clock_in) { apply(pow2) }
    }

rmac-sifive:
@aswaterman with that solution, I think I'd have to modify the each instantiation of the divider in the code-base from `Module(new Pow2ClockDivider(n))` to just`Pow2ClockDivider`, right?

aswaterman:
I'm proposing keeping the existing module entirely as-is, so old code doesn't have to change (but new code should probably favor `Pow2ClockDivider(n)`)

rmac-sifive:
Ah, understood. I'll leave it to @hcook which implementation to take

rmac-sifive:
I talked to Henry offline and he prefers the object implementation, and I updated the PR to that implementation

PR #1568, opened 2018-08-08
Fix AsyncResetReg

rmac-sifive:
This fixes a reset race condition caused when the reset of an AsyncResetReg_1 is driven by another AsyncResetReg_2. In that case, the AsyncResetReg_2's initialization behavior could cause AsyncResetReg_1 to be undefined

### DIFF
--- a/src/main/resources/vsrc/AsyncResetReg.v
+++ b/src/main/resources/vsrc/AsyncResetReg.v
@@ -58,16 +58,14 @@ input  wire rst;
       integer                    initvar;
       reg [31:0]                 _RAND;
       _RAND = {1{$random}};
-`ifdef verilator
-      q = RESET_VALUE;
-`else
  `ifdef RANDOMIZE
    `ifdef RANDOMIZE_REG_INIT
+    `ifndef verilator
       #0.002 begin end
+    `endif
       q = _RAND[0];
    `endif // RANDOMIZE_REG_INIT
  `endif // RANDOMIZE
-`endif // verilator
    end
 `endif // SYNTHESIS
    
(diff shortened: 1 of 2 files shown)


aswaterman:
I'm leery of the removal of the 0.002 delay - why's that OK here?

rmac-sifive:
This fixed an issue I see in simulation in which the reset of AsyncResetReg_1 was driven from AsyncResetReg_2. In that case, the assignments at t=0.002 raced, and AsyncResetReg_1's intial block at t=0.002 executed first, rst was not defined, and AsyncResetReg_1 starts simulation undefined

Now that I think of it, I think this PR only moves the same race to t=0, and it works by chance, since if AsyncResetReg_1's t=0 initial block executes first we'll have the same issue

Would just assigning q = _RAND at t=0.002 as we do with other registers work? Why do we ever need registers to have their reset value at t=0.002? I would expect some external reset should propagate through the design to do this

rmac-sifive:
@aswaterman, @jackkoenig, I've updated the PR to get rid of the initialization based on `rst` as we discussed offline. Now this unit is only initialized to `_RAND`, which is consistent with methodology for other registers

This fixed the issue I was seeing, but I will run a few more test cases before merging

rmac-sifive:
After running into issues with the previous version, it became apparent that the DMOuter domain (clocked externally by JTAG, SimJTAG in simulation), was relying on AsyncResetReg's previous behavior which initialization flops to their reset value at t = 0. This was masking the bug that the DMOuter clock domain was not reset properly, because TCK was never toggled during reset

PR #1570, opened 2018-08-10
removed BuildCore because of deprecated usage and straight up calls c…

jieltan:
…ore = new Rocket instead


hcook:
@ccelio Boom isn't still using this, right?

jieltan:
@aswaterman how about now

ccelio:
@hcook Thanks for checking! I am fine with this removal. BOOM now extends the BaseTile and uses the HasBoomTile mixin. 

https://github.com/ucb-bar/riscv-boom/blob/41cd1b8d55d98b26ff43e4fff3459747589acc8f/src/main/scala/common/tile.scala#L109
https://github.com/ucb-bar/riscv-boom/blob/41cd1b8d55d98b26ff43e4fff3459747589acc8f/src/main/scala/system/BoomSubsystem.scala#L21

PR #1577, opened 2018-08-14
Adding diplomatic interfaces between RoCC and FPU

jieltan:
Also wrapping FPU with lazy wrapper. This will be a simple showcase for diplomacy outside of existing protocols


aswaterman:
Once this is done, it would be good to reorganize it into logically distinct commits; remove commented-out stuff; and remove unintentional whitespace changes.  That will make it much easier to code-review.

PR #1586, opened 2018-08-17
Converting HellaCacheIO into diplomatic interface

jieltan:
also wrapping PTW and Rocket with Lazy wrappers. the whole diplomacy package can be found in rocket/hellacache/, removed existing way of using listbuffers and vars to calculate the port width of HellaCacheArbiter. removed trait HasCoreIO since it was not reused anywhere else


jieltan:
@hcook Henry did you have a chance to look at this yet?

PR #1601, opened 2018-08-24
Remove riscv tools

terpstra:

**Impact**: API modification

**Release Notes**
Removed riscv-tools from rocket-chip

The software ecosystem is now mostly stable. There are work-arounds to avoid bringing the giant riscv-tools into the tree for most projects using rocket. This PR makes their life easier.


terpstra:
If you are talking about how we pull in riscv-tools for travis, that could be a follow-up commit. If you are talking about how this works for end-users, I think the answer should be: package management.

PR #1603, opened 2018-08-25
Fix recent regressions in master

terpstra:
This PR fixes two bugs:
- AsyncQueue: singleton's still need to synchronize counters on reset; removed by the recent refactor
- PhysicalFilter: describe in diplomacy the fact that Grant.toB is now a possibility


terpstra:
I've not written that part of the PR, yet. ;)

PR #1622, opened 2018-09-07
DecoupledHelper require Excluded Bool is actually contained

mehnadnerd:
Meant to address, https://github.com/freechipsproject/rocket-chip/issues/1616, written by Jack. Also helps to catch other errors that are made when using DecoupledHelper.

**Related issue**: 

**Release Notes**


aswaterman:
Thanks for making this component safer to use.

PR #1626, opened 2018-09-10
Add missing RVC and RVQ instruction opcodes to list

aswaterman:


solomatnikov:
Is there a distinction between `C_SLLI` and `C_SLLI_RV32` like between `SLLI` and `SLLI_RV32`?

The same question for `SRLI` and `SRAI`.

aswaterman:
Technically yes, but for implementations that explicitly expand RVC instructions into RVI instructions, there's no reason to distinguish them: the RVI decoder will generate the illegal-instruction trap.

But it wouldn't hurt to include them in this list... I'll add them.

PR #1637, opened 2018-09-22
Add HierarchicalNode

grebe:
**Related issue**: 

**Release Notes**

Diplomatic node assignments can be chained together. This has many use-cases, including many simple adapters being composed to make a complex adapter.

It would be useful sometimes to treat a chain of connected nodes as a single diplomatic node. For example, a crossbar with a fixed parameterization could be implemented as
```
 AdapterA := FixedXBar := AdapterB
```
Ideally, this chain would be encapsulated as a single node (say, XBar) so you can do both
```
XBar := rhs
lhs := XBar
```
However, in the current version of diplomacy you would need two nodes for the XBar: one for when it is the lhs of assignments and another when it is the rhs.

This commit adds a HierarchicalNode that encapsulates other nodes. A HierarchicalNode passes connect operations to the appropriate side of a chain of diplomatic nodes.

### DIFF
--- a/src/main/scala/diplomacy/Nodes.scala
+++ b/src/main/scala/diplomacy/Nodes.scala
@@ -569,3 +569,39 @@ class MixedTestNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data] protected[di
     dangles
   }
 }
+
+
+class MixedHierarchicalNode[DI, UI, EI, BI <: Data, DM, UM, EM, BM <: Data, DO, UO, EO, BO <: Data]
+(
+  lhs: MixedNode[DI, UI, EI, BI, DM, UM, EM, BM],
+  rhs: MixedNode[DM, UM, EM, BM, DO, UO, EO, BO]
+)(implicit valName: ValName) extends MixedNode(lhs.inner, rhs.outer) {
+  override def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
+    (0, 0)
+  }
+  override def mapParamsD(n: Int, p: Seq[DI]): Seq[DO] = {
+    rhs.mapParamsD(n, lhs.mapParamsD(n, p))
+  }
+  override def mapParamsU(n: Int, p: Seq[UO]): Seq[UI] = {
+    lhs.mapParamsU(n, rhs.mapParamsU(n, p))
+  }
+}
+
+class HierarchicalNode[D, U, E, B <: Data]
+(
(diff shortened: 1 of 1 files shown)


hcook:
@terpstra can you contrast this approach with using `NodeHandle`s?

terpstra:
Well, his example is exactly what NodeHandles are for:
```
val AdapterA = ...
val AdapterB = ...
val Xbar = ...

AdapterA := Xbar := AdapterB

val handle = NodeHandle(AdapterB, AdapterA)

handle := rhs
lhs := handle

Indeed, the := operator ALREADY supports automatic creation of a node handle:
AdapterA := Xbar := AdapterB
val handle = NodeHandle(AdapterB, AdapterA)
is the same as:
val handle = AdapterA := Xbar := AdapterB
```

terpstra:
Also, I believe the implementation of HierarchicalNode posted will not connect the wires of the two sides in the way IdentityNodes do.

grebe:
My reading of NodeHandle is that
```
val handle = AdapterA := XBar := AdapterB
// equivalent?
val handle = NodeHandle(XBar, AdapterB)
```
In other words, my understanding was that the NodeHandle always tracks adjacent pairs in a chain of assignments. Is that wrong? It's a bit hard for me to track through by reading. I think I was also thinking it was right associative (I don't think so now?) and that it even matter (it doesn't?).

If node handles satisfy my use case, that's great. If it works like this:

```
val handle = AdapterB := Xbar := AdapterC
AdapterA := handle := AdapterD
// equivalent
AdapterA := AdapterB := Xbar := AdapterC := AdapterD
```
then I don't think there's much of a reason to have what I'm proposing.

I'm not sure what you mean about not connecting the two wires- I don't think the HierarchicalNode should connect anything (unless it is in a `:=` statement). It just encapsulates other nodes.

terpstra:
I don't know if it's right- or left- associative. Since the operator in question IS associative, it does not matter. You are correct that NodeHandle takes the LHS and RHS and combines them together, but when you apply this repeatedly to a chain of connections that recursively becomes the left-most node and the right-most node.

grebe:
I see. I had tried using NodeHandles this way before but I had some problems- I must have misdiagnosed them. I've constructed an example for myself that replaces this new thing with NodeHandles and things are working as expected. Thanks!

jerryhethatday:
> I see. I had tried using NodeHandles this way before but I had some problems- I must have misdiagnosed them. I've constructed an example for myself that replaces this new thing with NodeHandles and things are working as expected. Thanks!

can you share your new examples? thanks

PR #1641, opened 2018-09-24
Remove riscv-tools, use debian (or other) tools instead

tmagik:
**Related issue**: 

**Release Notes**
Removes the 'riscv-tools' dependency for building verilog targets, which seems better handled by using either SiFive freedom-sdk tools, or distribution tools (debian riscv compilers)


aswaterman:
There is an open (but failed) PR trying to do the same from @terpstra: https://github.com/freechipsproject/rocket-chip/pull/1601

Not sure why that one's failing.  Either way, it would be great to finally do this.

PR #1648, opened 2018-09-28
More Rocket clock gating

aswaterman:
Support clock-gating the FPU when no FP instructions are in the pipeline.

Support clock-gating the integer pipeline during I$ misses and during WFI.


src/main/scala/rocket/Frontend.scala
```diff
@@ -306,6 +320,14 @@ class FrontendModule(outer: Frontend) extends LazyModuleImp(outer)
   io.cpu.perf.tlbMiss := io.ptw.req.fire()
   io.errors := icache.io.errors
 
+  // gate the clock
+  clock_en_reg := io.cpu.might_request || // chicken bit
```
jerryhethatday, line 324: what does chicken bit mean?

jerryhethatday: > what does chicken bit mean?

In case some rookie wants to know : https://lwn.net/Articles/744840/

PR #1652, opened 2018-10-02
Clock-gate the debug module using the dmactive signal

aswaterman:

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -3,6 +3,7 @@
 package freechips.rocketchip.devices.debug
 
 import Chisel._
+import chisel3.experimental._
 import freechips.rocketchip.config._
 import freechips.rocketchip.diplomacy._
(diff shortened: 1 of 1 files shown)


aswaterman:
@mwachs5 I've updated the PR to once again reset the DM state when dmactive falls.

One note of caution, I've added a config option to engage DM clock gating, but I've defaulted it to on.  I think that's what we want in general, just keep in mind if trying to be conservative.

aswaterman:
Thanks for the feedback; I addressed it.

mwachs5:
See https://github.com/riscv/riscv-debug-spec/issues/395

What happens if DEBUGGER tries to read a debug module register when dmactive=0? It seems that this would not get through the Clock Domain Crossing into the DebugModuleInner, yet, the debug tests are passing in regression... I'm not sure how this is working

aswaterman:
Because we are still clocking the DMI crossing.  (Whether we should be doing that is up for debate.)

PR #1654, opened 2018-10-02
Better HintHandler

terpstra:
**Impact**: functional change

**Release Notes**
The old HintHandler was buggy because it reordered responses despite claiming that it would not. It was also grossly inefficient due to the use of Arbiters and Queues. This new HintHandler transforms Hints into empty PutPartials, avoiding the need to recombine the response stream. It can be placed between a device and it's Fragmenter as a very cheap stateless adapter, or higher up the tree like the old HintHandler, in which case it will use flops to hold the address steady.

### DIFF
--- a/src/main/scala/tilelink/Arbiter.scala
+++ b/src/main/scala/tilelink/Arbiter.scala
@@ -42,6 +42,8 @@ object TLArbiter
   def apply[T <: Data](policy: Policy)(sink: DecoupledIO[T], sources: (UInt, DecoupledIO[T])*) {
     if (sources.isEmpty) {
       sink.valid := Bool(false)
+    } else if (sources.size == 1) {
+      sink <> sources.head._2
     } else {
       val pairs = sources.toList
       val beatsIn = pairs.map(_._1)
@@ -77,15 +79,10 @@ object TLArbiter
       val muxState = Mux(idle, winner, state)
       state := muxState
 
-      if (sources.size > 1) {
-        val allowed = Mux(idle, readys, state)
-        (sourcesIn zip allowed) foreach { case (s, r) =>
-          s.ready := sink.ready && r
-        }
-      } else {
-        sourcesIn(0).ready := sink.ready
+      val allowed = Mux(idle, readys, state)
+      (sourcesIn zip allowed) foreach { case (s, r) =>
+        s.ready := sink.ready && r
       }
-
       sink.valid := Mux(idle, valids.reduce(_||_), Mux1H(state, valids))
       sink.bits := Mux1H(muxState, sourcesIn.map(_.bits))
     }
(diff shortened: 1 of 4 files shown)


terpstra:
There is of course a problem with this approach. It does not add Hint support to ROMs. :-/

aswaterman:
Isn't it trivial to add Hint support to ROMs manually?

terpstra:
It's trivial to add Hint support to any slave manually. That doesn't mean we want to hand-write it everywhere.

It seems to me that there are 3 possible ways to implement a Hint adapter:
1. conversion to PutPartial (mask=0)
2. conversion to Get (if device has no side-effects)
3. redirected to an error device which is kept ordered via the interconnect.

The third option is the most robust, but also most expensive. The old HintHandler was a buggy version of this option. I think this adapter could easily support both options 1 and 2 at the same time.

solomatnikov:
If ROM acks PutPartial w/ mask==0, then option 1 works.

terpstra:
We don't want to declare that ROMs support PutPartial as then the core will not trap on writes to them.

solomatnikov:
ROMs shouldn't declare support PutPartial, however, they can treat tPartial w/ mask==0 as NOP.

terpstra:
I think the only components in our designs where neither option 1 nor option 2 apply are AHB-connected MMIO devices. However, we've been planning to fix AHB somehow in the future, anyway.

terpstra:
If you don't declare support for PutPartial, it does not matter whether you handle some PutPartials properly. It is illegal to send you those PutPartials and the monitors will assert.

aswaterman:
The cores will trap if PutFull is not supported.  PutPartial doesn't matter.

terpstra:
@aswaterman fair enough, but support for PutPartial requires support of PutFull.

aswaterman:
true

solomatnikov:
Monitors can ignore tPartial w/ mask==0 as NOP

terpstra:
We would have to add a new category of declared support for your proposal. A "supportsPutPartial0". The attendant complexity of correctly setting 3 types of Put support affecting every slave does not seem a good trade-off to me.

terpstra:
This revised version can now translate to either Get or PutPartial based on whatever is safe for the target device. The hardware cost is quite minimal even in the dynamic case. In the degenerate case (no multi-beat Hints), the hardware cost dwindles to almost nothing.

My plan to solve the AHB problem (GetEffects + !PutPartial) is to just make the AHB converter support Hints directly. That adapter is somewhat exceptional already.

aswaterman:
Sounds good!

Separately, we might eventually want to teach the AHB converter about locked atomics, in which case eventual PutPartial support might fold into that effort.

terpstra:
I will re-open this PR with AHB hint support attached.

PR #1662, opened 2018-10-09
Off-chip coherence

terpstra:
**Release Notes**
Added RegionReplicator, which presents the same physical devices at multiple addresses.
Added the AddressAdjuster, which can select to which physical address route local traffic is routed.


src/main/scala/devices/tilelink/BusBypass.scala
```diff
@@ -120,30 +88,15 @@ class TLBusBypassBar(dFn: TLManagerPortParameters => TLManagerPortParameters)(im
     out0.d.ready := in.d.ready &&  bypass
     out1.d.ready := in.d.ready && !bypass
     in.d.valid   := Mux(bypass, out0.d.valid, out1.d.valid)
+    def cast(x: TLBundleD) = { val out = Wire(in.d.bits); out <> x; out }
```
hcook: We should turn this into a feature instead of a hack (cc/ @jackkoenig)

PR #1664, opened 2018-10-11
Banked TileLink

terpstra:
**Impact**: API modification
**Release Notes**
This PR adds the BankBinder and ProbePicker which can be combined with an enhanced TLXbar to create a multi-bank interconnect between cache layers.

This removes memBuses and replaces it with a singular 'mbus'. This affects memory bus attachments in other repositories. The change is necessary because now the singular Xbar in the mbus now links multi-channel memory with variable bank-count masters.


aswaterman:
Don’t have time to CR, but this is cool! I’ll rubber-stamp, but hopefully @hcook can review

PR #1671, opened 2018-10-20
Power queues

terpstra:
**Impact**: API addition
**Release Notes**
Added MultiPort and MultiLane Queues for use in packet processing applications.


src/main/scala/util/HellaQueue.scala
```diff
@@ -26,6 +26,8 @@ class HellaFlowQueue[T <: Data](val entries: Int)(data: => T) extends Module {
   val ram = SeqMem(entries, data)
   when (do_enq) { ram.write(enq_ptr, io.enq.bits) }
 
+  // BUG! does not hold the output of the SRAM when !ready
+  // ... However, HellaQueue is correct due to the pipe stage
```
hcook: Maybe make this class constructor `private`?

terpstra: If we go that route, we might as well just completely remove HellaQueue, since LanePositionedMultiQueue can do the same thing HellaQueue does with 1RW instead of 1R1W ports. I didn't want to take that radical approach immediately. ;)

hcook: Can we turn HellaFlowQueue/HellaQueue into thin wrappers that instantiate correctly parameterized LanePositionedMultiQueues?

terpstra:
I've just rebased this against master. I intend to merge it shortly.

PR #1672, opened 2018-10-23
Add WidthWidget adapter to toTile

rmac-sifive:

**Impact**: API modification

**Release Notes**
Adding TLWidthWidget to toTile, since current API assumes identical widths between master/slave


terpstra:
I'm a bit confused. Is this for slave devices that live in the tile?

rmac-sifive:
This is for connecting to the cbus

PR #1675, opened 2018-10-26
Handle subsystem bus interconnectivity in child classes of BaseSubsystem

hcook:
`BaseSubsystem` no longer mandates how its tilelink buses are to be interconnected to one another. Creating these connections are now the responsibility of child classes, e.g. `ExampleRocketSystem`, `GroundtestSubsystem`. I also added some helper functions for cross-connecting the buses.


hcook:
Ugh I merged this into the wrong branch

PR #1713, opened 2018-11-20
Fix non-determinism in TLMonitor generation

jackkoenig:
groupBy considered harmful

**Related issue**: 

**Type of change**: bug fix

**Impact**: no functional change / large impact on generated code order

**Release Notes**

Fix non-determinism in TLMonitor generation

### DIFF
--- a/src/main/scala/tilelink/Parameters.scala
+++ b/src/main/scala/tilelink/Parameters.scala
@@ -161,9 +161,11 @@ case class TLManagerPortParameters(
       lgSize:  UInt,
       range:   Option[TransferSizes]): Bool = {
     def trim(x: TransferSizes) = range.map(_.intersect(x)).getOrElse(x)
+    // groupBy returns an unordered map, convert back to Seq and sort the result for determinism
     val supportCases = managers.groupBy(m => trim(member(m))).mapValues(_.flatMap(_.address))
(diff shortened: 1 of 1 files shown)


jackkoenig:
*note* This has a large impact on the order, perhaps worth playing with to have smaller diff

src/main/scala/tilelink/Parameters.scala
```diff
@@ -161,9 +161,11 @@ case class TLManagerPortParameters(
       lgSize:  UInt,
       range:   Option[TransferSizes]): Bool = {
     def trim(x: TransferSizes) = range.map(_.intersect(x)).getOrElse(x)
+    // groupBy returns an unordered map, convert back to Seq and sort the result for determinism
     val supportCases = managers.groupBy(m => trim(member(m))).mapValues(_.flatMap(_.address))
```
terpstra: Why not just ask it to return a SortedMap instead?

jackkoenig: There's no implicit ordering for TransferSizes so I thought it easier just to go to Seq and then use `.sortWith`. Further more, supportCases is not actually used as a Map so I thought it more clear to make it a Seq

terpstra: Ok, fine.

PR #1715, opened 2018-11-20
Added hart array feature

ernie-sifive:

**Release Notes**
Implemented hart array feature as detailed in the RISC-V Debug Spec, v 0.13

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -92,7 +92,7 @@ import DebugAbstractCommandType._
   *  hasBusMaster: Whether or not a bus master should be included
   *  maxSupportedSBAccess: Maximum transaction size supported by System Bus Access logic.
   *  supportQuickAccess : Whether or not to support the quick access command.
-  *  supportHartArray : Whether or not to implement the hart array register.
+  *  supportHartArray : Whether or not to implement the hart array register (if >1 hart).
   *  hasImplicitEbreak: There is an additional RO program buffer word containing an ebreak
   **/
 case class DebugModuleParams (
@@ -103,7 +103,7 @@ case class DebugModuleParams (
   hasBusMaster : Boolean = false,
   maxSupportedSBAccess : Int = 32,
   supportQuickAccess : Boolean = false,
-  supportHartArray   : Boolean = false,
+  supportHartArray   : Boolean = true,
   hasImplicitEbreak : Boolean = false
 ) {
 
@@ -151,10 +151,12 @@ case object DebugModuleHartSelKey extends Field(DebugModuleHartSelFuncs())
 /* structure for passing hartsel between the "Outer" and "Inner"
  */
 
-class DebugInternalBundle ()(implicit val p: Parameters) extends ParameterizedBundle()(p) {
+class DebugInternalBundle (val nComponents: Int)(implicit val p: Parameters) extends ParameterizedBundle()(p) {
   val resumereq    = Bool()
   val hartsel      = UInt(10.W)
   val ackhavereset = Bool()
+  val hasel        = Bool()
+  val hamask       = Vec(nComponents, Bool())
 }
 
 /* structure for top-level Debug Module signals which aren't the bus interfaces.
@@ -187,8 +189,8 @@ class DebugCtrlBundle (nComponents: Int)(implicit val p: Parameters) extends Par
   *  Debug ROM. It runs partially off of the dmiClk (e.g. TCK) and
   *  the TL clock. Therefore, it is divided into "Outer" portion (running
   *  of off dmiClock and dmiReset) and "Inner" (running off tlClock and tlReset).
-  *  This allows DMCONTROL.haltreq, hartsel, dmactive, and ndreset to be
-  *  modified even while the Core is in reset or not being clocked. 
+  *  This allows DMCONTROL.haltreq, hartsel, hasel, hawindowsel, hawindow, dmactive,
+  *  and ndreset to be modified even while the Core is in reset or not being clocked.
   *  Not all reads from the Debugger to the Debug Module will actually complete
   *  in these scenarios either, they will just block until tlClock and tlReset
   *  allow them to complete. This is not strictly necessary for 
@@ -226,13 +228,17 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
   // For Shorter Register Names
   import DMI_RegAddrs._
 
+  val cfg = p(DebugModuleParams)
+
   val intnode = IntNexusNode(
     sourceFn       = { _ => IntSourcePortParameters(Seq(IntSourceParameters(1, Seq(Resource(device, "int"))))) },
     sinkFn         = { _ => IntSinkPortParameters(Seq(IntSinkParameters())) },
     outputRequiresInput = false)
 
   val dmiNode = TLRegisterNode (
-    address = AddressSet.misaligned(DMI_DMCONTROL << 2, 4),
+    address = AddressSet.misaligned(DMI_DMCONTROL   << 2, 4) ++
+              AddressSet.misaligned(DMI_HAWINDOWSEL << 2, 4) ++
+              AddressSet.misaligned(DMI_HAWINDOW    << 2, 4),
     device = device,
     beatBytes = 4,
     executable = false
@@ -242,10 +248,11 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
     require (intnode.edges.in.size == 0, "Debug Module does not accept interrupts")
 
     val nComponents = intnode.out.size
+    val supportHartArray = cfg.supportHartArray && (nComponents > 1)    // no hart array if only one hart
 
     val io = IO(new Bundle {
       val ctrl = (new DebugCtrlBundle(nComponents))
-      val innerCtrl = new DecoupledIO(new DebugInternalBundle())
+      val innerCtrl = new DecoupledIO(new DebugInternalBundle(nComponents))
     })
 
     //----DMCONTROL (The whole point of 'Outer' is to maintain this register on dmiClock (e.g. TCK) domain, so that it
@@ -281,6 +288,7 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
         DMCONTROLNxt.haltreq      := DMCONTROLWrData.haltreq
         DMCONTROLNxt.resumereq    := DMCONTROLWrData.resumereq
         DMCONTROLNxt.ackhavereset := DMCONTROLWrData.ackhavereset
+        DMCONTROLNxt.hasel        := (if (supportHartArray) DMCONTROLWrData.hasel else false.B)
       }
     }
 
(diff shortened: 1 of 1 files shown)


src/main/scala/devices/debug/Debug.scala
```diff
@@ -187,7 +189,7 @@ class DebugCtrlBundle (nComponents: Int)(implicit val p: Parameters) extends Par
   *  Debug ROM. It runs partially off of the dmiClk (e.g. TCK) and
   *  the TL clock. Therefore, it is divided into "Outer" portion (running
   *  of off dmiClock and dmiReset) and "Inner" (running off tlClock and tlReset).
-  *  This allows DMCONTROL.haltreq, hartsel, dmactive, and ndreset to be
+  *  This allows DMCONTROL.haltreq, hartsel, hasel, dmactive, and ndreset to be
```
mwachs5: what about hawindowsel and hawindow?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -151,10 +151,12 @@ case object DebugModuleHartSelKey extends Field(DebugModuleHartSelFuncs())
 /* structure for passing hartsel between the "Outer" and "Inner"
  */
 
-class DebugInternalBundle ()(implicit val p: Parameters) extends ParameterizedBundle()(p) {
+class DebugInternalBundle (val nComponents: Int)(implicit val p: Parameters) extends ParameterizedBundle()(p) {
   val resumereq    = Bool()
   val hartsel      = UInt(10.W)
   val ackhavereset = Bool()
+  val hasel        = Bool()
+  val hamask       = Vec(nComponents, Bool())
```
mwachs5: Hmm... this is all going across a not-so-clean asynchronous crossing. Perhaps it makes sense to shadow the HASEL and HAMASK registers in the proper domain vs doing it this way? Or, to just declare that these will be static so that they don't need to be synchronized?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -281,6 +288,7 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
         DMCONTROLNxt.haltreq      := DMCONTROLWrData.haltreq
         DMCONTROLNxt.resumereq    := DMCONTROLWrData.resumereq
         DMCONTROLNxt.ackhavereset := DMCONTROLWrData.ackhavereset
+        DMCONTROLNxt.hasel        := supportHartArray.B & DMCONTROLWrData.hasel
```
mwachs5: can we keep the logic in the chisel domain, vs generating logic for things we know at elaboration time? I'd write this as:
```
DMCONTROLNxt.hasel := if supportHartArray DMCONTROLWrData.hasel else false.B
```

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+      val HAWINDOWSELNxt = Wire(init = (new HAWINDOWSELFields().fromBits(0.U)))
+      val HAWINDOWSELReg = Wire(init = new HAWINDOWSELFields().fromBits(AsyncResetReg(updateData = HAWINDOWSELNxt.asUInt,
+        resetData = 0,
+        enable = true.B,
+        name = "HAWINDOWSEL"
+      )))
+
+      HAWINDOWSELRdData := Wire(init = HAWINDOWSELReg)
+      HAWINDOWSELNxt := HAWINDOWSELReg
+      when (~dmactive | ~supportHartArray.B) {
```
mwachs5: Why do we need to check for `supportHartArray` here? We already have `if(supportHartArray)` above so none of this logic should even be generated in this case

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+          } else {
+            HAWINDOWSELNxt.hawindowsel := 0.U
+          }
+        }
+      }
+
+      val numHAMASKSlices = ((nComponents - 1)/32)+1
+      HAWINDOWRdData.maskdata := 0.U     // default, overridden below
+      for (ii <- 0 until numHAMASKSlices) {
+        val SliceMask = if (nComponents > ii*32 + 31) 0xFFFFFFFF  // All harts in this slice exist
```
mwachs5: `SliceMask` -> `sliceMask` (scalastyle nit). Actually all these all-caps variables are very bad practice, but understand you're just following what's there ;)

mwachs5: can you put more parenthesis in the `nComponents > i*32 + 31`... I don't really know the precedence of `>` over `+` for all languages by heart

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+                        else (1<<(nComponents - ii*32)) - 1       // Partial last slice
+        val HAMASKRst = Wire(init = (new HAWINDOWFields().fromBits(0.U)))
+        val HAMASKNxt = Wire(init = (new HAWINDOWFields().fromBits(0.U)))
+        val HAMASKReg = Wire(init = Vec(AsyncResetReg(updateData = HAMASKNxt.asUInt,
+          resetData = 0,
+          enable = true.B,
+          name = "HAMASK")))
```
mwachs5: this will make a bunch of registers called "HAMASK" which then will rely on Chisel/FIRRTL to give them different names. Can you add `ii` to the name, ie `name = s"HMASKReg${ii}"`

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+    val HAWINDOWWrEn   = Wire(init = false.B)
+    val HAWINDOWRdEn   = Wire(init = false.B)
+
+    if (supportHartArray) {
+      val HAWINDOWSELReset = Wire(init = (new HAWINDOWSELFields().fromBits(0.U)))
+      val HAWINDOWSELNxt = Wire(init = (new HAWINDOWSELFields().fromBits(0.U)))
+      val HAWINDOWSELReg = Wire(init = new HAWINDOWSELFields().fromBits(AsyncResetReg(updateData = HAWINDOWSELNxt.asUInt,
+        resetData = 0,
+        enable = true.B,
+        name = "HAWINDOWSEL"
```
mwachs5: I forget if this macro automatically adds the 'Reg' at the end, i guess you are probably just following the pattern

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+        when (~dmactive) {
+          HAMASKNxt := HAMASKRst
+        }.otherwise {
+          when (HAWINDOWWrEn && (ii.U === HAWINDOWSELReg.hawindowsel)) {
+            HAMASKNxt.maskdata := HAWINDOWWrData.maskdata
+          }
+        }
+
+         // drive each slice of hamask with HAMASKReg or with new value being written
+        for (jj <- 0 until 32) {
```
mwachs5: `32` is a magic value in all this code. Can we come up with a variable and use it all the relevant places so the code is more clear? `maxHASlices`?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+          if (ii*32 + jj < nComponents) {
```
mwachs5: more parens here please

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+            hamask(ii*32 + jj) := Mux(HAWINDOWWrEn && (ii.U === HAWINDOWSELReg.hawindowsel),
+                                      (HAWINDOWWrData.maskdata >> jj) & 1.U,
```
mwachs5: You can do this more clearly by doing `HAWINDOWWrData.maskData.toBools(jj)`

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+                                      (HAMASKReg.asUInt >> jj) & 1.U)
```
mwachs5: same here, to me this is more clear as `HAMASKReg.asUInt.toBools(jj)`

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,11 +297,110 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+    // DMI register decoder for Outer
+    //--------------------------------------------------------------
+      // regmap addresses are byte offsets from lowest address
+    def DMI_DMCONTROL_OFFSET   = 0
+    def DMI_HAWINDOWSEL_OFFSET = ((DMI_HAWINDOWSEL - DMI_DMCONTROL) << 2)
+    def DMI_HAWINDOW_OFFSET    = ((DMI_HAWINDOW - DMI_DMCONTROL) << 2)
+
+    if (supportHartArray) {
+      dmiNode.regmap(
+        DMI_DMCONTROL_OFFSET   -> Seq(RWNotify(32, DMCONTROLRdData.asUInt(),
```
mwachs5: can you build up this sequence separately, and conditionally add to it, so that you don't have to copy the code for DMI_DMCONTROL_OFFSET. Then you can just call `dmiNode.regmap(prebuild_sequence)`

src/main/scala/devices/debug/Debug.scala
```diff
@@ -324,16 +431,20 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
       when (~dmactive) {
         debugIntNxt(component) := false.B
       }. otherwise {
-        when (DMCONTROLWrEn && DMCONTROLWrData.hartsello === component.U) {
+        when (DMCONTROLWrEn && ((DMCONTROLWrData.hartsello === component.U) || (supportHartArray.B && DMCONTROLWrData.hasel && hamask(component)))) {
```
mwachs5: (Review ends here, will review the rest in another pass)

src/main/scala/devices/debug/Debug.scala
ernie-sifive: Both of these new items are needed in both domains.  They can't be static because a change to hamask will affect the read values of fields in dmstatus.  hartsel, hasel, and hamask all need to be transmitted from Outer to Inner whenever any one of them changes.

mwachs5: yeah, makes sense. Just wanted to be sure you understood how this bundle was used

src/main/scala/devices/debug/Debug.scala
ernie-sifive: I couldn't (and still can't) get something like this (where scala is embedded within a line of chisel) to compile.  The logic is removed eventually since supportHartArray is a constant, but I agree not generating any logic in the first place would be preferable.  Any suggestions?

src/main/scala/devices/debug/Debug.scala
ernie-sifive:
```
[error] /scratch/erniee/federation/rocket-chip/src/main/scala/devices/debug/Debug.scala:375:63: overloaded method value toBools with alternatives:
[error]   (implicit sourceInfo: chisel3.internal.sourceinfo.SourceInfo,implicit compileOptions: chisel3.core.CompileOptions)Seq[chisel3.core.Bool] <and>
[error]   ()Seq[chisel3.core.Bool]
[error]  cannot be applied to (Int)
[error]                                       HAWINDOWWrData.maskdata.toBools(jj),
[error]                                                               ^

```

src/main/scala/devices/debug/Debug.scala
ernie-sifive: I could use some help with the code for this suggestion...

mwachs5: Maybe simpler, you can always have the offset but put the Seq of RegField there or not:

`DMI_HAWINDOWSEL_OFFSET -> (if (supportsHartArray) Seq(...) else Nil)`

src/main/scala/devices/debug/Debug.scala
ernie-sifive:
```
[error] /scratch/erniee/federation/rocket-chip/src/main/scala/devices/debug/Debug.scala:291:38: illegal start of simple expression
[error]         DMCONTROLNxt.hasel        := if supportHartArray DMCONTROLWrData.hasel else false.B
[error]                                      ^
[error] one error found

```

mwachs5: try throwing parentheses around it... `DMCONTROLNxt.hasel := (if ...)`

mwachs5: Oh, also maybe `if (supportsHartArray) ... `

ernie-sifive: Parens around the entire if statement worked...makes sense!
`DMCONTROLNxt.hasel        := (if (supportHartArray) DMCONTROLWrData.hasel else false.B)`

src/main/scala/devices/debug/Debug.scala
```diff
@@ -476,32 +608,32 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
...
-    DMSTATUSRdData.allhavereset := haveResetBitRegs(selectedHartReg)
-    DMSTATUSRdData.anyhavereset := haveResetBitRegs(selectedHartReg)
-
     val resumereq = io.innerCtrl.fire() && io.innerCtrl.bits.resumereq
 
-    when (io.innerCtrl.fire()){
-      when (io.innerCtrl.bits.ackhavereset) {
-        haveResetBitRegs(io.innerCtrl.bits.hartsel) := false.B
+    DMSTATUSRdData.anynonexistent := (selectedHartReg >= nComponents.U)   // only hartsel can be nonexistent
+    DMSTATUSRdData.allnonexistent := (selectedHartReg >= nComponents.U) & ~hamaskFull.reduce(_ | _)
```
mwachs5: can you put a parens around the `~` and the `reduce`? I'm not even quite sure the order you mean here

src/main/scala/devices/debug/Debug.scala
ernie-sifive: Still no luck with this one...the toBools part seems to be working but the problem seems to be the element selector (jj).  The error seems to be saying that a Vec of Bool "cannot be applied" to an Int (meaning jj).  Parens suggested don't change anything.

ernie-sifive: Resolved as suggested in Slack - assign .toBools first to a temp, then index from that.

ernie-sifive:
@mwachs, I think I've addressed all your comments so far.  Take another quick look?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -289,10 +297,103 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+            HAMASKNxt.maskdata := HAWINDOWWrData.maskdata
+          }
+        }
+
+         // drive each slice of hamask with stored HAMASKReg or with new value being written
+        for (jj <- 0 until haWindowSize) {
+          if (((ii*haWindowSize) + jj) < nComponents) {
+            val tempWrData = HAWINDOWWrData.maskdata.toBools
+            val tempMaskReg = HAMASKReg.asUInt.toBools
+            hamask((ii*haWindowSize) + jj) := Mux(HAWINDOWWrEn && (ii.U === HAWINDOWSELReg.hawindowsel), tempWrData(jj), tempMaskReg(jj))
```
mwachs5: this is fine/correct, but you could perhaps more clearly write it with a `when`:

```
when (HAWINDOWWrEn && (ii.U ...) {
  hamask((ii*haWindowSize) + jj) := tempWrData(jj)
}
```

src/main/scala/devices/debug/Debug.scala
```diff
@@ -324,16 +425,20 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
       when (~dmactive) {
         debugIntNxt(component) := false.B
       }. otherwise {
-        when (DMCONTROLWrEn && DMCONTROLWrData.hartsello === component.U) {
+        when (DMCONTROLWrEn && ((DMCONTROLWrData.hartsello === component.U) || (supportHartArray.B && DMCONTROLWrData.hasel && hamask(component)))) {
```
mwachs5: It looks like you pulled out the `supportHartArray.B ... ` clause elsewhere, at least I don't see it anymore. Is there a reason why this one is still here?

ernie-sifive:
@mwachs5, any further comments?  Ok to merge?

mwachs5:
LGTM. If you wanted to be super-sure you could test this PR on the master regression with test-long-manual-trigger

mwachs5:
Ah, becasue you already merged this :)

PR #1728, opened 2018-11-29
Tile Notification Nodes

hcook:
Various changes to the Tile APIs related to notifying external observers of tile status. Tile notifications are expressed as outward-going interrupts, but are handled differently (not routed to PLIC, not clock crossed). The nodes emitting these interrupts unconditionally exist, but their wire values may be driven low for tile configurations for which no such notification could ever be emitted.

The three current types of notifications are:
- `halt`: an unrecoverable error occurred (e.g. uncorrectable ECC error in a cache)
- `cease`: the tile has temporarily ceased to retire instructions
- `wfi`: the tile is waiting to receive an external interrupt

Other miscellaneous API changes to tile parameterization:
- the BEU is now in package `tile` and is enabled via a `TileParam` called `beuAddr`
- tiles unconditionally have output wires for their traced instructions; these don't have have to be connected to anything

Also:
- IntXbar factory helper

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/model/OMCore.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMCore.scala
@@ -9,7 +9,6 @@ trait OMCore extends OMComponent{
   def pmp: Option[OMPMP]
   def documentationName: String
   def hartIds: Seq[Int]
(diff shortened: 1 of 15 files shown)


aswaterman:
Cool

PR #1733, opened 2018-12-03
adding dcache,icache,rocket

depappas-sifive:
Adding object models for dcache, icache, and rocket

feature request

API addition (no impact on existing code)

implementation

**Release Notes**


src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,111 @@ class RocketTile(
...
+          nRegions = coreParams.nPMPs,
+          granularity = coreParams.pmpGranularity
+        ))
+      }
+      else { None }
+
+      val mulDiv = coreParams.mulDiv.map{ md => OMMulDiv.makeOMI(md, xLen)}
+
+      val baseInstructionSet = xLen match {
+        case 32 => if (XLen == 32) RV32E else RV32I // TODO coreParams.useRVE
```
richardxia: Why is this here? Rocket doesn't support RVE.

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,111 @@ class RocketTile(
...
+        xLen = xLen,
+        baseSpecification = baseSpec(baseInstructionSet, baseISAVersion),
+        base = baseInstructionSet,
+        m = coreParams.mulDiv.map { x => isaExtSpec(M, "2.0") },
+        a = coreParams.useAtomics.option(isaExtSpec(A, "2.0")),
+        f = coreParams.fpu.map { x => isaExtSpec(F, "2.0") },
+        d = d,
+        c = coreParams.useCompressed.option(isaExtSpec(C," 2.0")),
+        u = coreParams.useUser.option(isaExtSpec(U,"1.10")),
+        s = None,
```
richardxia: This isn't correct. A Rocket core may support S mode depending on the parameters passed in.

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,111 @@ class RocketTile(
...
+        addressTranslationModes = Nil
```
richardxia: Similarly, it will support certain address translation modes if S mode is enabled.

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,111 @@ class RocketTile(
...
+      )
+
+      val btb = rocketParams.btb.map(BTB.makeOMI(_))
+
+      val omICache = frontend.icache.device.getOMComponents(resourceBindingsMap).headOption.getOrElse(None).asInstanceOf[OMICache]
```
richardxia: I'm not sure I understand what's going on here. What does it mean to `.getOrElse(None)`? That ends up returning either an `A` or a `None` rather than a `Some[A]`, which is a very odd type to return.

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,111 @@ class RocketTile(
...
+
+      def dcache(p: DCacheParams): OMDCache = {
+        OMDCache(
+          memoryRegions = Nil,
```
richardxia: If the DCache is configured as a DTIM, then you should populate the memory regions here.

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,111 @@ class RocketTile(
...
+      Seq(OMRocketCore(
+        isa = omIsa,
+        mulDiv = mulDiv,
+        fpu = fpu,
+        performanceMonitor = perfMon,
+        pmp = pmp,
+        documentationName = "TODO",
+        hartIds = Seq(hartId),
+        hasVectoredInterrupts = true,
+        interruptLatency = 6,
```
richardxia: Where did this number come from and is it even correct?

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,111 @@ class RocketTile(
...
+        nLocalInterrupts = coreParams.nLocalInterrupts,
+        nBreakpoints = coreParams.nBreakpoints,
+        branchPredictor = btb,
+        dcache = omDCache,
+        icache = Some(omICache)
```
richardxia: Why is the ICache a Option but the DCache is not?

src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -53,6 +54,15 @@ trait HasRocketTiles extends HasTiles
 
     rocket
   }
+
+  abstract override def getSubsystemOMComponents(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
+    val rockets = getOMRocketCores(resourceBindingsMap)
+    val plics = plicOpt.map(_.device.getOMComponents(resourceBindingsMap))
+    val clints = clintOpt.map(_.device.getOMComponents(resourceBindingsMap)) // super.getSubsystemOMComponents(resourceBindingsMap)
```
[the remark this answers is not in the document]
depappas-sifive: No

src/main/scala/diplomaticobjectmodel/model/OMCaches.scala
```diff
@@ -2,6 +2,10 @@
 
 package freechips.rocketchip.diplomaticobjectmodel.model
 
+import freechips.rocketchip.diplomacy.ResourceBindings
+import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelAddressing
+import freechips.rocketchip.rocket.{DCacheParams, ICacheParams}
+
```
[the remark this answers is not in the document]
depappas-sifive: removed

src/main/scala/rocket/ICache.scala
```diff
@@ -54,7 +56,34 @@ class ICache(val icacheParams: ICacheParams, val hartId: Int)(implicit p: Parame
...
   val size = icacheParams.nSets * icacheParams.nWays * icacheParams.blockBytes
-  val device = new SimpleDevice("itim", Seq("sifive,itim0"))
+  val device = new SimpleDevice("itim", Seq("sifive,itim0")) {
+    override def getOMComponents(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
+      require(resourceBindingsMap.map.contains(this))
+      val resourceBindings = resourceBindingsMap.map.get(this)
+      resourceBindings.map {
+        rb => icache(icacheParams, rb)
+      }
+    }.getOrElse(throw new NoSuchElementException)
```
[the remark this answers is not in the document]
depappas-sifive: moved inside of the braces

src/main/scala/rocket/ICache.scala
```diff
@@ -54,7 +56,34 @@ class ICache(val icacheParams: ICacheParams, val hartId: Int)(implicit p: Parame
...
+  }
+
+  def icache(p: ICacheParams, resourceBindings: ResourceBindings): Seq[OMComponent] = {
```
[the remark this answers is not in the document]
depappas-sifive: renamed

src/main/scala/rocket/ICache.scala
```diff
@@ -54,7 +56,34 @@ class ICache(val icacheParams: ICacheParams, val hartId: Int)(implicit p: Parame
...
+    val regions = DiplomaticObjectModelAddressing.getOMMemoryRegions("ICache", resourceBindings)
+    Seq[OMComponent](
+      OMICache(
+        memoryRegions = DiplomaticObjectModelAddressing.getOMMemoryRegions("ICache", resourceBindings),
```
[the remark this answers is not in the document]
depappas-sifive: removed

src/main/scala/rocket/ICache.scala
```diff
@@ -54,7 +56,34 @@ class ICache(val icacheParams: ICacheParams, val hartId: Int)(implicit p: Parame
...
+        interrupts = Nil,
+        nSets = p.nSets,
+        nWays = p.nWays,
+        blockSizeBytes = p.blockBytes,
+        dataMemorySizeBytes = p.nSets * p.nWays * p.blockBytes,
+        dataECC = p.dataECC.map{case code => OMECC.getCode(code)},
+        tagECC = p.tagECC.map{case code => OMECC.getCode(code)},
```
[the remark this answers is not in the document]
depappas-sifive: fixed

src/main/scala/tile/RocketTile.scala
depappas-sifive: case class OMRocketCore(
  dcache: Option[OMDCache],
  icache: Option[OMICache]

src/main/scala/tile/RocketTile.scala
depappas-sifive: comment removed

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,111 @@ class RocketTile(
...
+      val icache = getOMICacheFromBindings(resourceBindingsMap)
+      cores ++: icache
+    }
+    def getOMICacheFromBindings(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
+      frontend.icache.device.getOMComponents(resourceBindingsMap)
+    }
+    def getOMRocketCores(resourceBindingsMap: ResourceBindingsMap): Seq[OMRocketCore] = {
+      val coreParams = rocketParams.core
+
+      val fpu= coreParams.fpu.map{f => OMFPU(fLen = f.fLen)}
```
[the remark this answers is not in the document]
depappas-sifive: added.

src/main/scala/tile/RocketTile.scala
depappas-sifive: @hcook says we should remove the interruptLatency number from the docs.

src/main/scala/tile/RocketTile.scala
depappas-sifive: fixed

src/main/scala/tile/RocketTile.scala
depappas-sifive: I was told by @hcook and @jsmithsf to use useVM to condition the S mode.

src/main/scala/rocket/ICache.scala
```diff
@@ -54,7 +56,33 @@ class ICache(val icacheParams: ICacheParams, val hartId: Int)(implicit p: Parame
...
+    override def getOMComponents(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
+      require(resourceBindingsMap.map.contains(this))
+      val resourceBindings = resourceBindingsMap.map.get(this)
+      resourceBindings.map {
+        rb => makeOMI(icacheParams, rb)
+      }.getOrElse(throw new NoSuchElementException)
+    }
+  }
+
+  def makeOMI(p: ICacheParams, resourceBindings: ResourceBindings): Seq[OMComponent] = {
```
hcook: What is the rationale behind making this a separate function when it is only called once?

hcook: Maybe this should be handled like the BTB, where it is a method on a helper object in the `diplomaticobjectmodel.model` package? I like that pattern

src/main/scala/diplomaticobjectmodel/model/OMRocketCore.scala
```diff
@@ -26,3 +28,13 @@ case class OMRocketCore(
   icache: Option[OMICache],
   _types: Seq[String] = Seq("OMRocketCore", "OMCore", "OMComponent", "OMCompoundType")
 ) extends OMCore
+
+object BTB {
```
hcook: OMBTB?

src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -53,6 +54,15 @@ trait HasRocketTiles extends HasTiles
...
     rocket
   }
+
+  abstract override def getSubsystemOMComponents(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
+    val rockets = getOMRocketCores(resourceBindingsMap)
+    val plics = plicOpt.map(_.device.getOMComponents(resourceBindingsMap))
+    val clints = clintOpt.map(_.device.getOMComponents(resourceBindingsMap))
+    rockets ++ plics.getOrElse(Nil) ++ clints.getOrElse(Nil)
+  }
+  def getOMRocketCores(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] =
```
hcook: Same could use-a-helper-method/only-called-once questions as above

src/main/scala/tile/RocketTile.scala
hcook: Translation modes are an addition consideration, should not just be Nil

src/main/scala/tile/RocketTile.scala
hcook: Yeah I have no idea what this number means

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,117 @@ class RocketTile(
...
+      val d = coreParams.fpu.filter(_.fLen > 32).map(x => isaExtSpec(D, "2.0"))
+
+      val omIsa = OMISA(
+        xLen = xLen,
+        baseSpecification = baseSpec(baseInstructionSet, baseISAVersion),
+        base = baseInstructionSet,
+        m = coreParams.mulDiv.map { x => isaExtSpec(M, "2.0") },
+        a = coreParams.useAtomics.option(isaExtSpec(A, "2.0")),
+        f = coreParams.fpu.map { x => isaExtSpec(F, "2.0") },
+        d = d,
```
hcook: Why is only `d` its own val?

src/main/scala/tile/RocketTile.scala
hcook: @aswaterman is there an straightforward way to calculate the mode from `coreParams`?

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,117 @@ class RocketTile(
...
+        addressTranslationModes = Nil
+      )
+
+      val btb = rocketParams.btb.map(BTB.makeOMI(_))
+
+      val omICache = rocketParams.icache.map(i => frontend.icache.device.getOMComponents(resourceBindingsMap) match {
+        case Seq() => throw new IllegalArgumentException
+        case Seq(h) => h.asInstanceOf[OMICache]
+        case _ => throw new IllegalArgumentException
+      })
```
hcook: I'm confused about the relationship between `omICache`, `getOMICacheFromBindings` and `frontend.icache.device.getOMComponents`

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,117 @@ class RocketTile(
       val Description(name, mapping) = super.describe(resources)
       Description(name, mapping ++ cpuProperties ++ nextLevelCacheProperty ++ tileProperties ++ dtimProperty ++ itimProperty)
     }
+
+    override def getOMComponents(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
+      val cores = getOMRocketCores(resourceBindingsMap)
+      val icache = getOMICacheFromBindings(resourceBindingsMap)
```
hcook: Why are icache and dcache handled differently here?

src/main/scala/tile/RocketTile.scala
```diff
@@ -80,6 +81,117 @@ class RocketTile(
...
+      val btb = rocketParams.btb.map(BTB.makeOMI(_))
+
+      val omICache = rocketParams.icache.map(i => frontend.icache.device.getOMComponents(resourceBindingsMap) match {
+        case Seq() => throw new IllegalArgumentException
+        case Seq(h) => h.asInstanceOf[OMICache]
+        case _ => throw new IllegalArgumentException
+      })
+
+//      //If DCache is configured as a DTIM, then you should populate the memory regions
+      def dcache(p: DCacheParams): OMDCache = {
```
hcook: Similarly, seems like a good candidate for a method on an `object OMDCache` factory method, maybe
`def make OMI(p: DCacheParams, device: OMComponent)` or something like that?

src/main/scala/rocket/ICache.scala
depappas-sifive: Made the method on a helper object in the diplomaticobjectmodel.model package

src/main/scala/tile/RocketTile.scala
depappas-sifive: What is the equation for addressTranslationModes?

src/main/scala/tile/RocketTile.scala
depappas-sifive: I spent about an hour looking for the DTIM device but could not find it.

src/main/scala/diplomaticobjectmodel/model/OMRocketCore.scala
depappas-sifive: changed

src/main/scala/tile/RocketTile.scala
depappas-sifive: moved the val d into line 143

src/main/scala/tile/RocketTile.scala
depappas-sifive: I copied the pattern from line 75

      val omICache = rocketParams.icache.map(i => frontend.icache.device.getOMComponents(resourceBindingsMap) match {
        case Seq() => throw new IllegalArgumentException
        case Seq(h) => h.asInstanceOf[OMICache]
        case _ => throw new IllegalArgumentException
      })

line 75
  val itimProperty = tileParams.icache.flatMap(_.itimAddr.map(i => Map(
    "sifive,itim" -> frontend.icache.device.asProperty))).getOrElse(Nil)

src/main/scala/tile/RocketTile.scala
richardxia: I believe that Rocket does not have an implementation of Sv48, so the rule is:

if Xlen == 32 then Sv32
elif Xlen == 64 then Sv39

depappas-sifive: fixed

src/main/scala/tile/RocketTile.scala
depappas-sifive: Should interruptLatency be removed from the schema?

src/main/scala/tile/RocketTile.scala
depappas-sifive: fixed

src/main/scala/tile/RocketTile.scala
depappas-sifive: fixed

src/main/scala/subsystem/RocketSubsystem.scala
depappas-sifive: This is a collection point for the lower level OM*'s. 
I feel that a helper function would make the code less clear in this case.

src/main/scala/diplomaticobjectmodel/model/OMISA.scala
```diff
@@ -36,3 +39,50 @@ case class OMISA(
...
+object OMISA {
+  // TODO the package object util has this code but is not seen in this scope
+  implicit class BooleanToAugmentedBoolean(val x: Boolean) extends AnyVal {
+    // this one's snagged from scalaz
+    def option[T](z: => T): Option[T] = if (x) Some(z) else None
+  }
+
+  def rocketISA(coreParams: RocketCoreParams, xLen: Int): OMISA = {
+    val baseInstructionSet = xLen match {
+      case 32 => if (XLen == 32) RV32E else RV32I
```
richardxia: Rocket *does not* currently implement RV32E, and therefore this should never return RV32E.

src/main/scala/tile/RocketTile.scala
richardxia: No, this is important information to capture. The value should be 4 for rocket-chip.

depappas-sifive: @hcook says that the number varies depending on the interrupt set. Lets go with 4 and then in rev2 lets revisit the way that the interrupt latency's are tracked in the OM.

richardxia: I just talked to @terpstra and @dbarbi1 and showed them this thread, and we think there's some confusion over what this value is supposed to be. This value is supposed to measure the latency between when an interrupt arrives at a core to when the core reacts to it, not the latency from the interrupt slaves to the core/interrupt controller.

depappas-sifive: How about if we change the name of this variable to interruptCoreResponseLatency?
Is the latency 4 for rocket?
What is the latency for Caboose?

src/main/scala/diplomaticobjectmodel/model/OMISA.scala
depappas-sifive: fixed

src/main/scala/diplomaticobjectmodel/model/OMCaches.scala
richardxia: These imports aren't used anymore.

src/main/scala/diplomaticobjectmodel/model/OMISA.scala
```diff
@@ -2,6 +2,10 @@
 
 package freechips.rocketchip.diplomaticobjectmodel.model
 
+import freechips.rocketchip.rocket.RocketCoreParams
+import freechips.rocketchip.tile.XLen
```
richardxia: This import isn't used.

depappas-sifive: fixed

src/main/scala/diplomaticobjectmodel/model/OMCaches.scala
depappas-sifive: fixed

PR #1738, opened 2018-12-07
Plic clint optional

ShuyunJia:
**Related issue**:  None

**Release Notes**

Add explicit config option `WithoutPlic()` which both configures `PLICKEY` and `NExtTopInterrupts`. Thus one could remove the PLIC from the core completely if no interrupt controller needed or extern interrupt controller is used.   
Fix bug in `connectInterrupts` when Clint is absent. Since no clint is implemented, no intsource should be connected to the intnode.  
Add explicit config option `WithoutClint()` for easy use.

### DIFF
--- a/src/main/scala/subsystem/HasTiles.scala
+++ b/src/main/scala/subsystem/HasTiles.scala
@@ -103,7 +103,7 @@ trait HasTiles { this: BaseSubsystem =>
     //    From CLINT: "msip" and "mtip"
     tile.crossIntIn() :=
       clintOpt.map { _.intnode }
-        .getOrElse { NullIntSource(sources = CLINTConsts.ints) }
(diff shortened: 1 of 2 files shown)


src/main/scala/subsystem/HasTiles.scala
```diff
@@ -103,7 +103,7 @@ trait HasTiles { this: BaseSubsystem =>
     //    From CLINT: "msip" and "mtip"
     tile.crossIntIn() :=
       clintOpt.map { _.intnode }
-        .getOrElse { NullIntSource(sources = CLINTConsts.ints) }
+        .getOrElse { NullIntSource() }
```
solomatnikov: Not clear why this change is needed - no CLINT option works even without this change.

PR #1740, opened 2018-12-08
Bundle bridge trace interface

rmac-sifive:

**Release Notes**
Use BundleBridgeSource for tile's trace interface

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -146,6 +146,10 @@ abstract class BaseTile(tileParams: TileParams, val crossing: ClockCrossingType)
   protected val tlSlaveXbar = LazyModule(new TLXbar)
   protected val intXbar = LazyModule(new IntXbar)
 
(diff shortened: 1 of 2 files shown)


src/main/scala/tile/BaseTile.scala
```diff
@@ -146,6 +146,10 @@ abstract class BaseTile(tileParams: TileParams, val crossing: ClockCrossingType)
   protected val tlSlaveXbar = LazyModule(new TLXbar)
   protected val intXbar = LazyModule(new IntXbar)
 
+  val traceNode = BundleBridgeSource(() => Vec(tileParams.core.retireWidth, new TracedInstruction()))
+  //val traceNode = BundleBroadcast[Vec[TracedInstruction]]
+  //traceNode := traceSourceNode
```
hcook: I think switching to the Broadcast version is necessary, because otherwise diplomacy will error out for configurations that have no further nodes connected to the BundleBridgeSource

PR #1741, opened 2018-12-08
Implement debug trigger-sharing and chain-limiting proposals

aswaterman:
See https://github.com/riscv/riscv-debug-spec/pull/282/files for relevant specification.

- Don't allow chain to be set if this trigger doesn't belong to D-mode but the next trigger does.

- Don't allow dmode to be set if the previous trigger doesn't belong to D-mode and has chain set.

- Don't allow chains longer than 2.

### DIFF
--- a/src/main/scala/rocket/Breakpoint.scala
+++ b/src/main/scala/rocket/Breakpoint.scala
@@ -70,14 +70,14 @@ class BreakpointUnit(n: Int)(implicit p: Parameters) extends CoreModule()(p) {
 
   io.bp.foldLeft((Bool(true), Bool(true), Bool(true))) { case ((ri, wi, xi), bp) =>
     val en = bp.control.enabled(io.status)
-    val r = en && ri && bp.control.r && bp.addressMatch(io.ea)
-    val w = en && wi && bp.control.w && bp.addressMatch(io.ea)
-    val x = en && xi && bp.control.x && bp.addressMatch(io.pc)
+    val r = en && bp.control.r && bp.addressMatch(io.ea)
+    val w = en && bp.control.w && bp.addressMatch(io.ea)
+    val x = en && bp.control.x && bp.addressMatch(io.pc)
     val end = !bp.control.chain
 
-    when (end && r) { io.xcpt_ld := !bp.control.action; io.debug_ld := bp.control.action }
-    when (end && w) { io.xcpt_st := !bp.control.action; io.debug_st := bp.control.action }
-    when (end && x) { io.xcpt_if := !bp.control.action; io.debug_if := bp.control.action }
+    when (end && r && ri) { io.xcpt_ld := !bp.control.action; io.debug_ld := bp.control.action }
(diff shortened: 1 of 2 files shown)


src/main/scala/rocket/CSR.scala
```diff
@@ -851,6 +856,7 @@ class CSRFile(
     when (reset) {
       bpc.action := false
       bpc.dmode := false
+      bpc.chain := false
```
mwachs5: is this the bit that prevents longer chains? Is the idea that only even/odd are allowed to have the chain bit set?

src/main/scala/rocket/Breakpoint.scala
```diff
@@ -70,14 +70,14 @@ class BreakpointUnit(n: Int)(implicit p: Parameters) extends CoreModule()(p) {
...
+    val r = en && bp.control.r && bp.addressMatch(io.ea)
+    val w = en && bp.control.w && bp.addressMatch(io.ea)
+    val x = en && bp.control.x && bp.addressMatch(io.pc)
     val end = !bp.control.chain
 
-    when (end && r) { io.xcpt_ld := !bp.control.action; io.debug_ld := bp.control.action }
-    when (end && w) { io.xcpt_st := !bp.control.action; io.debug_st := bp.control.action }
-    when (end && x) { io.xcpt_if := !bp.control.action; io.debug_if := bp.control.action }
+    when (end && r && ri) { io.xcpt_ld := !bp.control.action; io.debug_ld := bp.control.action }
+    when (end && w && ri) { io.xcpt_st := !bp.control.action; io.debug_st := bp.control.action }
```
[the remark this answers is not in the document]
aswaterman: Nice catch, was supposed to be ri/wi/xi, not ri/ri/ri.

(the *i variables are the chain input from the previous breakpoint)

src/main/scala/rocket/CSR.scala
aswaterman: This is just the bit indicating the breakpoint is chained.  I didn't restrict it to even/odd pairs; chains can still begin on odd numbers but are always limited to 2.  The reason I'm resetting the bits is to prevent them from coming up in an inconsistent state where >2 breakpoints appear to be chained.

The hardware is a little simpler if it's restricted to even-odd pairs, so if you think that won't cause problems, I could revise this PR to incorporate that restriction.  I think the only case where the even-odd restriction actually reduces the usefulness of the breakpoints is when there is an odd number of breakpoints, and D-mode wants to chain two of them but share some with M-mode.

mwachs5: Nah, I don't think you need to do the even/odd thing. So how does it work to limit the chain length? Does it prohibit you from setting the bit if the previous one is set, or something?

aswaterman: Right, the line

    bp.control.chain := newBPC.chain && !(prevChain || nextChain) && (dMode || !nextDMode)

prevents setting the chain bit when either the previous or next breakpoint has its chain bit set.

mwachs5: got it

PR #1743, opened 2018-12-10
Provide a modicum of TL.A buffering in the NoCrossing case

### LINKED ISSUE #1722: Data access replay error when crossingType=NoCrossing
I'm using a design based on the SiFive Freedom E310, and I'm seeing 2 cycles of latency between the Rocket core issuing a data access and it appearing on the tile's TileLink interface.

The subsystem's config had RocketCrossingParams with crossingType = SynchronousCrossing(). I switched to NoCrossing to see if this would reduce the latency, and if so by how much.

It did indeed reduce the latency by 1 cycle each way.

However I then noticed an issue where the DCache would issue a valid pulse whilst the TileLink address bus ready was deasserted, then issue an s2_nack back to the rocket core, but when it replayed the access the address would be incorrect.

image

What happened is an offset calculation has been repeated, corrupting the address. My disassembly looks like so:

```
  1c007e:       10050537                lui     a0,0x10050
  1c0082:       0b050513                addi    a0,a0,176 # 100500b0 <spi_load+0xfe9001e>
  1c0086:       f14025f3                csrr    a1,mhartid
  1c008a:       058a                    slli    a1,a1,0x2
  1c008c:       95aa                    add     a1,a1,a0
  1c008e:       4188                    lw      a0,0(a1)
  1c0090:       8502                    jr      a0
```

The issue is that the ```add a1,a1,a0``` occurs again (the core executes and completes the instruction at 0x1c008c a second time), resulting in an incorrect address in a1.

I've often wondered why the rocket core seems to replay instead of stalling while the various hazards or buses clear. Anyway, does this look like a real bug, and not a configuration issue? Is rocket supposed to work with NoCrossing?

aswaterman:
This prevents livelock in a multi-master system with NoCrossing.  Resolves #1722

### DIFF
--- a/src/main/scala/groundtest/Tile.scala
+++ b/src/main/scala/groundtest/Tile.scala
@@ -40,7 +40,7 @@ abstract class GroundTestTile(params: GroundTestTileParams)
   val haltNode: IntOutwardNode = IntIdentityNode()
   val wfiNode: IntOutwardNode = IntIdentityNode()
 
-  val dcacheOpt = params.dcache.map { dc => LazyModule(new DCache(0)) }
+  val dcacheOpt = params.dcache.map { dc => LazyModule(new DCache(0, crossing)) }
 
   override lazy val module = new GroundTestTileModuleImp(this)
 }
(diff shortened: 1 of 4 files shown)


hcook:
Closes #1722

aswaterman:
Async case now handled.

PR #1746, opened 2018-12-11
Core monitor bundle

p12nGH:

**Release Notes**

This PR moves CoreMonitorBundle class to package level and removes unused mechanism of instantiating CoreMonitors via Parameters mechanism. The new way to access CoreMonitor is to use cross module references without instantiating anything in the core itself.


src/main/scala/rocket/RocketCore.scala
```diff
@@ -802,24 +819,10 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
...
-    val rd0val = UInt(width = xLen)
-    val rd1src = UInt(width = 5)
-    val rd1val = UInt(width = xLen)
-    val inst = UInt(width = 32)
-  }
-  val coreMonitorBundle = Wire(new CoreMonitorBundle)
+  val coreMonitorBundle = Wire(new CoreMonitorBundle(xLen))
 
   coreMonitorBundle.hartid := io.hartid
-  coreMonitorBundle.time := csr.io.time(31,0)
```
p12nGH: Had to rename this field because time is a Verilog keyword

aswaterman: That sounds like a bug in a Firrtl pass. Firrtl is not supposed to emit Verilog keywords as wire names.  Unfortunately, "time" is already in the keyword list, so it's not obvious to me where the bug is.  attn @albertchen-sifive

seldridge: Possibly a bug in the by me in the Verilog keyword rename refactor: https://github.com/freechipsproject/firrtl/pull/909. I can check how `time` is handled.

It is supposed to be entirely safe to use Verilog keywords....

p12nGH: It is a bug(or rather a limitation) in the code where we consume this bundle. One of the monitor implementations generates a DPI function call and uses bundle field names as parameter names in that function. And SystemVerilog doesn't allow a function parameter to be named `time`.

seldridge: Thanks for the clarification @p12nGH. If you have code that assumes that bundle fields will not conflict with Verilog keywords, then it may make sense to generalize `RemoveKeywordCollisions` to also work on `HighForm` so that it can be _optionally_ run to clean these things up. There would have to be an annotation on that field and the DPI code generator would have to consume that to track the renamed name.

PR #1756, opened 2018-12-17
Typo: syntax error

colin4124:

### DIFF
--- a/src/main/scala/devices/tilelink/DevNull.scala
+++ b/src/main/scala/devices/tilelink/DevNull.scala
@@ -13,7 +13,7 @@ case class DevNullParams(
   region: RegionType.T = RegionType.UNCACHEABLE,
   executable: Boolean = true,
   mayDenyGet: Boolean = true,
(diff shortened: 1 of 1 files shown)


hcook:
@jackkoenig What is the status of trailing commas? I know support was added but since what Scala version? Seems like the build itself is working but we're encountering various failures in e.g. linters etc.

@colin4124 What Scala version are you using?

colin4124:
Trailing commas is awesome, Thanks to tell me about that.

It should be the old version used when I manual change the scala version in `build.sbt` to` 2.11.12`
https://github.com/freechipsproject/rocket-chip/blob/f1acf8c419240dec668682761466cb17ccbcb23b/build.sbt#L13

But if you use the feature only in 2.12 version, what is the meaning of crossScalaVersions? I assume that it can be compiled by 2.12.4 and 2.11.12 version

https://github.com/freechipsproject/rocket-chip/blob/f1acf8c419240dec668682761466cb17ccbcb23b/build.sbt#L14

PR #1765, opened 2018-12-29
Independent AXI4 port for Accelerators

tampler:
1. Added  a direct port to the next level Memory for Rocc Accelerators
2. Showed how to connect a generic ROCC accelerator to internal buses

Details here: http://i63.tinypic.com/wgt261.png

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -14,14 +14,14 @@ VLSI_MEM_GEN ?= $(base_dir)/scripts/vlsi_mem_gen
 
 CXX ?= g++
 CXXFLAGS := -O1
-JVM_MEMORY ?= 2G
+JVM_MEMORY ?= 4G
 
-SBT ?= java -Xmx$(JVM_MEMORY) -Xss8M -XX:MaxPermSize=256M -jar $(base_dir)/sbt-launch.jar ++2.12.4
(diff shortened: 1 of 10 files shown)


hcook:
Can you split part 3 into three or four separate commits on their own PR branch? I would like to evaluate them independently of the RTL changes

hcook:
It is somewhat unclear to me what your overall design intention is with these additions. You did not create a direct connection between a new port top-level port and the RoCC accelerator. The RoCC accelerator's `tlNode` is still connected to the SystemBus (which is in turn connected to the MemoryBus or new AccBus, which in turn generates a top-level port). But adding an entire new bus behind the SystemBus seems like overkill just to get a single second port.

Some questions:
1. Why can't you just use the regular memory port? If you need more bandwidth, you can increase the number of channels.
2. If you really need two distinct ports with different address ranges, why can't you use the MMIO port on the SystemBus as the second port? Does your accelerator contain a cache or does it issue only MMIO traffic?
3. What is the intended behavior when there are many tiles, each with multiple accelerators? Is each accelerator supposed to have its own memory port?

Overall, I think we can probably come up with a smaller-footprint solution to address your use case.

tampler:
Follow #1768 

Started from scratch to provide smaller impact commits. All was verified and built with Verilator 4.009 for the current master, sbt 1.2.8 and scala 2.12.8

PR #1793, opened 2019-01-21
[TLB] relax paddrBits to be stricly less than vaddrBits

ss2783:

This commit allows supporting larger physical adddress bits when
compared to the virtual address bits.

### DIFF
--- a/src/main/scala/rocket/TLB.scala
+++ b/src/main/scala/rocket/TLB.scala
@@ -205,7 +205,7 @@ class TLB(instruction: Boolean, lgMaxSize: Int, cfg: TLBConfig)(implicit edge: T
   val hitsVec = all_entries.map(vm_enabled && _.hit(vpn))
   val real_hits = hitsVec.asUInt
   val hits = Cat(!vm_enabled, real_hits)
-  val ppn = Mux1H(hitsVec :+ !vm_enabled, all_entries.map(_.ppn(vpn)) :+ vpn(ppnBits-1, 0))
+  val ppn = Mux1H(hitsVec :+ !vm_enabled, all_entries.map(_.ppn(vpn)) :+ io.req.bits.vaddr(paddrBits-1, pgIdxBits))
 
   // permission bit arrays
   when (do_refill && !invalidate_refill) {
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -53,7 +53,7 @@ trait HasTileParameters {
   def vaddrBits: Int =
     if (usingVM) {
       val v = pgIdxBits + pgLevels * pgLevelBits
-      require(v == xLen || xLen > v && v > paddrBits)
+      require(v == xLen || xLen > v)
       v
     } else {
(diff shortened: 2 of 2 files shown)


aswaterman:
Thanks for looking into this; it's a great feature addition.  The patch seems deceptively simple.  How have you tested it?  Is there any software that's capable of taking advantage of it?

I notice the pc, mepc, mbadaddr, etc. are still vaddrBitsExtended-bits wide.  Is the idea that physical memory past 2^38 is only accessible when mapped into virtual memory?  I think this is a very reasonable configuration, as it reduces HW cost.

I think this also require that page tables lie below 2^38 as well, as the address bus between the page-table walker and D$ is still limited to vaddrBitsExtended.  That might or might not be reasonable.  Is Linux OK with that restriction?  Either way, we need to test that something sane happens if you try to put a page table at 2^38 or 2^39 - it should cleanly generate an access exception, rather than blowing up.

aswaterman:
It appears Linux is OK with putting page tables only in low memory.  So if that's really the route we want to go (and I'm still not sure that it is), we still need (at minimum) the following changes:

- The DeviceTree needs to split memory is into low and high parts at address 2^(vaddrBits-1), whenever `paddrBits >= vaddrBits && paddrBits != xLen`.  The high part of memory needs to be described in the DeviceTree in such a way that Linux knows not to allocate page tables there.  (That might just work out of the box, but might require a new annotation.)

- The PTW should report an access exception in trying to fetch a PTE at those addresses, whenever `paddrBits >= vaddrBits && paddrBits != xLen`.

- Verify that pure physical-memory accesses and instruction fetches from addresses 2^(vaddrBits-1) and above correctly raise access exceptions and don't have unexpected side effects.

ss2783:
@aswaterman  I'm closing this PR as this is only a partial solution. 

A more complete solution could be along the following lines. We would need to define a `coreMaxAddrBitsExtended` parameter in `Core.scala` which compares the `vaddrBits` vs. `paddrBits` and selects the maximum of the two and stores a value which is one more than the maximum. All uses of `vaddrBitsExtended` in the pipeline and related files should instead use `coreMaxAddrBitsExtended`. Any address that is generated from the processor would continue to use the existing algorithm to compress out the higher bits and the TLB checks for `badvaddr` now has to consider all the cases i.e.:  `1.) vaddrBits > paddrBits`, `2.) vaddrBits > paddrBits`, and `3.) vaddrBits == paddrBits`. For the (2) TLB would have to check the extra-bit which encodes the higher bits (>`coreAddrBits`) and reduce the excess bits (i.e., `paddrBits-vaddrBits`) to compute a `badvaddr` check. 

Instead of this solution, I think dynamic support of Sv32 on Sv48 hardware works for a larger `paddrBits` (cop out)

aswaterman:
Makes sense.  Actually @ss2783 I think the dynamic-Sv48 thing is _not_ a total cop-out.  Linux performs better when it's able to map the entire physical address space into its virtual address space in an affine fashion, so there's a solid justification for coupling the two.  At least for Linux :)

Thanks for getting the ball rolling on this.

PR #1795, opened 2019-01-21
Diplomatic Clock Primitives

hcook:
This PR adds some primitives that can be used to pipe pairs of clock+reset signals through the design diplomatically. Adding this package upstreams some features from https://github.com/sifive/fpga-shells/tree/master/src/main/scala/clocks, which it is intended to (eventually) replace. There are node and edge types for both clock/reset pairs, and groups of such pairs that then must share a synchronous or rational relationship. Mutually asynchronous clocks should always be expressed as individual clock groups (which may then have only a single member).

The PR also adds some test harness clock generation functionality (which is not verilator compatible and so not currently added to regressions here), and includes a set of example clocking adapters, and updates some clocking utility blackbox wrappers to `import chisel3`.

Right now these features are only employed to drive the clocks of the TileLink buses included by default in `BaseSubsystem`. It maintains backwards compatibility with devices that tap into the bus clock directly by providing `clock` and `reset` outputs from the `TLBusWrapper` module instances.

This PR also adds the `trait Attachable`, which wraps the different nodes and members of the subsystem to which devices, tiles and other peripherals might which to attach themselves, including a point for registering additional clocks asynchronous to the tilelink bus clock. However, no code is yet changed to depend on this new `trait`.

Both `Attachable` and all the graph types added in the `prci` package should be considered experimental and are still under active development.

**Impact**: API change

While this PR is mostly additive, it does change how the clock/resets of `TLBusWrapper` subclasses are to be driven. Mostly, they drive each other's clocks based on the type of TL clock crossing adapter  that is inserted between them in the `crossToBus` and `crossFromBus` methods. However, the "root" bus needs to have its `clockGroupNode` driven, which the example `HasHierarchicalBusTopology` does via `BaseSubsystem.asyncClockGroupsNode`, which is in turn (by configurable default) driven by the implicit Chisel `clock` of `BaseSubsystem`'s module implementation, here https://github.com/chipsalliance/rocket-chip/pull/1795/files#diff-3d87f78ef4648ef6b47c853c4581a47bR95

**Future Work**:
There are already a set of requested features for future work. These cover both internal functionality of the clocking graph and external application of the graph to use cases in the subsystem. Logging them here for future reference:
Internal features:
- The type of `ClockBundle.reset` is currently `Bool`, not the new `chisel3.Reset` wrapper type. Working theory is that every instance of a `ClockBundle` should be explicitly told whether it is to be an instance of `Bool` or an instance of `AsyncReset`, but this needs to be proven out and won't necessarily be compatible with the version in `fpga-shells`
- The way in which members of ClockGroupBundles are named is just via `HeterogenousBag` indexes, this should be improved to be based on the eventual clock sink name.
- Mapping from `ClockGroupSourceParameters` to `ClockSourceParameters` needs to be fleshed out. ClockGroups should be capable of resolving into single pairs when only a single source is provided to satisfy all the clocks in a group.
External features:
- The `InterruptBusWrapper` should have its own `ClockSinkDomain`. 
- Tiles' crossing type relative to the system bus should be captured diplomatically
- Downstream oeripheral devices in e.g. `sifive-blocks` should be updated to rely on only the `Attachable` trait

### DIFF
--- a/src/main/scala/clocks/Bundles.scala
+++ b/src/main/scala/clocks/Bundles.scala
@@ -1,17 +1,16 @@
 // See LICENSE.SiFive for license details.
 package freechips.rocketchip.clocks
 
-import Chisel._
-import freechips.rocketchip.diplomacy._
-import freechips.rocketchip.util._
+import chisel3._
+import freechips.rocketchip.util.HeterogeneousBag
 
-class ClockBundle(params: ClockBundleParameters) extends GenericParameterizedBundle(params)
+class ClockBundle(val params: ClockBundleParameters) extends Bundle
 {
   val clock = Clock()
   val reset = Bool()
 }
 
-class ClockGroupBundle(params: ClockGroupBundleParameters) extends GenericParameterizedBundle(params)
+class ClockGroupBundle(val params: ClockGroupBundleParameters) extends Bundle
 {
   val member = HeterogeneousBag(params.members.map(p => new ClockBundle(p)))
 }
--- a/src/main/scala/clocks/Bundles.scala
+++ b/src/main/scala/clocks/Bundles.scala
@@ -1,0 +1,17 @@
+// See LICENSE.SiFive for license details.
+package freechips.rocketchip.clocks
+
+import Chisel._
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.util._
+
+class ClockBundle(params: ClockBundleParameters) extends GenericParameterizedBundle(params)
+{
+  val clock = Clock()
+  val reset = Bool()
+}
+
+class ClockGroupBundle(params: ClockGroupBundleParameters) extends GenericParameterizedBundle(params)
+{
+  val member = HeterogeneousBag(params.members.map(p => new ClockBundle(p)))
+}
--- a/src/main/scala/clocks/ClockDomain.scala
+++ b/src/main/scala/clocks/ClockDomain.scala
@@ -1,0 +1,31 @@
+package freechips.rocketchip.clocks
+
+import chisel3._
+import freechips.rocketchip.config.Parameters
+import freechips.rocketchip.diplomacy._
+
+class ClockSinkDomain(take: Option[ClockParameters] = None)(implicit p: Parameters) extends LazyModule
+    with LazyScope
+    with HasClockDomainCrossing {
+
+  val node = ClockSinkNode(Seq(ClockSinkParameters(take = take)))
+
+  lazy val module = new LazyRawModuleImp(this) {
+    val (bundle, edge) = node.in.head
+    childClock := bundle.clock
+    childReset := bundle.reset
+  }
+}
+
+class ClockSourceDomain(give: Option[ClockParameters] = None)(implicit p: Parameters) extends LazyModule
+    with LazyScope
+    with HasClockDomainCrossing {
+
+  val node = ClockSourceNode(Seq(ClockSourceParameters(give = give)))
+
+  lazy val module = new LazyRawModuleImp(this) {
+    val (bundle, edge) = node.out.head
+    childClock := bundle.clock
+    childReset := bundle.reset
+  }
+}
--- a/src/main/scala/clocks/ClockDomain.scala
+++ b/src/main/scala/clocks/ClockDomain.scala
@@ -8,10 +8,10 @@ class ClockSinkDomain(take: Option[ClockParameters] = None)(implicit p: Paramete
     with LazyScope
     with HasClockDomainCrossing {
 
-  val node = ClockSinkNode(Seq(ClockSinkParameters(take = take)))
+  val clockNode = ClockSinkNode(Seq(ClockSinkParameters(take = take)))
 
   lazy val module = new LazyRawModuleImp(this) {
-    val (bundle, edge) = node.in.head
+    val (bundle, _) = clockNode.in.head
     childClock := bundle.clock
     childReset := bundle.reset
   }
@@ -21,10 +21,10 @@ class ClockSourceDomain(give: Option[ClockParameters] = None)(implicit p: Parame
     with LazyScope
     with HasClockDomainCrossing {
 
-  val node = ClockSourceNode(Seq(ClockSourceParameters(give = give)))
+  val clockNode = ClockSourceNode(Seq(ClockSourceParameters(give = give)))
 
   lazy val module = new LazyRawModuleImp(this) {
-    val (bundle, edge) = node.out.head
+    val (bundle, _) = clockNode.out.head
     childClock := bundle.clock
     childReset := bundle.reset
   }
--- a/src/main/scala/clocks/ResetWrangler.scala
+++ b/src/main/scala/clocks/ResetWrangler.scala
@@ -18,10 +18,11 @@ class ResetWrangler(debounceNs: Double = 100000)(implicit p: Parameters) extends
 
     val status = IO(Output(UInt(in.size.W)))
     status := Cat(in.map(_.reset).reverse)
-
     val causes = in.map(_.reset).foldLeft(false.B)(_ || _)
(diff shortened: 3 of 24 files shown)


src/main/scala/clocks/ClockDomain.scala
```diff
@@ -0,0 +1,31 @@
...
+    with HasClockDomainCrossing {
+
+  val node = ClockSourceNode(Seq(ClockSourceParameters(give = give)))
+
+  lazy val module = new LazyRawModuleImp(this) {
+    val (bundle, edge) = node.out.head
+    childClock := bundle.clock
+    childReset := bundle.reset
+  }
+}
```
rmac-sifive: what's the use case for these two classes? is the idea you'd use `LazyScope` to put modules sharing a clock into the same `ClockSourceDomain` hierarchy, for example?

src/main/scala/clocks/Bundles.scala
```diff
@@ -0,0 +1,16 @@
...
+class ClockBundle(val params: ClockBundleParameters) extends Bundle
+{
+  val clock = Clock()
+  val reset = Bool()
+}
+
+class ClockGroupBundle(val params: ClockGroupBundleParameters) extends Bundle
+{
+  val member = HeterogeneousBag(params.members.map(p => new ClockBundle(p)))
+}
```
rmac-sifive: Why does this have to be a `HeterogeneousBag` and not `Seq[ClockBundle]`?

src/main/scala/clocks/ClockDomain.scala
hcook: Yes, you can wrap arbitrary subgraphs of adapters with it and give them all the same clock, similar to what the `couple` methods do in the `TLBusWrapper`s but with added clock assignments.

src/main/scala/clocks/Bundles.scala
hcook: "Have to" is a bit strong but the idea is that you are making a single port that has many clocks, but each clock bundle may have different edge parameters from the others (unlike `Vec`). At the moment it is a bit moot since nothing about the edge parameters actually effects the clock bundle wires themselves, but if we add e.g. optional clock enables in the future, this would let some members have it and some not. As opposed to `Seq`, HB is an actual hardware Record type, so you can make a single port that actually has that Data type, put it in a different container`[T<:Data]`, etc. (Whereas `Seq` is not `<: Data`).

src/main/scala/clocks/Bundles.scala
```diff
@@ -0,0 +1,16 @@
+// See LICENSE.SiFive for license details.
+package freechips.rocketchip.clocks
+
+import chisel3._
+import freechips.rocketchip.util.HeterogeneousBag
+
+class ClockBundle(val params: ClockBundleParameters) extends Bundle
+{
+  val clock = Clock()
+  val reset = Bool()
```
mwachs5: could this be of type `Reset`? instead of hard-coding as `Bool`?

src/main/scala/clocks/ResetWrangler.scala
```diff
@@ -0,0 +1,45 @@
...
+package freechips.rocketchip.clocks
+
+import chisel3._
+import chisel3.util._
+import chisel3.experimental.{withClockAndReset}
+import freechips.rocketchip.config._
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.util._
+
+class ResetWrangler(debounceNs: Double = 100000)(implicit p: Parameters) extends LazyModule
```
mwachs5: Can we at least put a one line comment description of what this does....

mwachs5: What I mean is, doesn't this take an async reset and turn it into a sync reset?

hcook: I just copied this adapter from FPGA shells, but I believe you are correct. None of these have had the notion of the new chisel3.AsyncReset feature added to them, but it is a good suggestion to incorporate.

src/main/scala/clocks/Bundles.scala
hcook: Good idea. I think it could be. What I need to work out is, is it just of type `Reset()`, or should it actually be parameterized specifically to be either `AsyncReset` or `Bool`? Or maybe we actually want it to be tri-valued like `knownToBeSynchronized: Option[Boolean]`.

Also, it seems that synchronous in this context must mean "synchronized with respect to this particular clock", but is that more specific than the general chisel3 concept of `AsyncReset`? Or is that always defined relative to a particular clock too? @jackkoenig any opinions?

hcook: Talking with @jackkoenig IRL our current feeling is that actually this should be explicitly parameterized as a handshake between source and sinks, which then makes each bundle emitted concretely contain either Bool or AsyncReset. Argument by analogy here is avoidance of width inference across diplomatically emitted `Record` subtypes.

hcook: And to answer my above question, reset is always assumed to be synchronously deasserted with respect to the clock of the "module" in question.

src/main/scala/clocks/ResetWrangler.scala
hcook: I'm going to do it as a follow-on PR though (possibly after thanksgiving).

src/main/scala/prci/ClockBundles.scala
```diff
@@ -0,0 +1,16 @@
+// See LICENSE.SiFive for license details.
+package freechips.rocketchip.prci
+
+import chisel3._
+import freechips.rocketchip.util.HeterogeneousBag
+
+class ClockBundle(val params: ClockBundleParameters) extends Bundle
+{
+  val clock = Clock()
+  val reset = Bool()
```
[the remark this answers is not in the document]
hcook: Future work. I think we will always pick between Bool/AsyncReset based on a parameter that flows from source to sink, and never rely on `Reset` inference, but that needs to be finalized. It would also break backwards compatibility with the version in fpga-shells, so we might want to reason about whether merging them first and then change it is the right order of operations or not

hcook:
One year birthday of opening this PR 🎉

jerryhethatday:
clockGroupNode what does this class do?

jerryhethatday:
what does "each other" refer to?
> While this PR is mostly additive, it does change how the clock/resets of TLBusWrapper subclasses are to be driven. Mostly, they drive **each other**'s clocks based on the type of TL clock crossing adapter that is inserted between them in the crossToBus and crossFromBus methods.
@hcook

src/main/scala/clocks/Bundles.scala
jerryhethatday: > "Have to" is a bit strong but the idea is that you are making a single port that has many clocks, but each clock bundle may have different edge parameters from the others (unlike `Vec`). At the moment it is a bit moot since nothing about the edge parameters actually effects the clock bundle wires themselves, but if we add e.g. optional clock enables in the future, this would let some members have it and some not. As opposed to `Seq`, HB is an actual hardware Record type, so you can make a single port that actually has that Data type, put it in a different container`[T<:Data]`, etc. (Whereas `Seq` is not `<: Data`).

now it becomes `val member: RecordMap[ClockBundle] = RecordMap(params.members.map { case (k, v) =>
    k -> new ClockBundle(v)
  })`
 _As opposed to Seq, HB is an actual hardware Record type, so you can make a single port that actually has that Data type, put it in a different container[T<:Data], etc. (Whereas Seq is not <: Data)._
I still felt confused about this, you mean I can make a single port that actually has HeterogeneousBag(Or RecordMap) type? what do you mean by  put it in a different container[T<:Data], etc.

src/main/scala/clocks/ResetWrangler.scala
jerryhethatday: what specificly does the ResetWrangler do?

PR #1799, opened 2019-01-24
Support delegation of misaligned ld/st and illegal instruction traps

aswaterman:
This facilitates moving emulation code out of the M-mode trusted codebase.

**Release Notes**


aswaterman:
Applying "don't merge" because I want to first find and fix any software that was assuming these settings were hard-wired to "don't delegate."

PR #1800, opened 2019-01-24
Om scala plic fix1

depappas-sifive:
**Related issue**: 

**Release Notes**

Adding OMInterrupt and OMInterruptTarget


src/main/scala/diplomacy/Resources.scala
```diff
@@ -2,6 +2,8 @@
 
 package freechips.rocketchip.diplomacy
 
+import java.awt.Label
```
richardxia: Is this needed?

src/main/scala/diplomacy/Resources.scala
```diff
@@ -208,6 +212,8 @@ class SimpleDevice(devname: String, devcompat: Seq[String]) extends Device
     val names = optDef("reg-names", named.map(x => ResourceString(DiplomacyUtils.regName(x._1).get)).toList) // names of the named address space
     val regs = optDef("reg", (named ++ bulk).flatMap(_._2.map(_.value)).toList) // address ranges of all spaces (named and bulk)
 
+    devNamePlusAddress = name
```
richardxia: Although doing it this way makes it a bit easier for you to write this code now, this will in the long run make the code much more complicated since you are relying on state being mutated at a specific point in time, and that means if you try accessing this value too early, you don't get the result that you want.

I think a better way of doing this would be to not rely on mutability and instead turn this mutable property into a purely functional method. The method should require the caller to pass in all the information that is needed to compute this value, and then it will not be a game of "find the needle in the haystack" for the next person to figure out when this property is set.

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
```diff
@@ -186,4 +186,34 @@ object DiplomaticObjectModelAddressing {
...
+  def describeInterrupts(name: String, resources: ResourceBindings): Seq[OMInterrupt] = {
+    val int = resources("int")
+    int.flatMap {
+      b =>
+        val grandParentOpt = b.device.get.parent
+        grandParentOpt.map {
+         gp =>
+          OMInterrupt(
+            receiver = getDeviceName(gp),
+            numberAtReceiver = getInterruptNumber(b.value).toInt,
```
richardxia: Maybe `numberAtReceiver` should be a BigInt instead of you casting it to an int?

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
```diff
@@ -186,4 +186,34 @@ object DiplomaticObjectModelAddressing {
...
+    }
+  }
+
+  def describeInterrupts(name: String, resources: ResourceBindings): Seq[OMInterrupt] = {
+    val int = resources("int")
+    int.flatMap {
+      b =>
+        val grandParentOpt = b.device.get.parent
+        grandParentOpt.map {
+         gp =>
```
richardxia: This is another instance where using a `for` comprehension may improve readability by reducing the level of nesting required:

```scala
for {
  b <- int
  grandParentOpt = b.device.get.parent
  gp <- grandParentOpt
} yield OMInterrupt(...)
```

src/main/scala/diplomacy/Resources.scala
depappas-sifive: Wes suggested mutability.

The caller does not have the info.

@terpstra ????

src/main/scala/diplomacy/Resources.scala
depappas-sifive: No longer present.

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
depappas-sifive: Sure

src/main/scala/diplomacy/Resources.scala
richardxia: I still stand by my statement that immutability will make this easier to read in the long run.

The caller does have the information. The caller has the resource bindings, which is the only thing you need to pass into the method I'm suggesting you create:

```scala
class SimpleDevice(devname: String, devcompat: Seq[String]) extends Device ... {
  ...

  def devNamePlusAddress(resources: ResourceBindings) = describeName(devname, resources)

  ...
}
```

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
depappas-sifive: This coding style should be brought up at the engineering forum

src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -88,21 +88,21 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
...
+          interrupts = module.getSources(),
           specifications = List(
             OMSpecification(
               name = "The RISC‑V Instruction Set Manual, Volume II: Privileged Architecture",
               version = "1.10"
             )
           ),
           latency = 2, // TODO
-          nInterrupts = 3,
           nPriorities = params.maxPriorities,
```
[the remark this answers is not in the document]
depappas-sifive: This is what the code I checkedin last night looks like:

```scala
  def getOMPLIC(resourceBindings: ResourceBindings): Seq[OMComponent] = {
      val memRegions : Seq[OMMemoryRegion]= DiplomaticObjectModelAddressing.getOMMemoryRegions("PLIC", resourceBindings, Some(module.omRegMap))
      val ints = DiplomaticObjectModelAddressing.describeInterrupts(describe(resourceBindings).name, resourceBindings)
      val Description(name, mapping) = describe(resourceBindings)

      Seq[OMComponent](
        OMPLIC(
          memoryRegions = memRegions,
          interrupts = ints,
          specifications = List(
            OMSpecification(
              name = "The RISC‑V Instruction Set Manual, Volume II: Privileged Architecture",
              version = "1.10"
            )
          ),
          latency = 2, // TODO
          nPriorities = params.maxPriorities,
          targets = module.getTargets()
        )
      )
    }
```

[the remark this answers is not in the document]
depappas-sifive: fixed

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
depappas-sifive: changed

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
depappas-sifive: There are DiplomaticObjectModelUtils functions which receive params objects which need their signature refactored.

khannaudit:
Approved with the following observation/comments:
- "Interrupts" attribute is correct
- "Targets" attribute, which is actually just another way of representing the "interrupts" attribute is *incorrect*, and there doesn't seem to be a straightforward way to get this information from the PLIC alone.
- "nInterrupts" attribute that was removed from the object model PLIC schema may need to be added again later, although that should be a backward-compatible change.

Issue #1804, opened 2019-01-26
untethered simulation problems


gsomlo:
**Type of issue**: other

**Impact**: unknown

**Development Phase**: request

**Other information**
Attempting to simulate rocket-chip untethered (i.e., after replacing the SimDTM module with a stub) results in the design never properly initializing and entering the "hang" infinite loop at 0x10040.

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
1. Clone the repo (I used commit 1d947b66, but anything newer should also be OK), then build "DefaultFPGASmallConfig":
```
git clone --recursive https://github.com/freechipsproject/rocket-chip
make RISCV=$(HOME)/RISCV -C rocket-chip/vsim verilog CONFIG=DefaultFPGASmallConfig
```
2. Next, add a testbench [tb.v.txt](https://github.com/freechipsproject/rocket-chip/files/2799213/tb.v.txt) (top module named "testbench"), which also contains a stub module for SimDTM (returning 0 for debug_req_valid, debug_resp_ready, and exit)

3. Compile with either VCS:
```
vcs -top testbench tb.v \
  rocket-chip/src/main/resources/vsrc/AsyncResetReg.v \
  rocket-chip/src/main/resources/vsrc/EICG_wrapper.v \
  rocket-chip/src/main/resources/vsrc/plusarg_reader.v \
  rocket-chip/vsim/generated-src/freechips.rocketchip.system.DefaultFPGASmallConfig.v \
  rocket-chip/vsim/generated-src/freechips.rocketchip.system.DefaultFPGASmallConfig.behav_srams.v
```
or iverilog:
```
iverilog -s testbench -o tb.vvp tb.v \
  rocket-chip/src/main/resources/vsrc/AsyncResetReg.v \
  rocket-chip/src/main/resources/vsrc/EICG_wrapper.v \
  rocket-chip/src/main/resources/vsrc/plusarg_reader.v \
  rocket-chip/vsim/generated-src/freechips.rocketchip.system.DefaultFPGASmallConfig.v \
  rocket-chip/vsim/generated-src/freechips.rocketchip.system.DefaultFPGASmallConfig.behav_srams.v
```

4. Run simulation, either
`./simv`
or
`vvp -N tb.vvp`

**What is the current behavior?**
```
    0 ########################################
    result=0    
    pc=xxxxxxxxx
 
[...]

PR #1807, opened 2019-01-28
AsyncReset: iverilog requires named initial block with `RANDOMIZE

gsomlo:
When simulating with iverilog, we get an error message for the
assignment to _RAND in the module's initial block:
"Variable declaration in unnamed block requires SystemVerilog."

Provide a name (":B0") for the initial block, which should be a
syntac no-op for already supported simulators, and allow also
running simulations under iverilog.

Signed-off-by: Gabriel Somlo <gsomlo@gmail.com>

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/resources/vsrc/AsyncResetReg.v
+++ b/src/main/resources/vsrc/AsyncResetReg.v
@@ -54,7 +54,7 @@ input  wire rst;
    // that, yet Chisel codebase is absolutely intolerant
    // of Xs.
 `ifndef SYNTHESIS
-  initial begin
+  initial begin:B0
     `ifdef RANDOMIZE
     integer    initvar;
     reg [31:0] _RAND;


gsomlo:
This would allow simulation with iverilog as described in issue #1804

jackkoenig:
This looks fine to me.

Since you filed this PR against rocket-chip, I'm assuming you don't have an issue with the FIRRTL-generated Verilog. Am I correct that this is due to either the fact that FIRRTL surrounds the entire initial block with `ifdef RANDOMIZE` (which you presumably are not using) or perhaps that it doesn't define any variables in the initial blocks?

gsomlo:
On Thu, Jan 31, 2019 at 01:45:33PM -0800, Jack Koenig wrote:
> This looks fine to me.
> 
> Since you filed this PR against rocket-chip, I'm assuming you don't have an
> issue with the FIRRTL-generated Verilog.

Correct.

> Am I correct that this is due to either the fact that FIRRTL surrounds
> the entire initial block with ifdef RANDOMIZE (which you presumably are
> not using) or perhaps that it doesn't define any variables in the
> initial blocks?

I *am* using "ifdef RANDOMIZE", which surrounds the variable
declaration(s) within the initial block that iverilog is complaining
about.

jackkoenig:
Alright well this looks good to me! Thank you for your contribution!

PR #1821, opened 2019-02-06
Get L2 TLB parity check off the critical path

aswaterman:


hcook:
l2_valid was returned but not used?

Otherwise lgtm

aswaterman:
Yeah, it became disused last year but I forgot to prune it.

PR #1822, opened 2019-02-06
Add ITIM name

derekpappas:
**Related issue**: 

**Release Notes**

Changed cache name from ICache to ITIM if there is a memory region

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/model/OMRocketCore.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMRocketCore.scala
@@ -58,7 +58,7 @@ object OMCaches {
 
   def icache(p: ICacheParams, resourceBindings: Option[ResourceBindings]): OMICache = {
     OMICache(
-      memoryRegions = resourceBindings.map(DiplomaticObjectModelAddressing.getOMMemoryRegions("ICache", _)).getOrElse(Nil),
+      memoryRegions = resourceBindings.map(DiplomaticObjectModelAddressing.getOMMemoryRegions("ITIM", _)).getOrElse(Nil),
       interrupts = Nil,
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomaticobjectmodel/model/OMRocketCore.scala
```diff
@@ -58,7 +58,7 @@ object OMCaches {
 
   def icache(p: ICacheParams, resourceBindings: Option[ResourceBindings]): OMICache = {
     OMICache(
-      memoryRegions = resourceBindings.map(DiplomaticObjectModelAddressing.getOMMemoryRegions("ICache", _)).getOrElse(Nil),
+      memoryRegions = resourceBindings.map(DiplomaticObjectModelAddressing.getOMMemoryRegions("ITIM", _)).getOrElse(Nil),
```
jackkoenig: Is the implication that the ICache only has a memory region if it's configured as an ITIM? Is there a similar issue above with the DCache/DTIM?

richardxia: Yes to the first question and no to the second. When an I-Cache is (Chisel-time) configured as an ITIM, it allows software to lock some of the cache ways at runtime. Note that not all of the ways can be locked, so part of the I-Cache must always function like a cache rather than a TIM.

The D-Cache/DTIM does not function this way, since it is only a Chisel-time configuration and not a software runtime configuration. The DTIM is also configured fully as a scratchpad and does not have the same way-locking functionality as the ITIM.

jackkoenig: Thanks for the clarification!

PR #1823, opened 2019-02-06
Added ExportDisableDebug flag to enable optional disable input

ernie-sifive:

**Release Notes**

New ExportDisableDebug flag allows an implementation to add an additional input to the debug interface, disableDebug, which disables JTAG debug access.


mwachs5:
@richardxia @derekpappas do we need to update the object model for this?

PR #1824, opened 2019-02-06
Global om

derekpappas:
**Related issue**: 

**Release Notes**

Support the global object model


src/main/scala/diplomaticobjectmodel/DiplomaticObjectModel.scala
```diff
@@ -0,0 +1,25 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.diplomaticobjectmodel
+
+import freechips.rocketchip.diplomaticobjectmodel.model.OMComponent
+
+import scala.collection.mutable.ListBuffer
+
+class DomCollector {
```
richardxia: It's a little confusing that you use the acronym "DOM" in this file but "OM" everywhere else. Can you rename both the class and the instance variables to be consistent with the existing convention?

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
```diff
@@ -6,6 +6,7 @@ import java.io.{File, FileWriter}
 
 import Chisel.{Data, Vec, log2Ceil}
 import freechips.rocketchip.diplomacy.{AddressRange, AddressSet, Binding, Device, DiplomacyUtils, ResourceAddress, ResourceBindings, ResourceBindingsMap, ResourceInt, ResourceMapping, ResourcePermissions, ResourceValue, SimpleDevice}
+import freechips.rocketchip.diplomaticobjectmodel._
```
richardxia: This wildcard import presumably is unnecessary because nothing else in this file changed.

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -33,8 +33,10 @@ abstract class BareSubsystemModuleImp[+L <: BareSubsystem](_outer: L) extends La
   ElaborationArtefacts.add("plusArgs", PlusArgArtefacts.serialize_cHeader)
   println(outer.dts)
 
-  def addOMArtefacts(): Unit = {
-    ElaborationArtefacts.add("objectModel.json", outer.objectModelJson)
+  def addOMArtefacts(components: Seq[OMComponent]): Unit = {
```
richardxia: I'm not understanding this API. Why are you passing a Seq of OMComponents into this method, then getting OMComponents that the BareSubsystem knows about, and finally creating an ElaborationArtefact? The part that confuses me the most is that you expect to grab a collection of OMComponents from elsewhere, so why is the BareSubsystem so special that it gets to both grab its own OMComponents and also create the ElaborationArtefact but the other places that have OMComponents aren't that special?

What if you just have the BareSubsystem be in charge of grabbing and returning its OMComponents (or registering its OMComponents with the DomCollector) and then have some other module be responsible for emitting the ElaborationArtefact?

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModel.scala
derekpappas: Fixed

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
derekpappas: Fixed

src/main/scala/subsystem/BaseSubsystem.scala
derekpappas: The L2$ cache is created with a mixin above the EnterpriseBasSubSystem.
Lets talk in person.

derekpappas: changed the API

PR #1825, opened 2019-02-07
Add utility groupByIntoSeq and use instead of groupBy for determinism

jackkoenig:
`groupBy` considered harmful.
`groupByIntoSeq` considered wonderful.

Follow on to https://github.com/freechipsproject/rocket-chip/pull/1713

**Related issue**: 

**Type of change**: [non-determinism] bug fix

**Release Notes**


jackkoenig:
Scala 2.13 (not yet released but supposed to be this month) fixes iterating on the keys of a `LinkedHashMap`, but as far as I know, there is no change to `groupBy` returning a `Map`

Issue #1817, opened 2019-02-03
Remove obsolete WireInit. Use WireDefault instead


tampler:
**WireInit** is outphased from recent Chisel3. Update DCache.scala to use **WireDefault** instead

```shell
[error] /home/bku/work/ndyne/fusion/src/main/scala/rocket/DCache.scala:13:8: object WireInit is not a member of package chisel3.core
[error] import chisel3.core.{ DontCare, WireInit }
[error]        ^
[error] /home/bku/work/ndyne/fusion/src/main/scala/rocket/DCache.scala:176:24: not found: value WireInit
[error]     val uncachedResp = WireInit(new HellaCacheReq, DontCare)
```

PR #1827, opened 2019-02-07
Avoid import from chisel3.core

edwardcwang:
**Related issue**: #1817

**Type of change**: bugfix

Fix incompability with Chisel3 master by not importing from `chisel3.core`


edwardcwang:
This appears to build and elaborate locally.

PR #1829, opened 2019-02-09
DTM: Don't accept DMI response when busy

p12nGH:
This PR fixes the following problem:
Transition throgh JTAG DR Capture state accepts DMI response even when sticky DMI busy bit is set.
This causes subsequent transaction to accept DMI response even after DMI Busy error, which, in turn, loses DMI Read data from previous transaction. It makes it difficult to recover from DMI Busy error in some cases.
Additionally, this PR removes duplicate code in DTM.

**Type of change**: bug fix
**Release Notes**: None.

### DIFF
--- a/src/main/scala/devices/debug/DebugTransport.scala
+++ b/src/main/scala/devices/debug/DebugTransport.scala
@@ -234,7 +234,7 @@ class DebugTransportModuleJTAG(debugAddrBits: Int, c: JtagDTMConfig)
       // for write operations confirm resp immediately because we don't care about data
       io.dmi.resp.valid,
       // for read operations confirm resp when we capture the data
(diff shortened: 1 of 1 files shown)


src/main/scala/devices/debug/DebugTransport.scala
```diff
@@ -193,16 +190,6 @@ class DebugTransportModuleJTAG(debugAddrBits: Int, c: JtagDTMConfig)
   // Debug Access Chain Implementation
 
   dmiAccessChain.io.capture.bits := Mux(busy, busyResp, Mux(io.dmi.resp.valid, dmiResp, nopResp))
-  when (dmiAccessChain.io.update.valid) {
```
mwachs5: well that was unnecessary!

PR #1830, opened 2019-02-09
Support CFLUSH.D.L1 rs1 address argument

aswaterman:
When rs1 = x0, it flushes the whole cache.

When rs1 != x0, it flushes the line containing the address in rs1.


hcook:
Any particular reason you used the size/`typ` field instead of the anciently-preallocated M_FLUSH `cmd`?  e.g. `val s1_flush_line = s1_req.cmd === M_FLUSH_ALL && s1_req.typ(0)` vs `s1_req.cmd === M_FLUSH`?

aswaterman:
Just for control-logic cost reasons because it's symmetric with SFENCE. `rs1 != x0` is already wired up to `typ(0)` for that instruction, and SFENCE also follows the pattern of using the same `M_SFENCE` command for both the full-flush and by-address variants of the instruction.

It would be less confusing of both variants of CFLUSH used `M_FLUSH` and we deleted the `M_FLUSH_ALL` command, but that is a more involved refactoring that I would like to separate into a future PR that makes no functional changes.

PR #1836, opened 2019-02-12
OMISA user fix

derekpappas:
**Related issue**: 

**Impact**: API modification

**Release Notes**

The OMISA was missing the user string in the case the useVM was set

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/model/OMISA.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMISA.scala
@@ -74,7 +74,7 @@ object OMISA {
       f = coreParams.fpu.map(x => isaExtSpec(F, "2.0")),
       d = coreParams.fpu.filter(_.fLen > 32).map(x => isaExtSpec(D, "2.0")),
       c = coreParams.useCompressed.option(isaExtSpec(C, " 2.0")),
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomaticobjectmodel/model/OMISA.scala
```diff
@@ -74,7 +74,7 @@ object OMISA {
       f = coreParams.fpu.map(x => isaExtSpec(F, "2.0")),
       d = coreParams.fpu.filter(_.fLen > 32).map(x => isaExtSpec(D, "2.0")),
       c = coreParams.useCompressed.option(isaExtSpec(C, " 2.0")),
-      u = coreParams.useUser.option(isaExtSpec(U, "1.10")),
+      u = (coreParams.useVM || coreParams.useUser).option(isaExtSpec(U, "1.10")),
```
richardxia: It seems like this is duplicating the logic encoded in https://github.com/freechipsproject/rocket-chip/blob/ad61beed8a4c7593bc91f0c8ea51e927616b50f8/src/main/scala/tile/BaseTile.scala#L41. It would be nice if only one place described that logic, but I don't think there's a quick fix to do that because this object uses just the RocketCoreParams and the line I linked to uses the RocketTileParams.

PR #1838, opened 2019-02-12
Vivado bram fix

erikdanie:
**Related issue**: Vivado cannot properly infer large Block Rams

**Release Notes**

### DIFF
--- a/scripts/vlsi_mem_gen
+++ b/scripts/vlsi_mem_gen
@@ -167,12 +167,15 @@ def gen_mem(name, width, depth, mask_gran, mask_seg, ports):
     prefix = 'RW%d_' % idx
     sequential.append('always @(posedge %sclk)' % prefix)
     sequential.append("  if (%sen && %swmode) begin" % (prefix, prefix))
-    if mask_seg > 0:
-      sequential.append("      for(i=0;i<%d;i=i+1) begin" % mask_seg)
-      sequential.append("        if(%swmask[i]) begin" % prefix)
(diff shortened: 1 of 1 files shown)


scripts/vlsi_mem_gen
```diff
@@ -167,10 +167,15 @@ def gen_mem(name, width, depth, mask_gran, mask_seg, ports):
     prefix = 'RW%d_' % idx
     sequential.append('always @(posedge %sclk)' % prefix)
     sequential.append("  if (%sen && %swmode) begin" % (prefix, prefix))
-    for i in range(mask_seg):
-      mask = ('if (%swmask[%d]) ' % (prefix, i)) if pid in maskedports else ''
-      ram_range = '%d:%d' % ((i+1)*mask_gran-1, i*mask_gran)
-      sequential.append("    %sram[%saddr][%s] <= %swdata[%s];" % (mask, prefix, ram_range, prefix, ram_range))
+    if mask_seg > 0:
```
aswaterman: Is it actually possible for mask_seg to be less than 1?  If it's not supposed to be, I think this should be an assertion rather than an `if`.

aswaterman:
when it comes time to merge, please squash & merge rather than creating a merge commit

scripts/vlsi_mem_gen
erikdanie: Well, because the previous one did `for i in range (mask_seg)` I essentially wanted to only do this if maskseg is nonzero. I could just do `if maskseg`

aswaterman: I see, no it's fine

PR #1845, opened 2019-02-15
Add external global interrupts

derekpappas:
**Related issue**: 

**Release Notes**

This PR adds support for generating the object model global external interrupts in the DiplomaticObjectModelAddressing object.

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
+++ b/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
@@ -204,7 +204,7 @@ object DiplomaticObjectModelAddressing {
       )
     }
 
-  private def getInterruptNumber(r: ResourceValue): BigInt = {≥…
+  private def getInterruptNumber(r: ResourceValue): BigInt = {
     r match {
       case ResourceInt(value: BigInt) => value
       case _ => throw new IllegalArgumentException
--- a/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
+++ b/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
@@ -204,7 +204,7 @@ object DiplomaticObjectModelAddressing {
       )
     }
 
-  private def getInterruptNumber(r: ResourceValue): BigInt = {
+  private def getInterruptNumber(r: ResourceValue): BigInt = {≥…
     r match {
       case ResourceInt(value: BigInt) => value
       case _ => throw new IllegalArgumentException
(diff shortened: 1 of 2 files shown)


src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
```diff
@@ -225,9 +225,21 @@ object DiplomaticObjectModelAddressing {
...
+      numberAtReceiver = getInterruptNumber(b.value),
+      name = name
+    )
+  }
+
+  def describeGlobalInterrupts(name: String, resources: ResourceBindings): Seq[OMInterrupt] = {
+    val bindings = resources("int")
+    for {
+      binding <- bindings
+      device = binding.device.get
```
richardxia: Is `binding.device` an `Option`? I'm trying to infer based on the call to `.get` here. If it is, you probably want to write this like:

```scala
for {
  binding <- bindings
  device <- binding.device
} yield ...
```

The difference is that the way you wrote it, if the `binding.device` is `None`, it'll throw an exception because you're calling `.get` on it, but the way I wrote it, if it's a `None`, it'll simply skip over to the next `binding`, treating it essentially like an empty `Seq`.

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
```diff
@@ -225,9 +225,21 @@ object DiplomaticObjectModelAddressing {
...
-    }
-}
+      receiver = getDeviceName(gp, resources),
+      numberAtReceiver = getInterruptNumber(b.value),
+      name = name
+    )
+  }
+
+  def describeGlobalInterrupts(name: String, resources: ResourceBindings): Seq[OMInterrupt] = {
+    val bindings = resources("int")
```
richardxia: This method name includes "global", but I'm not sure I see where you separate global interrupts from local ones.

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
khannaudit: I don't follow why `describeInterrupts` uses `device.get.parent`, but  `descibeGlobalInterrupts` stops at `device.get`. We need the parent device in all cases I can think of. The same also makes the `receiver` property inconsistent.

derekpappas: I mentioned that the Device that these interrupts are extracted from is not the same type of device as the PLIC device which gets the parent. I do not see a way to get the target (PLIC) device using the Interrupts device.

PR #1847, opened 2019-02-16
BundleBridge IOs

grebe:
I frequently use chisel-testers to unit test blocks with diplomatic interfaces. Making IOs for these modules that usually live deep in the hierarchy is tedious and error prone, and means that I'm not really testing the block that I will be using.

`BundleBridge` is a feature that helps automate this process- a `BundleBridge` uses diplomacy for arbitrary bundles. In this PR, I create a diplomatic converter that converts a AXI4 or TL node to a BundleBridge of the `AXI4Bundle` or `TLBundle`. In terms of hardware, this just adds wires to rename signals.

Here, I only have TL and AXI4 implementations to get feedback. If reviews are favorable, I'll implement APB and AHB.

To test the functionality yourself, add one of these and run

```
package freechips.rocketchip.amba.axi4

import chisel3._
import chisel3.internal.sourceinfo.SourceInfo
import freechips.rocketchip.util.AsyncQueueParams

class AXI4IOModule extends LazyModule()(Parameters.empty) {
  val mparams = AXI4MasterPortParameters(Seq(AXI4MasterParameters("slave")))
  val sparams = AXI4SlavePortParameters(Seq(AXI4SlaveParameters(Seq(AddressSet(0x0, 0xff)), supportsRead = TransferSizes(8, 8))), 8)
  val s = AXI4IO.slave(sparams)
  val m = AXI4IO.master(mparams)

  s._2 := m._2

  lazy val module = new LazyModuleImp(this)
}

object AXI4IOModuleMain extends App {
  val lm = LazyModule(new AXI4IOModule)
  chisel3.Driver.execute(Array[String](), () => lm.module)
}
```

```
package freechips.rocketchip.tilelink

class TLIOModule extends LazyModule()(Parameters.empty) {
  val cparams = TLClientPortParameters(Seq(TLClientParameters("slave")))
  val mparams = TLManagerPortParameters(Seq(TLManagerParameters(Seq(AddressSet(0x0, 0xff)), supportsGet = TransferSizes(8, 8))), 8)
  val c = TLIO.client(cparams)
  val m = TLIO.manager(mparams)

  m._2 := c._2

  lazy val module = new LazyModuleImp(this)
}

object TLIOModuleMain extends App {
  val lm = LazyModule(new TLIOModule)
  chisel3.Driver.execute(Array[String](), () => lm.module)
}
```

**Release Notes**

Add BundleBridge converters for AXI4 and TileLink.


hcook:
All diplomatic edge types have the ability to be bridged across module hierarchies as a built-in feature. The mechanism to do the wiring is built into `LazyModuleImp` itself (https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/diplomacy/LazyModule.scala#L171). In other words, `BundleBridge` just takes a capability that was already available for TileLink and AXI4 and makes it available to any generic Bundle, without the extra parameter negotiation and node definition boilerplate that those more advanced protocol types require. BundleBridge is useful, but using it to add a capability to TileLink and AXI that they already have built-in does not seem to make sense to me.

Can you provide some examples of use cases with chiselTesters code?

grebe:
I understand that diplomatic edges can bridge module boundaries, but unless I'm missing something I believe what you've referenced only works if there are two diplomatic endpoints to form an edge where the edge spans a module boundary. Is that incorrect? The use-case I'm interested in is for modules that are normally (i.e. in a full design) child modules with nodes that connect to parents. However, when you want to unit test blocks like this, there is no parent module with edges to connect to.

You need to create a master or slave node in the top level module and then make IOs for the new node. I don't believe the auto bundle includes nodes like this- doesn't it only include nodes if they have an edge to a parent?

[Here's](https://github.com/ucb-bar/dsptools/blob/master/rocket/src/test/scala/dspblocks/BasicBlockTesters.scala#L11) an example from dsptools. It uses the mixin [StandaloneBlock](https://github.com/ucb-bar/dsptools/blob/master/rocket/src/main/scala/dspblocks/DspBlock.scala#L44) that has a simple version of what is in this PR. The test I've included then has a trait that defines `resetMem()`, `writeAddr`, `readAddr`, etc. to drive the top level axi interface.

I suppose an alternative that doesn't use bundle bridges is to add `makeIO()` to AXI4, TL, etc. nodes, but I think it would have the same API and I think this approach is a bit DRYer.

hcook:
I think the diplomatic node abstract types `SourceNode` and `SinkNode` already fulfill the "make a node that doesn't have to have a parental := in order to produce an IO bundle" functionality. See e.g. https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/amba/axi4/Nodes.scala#L23, which we use to make ports at the top of the SoC DUT without requiring the test harness module to have to be a `LazyModule` itself. You don't have to add `makeIO` to every existing node type in order to use this functionality.

So, assuming you have SourceNodes and SinkNodes defined, couldn't `StandaloneBlock` just have: 
```
  val inNode = AXIStreamSourceNode(AXI4StreamBundleParameters(n = 8))
  val outNode = AXIStreamSinkNode(...)
  outNode := streamNode := inNode
  val in = InModuleBody { inNode.bundle }
  val out = InModuleBody { outNode.bundle }
```

Sidebar: It doesn't seem like you are even using the BundleBridge to cross module boundaries, but rather just to make new IOs in the DUT? Aren't you concerned that you are changing the DUT itself by making it a `StandaloneBlock` instead of whatever it was before? It seems like `StandaloneBlock` should be a parent `Module` capable of wrapping any DUT with a certain node type, and providing sources and/or sinks connected to the DUT's node. That way the DUT RTL will be identical, which is helpful when trying to collect metrics such as verilog line coverage.

grebe:
I think the example code you have written is close to what I'd like, but unfortunately I don't think it works. As written, `inNode.bundle` won't get an IO binding (I think it should actually be `inNode.out.head._1` b/c `bundle` is a function taking edge params). The linked `AXI4MasterNode` won't give you an `IO` node without doing something like [what `SimAXIMem` does here](https://github.com/freechipsproject/rocket-chip/blob/53dc5d2b8bcef41616373a751f6e9f0ae78d5eeb/src/main/scala/subsystem/Ports.scala#L226). 

On your sidebar, you are correct: I'm not using the bundlebridge to cross boundaries, I'm just using it because it defines `makeIO`. Perhaps the base sink nodes and source nodes should just define `makeIO()` too- I may be using an overly powerful tool for this job. The converters added by using bundlebridges are annoying.

Ideally the DUT RTL would be identical, but adding a wrapper module (which was my original approach) has annoying consequences too- boilerplate and indirection that made writing the tests more tedious, plus if your module has some non-diplomatic IOs then passing them up and down is annoying. Ideally, I think it would be nice to have a `LazyModule` with diplomatic nodes be OK as a top level module for testing.

The main issues are:
1) Need to add source/sink nodes
2) Need to make IOs for the added source/sink nodes (I think exists at the top level only for bundle bridge)
3) For RTL to be same as the full design (barring const-prop, etc.), standalone names need to be same as non-standalone names. If that means everything gets an auto name, I think that's OK. I currently give up on this issue b/c I thought it would be tricky to add `makeIO` to all source and sink nodes, but now I'm not so sure.

src/main/scala/diplomacy/BundleBridge.scala
```diff
@@ -30,7 +41,7 @@ case class BundleBridgeSink[T <: Data]()(implicit valName: ValName) extends Sink
   }
 }
 
-case class BundleBridgeSource[T <: Data](gen: () => T)(implicit valName: ValName) extends SourceNode(new BundleBridgeImp[T])(Seq(BundleBridgeParams(gen)))
+trait BundleBridgeBaseSource[T <: Data] extends SourceNode[BundleBridgeNull, BundleBridgeNull, BundleBridgeParams[T], BundleBridgeParams[T], T]
```
jerryhethatday, line 44: why you change from `case class` to `trait`

PR #1859, opened 2019-03-01
adding rangefilter to the getOMMemoryRegions gating expression

derekpappas:
**Related issue**: 

**Impact**: API modification

**Release Notes**

Added an additional gating condition to the  getOMMemoryRegions memory region creation function. This condition fixes a bug where certain memories were not included in the memory regions.

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
+++ b/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
@@ -168,7 +168,7 @@ object DiplomaticObjectModelAddressing {
 
   def getOMMemoryRegions(name: String, resourceBindings: ResourceBindings, omRegMap: Option[OMRegisterMap] = None): Seq[OMMemoryRegion]= {
     resourceBindings.map.collect {
-      case (x: String, seq: Seq[Binding]) if (DiplomacyUtils.regFilter(x)) || (DiplomacyUtils.rangeFilter(x))=>
+      case (x: String, seq: Seq[Binding]) if (DiplomacyUtils.regFilter(x) || DiplomacyUtils.rangeFilter(x)) =>
         seq.map {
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
```diff
@@ -168,7 +168,7 @@ object DiplomaticObjectModelAddressing {
 
   def getOMMemoryRegions(name: String, resourceBindings: ResourceBindings, omRegMap: Option[OMRegisterMap] = None): Seq[OMMemoryRegion]= {
     resourceBindings.map.collect {
-      case (x: String, seq: Seq[Binding]) if (DiplomacyUtils.regFilter(x)) =>
+      case (x: String, seq: Seq[Binding]) if (DiplomacyUtils.regFilter(x)) || (DiplomacyUtils.rangeFilter(x))=>
```
richardxia: Nit: Put a space before the `=>` arrow.
Nit 2: I think the parentheses are unnecessary. Does the following suggestion compile?

```suggestion
      case (x: String, seq: Seq[Binding]) if DiplomacyUtils.regFilter(x) || DiplomacyUtils.rangeFilter(x) =>
```

derekpappas: Does the gating expression in getOMPortMemoryRegions need the same set of conditions?

richardxia: Oh, sorry, I didn't read this PR carefully enough and mixed it up with the other one. This change should *only* go on the ports, although it's probably not too harmful to put it on the memories.

derekpappas: Added

PR #1860, opened 2019-03-02
Change target of reg desc annotation to IO.

grebe:
Currently, the target of the reg desc annotation is the module with the register mapping. However, a module could have multiple memory interfaces with their own distinct regmaps, and downstream tools (especially for unit verification) may make a distinction between the interfaces.

Currently, the target of the annotation is the module so you can't tell which regmap would go with a particular memory interface. This commit resolves this issue by making the target of the annotation the bundle corresponding to the memory interface.

My particular motivation is work @chick and I are making to add firrtl passes that generate IPXact.

**Impact**: API modification

### DIFF
--- a/src/main/scala/tilelink/RegisterRouter.scala
+++ b/src/main/scala/tilelink/RegisterRouter.scala
@@ -85,7 +85,7 @@ case class TLRegisterNode(
     bundleIn.c.ready := Bool(true)
     bundleIn.e.ready := Bool(true)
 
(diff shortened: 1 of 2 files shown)


mwachs5 (commented):
i think this makes sense... will take a deeper look later today

mwachs5:
oh, this is really old. Sorry I just got tagged on this. Do you still want this reviewed @grebe , @chick or should I close this PR?

PR #1864, opened 2019-03-06
Logical tree

derekpappas:
**Related issue**: 

**Release Notes**


src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
```diff
@@ -0,0 +1,72 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.tile
```
hcook: This should be in package `diplomaticobjectmodel`

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
```diff
@@ -0,0 +1,72 @@
...
+  *
+  * @param parent The parent LogicalTree node
+  * @param child The child LogicalTree node
+  */
+case class LogicalTreeEdge(
+  parent: LogicalTree,
+  child: LogicalTree
+)
+
+object LogicalModuleTree {
```
hcook: What is Module-specific about this? Are logical tree different from the Module hierarchy tree, or not?

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
```diff
@@ -0,0 +1,72 @@
...
+    val roots = values.diff(keys)
+
+    assert(roots.size == 1, "Logical Tree contains more than one root.")
+    roots.head
+  }
+}
+
+case class Tree[A](parent: A, children: List[Tree[A]])
+
+object OMLogicalTreeTree {
```
hcook: Why is this its own object instead of just some more methods on `LogicalModuleTree`?

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
```diff
@@ -0,0 +1,72 @@
...
+  def makeTree(): Tree[LogicalTree] = {
+    val root: LogicalTree = LogicalModuleTree.findRoot()
+    val treeMap = LogicalModuleTree.getTreeMap()
+
+    def recurse(treeNode: LogicalTree): Tree[LogicalTree] = {
```
hcook: Shouldn't this be a method of `LogicalTree`? What is it used for?

src/main/scala/diplomaticobjectmodel/OMProcessor.scala
```diff
@@ -0,0 +1,16 @@
...
+
+package freechips.rocketchip.tile
+
+import freechips.rocketchip.diplomacy.ResourceBindingsMap
+import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelUtils
+import freechips.rocketchip.diplomaticobjectmodel.model.OMComponent
+import freechips.rocketchip.util.ElaborationArtefacts
+
+case object OMProcessor {
+  def process(resourceBindingsMap: ResourceBindingsMap): Unit = {
```
hcook: Why `process`?

src/main/scala/tile/RocketTile.scala
```diff
@@ -137,6 +137,14 @@ class RocketTile(
     }
   }
 
+  class RocketLogicalTree extends LogicalTree {
+    override def getOMComponents(resourceBindingsMap: ResourceBindingsMap, components: Seq[OMComponent]): Seq[OMComponent] = {
```
hcook: Why is `components` an argument here?

src/main/scala/tile/RocketTile.scala
```diff
@@ -137,6 +137,14 @@ class RocketTile(
     }
   }
 
+  class RocketLogicalTree extends LogicalTree {
```
hcook: Instead of calling `cpuDevice.getOMComponents` below, could you instead move the code from `getOMRocketCores` into `getOMComponents` here, make the other OM functions be methods on `RocketLogicalTree`, and finally make `RocketLogicalTree` take `(rocketParams: RocketTileParams, xlen: Int, ...)` as constructor arguments? 


That way, rather than adding another layer of shallow indirection, you would
1) be making the API for the OM metadata layer more explicit
2) encapsulate all functions it uses to generate information in one class
3) decouple yourself from the `Device`/`Description`/`Property` API, which AFAICT you aren't actually using / don't actually need

src/main/scala/diplomaticobjectmodel/OMProcessor.scala
```diff
@@ -0,0 +1,16 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.tile
+
+import freechips.rocketchip.diplomacy.ResourceBindingsMap
+import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelUtils
+import freechips.rocketchip.diplomaticobjectmodel.model.OMComponent
+import freechips.rocketchip.util.ElaborationArtefacts
+
+case object OMProcessor {
```
hcook: Is a processor different from a core? How so?

src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -335,6 +336,14 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
     def ccover(cond: Bool, label: String, desc: String)(implicit sourceInfo: SourceInfo) =
       cover(cond, s"PLIC_$label", "Interrupts;;" + desc)
   }
+
+  class PLICLogicalTree extends LogicalTree {
+    override def getOMComponents(resourceBindingsMap: ResourceBindingsMap, components: Seq[OMComponent]): Seq[OMComponent] = {
+      device.getOMComponents(resourceBindingsMap)
```
hcook: Seems like this should just have the current contents of `getOMPLIC`?

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
derekpappas: LogicalTree's are elements of the LogicalModuleTree.

I am open to any names that people feel are appropriate.

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
derekpappas: fixed

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
derekpappas: 1. First a list of   val edges = ArrayBuffer[LogicalTreeEdge]() is constructed.

2. Then def getTreeMap(): Map[LogicalTree, List[LogicalTree]] converts the list into a map which represents the tree.

3. Then the tree map is converted into a OMLogicalTree by the  def makeTree(): Tree[LogicalTree].

The division between the tree map and OMLogical tree was done to separate the concerns.

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
derekpappas: See the above comment.

src/main/scala/diplomaticobjectmodel/OMProcessor.scala
derekpappas: I had pipeline() before and was asked by a reviewer in a meeting to change this to process.

src/main/scala/tile/RocketTile.scala
derekpappas: Child components are the OM children of the current node and are added to the current node.

  def tree(t: Tree[LogicalTree], resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
    val childComponents =  t.children.flatMap(tree(_, resourceBindingsMap))
    t.parent.getOMComponents(resourceBindingsMap, childComponents)
  }

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTree.scala
```diff
@@ -0,0 +1,58 @@
...
+import freechips.rocketchip.diplomacy.{LazyModule, ResourceBindingsMap, SimpleDevice}
+import freechips.rocketchip.diplomaticobjectmodel.model._
+import freechips.rocketchip.rocket.{DCacheParams, Frontend, ScratchpadSlavePort}
+import freechips.rocketchip.tile.{RocketTileParams, TileParams, XLen}
+
+
+class RocketLogicalTree(
+  device: SimpleDevice,
+  tileParams: TileParams,
+  rocketParams: RocketTileParams,
```
hcook: `RocketTileParams extends TileParams`, so you should be able to use `rocketParams` anywhere you are using `tileParams` in the below code

src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -56,7 +56,7 @@ trait HasRocketTiles extends HasTiles
   }).toList
 
   def getOMRocketInterruptTargets(): Seq[OMInterruptTarget] =
-    rocketTiles.flatMap(c => c.cpuDevice.getInterruptTargets())
+    rocketTiles.flatMap(c => c.rocketLogicalTree.getInterruptTargets())
 
   def getOMRocketCores(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] =
     rocketTiles.flatMap(c => c.cpuDevice.getOMComponents(resourceBindingsMap))
```
hcook: Could this also come from `rocketLogicalTree`?

src/main/scala/tile/RocketTile.scala
```diff
@@ -75,68 +76,27 @@ class RocketTile(
...
       val Description(name, mapping) = super.describe(resources)
       Description(name, mapping ++ cpuProperties ++ nextLevelCacheProperty ++ tileProperties ++ dtimProperty ++ itimProperty)
     }
 
+    /**
+      * This function is for backwards compatiblity and will be removed in the future
+      *
+      * @param resourceBindingsMap
+      * @return
+      */
```
hcook: Are you going to remove this as part of this PR? Why is it necessary to keep it?

Instead of having:
`RocketSubsystem. getOMRocketCores` call 
`RocketTile.cpuDevice.getOMComponents` which makes a private `RocketLogicalTree` and then calls
 `RocketLogicalTree.getOMComponents` on it,
couldn't you just have `RocketSubsystem. getOMRocketCores` call
`rocketTiles.flatMap(c => c.rocketLogicalTree.getOMComponents(resourceBindingsMap))` directly?

src/main/scala/rocket/Frontend.scala
```diff
@@ -62,6 +63,15 @@ class Frontend(val icacheParams: ICacheParams, hartid: Int)(implicit p: Paramete
   val icache = LazyModule(new ICache(icacheParams, hartid))
   val masterNode = icache.masterNode
   val slaveNode = icache.slaveNode
+
+
+  def getOMICache(resourceBindingsMap: ResourceBindingsMap): OMICache = {
```
hcook: Why are the ICache and Dcache different from each other?

src/main/scala/diplomaticobjectmodel/logicaltree/OMLogicalTree.scala
```diff
@@ -0,0 +1,85 @@
+// See LICENSE.SiFive for license details.
+
+package diplomaticobjectmodel.logicaltree
+
+import freechips.rocketchip.diplomacy.ResourceBindingsMap
+import freechips.rocketchip.diplomaticobjectmodel.model.OMComponent
+import scala.collection.mutable.ArrayBuffer
+
+
+trait LogicalTree {
```
hcook: Should this be call `LogicalTreeNode`>

src/main/scala/diplomaticobjectmodel/logicaltree/OMLogicalTree.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  /**
+    * The getTreeMap function constructs a tree by converting the list of LogicalTreeEdges into a Map which
+    * contains each of the parents and their children
+    *
+    * @return
+    */
+  def getTreeMap(): Map[LogicalTree, List[LogicalTree]] = {
+    edges.groupBy(_.parent).map{ case (k, v) => (k, v.map(_.child).toList)}
```
hcook: Does this actually add all the children to the parent? Seems like it will produce a single key per parent, with a list holdings a single child, whichever one happened to be last in the ArrayBuffer

src/main/scala/tile/RocketTile.scala
derekpappas: I changed the API

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
hcook: I would say `LogicalTreeNodes` are part of a `LogicalTree`

src/main/scala/diplomaticobjectmodel/OMProcessor.scala
hcook: What is a "process" in this context? Not "processor"?

src/main/scala/diplomaticobjectmodel/logicaltree/RocketComponentLogicalTrees.scala
```diff
@@ -0,0 +1,87 @@
+package diplomaticobjectmodel.logicaltree
```
hcook: None of these component have anything to do with "Rocket" the core pipeline, so I would rename this file to DeviceLogicalTrees.scala

src/main/scala/diplomaticobjectmodel/logicaltree/RocketComponentLogicalTrees.scala
```diff
@@ -0,0 +1,87 @@
...
+          OMSpecification(
+            name = "The RISC-V Instruction Set Manual, Volume II: Privileged Architecture",
+            version = "1.10"
+          )
+        )
+      )
+    )
+  }
+
+  override def getOMComponents(resourceBindingsMap: ResourceBindingsMap, components: Seq[OMComponent]): Seq[OMComponent] = {
```
hcook: You pass in components as an argument and never use it.

src/main/scala/diplomaticobjectmodel/logicaltree/RocketComponentLogicalTrees.scala
```diff
@@ -0,0 +1,87 @@
...
+          )
+        ),
+        latency = 2, // TODO
+        nPriorities = nPriorities,
+        targets = Nil
+      )
+    )
+  }
+
+  override def getOMComponents(resourceBindingsMap: ResourceBindingsMap, components: Seq[OMComponent]): Seq[OMComponent] = {
```
hcook: You pass in components as an argument and never use it.

src/main/scala/diplomaticobjectmodel/logicaltree/RocketComponentLogicalTrees.scala
```diff
@@ -0,0 +1,87 @@
...
+          )
+        ),
+        nAbstractDataWords = cfg.nAbstractDataWords,
+        nProgramBufferWords = cfg.nProgramBufferWords,
+        interfaceType = OMDebug.getDebugInterfaceType(exportDebugJTAG, exportDebugCJTAG, exportDebugDMI),
+      )
+    )
+  }
+
+  override def getOMComponents(resourceBindingsMap: ResourceBindingsMap, components: Seq[OMComponent]): Seq[OMComponent] = {
```
hcook: You pass in components as an argument and never use it.

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTree.scala
```diff
@@ -0,0 +1,58 @@
...
+  }
+
+  def getInterruptTargets(): Seq[OMInterruptTarget] = {
+    Seq(OMInterruptTarget(
+      hartId = rocketParams.hartId,
+      modes = OMModes.getModes(rocketParams.core.useVM)
+    ))
+  }
+
+  override def  getOMComponents(resourceBindingsMap: ResourceBindingsMap, components: Seq[OMComponent]): Seq[OMComponent] = {
```
hcook: You pass in components as an argument and never use it.

src/main/scala/diplomaticobjectmodel/logicaltree/OMLogicalTree.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  def findRoot(): LogicalTree = {
```
hcook: Only put `()` after functions that have side effects

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
hcook: So there are three or four data structures here, some of which have their own object whose only function is to convert it to another. I don't understand why there are so many.

1. Couldn't you just serialize out the edge-map structure itself? Why do we need to turn it into a actual immutable nested tree data structure? If downstream tools are going to ask graphical questions about it, they are probably just going to turn it back into an edge map. Plus, it looks like only the final `Seq[OMComponents]` gets serialized anyways, so why even do all these conversions?
2. What value does `Tree[LogicalTree]` add over `Map[LogicalTree, List[LogicalTree]]`? If the latter is too long a type name, you can define a local type member `type LogicalTree = Map[LogicalTreeNode, Seq[LogicalTreeNode]]`
3. Why doesn't `OMLogicalTree.makeTree()` just take the `resourceBindings` as an argument? As is, both `OMLogicalTree.makeTree` and `OMTree.tree` both have to recurse through the whole tree, once each.

Overall I think `Map[LogicalTreeNode, List[LogicalTreeNode]]` is probably the fundamental data being captured here, and I think if you used a mutable Map and re-wrote the `add` function you could build it as you go instead of computing it later. Ultimately, the only thing being serialized is the `Seq[OMComponent]`, so it seems sufficient to do:

```
object OMLogicalTree {
  private val tree  = mutable.Map[LogicalTreeNode, Seq[LogicalTreeNode]].empty
  def add(edge: (LogicalTreeNode, LogicalTreeNode)): Unit = {
    tree += ((edge._1, tree(edge._1) + edge._2) // ... or something like this 
  }
  def roots: Seq[LogicalTreeNode] = { ... }
  def checkIsTree: Boolean = roots.size == 1
  def bind(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
    def recurse(node: LogicalTreeNode): Seq[OMComponent] = 
       node.getOMComponents(resourceBindingsMap, tree(node).flatMap(recurse))
    recurse(root)
  }
}
```

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTree.scala
hcook: What I mean is, shouldn't `OMICache`, `OMDCache`, etc, be passed in here via `components: Seq[OMComponent]`? Otherwise what was the point of recursing through the tree?

src/main/scala/diplomaticobjectmodel/OMProcessor.scala
derekpappas: Not a processor (cpu/core/micro engine). This is the function that runs the pipeline that creates the OM.

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
derekpappas: Certainly your solution works given the current of requirements. I am leaving room for expansion later to create graphs, such as interrupt, clock, power,  between the tree nodes.

hcook: If they are Graphs, why are they called Trees?

src/main/scala/diplomaticobjectmodel/logicaltree/RocketComponentLogicalTrees.scala
derekpappas: fixed

src/main/scala/diplomaticobjectmodel/OMLogicalTree.scala
derekpappas: Changed to LogicalTreeNode

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTree.scala
derekpappas: fixed

src/main/scala/tile/RocketTile.scala
```diff
@@ -75,68 +76,27 @@ class RocketTile(
...
-        hartIds = Seq(hartId),
-        hasVectoredInterrupts = true,
-        interruptLatency = 4,
-        nLocalInterrupts = coreParams.nLocalInterrupts,
-        nBreakpoints = coreParams.nBreakpoints,
-        branchPredictor = rocketParams.btb.map(OMBTB.makeOMI),
-        dcache = omDCache,
-        icache = omICache
-      ))
+      val rocketLogicalTree: RocketLogicalTree = new RocketLogicalTree(cpuDevice, tileParams, rocketParams, frontend, dtim_adapter, p(XLen))
```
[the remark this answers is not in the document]
derekpappas: fixed

src/main/scala/diplomaticobjectmodel/logicaltree/OMLogicalTree.scala
derekpappas: fixed

src/main/scala/diplomaticobjectmodel/logicaltree/OMLogicalTree.scala
derekpappas: fixed

src/main/scala/diplomaticobjectmodel/logicaltree/RocketComponentLogicalTrees.scala
richardxia: What is "fixed" about this? The file still isn't renamed.

src/main/scala/diplomaticobjectmodel/logicaltree/OMLogicalTree.scala
richardxia: What is "fixed" about this? The parentheses still exist.

PR #1865, opened 2019-03-07
Bump hardfloat, clean up build.

grebe:
Hardfloat used to require Chisel 2 for testing, which is only published for Scala 2.11. Rocket uses Scala 2.12, which meant the build system needed a workaround to make hardfloat use Scala 2.12 when being used as a part of rocket.

Hardfloat's tests have been updated to use Chisel 3, so the workaround is no longer needed. This commit bumps the hardfloat submodule and cleans up build.sbt to use Scala 2.12 only and Makefrag to not use `++2.12.4`.

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -16,7 +16,7 @@ CXX ?= g++
 CXXFLAGS := -O1
 JVM_MEMORY ?= 2G
 
-SBT ?= java -Xmx$(JVM_MEMORY) -Xss8M -XX:MaxPermSize=256M -jar $(base_dir)/sbt-launch.jar ++2.12.4
+SBT ?= java -Xmx$(JVM_MEMORY) -Xss8M -XX:MaxPermSize=256M -jar $(base_dir)/sbt-launch.jar
 SHELL := /bin/bash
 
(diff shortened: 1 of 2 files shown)


jackkoenig:
Hooray for ditching the annoying `++2.12.4`! May I ask why we're getting rid of the 2.11 crossbuild? I don't care that much, but there are likely people using rocket-chip with Scala 2.11

ucbjrl:
Unfortunately, it looks like some `2.12`isms have crept into rocket-chip (notably trailing commas), and it won't build under `2.11`.

I'm all for keeping it compatible with `2.11`, but this requires some diligence (i.e. test with both `2.11` and `2.12`) before accepting PRs.

jackkoenig:
If it doesn't work with 2.11 anyway then removing it from crossScalaVersions makes sense, at least it will be more explicit about it!

azidar:
Can this be merged?

aswaterman:
Still testing it.

aswaterman:
Looks good, thanks again @grebe

PR #1867, opened 2019-03-07
Refactor BaseTile again (to obviate augmentedTileParameters)

hcook:
This PR makes a number of modifications to the BaseTile class, with the goal of encapsulating all the legacy weirdness around propagating the keys from `augmentedTileParameters` through all tile submodules.

Specific changes:
- adds `BaseTile.visibilityNode`
- adds `HasNonDiplomaticTileParameters` trait
- replace `SharedMemoryTLEdge` with `TileVisibilityNodeKey`
- encapsulate tile Parameter augmentation within `BaseTile`
- simplify tile input constants
- provide some `LookupByHartImpl` implementations
- remove `subsystem.augmentedTileParams`

The main trick to getting the legacy key/values set inside the tile is to use a public constructor that captures and alters the implicit parameters to contain them, before handing it off to a private constructor that in turn propagates the altered parameters to the `LazyModule` base class, which then implicitly supplies them to all the child module constructors:
```
abstract class Tile private (..., q: Parameters) extends LazyModule()(q) {
  def this(...)(implicit p: Parameters) = this(..., p.alterMap(Map(Legacy -> ...)))
}
```

### DIFF
--- a/src/main/scala/groundtest/GroundTestSubsystem.scala
+++ b/src/main/scala/groundtest/GroundTestSubsystem.scala
@@ -18,12 +18,7 @@ case object TileId extends Field[Int]
 class GroundTestSubsystem(implicit p: Parameters) extends BaseSubsystem
     with CanHaveMasterAXI4MemPort {
   val tileParams = p(GroundTestTilesKey)
-  val tiles = tileParams.zipWithIndex.map { case(c, i) => LazyModule(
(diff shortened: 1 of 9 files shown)


hcook:
@terpstra I agree with you but I want to emphasize how much work that is going to be. Especially since it is an API incompatible change and  `CoreBundle` has propagated quite far through all processor code bases.

PR #1869, opened 2019-03-12
logical-tree1

derekpappas:
**Related issue**: 

**Release Notes**


hcook:
Why did you open a new PR? I don't think you addressed my concerns on the previous version #1864.
1.  Your implementation of three different tree data structures is confusing and more complicated than necessitated by the current requirements.
2. You left in a bunch of legacy `getOMComponents` on various devices that I think could be removed.
3. You treat the `OMICache` different than the `OMDCache`.
4. I'm not convinced that this `add` functionality is actually sufficient to decouple the designs, because you don't e.g., use it to add the ICache and DCache to the `RocketLogicalTree`. In fact no `components` are returned to their parent via the recursive tree walk.

derekpappas:
> Why did you open a new PR? I don't think you addressed my concerns on the previous version #1864.
> 
> 1. Your implementation of three different tree data structures is confusing and more complicated than necessitated by the current requirements.
> 2. You left in a bunch of legacy `getOMComponents` on various devices that I think could be removed.
> 3. You treat the `OMICache` different than the `OMDCache`.
> 4. I'm not convinced that this `add` functionality is actually sufficient to decouple the designs, because you don't e.g., use it to add the ICache and DCache to the `RocketLogicalTree`. In fact no `components` are returned to their parent via the recursive tree walk.

src/main/scala/devices/tilelink/CLINT.scala
```diff
@@ -112,13 +101,19 @@ class CLINT(params: CLINTParams, beatBytes: Int)(implicit p: Parameters) extends
       timeOffset       -> RegFieldGroup("mtime", Some("Timer Register"),
         RegField.bytes(time, Some(RegFieldDesc("mtime", "", reset=Some(0), volatile=true))))
     )
+
+    def getOMRegMap(): OMRegisterMap = omRegMap
   }
+
+  val clintLogicalTree: CLINTLogicalTreeNode = new CLINTLogicalTreeNode(device, module.getOMRegMap)
```
hcook: This could just be:
`val clintLogicalTree: CLINTLogicalTreeNode = new CLINTLogicalTreeNode(device, module.omRegMap)`

You don't need the extra getter function, and pretty much never should in Scala.

src/main/scala/rocket/Frontend.scala
```diff
@@ -14,6 +14,7 @@ import freechips.rocketchip.tile._
 import freechips.rocketchip.util._
 import freechips.rocketchip.util.property._
 import chisel3.internal.sourceinfo.SourceInfo
+import freechips.rocketchip.diplomaticobjectmodel.model.OMICache
```
hcook: Unneeded import

src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -50,12 +51,19 @@ trait HasRocketTiles extends HasTiles
...
+    * getOMRocketInterruptTargets and getOMRocketCores are deprecated
+    *
+    * @return
+    */
   def getOMRocketInterruptTargets(): Seq[OMInterruptTarget] =
-    rocketTiles.flatMap(c => c.cpuDevice.getInterruptTargets())
+    rocketTiles.flatMap(c => c.rocketLogicalTree.getInterruptTargets())
 
   def getOMRocketCores(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] =
     rocketTiles.flatMap(c => c.cpuDevice.getOMComponents(resourceBindingsMap))
```
hcook: I still don't understand why `getOMRocketCores` is using `cpuDevice`, and `getOMRocketInterruptTargets` is using `getOMRocketInterruptTargets`.

I think you should go ahead and remove all the deprecated functions now. If you need to compare the outputs to the old one, you can just generate the file on `master` and compare that way.

src/main/scala/tile/RocketTile.scala
```diff
@@ -111,13 +118,6 @@ class RocketTile private(
...
     }
 
-    def getInterruptTargets(): Seq[OMInterruptTarget] = {
-      Seq(OMInterruptTarget(
-        hartId = rocketParams.hartId,
-        modes = OMModes.getModes(rocketParams.core.useVM)
-      ))
-    }
-
     def getOMRocketCores(resourceBindingsMap: ResourceBindingsMap): Seq[OMRocketCore] = {
```
hcook: How come you were able to remove `getInterruptTargets` here but not `getOMRocketCores ` or `getOMICacheFromBindings`? Can't you just use the `RocketLogicalTreeNode` version like you did for the CLINT and Debug?

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
```diff
@@ -0,0 +1,53 @@
...
+package freechips.rocketchip.diplomaticobjectmodel.logicaltree
+
+import freechips.rocketchip.diplomacy.ResourceBindingsMap
+import freechips.rocketchip.diplomaticobjectmodel.model.OMComponent
+
+import scala.collection.mutable
+
+trait LogicalTreeNode {
+  def getOMComponents(resourceBindingsMap: ResourceBindingsMap, children: Seq[OMComponent]): Seq[OMComponent]
+}
```
hcook: I think we should add a trait: 
```
trait HasLogicalTreeNode { this: LazyModule =>
  val logicalTree: LogicalTreeNode
  def addLogicalTreeNode(child: LogicalTreeNode): Unit = { LogicalModuleTree.add(logicalTree, child) }
}
```
that the various LazyModule classes extend. This will be the hook by which we could later automatically add edges to the logical tree graph. It will be easier if all the LogicalModules have a member with the same name, instead of `clintLogicalTree`, `rocketLogicalTree` &etc.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
```diff
@@ -0,0 +1,61 @@
...
+
+package freechips.rocketchip.diplomaticobjectmodel.logicaltree
+
+import freechips.rocketchip.devices.debug._
+import freechips.rocketchip.devices.tilelink.{CLINT, TLPLIC}
+import freechips.rocketchip.diplomacy.{ResourceBindings, ResourceBindingsMap, SimpleDevice}
+import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelAddressing
+import freechips.rocketchip.diplomaticobjectmodel.model._
+
+class CLINTLogicalTreeNode(device: SimpleDevice, f: () => OMRegisterMap) extends LogicalTreeNode {
```
hcook: I think if you make this:
```
class CLINTLogicalTreeNode(device: SimpleDevice, f: => OMRegisterMap) extends LogicalTreeNode {
```
you can avoid having issues with lazy `module` evaluation without having to add in the the extra `CLINT.getOMRegMap` getter function

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -71,6 +73,14 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem {
       }
     }
   }
+
+  class BaseSubsystemLogicalTreeNode extends LogicalTreeNode {
+    override def getOMComponents(resourceBindingsMap: ResourceBindingsMap, cs: Seq[OMComponent]): Seq[OMComponent] = {
```
hcook: Should this be the default implementation of `getOMComponents` in `LogicalTreeNode`?

src/main/scala/rocket/Frontend.scala
derekpappas: fixed

src/main/scala/subsystem/RocketSubsystem.scala
derekpappas: I will remove all of the deprecated functions.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
derekpappas: I will check.

src/main/scala/subsystem/RocketSubsystem.scala
derekpappas: getOMRocketCores was removed.

The  deprecated functions should have been removed.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
derekpappas: ARE you suggesting that all LazyModules extend the trait HasLogicalTreeNode?

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
derekpappas: val clintLogicalTree: CLINTLogicalTreeNode = new CLINTLogicalTreeNode(device, module.getOMRegMap)
 works

src/main/scala/subsystem/BaseSubsystem.scala
derekpappas: I don't think so. There are only a few instances where the children are included.

src/main/scala/subsystem/RocketSubsystem.scala
derekpappas: WRT to the question about "getOMRocketInterruptTargets is using getOMRocketInterruptTargets." we should talk about the following:

```
  val ebssLogicalTree = new EBSSLogicalTreeNode(getOMCabooseInterruptTargets, getOMRocketInterruptTargets, getOMInterruptDevice)
  LogicalModuleTree.add(ebssLogicalTree, logicalTree)
```

```
 def getOMCabooseInterruptTargets(): Seq[OMInterruptTarget] =
    cabooseTiles.flatMap(c => c.cabooseLogicalTree.getOMInterruptTargets())
  cabooseTiles.map(c => LogicalModuleTree.add(logicalTree, c.cabooseLogicalTree))
```

```
  rocketTiles.map(r => LogicalModuleTree.add(logicalTree, r.rocketLogicalTree))
```

src/main/scala/tile/RocketTile.scala
derekpappas: Removed getOMRocketCores

PR #1873, opened 2019-03-12
Sba 64bit

ernie-sifive:

**Release Notes**
Address bits above 32 driven from the SBA into the system bus were incorrect.  This change uses the registered version of sbaddress1, which had to have been written earlier.  The write to sbaddress0 actually kicks off the TileLink transaction.

### DIFF
--- a/src/main/scala/devices/debug/SBA.scala
+++ b/src/main/scala/devices/debug/SBA.scala
@@ -101,7 +101,9 @@ object SystemBusAccessModule
       }
     }
 
-    sb2tl.module.io.addrIn := Mux(sb2tl.module.io.rdEn,Cat(SBADDRESSWrData.reverse),Cat(SBADDRESSFieldsReg.reverse))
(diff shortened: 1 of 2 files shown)


src/main/scala/devices/debug/SBA.scala
```diff
@@ -101,7 +101,9 @@ object SystemBusAccessModule
       }
     }
 
-    sb2tl.module.io.addrIn := Mux(sb2tl.module.io.rdEn,Cat(SBADDRESSWrData.reverse),Cat(SBADDRESSFieldsReg.reverse))
+    sb2tl.module.io.addrIn := Mux(sb2tl.module.io.rdEn,
+      Cat(SBADDRESSFieldsReg(3), SBADDRESSFieldsReg(2), SBADDRESSFieldsReg(1), SBADDRESSWrData(0)),
```
rmac-sifive: Good catch. But maybe instead of writing all the fields out by hand we could write:
`Cat(Cat(SBADDRESSFieldsReg.drop(1).reverse),SBADDRESSWrData(0))`

PR #1875, opened 2019-03-13
Make rocket publish a single jar.

grebe:
This puts macros + hardfloat in a single jar instead of being published as separate jars.

Is anyone relying on `rocket-macros` or `hardfloat` existing independently of rocket?

@ucbjrl, can you verify that this produces artifacts that look correct?


aswaterman:
Thanks @grebe. Works fine for us.  @ucbjrl feel free to merge if it accomplishes your goal

grebe:
Can someone merge this? I don't have the power

PR #1890, opened 2019-03-21
Logical tree3

derekpappas:
**Related issue**: 

**Release Notes**


src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -30,7 +31,8 @@ case object RocketCrossingKey extends Field[Seq[RocketCrossingParams]](List(Rock
 trait HasRocketTiles extends HasTiles
     with CanHavePeripheryPLIC
     with CanHavePeripheryCLINT
-    with HasPeripheryDebug { this: BaseSubsystem =>
+    with HasPeripheryDebug {
+  this: BaseSubsystem =>
```
hcook: This formatting doesn't match how we usually express self types; on the same line as the opening brace.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
```diff
@@ -0,0 +1,89 @@
...
+package freechips.rocketchip.diplomaticobjectmodel.logicaltree
+
+import freechips.rocketchip.devices.debug._
+import freechips.rocketchip.devices.tilelink.PLICParams
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelAddressing
+import freechips.rocketchip.diplomaticobjectmodel.model.{OMComponent, _}
+import freechips.rocketchip.rocket.{DCacheParams, ICacheParams, ScratchpadSlavePort}
+
+class CLINTLogicalTreeNode(device: SimpleDevice, f: () => OMRegisterMap) extends LogicalTreeNode {
```
hcook: Did you try making this: `class CLINTLogicalTreeNode(device: SimpleDevice, f: => OMRegisterMap)` and then getting rid of the extra `getOMRegMap` accessor?

hcook:
Can you close the older versions of this PR?

src/main/scala/subsystem/RocketSubsystem.scala
derekpappas: fixed

derekpappas: fixed

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
derekpappas: fixed. lets go over this on Monday for 30 seconds...

derekpappas:
@hcook we should talk about: "I think we should add a trait:

```trait HasLogicalTreeNode { this: LazyModule =>
  val logicalTree: LogicalTreeNode
  def addLogicalTreeNode(child: LogicalTreeNode): Unit = { LogicalModuleTree.add(logicalTree, child) }
}```

that the various LazyModule classes extend. This will be the hook by which we could later automatically add edges to the logical tree graph. It will be easier if all the LogicalModules have a member with the same name, instead of clintLogicalTree, rocketLogicalTree &etc."

File: src/main/scala/rocket/Frontend.scala [hunk 3]
```diff
@@ -343,12 +344,15 @@ class FrontendModule(outer: Frontend) extends LazyModuleImp(outer)
 }
 
 /** Mix-ins for constructing tiles that have an ICache-based pipeline frontend */
-trait HasICacheFrontend extends CanHavePTW { this: BaseTile =>
+trait HasICacheFrontend extends CanHavePTW {
```
hcook: Can you please undo this whitespace change

src/main/scala/subsystem/RocketSubsystem.scala
hcook: I don't see a change

src/main/scala/tile/RocketTile.scala
```diff
@@ -159,6 +105,9 @@ class RocketTile private(
     if (!rocketParams.boundaryBuffers) super.makeSlaveBoundaryBuffers
     else TLBuffer(BufferParams.flow, BufferParams.none, BufferParams.none, BufferParams.none, BufferParams.none)
   }
+
+  val rocketLogicalTree: RocketLogicalTreeNode = new RocketLogicalTreeNode(cpuDevice, rocketParams, dtim_adapter, p(XLen), iCacheLogicalTreeNode)
+  //LogicalModuleTree.add(rocketLogicalTree, iCacheLogicalTreeNode)
```
hcook: I think the commented out way is the ideal way of doing it; then it could be added to the tree at the point where the `ICache` is added, in `HasICacheFrontend`

src/main/scala/subsystem/RocketSubsystem.scala
derekpappas: Fixed.

src/main/scala/devices/tilelink/CLINT.scala
```diff
@@ -112,13 +91,19 @@ class CLINT(params: CLINTParams, beatBytes: Int)(implicit p: Parameters) extends
       timeOffset       -> RegFieldGroup("mtime", Some("Timer Register"),
         RegField.bytes(time, Some(RegFieldDesc("mtime", "", reset=Some(0), volatile=true))))
     )
+
+    def getOMRegMap(): OMRegisterMap = omRegMap
```
richardxia: This doesn't appear to be needed anymore.

PR #1897, opened 2019-03-23
Access regmaps

derekpappas:
**Related issue**: 

**Release Notes**

This PR will add access for regmaps.

### DIFF
--- a/src/main/scala/tilelink/RegisterRouter.scala
+++ b/src/main/scala/tilelink/RegisterRouter.scala
@@ -7,6 +7,7 @@ import chisel3.experimental.RawModule
 import firrtl.annotations.ModuleName
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.diplomaticobjectmodel.{DiplomaticObjectModelAddressing, DiplomaticObjectModelUtils}
 import freechips.rocketchip.diplomaticobjectmodel.model.{OMMemoryRegion, OMRegister, OMRegisterMap}
 import freechips.rocketchip.regmapper._
 import freechips.rocketchip.interrupts._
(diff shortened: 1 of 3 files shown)


src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -42,6 +43,13 @@ case class TLRegisterNode(
       s"TLRegisterNode addresses (${address}) must be aligned to its size ${size}")
   }
 
+  def getMemory(name: String, omRegMap : Option[OMRegisterMap], resourceBindingsMap: ResourceBindingsMap): Seq[OMMemoryRegion] = {
+    val resourceBindings = resourceBindingsMap.map.get(device)
+    assert(device != None, s"Device is null : ${device.label}")
```
richardxia: What's going on with this assertion? `device` is a `Device`, not an `Option[Device]`, so this check will never work. Note that `null` and `None` are two different concepts in Scala. `None` is one of the two implementations of the `Option` abstract class. `null` is the same thing as the Java `null`, and it's usually only encountered when interoperating with Java code.

src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -173,7 +181,7 @@ class TLRegisterRouter[B <: TLRegBundleBase, M <: LazyModuleImp](
 
 /** Mix this trait into a RegisterRouter to be able to attach its register map to a TL bus */
 trait HasTLControlRegMap { this: RegisterRouter[_] =>
-  protected val controlNode = TLRegisterNode(
+  val controlNode = TLRegisterNode(
```
hcook: I don't like that you had to remove this protection! That means you are making the API leakier. Why isn't `getMemory` a function here that accesses the node's device?

PR #1903, opened 2019-03-27
APBToTL Adapter

mwachs5:
There were already AHB2TL and AXI2TL adapters. Add one for APB.

Reviewers, please help with my "TODO" items, I am not sure I am following the pattern as intended.

**Related issue**: 

**Impact**: API addition (no impact on existing code) |

**Release Notes**

### DIFF
--- a/src/main/scala/amba/apb/ToTL.scala
+++ b/src/main/scala/amba/apb/ToTL.scala
@@ -1,0 +1,83 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.amba.apb
+
+import Chisel._
+import freechips.rocketchip.config.Parameters
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.util.MaskGen
+
+case class APBToTLNode()(implicit valName: ValName) extends MixedAdapterNode(APBImp, TLImp)(
+  dFn = { case APBMasterPortParameters(masters) =>
+    TLClientPortParameters(clients = masters.map { m =>
+      TLClientParameters(name = m.name, nodePath = m.nodePath)
+    })
+  },
+  uFn = { mp => APBSlavePortParameters(
+    slaves = mp.managers.map { m =>
+      APBSlaveParameters(
+        address       = m.address,
+        resources     = m.resources,
+        regionType    = m.regionType,
+        executable    = m.executable,
+        nodePath      = m.nodePath,
+        //TODO why and how should I do this adjust for APB?
+        // Also, shouldn't I use supportsPutPartial because of PSTRB?
+        supportsWrite = m.supportsPutFull,//adjust(m.supportsPutFull),
+        supportsRead  = m.supportsGet)},//adjust(m.supportsGet))},
+    beatBytes = mp.beatBytes)
+  }) {
+  //TODO: Shouldn't there be a requirement that TL side is wider than APB side?
+}
+
+class APBToTL()(implicit p: Parameters) extends LazyModule
+{
+  val node = APBToTLNode()
+
+  lazy val module = new LazyModuleImp(this) {
+    (node.in zip node.out) foreach { case ((in, edgeIn), (out, edgeOut)) =>
+      val beatBytes = edgeOut.manager.beatBytes
+
+      //-----------------------------------------------
+      // Flow Control
+      out.a.valid := in.psel && in.penable
+      in.pready := in.psel & in.penable & out.d.valid
+      out.d.ready := in.psel && in.penable // No backpressure to APB Master
+
+      // Write
+      // Question -- do I want PutPartial here or PutFull?
+      // I was thinking PutPartial because of the pstrb masking.
+      out.a.bits.opcode  := Mux(in.pwrite, TLMessages.PutPartialData, TLMessages.Get)
+      out.a.bits.param   := UInt(0)
+      out.a.bits.size    := UInt(log2Ceil(in.params.dataBits/8))
+      out.a.bits.source  := UInt(0)
+      out.a.bits.address := in.paddr
+      out.a.bits.data    := in.pwdata
+      val pstrbFill = beatBytes / (in.params.dataBits/8)
+      out.a.bits.mask    := MaskGen(out.a.bits.address, out.a.bits.size, beatBytes) | Cat(Seq.fill(pstrbFill)(in.pstrb).reverse)
+      out.a.bits.corrupt := Bool(false)
+      assert (in.pprot === 0.U, "Can't support APB PPROT != 0 in APB2TL conversion")
+      // Read
+      // TODO: do I need to slice this? What if they aren't same width?
+      in.prdata := out.d.bits.data
+
+      // Error
+      in.pslverr := out.d.bits.corrupt || out.d.bits.denied
+
+      // Unused channels
+      out.b.ready := Bool(true)
+      out.c.valid := Bool(false)
+      out.e.valid := Bool(false)
+    }
+  }
+}
+
+object APBToTL
+{
+  def apply()(implicit p: Parameters) =
+  {
+    val apb2tl = LazyModule(new APBToTL)
+    apb2tl.node
+  }
+}
--- a/src/main/scala/amba/apb/ToTL.scala
+++ b/src/main/scala/amba/apb/ToTL.scala
@@ -24,42 +24,50 @@ case class APBToTLNode()(implicit valName: ValName) extends MixedAdapterNode(APB
         nodePath      = m.nodePath,
         //TODO why and how should I do this adjust for APB?
         // Also, shouldn't I use supportsPutPartial because of PSTRB?
-        supportsWrite = m.supportsPutFull,//adjust(m.supportsPutFull),
-        supportsRead  = m.supportsGet)},//adjust(m.supportsGet))},
+        supportsWrite = m.supportsPutPartial.intersect(TransferSizes(1, mp.beatBytes)),
+        supportsRead  = m.supportsGet.intersect(TransferSizes(1, mp.beatBytes)))},
     beatBytes = mp.beatBytes)
-  }) {
-  //TODO: Shouldn't there be a requirement that TL side is wider than APB side?
-}
+  })
 
 class APBToTL()(implicit p: Parameters) extends LazyModule
 {
   val node = APBToTLNode()
 
+  
   lazy val module = new LazyModuleImp(this) {
     (node.in zip node.out) foreach { case ((in, edgeIn), (out, edgeOut)) =>
       val beatBytes = edgeOut.manager.beatBytes
 
       //-----------------------------------------------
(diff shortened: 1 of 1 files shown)


src/main/scala/amba/apb/ToTL.scala
```diff
@@ -0,0 +1,83 @@
...
+      out.a.bits.opcode  := Mux(in.pwrite, TLMessages.PutPartialData, TLMessages.Get)
+      out.a.bits.param   := UInt(0)
+      out.a.bits.size    := UInt(log2Ceil(in.params.dataBits/8))
+      out.a.bits.source  := UInt(0)
+      out.a.bits.address := in.paddr
+      out.a.bits.data    := in.pwdata
+      val pstrbFill = beatBytes / (in.params.dataBits/8)
+      out.a.bits.mask    := MaskGen(out.a.bits.address, out.a.bits.size, beatBytes) | Cat(Seq.fill(pstrbFill)(in.pstrb).reverse)
+      out.a.bits.corrupt := Bool(false)
+      assert (in.pprot === 0.U, "Can't support APB PPROT != 0 in APB2TL conversion")
```
ernie-sifive: Maybe ignore in.pprot rather than requiring a zero value?  Our own APB master drives pprot = 1.U.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  lazy val module = new LazyModuleImp(this) {
+    (node.in zip node.out) foreach { case ((in, edgeIn), (out, edgeOut)) =>
+      val beatBytes = edgeOut.manager.beatBytes
+
+      //-----------------------------------------------
+      // Flow Control
+      out.a.valid := in.psel && in.penable
+      in.pready := in.psel & in.penable & out.d.valid
+      out.d.ready := in.psel && in.penable // No backpressure to APB Master
```
terpstra: This is wrong. You can't just arbitrarily back-pressure D. I think what you actually want is:

 in.pready := out.d.valid
 out.d.ready := true.B
 assert (!out.d.valid || (in.penable && in.psel))

The APB spec guarantees that penable and psel stay high until pready is asserted by the slave.

src/main/scala/amba/apb/ToTL.scala
terpstra: Yeah, I'd just ignore it.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -0,0 +1,83 @@
...
+      // Question -- do I want PutPartial here or PutFull?
+      // I was thinking PutPartial because of the pstrb masking.
+      out.a.bits.opcode  := Mux(in.pwrite, TLMessages.PutPartialData, TLMessages.Get)
+      out.a.bits.param   := UInt(0)
+      out.a.bits.size    := UInt(log2Ceil(in.params.dataBits/8))
+      out.a.bits.source  := UInt(0)
+      out.a.bits.address := in.paddr
+      out.a.bits.data    := in.pwdata
+      val pstrbFill = beatBytes / (in.params.dataBits/8)
+      out.a.bits.mask    := MaskGen(out.a.bits.address, out.a.bits.size, beatBytes) | Cat(Seq.fill(pstrbFill)(in.pstrb).reverse)
```
terpstra: There are several things wrong with this logic.

First, the whole pstrbFill is pointless. out.manager.beatBytes must equal in.manager.beatBytes due to your (correct) Node rules. It then follows that beatBytes == in.params.dataBits/8.

Second, on a TileLink Get the mask must be the result you computed with MaskGen.

This is the correct line:
out.a.bits.mask := Mux(in.pwrite, in.pstrb, MaskGen(out.a.bits.address, out.a.bits.size, beatBytes))

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -0,0 +1,83 @@
...
+      out.a.bits.corrupt := Bool(false)
+      assert (in.pprot === 0.U, "Can't support APB PPROT != 0 in APB2TL conversion")
+      // Read
+      // TODO: do I need to slice this? What if they aren't same width?
+      in.prdata := out.d.bits.data
```
terpstra: They will always be the same width.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -0,0 +1,83 @@
...
+
+      //-----------------------------------------------
+      // Flow Control
+      out.a.valid := in.psel && in.penable
+      in.pready := in.psel & in.penable & out.d.valid
+      out.d.ready := in.psel && in.penable // No backpressure to APB Master
+
+      // Write
+      // Question -- do I want PutPartial here or PutFull?
+      // I was thinking PutPartial because of the pstrb masking.
```
terpstra: Your line of reasoning is correct.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -0,0 +1,83 @@
...
+{
+  val node = APBToTLNode()
+
+  lazy val module = new LazyModuleImp(this) {
+    (node.in zip node.out) foreach { case ((in, edgeIn), (out, edgeOut)) =>
+      val beatBytes = edgeOut.manager.beatBytes
+
+      //-----------------------------------------------
+      // Flow Control
+      out.a.valid := in.psel && in.penable
```
terpstra: This is completely wrong. psel and penable stay high for the whole access phase. TileLink a valid indicates a transfer each time you assert it. You need a state machine like this:

val sent = RegInit(false.B)
out.a.valid := in.psel && !sent
when (out.a.fire()) { sent := true.B }
when (out.d.fire()) { sent := false.B }

src/main/scala/amba/apb/ToTL.scala
terpstra: There's another problem, which is that TileLink might respond on the same cycle as the request. This is too early for APB to handle. However, the obvious fix of back-pressuring d.ready when !sent would violate the TileLink forward progress rules. The only way to be fully compliant would be a data-bus wide Queue to capture the D response. However, this corner case is probably not relevant most of the time, so I'd instead just add a requirement on diplomacy:

require (edgeOut.manager.minLatency >= 1)

mwachs5: It's not too early for APB to handle if you only look at PENABLE, right?  I thought that the first cycle (PSEL = 1) is sort of the "get ready" and "PENABLE = 1" means "go". And if that PREADY is high right away, then the read data is also ready.

In other words, PSEL is pretty much pointless for our conversion.

Though I could be reading the spec wrong.

src/main/scala/amba/apb/ToTL.scala
mwachs5: Ahhhh, the fact that penable stays high is the part I was missing from reading the spec. got it.

src/main/scala/amba/apb/ToTL.scala
terpstra: No. PSEL is "go" and PENABLE is "now it's your turn".

terpstra: The old APB and AHB specs are from an era when you had this whole bus-turn-around nonsense. You could imagine PENABLE to be like the direction of the data bus. PENABLE=0 means master controls it and PENABLE=1 means the slave controls it.

src/main/scala/amba/apb/ToTL.scala
terpstra: Actually, there is another problem I didn't notice on my first review pass: paddr. TileLink requests must be aligned to their size. I think you are correct to set size unilaterally to log2Ceil(beatBytes).U

However, this means you need to force alignment of the TileLink address:
out.a.bits.address := ~(~in.paddr | (beatBytes-1).U)
// The double negative here is to work around Chisel's broken implementation of widening ~x.

That also means that Gets are always full width:
out.a.bits.mask := Mux(in.pwrite, in.pstrb, ~0.U(beatBytes.W))

terpstra:
BTW, this is a new block. It should be written in chisel3. Don't copy-paste the legacy!

src/main/scala/amba/apb/ToTL.scala
terpstra: Check out Figure 4-1 in the APB spec. It explains it well.

src/main/scala/amba/apb/ToTL.scala
mwachs5: For the address, should I be adding an assertion that it's aligned when it comes in? Or will TL assertions already check that?

mwachs5: Oh I see above that you said I should force alignment. 

Shouldn't the user be required to do that? Othewise I'm not sure what the address means that I am receiving.

mwachs5: For now I am adding an assertion that the alignment logic above does nothing.

src/main/scala/amba/apb/ToTL.scala
mwachs5: I somewhat changed your suggested assertion above. I want the transaction to go out on PSEL=1, but not come back until PENABLE=1, so this is the assertion I wrote:

`      assert (!(in.psel && !in.penable && out.d.valid))
`
Make sense?

terpstra: By de morgan's laws, here is my assertion in your form: (!(out.d.valid && (!in.penable || !in.psel))
You can see it already covers your scenario, but additionally requires that psel is 1 when dvalid is 1.

src/main/scala/amba/apb/ToTL.scala
terpstra: I think you should silently accept a misaligned address. If the bus is 4-byte wide, you might get a write with address = 6 and strb=0100. Converting that to address = 4 and strb=0100 is fine. In the other AMBA protocols there is a 'size' field, but APB does not have one. For AXI, the equivalent transformation would be to align the address and round the size up to the bus width. That does not change the amount written. It marginally changes the semantics of the operation, because the size was adjusted. However, for APB, you don't have a size field. Your adapter is just assuming all operations are bus-width and setting size to a constant, a decision I agree with. Thus you need the other half of the transformation: round the address down.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -0,0 +1,91 @@
...
+      val beatBytes = edgeOut.manager.beatBytes
+
+      //-----------------------------------------------
+      // Flow Control
+      // We don't want to handle the case where the data is returned immediately
+      // on the same cycle that we send the transaction. Require and verify.
+      require (edgeOut.manager.minLatency >= 1)
+      assert (!(in.psel && !in.penable && out.d.valid))
+
+      val in_flight_reg = RegInit(false.B)
```
ernie-sifive: One possible issue I noticed:  The APB spec has an active-low reset, but this implementation appears to be active-high.

PR #1906, opened 2019-03-27
Ephemeral nodes

terpstra:
This PR introduces a new category of Node types: EphemeralNodes. These nodes act like IdentityNodes in the sense that anything connected to their inputs and outputs are just connected to each other pass-through. They differ, however, in that the node disappears from the final node graph. This makes it possible to define a logical connection point that can be used to interconnect pieces all around the design, and yet have the eventual connections be point-to-point.

**Related issue**: Remove wires plumbed in and out of tiles

**Release Notes**
Added Ephemeral Node type.


hcook:
Parentless `PSDTestModeBroadcastKey` worked as expected.

PR #1909, opened 2019-04-03
diplomacy: don't auto-punch ephemeral node wires

terpstra:
They don't go anywhere, so don't plumb them uselessly to the top!

**Release Notes**
Ephemeral nodes should not create wires


terpstra:
@hcook Can you try this with the designs where it was an issue? I couldn't reproduce the problem, but based on your description, I think this should fix it.

PR #1912, opened 2019-04-04
haslogicaltreenode

derekpappas:
**Related issue**: 

**Release Notes**

Adding a trait which will add the logical tree node from the parent to the child...

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
+++ b/src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
@@ -1,0 +1,9 @@
+package diplomaticobjectmodel
+
+import freechips.rocketchip.diplomaticobjectmodel.logicaltree._
+
+trait HasLogicalTreeNode {
+  def logicalTreeNode: LogicalTreeNode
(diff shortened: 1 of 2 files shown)


src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
```diff
@@ -0,0 +1,9 @@
+package diplomaticobjectmodel
```
hcook: This should match all the other packages and be
`package freechips.rocketchip.diplomaticobjectmodel`

hcook:
Why don't you you just go ahead and add this trait to everything that currently has a `val logicalTreeNode = ...` in it?

src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
derekpappas: fixed

derekpappas:
We need to talk about replacing the pattern for adding nodes. Lets get this PR approved so that I can add the composable cache.

PR #1923, opened 2019-04-16
Widen BP action to 3 bits and create bpwatch output

ernie-sifive:

**Release Notes**

Increase width of the action field in breakpoint mcontrol register.  This will support actions beyond the existing breakpoint exception and debug exception including codes 2-5 reserved for trace watchpoints in the RISC-V debug spec.  The action field and valid bit for each breakpoint is sent out via BundleBridge to a trace client.

In addition, this PR adds **with Clocked** to the TracedInstruction bundle so the trace client receives clock and reset along with the bundle.


aswaterman:
Given that the breakpoint unit can be a critical path, and the widened action field feeds into it, shouldn’t we conditionalize its widening on the presence of trace? The field can remain one bit wide when there’s no trace unit.

ernie-sifive:
@aswaterman So I can add a new "useBPWatch" item to RocketCoreParams, which would default to false and predicate all the new logic.  Does that sound right?

aswaterman:
It looks like all of the relevant logic will synthesize away if you conditionalize the width of the field.  So not all of the logic needs to be conditionalized, which will keep the code cleaner.   It should suffice to do something like

    val reserved = UInt(width = xLen-26 - (if (useBPWatch) 2 else 0))
    val action = UInt(width = 1 + (if (useBPWatch) 2 else 0))

ernie-sifive:
@aswaterman, I've made the changes you requested.  Please review at your convenience.

PR #1927, opened 2019-04-23
Om th

derekpappas:
**Related issue**: 

**Release Notes**

Adding support for logical tree nodes in diplomatic SRAM's
* This PR adds ResourceBindings to the test rams
* This PR does not remove the old omMem return value for now.
* Supports multiple roots - design and test harness roots are added under root now
* The Logical TreeNodes are passed down the DUT hierarchy through function parameters
* The DescribedSRAM now has a class and the object has two functions: apply and build
  * build is used by the test harness memories -Returns an object which has the information needed to construct an OMMemory - includes a hash function for matching the DescribedSRAM annotation that contains the same hash value. The DescribedSRAM anno will have the Veirlog path. The anno and the OM mem will be joined in a post process step. Another PR can add the hash to the design memories as well. 
  * apply is used by all legacy memories


src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
```diff
@@ -23,7 +29,7 @@ object LogicalModuleTree {
 
   def root: LogicalTreeNode = {
     val roots = tree.collect { case (k, _) if !tree.exists(_._2.contains(k)) => k }
-    assert(roots.size == 1, "Logical Tree contains more than one root.")
+    assert(roots.size <= 2, "Logical Tree contains more than two roots.")
```
hcook: "two or more"?

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
```diff
@@ -11,7 +12,12 @@ trait LogicalTreeNode {
   def getOMComponents(resourceBindingsMap: ResourceBindingsMap, children: Seq[OMComponent] = Nil): Seq[OMComponent]
 }
 
+class RootLogicalTreeNode extends LogicalTreeNode {
+def getOMComponents(resourceBindingsMap: ResourceBindingsMap, children: Seq[OMComponent] = Nil): Seq[OMComponent] = children
+}
+
 object LogicalModuleTree {
+  val rootLTN = new RootLogicalTreeNode()
```
hcook: I'd just call this `root`

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
```diff
@@ -2,6 +2,7 @@
 
 package freechips.rocketchip.diplomaticobjectmodel.logicaltree
 
+import freechips.rocketchip.config.Field
```
hcook: Unneeded import?

src/main/scala/amba/axi4/SRAM.scala
```diff
@@ -15,7 +17,7 @@ class AXI4RAM(
     errors: Seq[AddressSet] = Nil,
     wcorrupt: Boolean = false)
   (implicit p: Parameters) extends DiplomaticSRAM(address, beatBytes, devName)
-{
+  {
```
hcook: Why did you change all the whitespace in this one file?

src/main/scala/amba/axi4/SRAM.scala
```diff
@@ -5,6 +5,8 @@ package freechips.rocketchip.amba.axi4
 import Chisel._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.diplomaticobjectmodel.logicaltree._
+import freechips.rocketchip.diplomaticobjectmodel.model.{OMAPBRAM, OMAXI4RAM}
```
hcook: Unneeded import

src/main/scala/amba/apb/SRAM.scala
```diff
@@ -5,6 +5,8 @@ package freechips.rocketchip.amba.apb
 import Chisel._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.diplomaticobjectmodel.logicaltree._
+import freechips.rocketchip.diplomaticobjectmodel.model.{OMAPBRAM, OMTLRAM}
```
hcook: Unneeded import

src/main/scala/amba/ahb/SRAM.scala
```diff
@@ -5,6 +5,7 @@ package freechips.rocketchip.amba.ahb
 import Chisel._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.diplomaticobjectmodel.model.{OMAHBRAM, OMTLRAM}
```
hcook: Unneeded import

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
```diff
@@ -97,3 +99,71 @@ class SubSystemLogicalTreeNode(var getOMInterruptDevice: (ResourceBindingsMap) =
...
+
+case object DeviceKey extends Field[Option[Device]](
+  None
+)
+
+case object ParentLogicalTreeNodeKey extends Field[Option[LogicalTreeNode]](
+  None
+)
+
+case object ResourceBindingsMapLogicalTreeNodeKey extends Field[Option[ResourceBindingsMap]](
```
hcook: What does this have to do with "LogicalTreeNodes"? Also, why do you need it? Aren't the ResourceBindings supposed to be resolved when you walk the tree at the end?

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
```diff
@@ -97,3 +99,71 @@ class SubSystemLogicalTreeNode(var getOMInterruptDevice: (ResourceBindingsMap) =
...
+  None
+)
+
+object ParentTreeNode {
+  /**
+    * Add the parent tree node to the Parameters so that the children can access the tree node via the params
+    * @param p
+    * @return
+    */
+  def set(pltnOpt: Option[LogicalTreeNode], ltn: LogicalTreeNode)(implicit p: Parameters): Parameters = {
```
hcook: Is this used?

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -77,6 +77,7 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem with
   def getOMInterruptDevice(resourceBindingsMap: ResourceBindingsMap): Seq[OMInterrupt]
 
   val logicalTreeNode = new SubSystemLogicalTreeNode(getOMInterruptDevice)
+  LogicalModuleTree.add(LogicalModuleTree.rootLTN, logicalTreeNode)
```
hcook: Should this really always be added to the root? What if there are other layers in between?

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
```diff
@@ -97,3 +99,71 @@ class SubSystemLogicalTreeNode(var getOMInterruptDevice: (ResourceBindingsMap) =
...
+  data: T,
+  resourceBindings: ResourceBindings,
+  architecture: RAMArchitecture,
+  description: String,
+  depth: Int,
+  ecc: Option[OMECC] = None,
+  hasAtomics: Option[Boolean] = None
+)
+
+class SRAMLogicalTreeNode[T <: Data](description: String, architecture: RAMArchitecture, depth: Int, lanes: Int, bits: Int, device: () => Device,
```
hcook: Why does this class need both `rbm: () => ResourceBindingsMap` AND `def getOMComponents(resourceBindingsMap: ResourceBindingsMap,...)` Are these the same RBM, or different RBMs? Something seems wrong about the fact that there are two and you only use one...

src/main/scala/diplomaticobjectmodel/ConstructOM.scala
```diff
@@ -7,9 +7,14 @@ import freechips.rocketchip.diplomaticobjectmodel.logicaltree._
 import freechips.rocketchip.diplomaticobjectmodel.model.OMComponent
 import freechips.rocketchip.util.ElaborationArtefacts
 
-case object ConstructOM {
-  def constructOM(resourceBindingsMap: => ResourceBindingsMap): Unit = {
-    val om: Seq[OMComponent] = LogicalModuleTree.bind(resourceBindingsMap)
+object ConstructOM {
+  private var resourceBindingsMap: () => ResourceBindingsMap = _
+
+  def apply(rbm: () => ResourceBindingsMap): Unit = resourceBindingsMap = rbm
```
hcook: I'd call this method `bind` or `setBindingsMap` or something more explicit than just the apply method. Usually when object have an apply method they return an instance of the companion class.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
```diff
@@ -97,3 +99,71 @@ class SubSystemLogicalTreeNode(var getOMInterruptDevice: (ResourceBindingsMap) =
...
+    children
+}
+
+class TestSoCLogicalTreeNode extends LogicalTreeNode {
+  override def getOMComponents(resourceBindingsMap: ResourceBindingsMap, children: Seq[OMComponent] = Nil): Seq[OMComponent] = {
+    Seq(OMTestHarness( components = children))
+  }
+}
+
+case object DeviceKey extends Field[Option[Device]](
```
hcook: What is this for?

src/main/scala/diplomacy/Resources.scala
```diff
@@ -183,11 +183,14 @@ class SimpleDevice(devname: String, devcompat: Seq[String]) extends Device
   with DeviceClocks
   with DeviceRegName
 {
+  var resourceBindings: Option[ResourceBindings] = None
```
hcook: @terpstra Are devices supposed to have their own ResourceBindings?

src/main/scala/util/DescribedSRAM.scala
```diff
@@ -29,6 +24,8 @@ object DescribedSRAM {
       case d => d.getWidth
     }
 
+    val f = () => mem.toNamed.toString.hashCode
```
hcook: What is this for? Where is it used?

src/main/scala/diplomaticobjectmodel/ConstructOM.scala
derekpappas: Thanks.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
derekpappas: The structure of the tree is as follows: @terpstra @richardxia 

                      root
                       /   \
                     /       \
TestHarness        CoreComplex

richardxia: I don't really have much context for what is happening right now, but if we are already going from 1 to 2 roots, is there any downside to just allowing an unbounded number of roots? We essentially already have a forest.

src/main/scala/diplomaticobjectmodel/ConstructOM.scala
derekpappas: Changed to:

  def setBindingsMap(rbm: () => ResourceBindingsMap): Unit = resourceBindingsMap = rbm

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
```diff
@@ -97,3 +101,53 @@ class SubSystemLogicalTreeNode(var getOMInterruptDevice: (ResourceBindingsMap) =
...
+        memories = Nil,
+        architecture = ramArchitecture
+      )
+    )
+  }
+
+  def getOMComponents(resourceBindingsMap: ResourceBindingsMap, components: Seq[OMComponent]): Seq[OMComponent] = {
+    val c = DiplomaticObjectModelAddressing.getOMComponentHelper(device(), resourceBindingsMap, getOMTestMemory)
+    val cc = components.map(_.asInstanceOf[OMMemory])
+    val children = c.toList.head.asInstanceOf[OMTestMemory].copy( memories = cc).asInstanceOf[OMComponent]
```
richardxia: This feels awkward, having to take the `c` object, which we just happen to know is a Seq containing exactly one `OMTestMemory` and having to grab the first element and cast it to an `OMTestMemory`. I know that the type system only knows that `c` is a Seq of `OMComponent`s, but this seems like we are throwing away type safety.

I think there's a way to make this work without explicit type casts by adding a type parameter to `getOMComponentHelper`, but I don't know if that refactoring is a good idea here.

src/main/scala/tilelink/SRAM.scala
```diff
@@ -201,8 +207,8 @@ class TLRAM(
       if (code.canDetect) code.encode(d, p) else code.encode(d)
     })
 
-    d_raw_data := mem.read(addr, ren)
-    when (wen) { mem.write(addr, coded, sel) }
+    d_raw_data := mem.mem.read(addr, ren)
```
richardxia: This looks a little bit awkward and confusing now, with the double `mem`. Could we give the outer `mem` a different name up on line 61 here?

This kind of change will need to be replicated to all the other SRAM instances for the other protocols.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
```diff
@@ -11,27 +12,47 @@ trait LogicalTreeNode {
   def getOMComponents(resourceBindingsMap: ResourceBindingsMap, children: Seq[OMComponent] = Nil): Seq[OMComponent]
 }
 
+class RootLogicalTreeNode extends LogicalTreeNode {
```
derekpappas: Make one concrete generic LTN and replace this LTN and the ImpLTN with the single concrete LTN.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
```diff
@@ -11,27 +12,47 @@ trait LogicalTreeNode {
...
-  def add(parent: LogicalTreeNode, child: => LogicalTreeNode): Unit = {
+
+  def addResourceScope(ltn: LogicalTreeNode, rbm: () => ResourceBindingsMap): Unit = {
+    require(! resourceScopes.contains(ltn))
+    resourceScopes.put(ltn, rbm)
+  }
+
+  def add(parent: LogicalTreeNode, child: => LogicalTreeNode, resourceBindingsMap: Option[() => ResourceBindingsMap] = None): Unit = {
     val treeOpt = tree.get(parent)
     val treeNode = treeOpt.map{
```
derekpappas: Add a check that the parent does not contain the child already.

Add a check that the child has not been added to the tree more than once.

This can be done in the map add function.

This means when adding a parent child edge in the map that the child should not appear in any other parent child list. This can be checked with a second map of all children.

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
```diff
@@ -11,27 +12,47 @@ trait LogicalTreeNode {
...
     roots.head
   }
 
   def bind(resourceBindingsMap: ResourceBindingsMap): Seq[OMComponent] = {
     def getOMComponentTree(node: LogicalTreeNode): Seq[OMComponent] = {
-      node.getOMComponents(resourceBindingsMap, tree.get(node).getOrElse(Nil).flatMap(getOMComponentTree))
+      val rbm = resourceScopes.get(node) match {
+        case Some(r) =>
+          r()
+        case _ => resourceBindingsMap
```
derekpappas: Change to case None

src/main/scala/amba/ahb/Parameters.scala
```diff
@@ -28,6 +28,15 @@ case class AHBSlaveParameters(
...
   // The device had better not support a transfer larger than it's alignment
   require (minAlignment >= maxTransfer)
+
+  def toResource: ResourceAddress = {
+    ResourceAddress(address, ResourcePermissions(
+      r = supportsRead,
+      w = supportsWrite,
+      x = executable,
+      c = false,
+      a = false))
```
derekpappas: @terpstra @richardxia wants to know if the default = false is always true

PR #1928, opened 2019-04-24
tilelink: improve specificity of AtomicAutomata graph check

hcook:
The TLAtomicAutomata uses a diplomatic graph search to ensure that it is at the root of a tree of nodes. This property is necessary but not sufficient for correct behavior; no other master should be able to intervene between the AA's read and write operations. However, this check should actually only check tree violations for managers to which the AA is actually adding atomics support. For example, if a manager natively supports atomics, and `passthru` is enabled, then the AA does not have to be at the root of a tree with respect to that manager. This change makes the check handled on a per-manager basis.

**Related issue**: 

**Impact**: no RTL change, only further specificity of a diplomatic safeguard

### DIFF
--- a/src/main/scala/tilelink/AtomicAutomata.scala
+++ b/src/main/scala/tilelink/AtomicAutomata.scala
@@ -32,15 +32,8 @@ class TLAtomicAutomata(logical: Boolean = true, arithmetic: Boolean = true, conc
       val managers = edgeOut.manager.managers
       val beatBytes = edgeOut.manager.beatBytes
 
-      // This is necessary (though not sufficient) for correctness:
-      edgeOut.manager.findTreeViolation() match {
-        case None => ()
-        case Some(node) => require(edgeOut.manager.isTree, s"AtomicAutomata can only be placed infront of a tree of diplomatic nodes (${node.name} has parents ${node.inputs.map(_._1.name)})")
-      }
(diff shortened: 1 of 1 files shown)


src/main/scala/tilelink/AtomicAutomata.scala
```diff
@@ -32,22 +32,24 @@ class TLAtomicAutomata(logical: Boolean = true, arithmetic: Boolean = true, conc
...
       val managersNeedingHelp = managers.filter { m =>
         m.supportsPutFull.contains(ourSupport) &&
         m.supportsGet.contains(ourSupport) &&
         ((logical    && !m.supportsLogical   .contains(ourSupport)) ||
          (arithmetic && !m.supportsArithmetic.contains(ourSupport)) ||
          !passthrough) // we will do atomics for everyone we can
       }
+
+      // This is necessary (though not sufficient) for correctness:
+      // (You also need to know that the slaves don't have any non-diplomatic intermal masters that can get between the read and write)
```
aswaterman: typo on "intermal"

hcook:
@terpstra, I just copied the original text, which was pointing out that the property is not actually sufficient because of the possibility of private masters. I guess it is neither, depending on whether managers need help or not. I will attempt to rephrase.

It seemed wasteful to check the tree property on the edge when there might not even be any managers that need help..?

PR #1933, opened 2019-04-30
Add PMP to DeviceTree

aswaterman:

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -108,7 +108,9 @@ trait HasNonDiplomaticTileParameters {
         "tlb-split" -> Nil,
         "mmu-type"  -> s"riscv,sv$maxSVAddrBits".asProperty)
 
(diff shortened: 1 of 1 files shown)


ccelio:
FMI (for my information), what's the use-case here? I'm not very imaginative.

aswaterman:
For introspection from M-mode code.  I don't have a particular piece of code in mind; this is a feature request from @palmer-dabbelt and crew.

palmer-dabbelt:
The general idea is that you may want to know how many PMPs there are on the target at compile time, which currently isn't encoded anywhere.  It's hard to enumerate all the users of this sort of thing, but just for example:

Imagine you're writing a machine-mode security monitor and know you need two PMPs to support an enclave.  With this you could just to staticly allocate the enclaves because you'd know the number of PMPs at compile time.  Without this you'd need to either allocate the ISA-mandated maximum (which could eat up a meaningful amount of memory) or probe at runtime (which would require dynamic allocation).

PR #1934, opened 2019-05-02
Move SeqMems to BigInts and plumb through test mems

colinschmidt:
**Related issue**:
I would like to be able to enable ExtMem sizes larger than 2GB. Currently this only fails in the simulation memory. Now that chisel and firrtl support BigInt depth on memories we can support this cleanly.

**Type of change**: bug report / feature request.
Certain configs being invalid straddles the line between bug and feature request

**Impact**: API modification.
Certain internal(-only?) interfaces now use BigInt rather than Int

**Development Phase**: implementation.
I'm using this downstream and have tested it with 1,2,4,8,16GB ExtMem sizes. Note that past 1GB ExtMem sizes you will need special handling of the testharness memory in some simulators such as vcs

**Release Notes**
Enable ExtMem sizes greater than 2GB by including BigInt memories from a recent chisel3 and firrtl update.

### DIFF
--- a/src/main/scala/amba/axi4/SRAM.scala
+++ b/src/main/scala/amba/axi4/SRAM.scala
@@ -31,7 +31,7 @@ class AXI4RAM(
 
   lazy val module = new LazyModuleImp(this) {
     val (in, _) = node.in(0)
-    val (mem, omMem) = makeSinglePortedByteWriteSeqMem(1 << mask.filter(b=>b).size)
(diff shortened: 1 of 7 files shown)


hcook:
LGTM.

@derekpappas Acceptable schema change to `OMMemory`?

derekpappas:
LGTM. FYI I think some of the om-th PR will be in conflict with some of the modified files after this PR is approved.

colinschmidt:
Thanks for the review.
I'm going to merge this later tonight or tomorrow morning unless there is more feedback.

PR #1937, opened 2019-05-06
BlackBox RoCC Example

seldridge:
This adds a `BlackBoxExample` RoCC with a backing Verilog BlackBox resource in `vsrc/RoccBlackBox.v`. Rocket Chip parameters are passed as Verilog parameters. Nothing with diplomatic parameters / TL is dealt with here.

The backing example RoCC implements a simple accumulation of rs1 and rs2 into a persistent register that is returned if `xd` is set.

**Related Issue**: https://github.com/freechipsproject/rocket-chip/issues/1195#issuecomment-488546441chip/issues/1195

**Release Notes**

### DIFF
--- a/src/main/resources/vsrc/RoccBlackBox.v
+++ b/src/main/resources/vsrc/RoccBlackBox.v
@@ -1,0 +1,181 @@
+// See LICENSE.SiFive for license details.
+
+module RoccBlackBox
+  #( parameter xLen,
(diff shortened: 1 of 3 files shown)


hcook:
Nice. I think adding the TL port would only be a three additional parameters than have to be piped in. I'll try to find someone to work on it this summer.

PR #1940, opened 2019-05-08
diplomacy: split UNCACHEABLE region type into idempotent and volatile

hcook:
The `UNCACHEABLE` region type was conflating two somewhat different semantic meanings, which are now differentiated into `IDEMPOTENT` and `VOLATILE`. (Also the names `UNCACHED` vs `UNCACHEABLE` was confusing, and `GET_EFFECTS` and `PUT_EFFECTS` are also uncacheable).

The complete region type enumeration can be divided into two groups, one of which is cacheable and the other of which is uncacheable. Note that these region types denote the expected behavior of a memory region at a particular point in the diplomatic graph structure. The behavior of a particular address may change throughout the graph as a result of the interposition of certain adapter nodes. For example, inserting a caching adapter node between a processor and an `UNCACHED` memory region may cause its addresses to become `CACHED` but should have no effect on an address that is `IDEMPOTENT`. Furthermore, to determine practical cacheability and coherence the region types have to be combined with the actual capabilities offered by any particular manager on any particular link (e.g. `manager.supportsAcquireT`).

Cacheable region types:
- `CACHED`: an intermediate agent may have cached a copy of the region for you
- `TRACKED`: the region may have been cached by another master, but coherence is being provided
- `UNCACHED`: the region has not been cached or tracked yet, but should be cached when possible

Uncacheable region types:
- `IDEMPOTENT`: gets return most recently put content, but content should not be cached
 - `VOLATILE`: content may change without a put, but puts and gets have no side effects
 - `PUT_EFFECTS` : puts produce side effects and so must not be combined/delayed
 - `GET_EFFECTS` : gets produce side effects and so must not be issued speculatively

Neither `IDEMPOTENT` nor `VOLATILE` regions may be cached, as before. The `SystemBus.fromTile` method enforces FIFO ordering on accesses to `VOLATILE` regions, but not to `IDEMPOTENT` ones. SRAM scratchpads are `IDEMPOTENT` if they are not cacheable.


jwright6323:
I can't tell if I'm misreading this PR or if there's a title typo- should it be "diplomacy: split *uncachable* region type into idempotent and volatile"?

hcook:
@jwright6323 omg this is why we are changing the name

PR #1943, opened 2019-05-08
bump chisel3 for Vec of Analog fix

aswaterman:
@jackkoenig I created this to get the tests running.  It's based against the top of your chisel PR so is only safely mergeable if that one is merged with a merge commit and without additional force-pushes, hence the hopefully temporary DONT MERGE label.


aswaterman:
I expect that this will be fixed by merging https://github.com/freechipsproject/chisel3/pull/1093

PR #1944, opened 2019-05-09
Modifying SRAMAnnotations for OM purposes

debs-sifive:

**Impact**: API modification

**Release Notes**

* `SRAMAnnotation`s have the target type `ReferenceTarget` to allow for unique path generation
* `DescribedSRAM`s generate a unique ID that can be correlated to the OM

@derekpappas @jackkoenig

### DIFF
--- a/src/main/scala/util/Annotations.scala
+++ b/src/main/scala/util/Annotations.scala
@@ -19,11 +19,12 @@ import org.json4s.jackson.JsonMethods.{pretty, render}
 case class SRAMAnnotation(target: ReferenceTarget,
   address_width: Int,
   name: String,
+  wrap_target: ReferenceTarget,
   data_width: Int,
-  depth: BigInt,
+  depth: Int,
   description: String,
   write_mask_granularity: Int,
-  idhash: Int=0) extends SingleTargetAnnotation[ReferenceTarget] {
+  idhash: Int = 0) extends SingleTargetAnnotation[ReferenceTarget] {
   def duplicate(n: ReferenceTarget) = this.copy(n)
 }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/util/DescribedSRAM.scala
```diff
@@ -29,14 +29,17 @@ object DescribedSRAM {
       case d => d.getWidth
     }
 
+    val f = () => mem.toNamed.toString.hashCode
```
jackkoenig: The odds of a hash collision are low since we're likely to have 10s of these, but it's worth thinking about. It might be worth using a 64-bit hash

src/main/scala/util/Annotations.scala
```diff
@@ -15,14 +16,15 @@ import org.json4s.JsonDSL._
...
+case class SRAMAnnotation(target: ReferenceTarget,
   address_width: Int,
   name: String,
   data_width: Int,
   depth: BigInt,
   description: String,
-  write_mask_granularity: Int) extends SingleTargetAnnotation[Named] {
-  def duplicate(n: Named) = this.copy(n)
+  write_mask_granularity: Int,
+  idhash: Int=0) extends SingleTargetAnnotation[ReferenceTarget] {
```
jackkoenig: Style nit, but you generally put spaces around `=`

src/main/scala/util/Annotations.scala
```diff
@@ -104,22 +106,24 @@ case class ResetVectorAnnotation(target: Named, resetVec: BigInt) extends Single
...
+  def srams[T <: Data](
+    component: SyncReadMem[T],
     name: String,
     address_width: Int,
     data_width: Int,
     depth: BigInt,
     description: String,
-    write_mask_granularity: Int): Unit = {
+    write_mask_granularity: Int,
+    idhash: ()=>Int = ()=>0): Unit = {
```
jackkoenig: Similarly here, style nit, would be:
```scala
idhash: () => Int = () => 0): Unit = {
```

src/main/scala/util/DescribedSRAM.scala
debs-sifive: as discussed through Slack, the ID generator has been switched to a global counter

src/main/scala/util/Annotations.scala
```diff
@@ -104,22 +106,24 @@ case class ResetVectorAnnotation(target: Named, resetVec: BigInt) extends Single
...
+  def srams[T <: Data](
+    component: SyncReadMem[T],
     name: String,
     address_width: Int,
     data_width: Int,
     depth: BigInt,
     description: String,
-    write_mask_granularity: Int): Unit = {
+    write_mask_granularity: Int,
+    uid: () => Int = () => 0): Unit = {
```
jackkoenig: I would change this from `() => Int` to just `Int`. It doesn't matter that much since it's just an argument to a function, but we don't need to delay creation of the the id anymore and the order of id allocation will be the same whether we do it immediately here or delay it until the Firrtl Annotation is created.

debs-sifive: done!

PR #1950, opened 2019-05-14
Full AHB

terpstra:
**Impact**:  API modification

**Release Notes**
This PR splits the AHB bundle types into master and slave bundles. AHB is not a point-to-point protocol. It is a shared bus protocol, with different signals from master to arbiter and arbiter to slave. Trying to shoe-horn these into one Bundle type inhibited implementation of full AHB arbitration. Adding support for AHB arbitration was just a 3-line change to TLToAHB once the types were split.

### DIFF
--- a/src/main/scala/amba/ahb/Bundles.scala
+++ b/src/main/scala/amba/ahb/Bundles.scala
@@ -8,51 +8,116 @@ import freechips.rocketchip.util.GenericParameterizedBundle
 abstract class AHBBundleBase(params: AHBBundleParameters) extends GenericParameterizedBundle(params)
 
 // Signal directions are from the master's point-of-view
-class AHBBundle(params: AHBBundleParameters) extends AHBBundleBase(params)
+class AHBSlaveBundle(params: AHBBundleParameters) extends AHBBundleBase(params)
 {
-  // Flow control signals from the master
+  // Control signals from the arbiter to slave
   val hmastlock = Bool(OUTPUT)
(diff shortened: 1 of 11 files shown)


mwachs5:
Presumably this changes all I/O names of any AHB interfaces?

aswaterman:
Where's the part about asserting the lock to prevent split transactions?  Did I miss it, or are you not doing that yet?

terpstra:
@aswaterman Line 191 of ToAHB.scala

aswaterman:
Yeah, I saw the assertion that split/retry don't happen, but I thought you were also proposing asserting the lock signal so that the slave isn't even allowed to respond that way.  I still see the lock signal wired to false.

terpstra:
@mwachs5 Yes, It will change the IO names for ports where we are the AHB master. It actually makes us comply with the spec, whereas before we did not. Customers have fairly regularly asked about our non-standard port names, so this may or may not be a net positive.

terpstra:
@yunsup I was reading a document titled "AMBA Specification (Rev 2.0)". So this might actually be AMBA2, not AMBA3. :-/

terpstra:
@aswaterman Right. I don't do that yet. That will be a more invasive change, which I think we should make when we also make atomics drive the signal. As the current deliverable only needed hbusreq+hgrant, I wanted to minimize risk in this change set. That should definitely be follow-up work.

aswaterman:
Got it

PR #1953, opened 2019-05-15
APB2TL: Check for address legality

mwachs5:
**Related issue**: 

**Release Notes**

APB2TL Converter was not checking for address legality. Tile Link protocol forbids illegal addresses, so they need to be diverted at the source.

### DIFF
--- a/src/main/scala/amba/apb/ToTL.scala
+++ b/src/main/scala/amba/apb/ToTL.scala
@@ -45,13 +45,32 @@ class APBToTL()(implicit p: Parameters) extends LazyModule
       require (edgeOut.manager.minLatency >= 1)
       assert (!(in.psel && !in.penable && out.d.valid))
 
+
+      val beat = TransferSizes(1, beatBytes)
+      // The double negative here is to work around Chisel's broken implementation of widening ~x.
+      val alignedAddr =  ~(~in.paddr | (beatBytes-1).U)
+      val dataSize = UInt(log2Ceil(in.params.dataBits/8))
+
+      // Is this access allowed? Illegal addresses are a violation of tile link protocol.
+      // If an illegal address is provided, return an error instead of sending over tile link.
+      val a_legal =
+        Mux(in.pwrite,
+          edgeOut.manager.supportsPutPartialSafe(alignedAddr, dataSize, Some(beat)),
+          edgeOut.manager.supportsGetSafe       (alignedAddr, dataSize, Some(beat)))
+
       val in_flight_reg = RegInit(false.B)
-      out.a.valid := in.psel && !in_flight_reg
-      in.pready := out.d.valid
+      val error_in_flight_reg = RegInit(false.B)
+      val in_flight = in_flight_reg || error_in_flight_reg
+
+      out.a.valid := in.psel && !in_flight && a_legal
+      in.pready := out.d.valid  || error_in_flight_reg
       out.d.ready := true.B
 
       when (out.a.fire()){in_flight_reg := true.B}
       when (out.d.fire()){in_flight_reg := false.B}
+
+      when (in.psel && !error_in_flight_reg){error_in_flight_reg := !a_legal}
+      when (error_in_flight_reg){ error_in_flight_reg := false.B }
 
       // Write
       // PutPartial because of pstrb masking.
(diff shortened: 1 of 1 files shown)


src/main/scala/amba/apb/ToTL.scala
```diff
@@ -45,23 +45,41 @@ class APBToTL()(implicit p: Parameters) extends LazyModule
       require (edgeOut.manager.minLatency >= 1)
       assert (!(in.psel && !in.penable && out.d.valid))
 
+
+      val beat = TransferSizes(1, beatBytes)
```
terpstra: You could as well just say TransferSizes(beatBytes, beatBytes), which might hypothetically give you a better QoR.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -45,23 +45,41 @@ class APBToTL()(implicit p: Parameters) extends LazyModule
...
+      // The double negative here is to work around Chisel's broken implementation of widening ~x.
+      val alignedAddr =  ~(~in.paddr | (beatBytes-1).U)
+      val dataSize = UInt(log2Ceil(in.params.dataBits/8))
```
terpstra: Even though beatBytes=dataBits/8, since you use beatBytes everywhere, I think you should use it here too so it's easy to read and reason about.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -45,23 +45,41 @@ class APBToTL()(implicit p: Parameters) extends LazyModule
...
+      out.a.valid := in.psel && !in_flight && a_legal
+      in.pready := out.d.valid  || error_in_flight_reg
       out.d.ready := true.B
 
       when (out.a.fire()){in_flight_reg := true.B}
       when (out.d.fire()){in_flight_reg := false.B}
 
+      when (in.psel && !error_in_flight_reg){error_in_flight_reg := !a_legal}
+      when (error_in_flight_reg){ error_in_flight_reg := false.B }
+
```
terpstra: I agree that your formulation is correct. However, even though the spec guarantees it, I don't think we should rely on masters to hold paddr constant if we don't have to.

error_in_flight_reg := false.B
when (in.psel && !in_flight) {error_in_flight_reg := !a_legal}

src/main/scala/amba/apb/ToTL.scala
mwachs5: Ah, `TransferSizes` is a range. got it.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -45,23 +43,44 @@ class APBToTL()(implicit p: Parameters) extends LazyModule
       require (edgeOut.manager.minLatency >= 1)
       assert (!(in.psel && !in.penable && out.d.valid))
 
+
+      val beat = TransferSizes(beatBytes, beatBytes)
+      // The double negative here is to work around Chisel's broken implementation of widening ~x.
+      val aligned_addr =  ~(~in.paddr | (beatBytes-1).U)
+      require(beatBytes == log2Ceil(in.params.dataBits/8),
+              s"TL beatBytes(${beat_bytes}) doesn't match expected APB data width(${in.params.dataBits})")
```
terpstra: There's no need to put this requirement here. It's true by your definition of beatBytes. My comment was just that it's easier to read code when the same variable name is used consistently.

src/main/scala/amba/apb/ToTL.scala
```diff
@@ -45,23 +43,44 @@ class APBToTL()(implicit p: Parameters) extends LazyModule
...
+      // Default to false except when actually returning error.
+      error_in_flight_reg := false.B
+      when (in.psel && !in_flight) {error_in_flight_reg := !a_legal}
+      
       // Write
       // PutPartial because of pstrb masking.
       out.a.bits.opcode  := Mux(in.pwrite, TLMessages.PutPartialData, TLMessages.Get)
       out.a.bits.param   := UInt(0)
-      out.a.bits.size    := UInt(log2Ceil(in.params.dataBits/8))
+      out.a.bits.size    := data_size
```
terpstra: This is wrong. You lost the 'log2Ceil'.

src/main/scala/amba/apb/ToTL.scala
terpstra: Also, your requirement is just wrong. beatBytes == in.params.dataBits/8  ... there's no log2Ceil.

mwachs5: Sure, I understood your comment. But I added the require since I wasn't sure about the code at the top... if in.params.dataBits doesn't show up in this file, it's getting checked how...?

terpstra: Your adapter did not change the width of the bus.
Therefore, edgeIn.manager.beatBytes = edgeOut.manager.beatBytes = in.params.dataBits/8 = out.params.dataBits/8

mwachs5: ok. If you don't mind, I'd like to leave the require since it's not obvious to me (and potentially future code readers) otherwise.

terpstra: Your call. You should fix the compile error, though. (beat_bytes => beatBytes)

mwachs5: sigh. I need to stop editing code through GitHub

PR #1956, opened 2019-05-16
refactored the getOMComponents ResourceBindingsMap to ResourceBindings

derekpappas:
**Related issue**: 

**Release Notes**

1. Changes the getOMComponents signature - ResourceBindingsMap is now ResourceBindings

2. Registers the ResourceBindingsMaps for each BindingScope 

3. Modified the LogicalModuleTree bind function to look up the ResourceBindings for each device in the list of BindingScopes

### DIFF
--- a/src/main/scala/diplomacy/Resources.scala
+++ b/src/main/scala/diplomacy/Resources.scala
@@ -178,7 +178,7 @@ object DiplomacyUtils {
   * @param devname      the base device named used in device name generation.
   * @param devcompat    a list of compatible devices. See device tree property "compatible".
   */
-class SimpleDevice(devname: String, devcompat: Seq[String]) extends Device
+class SimpleDevice(val devname: String, devcompat: Seq[String]) extends Device
   with DeviceInterrupts
   with DeviceClocks
   with DeviceRegName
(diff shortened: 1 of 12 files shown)


src/main/scala/diplomacy/Resources.scala
```diff
@@ -392,6 +394,10 @@ object BindingScope
     case x: BindingScope => find(x.parent).orElse(Some(x))
     case x => find(x.parent)
   }
+
+  var bindingScopes = new collection.mutable.ArrayBuffer[BindingScope]()
```
jackkoenig: Why is it both an `ArrayBuffer` and a `var`? 

This should probably be a `private val`

src/main/scala/diplomacy/Resources.scala
```diff
@@ -392,6 +394,10 @@ object BindingScope
...
+
+  def add(bs: BindingScope) = BindingScope.bindingScopes.+=:(bs)
```
jackkoenig: Minor nit, but this could use infix notation
```suggestion
  def add(bs: BindingScope) = bs +=: BindingScope.bindingScopes
```

You could alternatively explicitly use `prepend`, I think either is better the `.+=:`

PR #1959, opened 2019-05-18
conditionally add cflush

derekpappas:
**Related issue**: 

**Impact**: API modification

**Release Notes**

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/model/OMISA.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMISA.scala
@@ -44,10 +44,8 @@ case class OMISA(
 ) extends OMCompoundType
 
 object OMISA {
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomaticobjectmodel/model/OMISA.scala
```diff
@@ -44,10 +44,8 @@ case class OMISA(
 ) extends OMCompoundType
 
 object OMISA {
-  def customExtensions(): List[OMCustomExtensionSpecification] = {
-    List(
-      Xsifivecflushdlone()
-    )
+  def customExtensions(coreParams: RocketCoreParams): List[OMCustomExtensionSpecification] = {
+    if (coreParams.haveCFlush) List (Xsifivecflushdlone()) else Nil
```
richardxia: Is there any reason why this is a List and not a Seq?

PR #1966, opened 2019-05-21
fix Vivado parsing errors

mikeyangsiv:
**Related issue**: #1961 

**Release Notes**

Vivado parsing failed on:
* declaration in an unnamed block --> added name
* multi-dimensional array --> flattened

### DIFF
--- a/scripts/vlsi_rom_gen
+++ b/scripts/vlsi_rom_gen
@@ -26,7 +26,7 @@ module {name}(
   initial begin: init_and_load
     integer i;
     // 256 is the maximum length of $readmemh filename supported by Verilator
-    reg [255*8:0] path;
+    reg [255*8-1:0] path;
 `ifdef RANDOMIZE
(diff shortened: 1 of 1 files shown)


scripts/vlsi_rom_gen
```diff
@@ -23,10 +23,10 @@ module {name}(
   reg [{output_width_minus_1}:0] rom [0:{depth_minus_1}];
 
 
-  initial begin
+  initial begin: init_and_load
     integer i;
     // 256 is the maximum length of $readmemh filename supported by Verilator
-    reg [255:0] [7:0] path;
+    reg [255*8:0] path;
```
richardxia: Wait, actually, I'm going to revoke my approval. This should be 256*8 - 1, no?

mikeyangsiv: yes you're right ... I thought I had rationalized this in my head, but incorrectly

PR #1967, opened 2019-05-21
Teach DTS generator about RV32E

aswaterman:
Right now, the DTS generator unconditionally prints RVxxI.  Make it so that it knows how to print RVxxE, too.

**Release Notes**

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -72,12 +72,13 @@ trait HasNonDiplomaticTileParameters {
 
   // TODO merge with isaString in CSR.scala
   def isaDTS: String = {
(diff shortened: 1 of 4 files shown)


src/main/scala/tile/BaseTile.scala
```diff
@@ -72,12 +72,13 @@ trait HasNonDiplomaticTileParameters {
 
   // TODO merge with isaString in CSR.scala
```
hcook: Does "isaString in CSR.scala" do the right thing?

aswaterman: No, and it didn't matter for what I was trying to do, but I see no reason not to fix it.

PR #1969, opened 2019-05-22
Add OMBusMemory

derekpappas:
**Related issue**: 

**Release Notes**

Adds OMBusMemory's and OMSRAM's

### DIFF
--- a/src/main/scala/amba/axi4/SRAM.scala
+++ b/src/main/scala/amba/axi4/SRAM.scala
@@ -5,17 +5,20 @@ package freechips.rocketchip.amba.axi4
 import Chisel._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.diplomaticobjectmodel.logicaltree.{BusMemoryLogicalTreeNode, LogicalModuleTree, LogicalTreeNode}
+import freechips.rocketchip.diplomaticobjectmodel.model.{AHB_Lite, AXI4_Lite}
 import freechips.rocketchip.util._
 
 class AXI4RAM(
     address: AddressSet,
     cacheable: Boolean = true,
+    parentLogicalTreeNode: Option[LogicalTreeNode] = None,
     executable: Boolean = true,
     beatBytes: Int = 4,
     devName: Option[String] = None,
     errors: Seq[AddressSet] = Nil,
     wcorrupt: Boolean = false)
-  (implicit p: Parameters) extends DiplomaticSRAM(address, beatBytes, devName)
+  (implicit p: Parameters) extends DiplomaticSRAM(address, beatBytes, parentLogicalTreeNode, devName)
 {
   val node = AXI4SlaveNode(Seq(AXI4SlavePortParameters(
     Seq(AXI4SlaveParameters(
@@ -32,7 +35,20 @@ class AXI4RAM(
 
   lazy val module = new LazyModuleImp(this) {
     val (in, _) = node.in(0)
-    val (mem, omMem) = makeSinglePortedByteWriteSeqMem(1L << mask.filter(b=>b).size)
+    val (mem, omSRAM, omMem) = makeSinglePortedByteWriteSeqMem(size = 1 << mask.filter(b=>b).size)
+
+    parentLogicalTreeNode.map {
+      case parentLTN =>
+        def sramLogicalTreeNode = new BusMemoryLogicalTreeNode(
+          device = device,
+          omSRAMs = Seq(omSRAM),
+          busProtocol = new AXI4_Lite(None),
+          dataECC = None,
+          hasAtomics = None,
+          busProtocolSpecification = None)
+        LogicalModuleTree.add(parentLTN, sramLogicalTreeNode)
+    }
+
     val corrupt = if (wcorrupt) Some(SeqMem(1 << mask.filter(b=>b).size, UInt(width=2))) else None
 
     val r_addr = Cat((mask zip (in.ar.bits.addr >> log2Ceil(beatBytes)).asBools).filter(_._1).map(_._2).reverse)
@@ -102,6 +118,7 @@ object AXI4RAM
   def apply(
     address: AddressSet,
     cacheable: Boolean = true,
+    parentLogicalTreeNode: Option[LogicalTreeNode] = None,
     executable: Boolean = true,
     beatBytes: Int = 4,
     devName: Option[String] = None,
@@ -115,6 +132,5 @@ object AXI4RAM
       beatBytes = beatBytes,
       devName = devName,
       errors = errors))
-    axi4ram.node
   }
 }
--- a/src/main/scala/amba/axi4/SRAM.scala
+++ b/src/main/scala/amba/axi4/SRAM.scala
@@ -6,7 +6,7 @@ import Chisel._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.diplomaticobjectmodel.logicaltree.{BusMemoryLogicalTreeNode, LogicalModuleTree, LogicalTreeNode}
-import freechips.rocketchip.diplomaticobjectmodel.model.{AHB_Lite, AXI4_Lite}
+import freechips.rocketchip.diplomaticobjectmodel.model.AXI4_Lite
 import freechips.rocketchip.util._
 
 class AXI4RAM(
@@ -18,7 +18,7 @@ class AXI4RAM(
     devName: Option[String] = None,
     errors: Seq[AddressSet] = Nil,
     wcorrupt: Boolean = false)
-  (implicit p: Parameters) extends DiplomaticSRAM(address, beatBytes, parentLogicalTreeNode, devName)
+  (implicit p: Parameters) extends DiplomaticSRAM(address, beatBytes, devName)
 {
   val node = AXI4SlaveNode(Seq(AXI4SlavePortParameters(
     Seq(AXI4SlaveParameters(
@@ -132,5 +132,6 @@ object AXI4RAM
       beatBytes = beatBytes,
       devName = devName,
       errors = errors))
+    axi4ram.node
   }
 }
--- a/src/main/scala/diplomacy/SRAM.scala
+++ b/src/main/scala/diplomacy/SRAM.scala
@@ -32,10 +32,7 @@ abstract class DiplomaticSRAM(
   def mask: List[Boolean] = bigBits(address.mask >> log2Ceil(beatBytes))
 
   // Use single-ported memory with byte-write enable
-  def makeSinglePortedByteWriteSeqMem(
-    size: BigInt,
-    lanes: Int = beatBytes,
-    bits: Int = 8) = {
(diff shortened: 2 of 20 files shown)


src/main/scala/diplomacy/SRAM.scala
```diff
@@ -34,16 +32,43 @@ abstract class DiplomaticSRAM(
...
+          device = device,
+          omSRAMs = Seq(omSRAM),
+          busProtocol = busProtocol.getOrElse(throw new IllegalArgumentException("Protocol not specified")),
+          dataECC = dataECC,
+          hasAtomics = hasAtomics,
+          busProtocolSpecification = None)
+        LogicalModuleTree.add(parentLTN, sramLogicalTreeNode)
+    }
+
+
```
hcook: Why is there both OMMem and OMSRAM?

hcook: Seems to me like you should unify them, return it/them, and add them to the logical tree in the parent

src/main/scala/diplomacy/SRAM.scala
```diff
@@ -34,16 +32,43 @@ abstract class DiplomaticSRAM(
...
+      data = Vec(lanes, UInt(width = bits)),
+      id = DescribedSRAMIdAssigner.genId()
     )
     devName.foreach(n => mem.suggestName(n.split("-").last))
 
+    val omSRAM: OMSRAM = DiplomaticObjectModelAddressing.makeOMSRAM(
+      desc = "mem", //lim._2.name.map(n => n).getOrElse(lim._1.name),
+      depth = size,
+      data = Vec(lanes, UInt(width = bits)),
+      id = DescribedSRAMIdAssigner.genId()
```
hcook: Is it correct to call this function twice? Shouldn't you call it once and use the result in both places?

src/main/scala/diplomacy/SRAM.scala
```diff
@@ -34,16 +32,43 @@ abstract class DiplomaticSRAM(
...
 
   // Use single-ported memory with byte-write enable
-  def makeSinglePortedByteWriteSeqMem(size: BigInt, lanes: Int = beatBytes, bits: Int = 8) = {
+  def makeSinglePortedByteWriteSeqMem(
+    size: BigInt,
+    lanes: Int = beatBytes,
+    bits: Int = 8,
+    busProtocol: Option[OMProtocol],
+    dataECC: Option[OMECC] = None,
+    hasAtomics: Option[Boolean] = None) = {
```
hcook: I think you should return `omSRAM` from this function and add it to logical tree at the call site, instead of having to pass all these extra arguments in to this function that are only used to make the logical tree node

src/main/scala/rocket/ICache.scala
```diff
@@ -182,7 +182,8 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
     name = "tag_array",
     desc = "ICache Tag Array",
     size = nSets,
-    data = Vec(nWays, UInt(width = tECC.width(1 + tagBits)))
+    data = Vec(nWays, UInt(width = tECC.width(1 + tagBits))),
+    id = DescribedSRAMIdAssigner.genId()
```
hcook: Why does this get passed in instead of just getting called inside `DescribedSRAM.apply`?

src/main/scala/util/DescribedSRAM.scala
```diff
@@ -46,7 +47,7 @@ object DescribedSRAM {
       depth = size,
       description = desc,
       write_mask_granularity = granWidth,
-      uid = DescribedSRAMIdAssigner.genId())
+      uid = id)
```
hcook: This change seems worse than returning `id` from this function along with `mem`

src/main/scala/tilelink/SRAM.scala
```diff
@@ -223,7 +234,7 @@ object TLRAM
     devName: Option[String] = None,
   )(implicit p: Parameters): TLInwardNode =
   {
-    val ram = LazyModule(new TLRAM(address, cacheable, executable, atomics, beatBytes, ecc, devName))
+    val ram = LazyModule(new TLRAM(address, parentLogicalTreeNode, cacheable, executable, atomics, beatBytes, ecc, devName))
```
hcook: Could you give TLRAM its own `logicalTreeNode`, and add that to parent here instead of having to pass the parent all that way in?

src/main/scala/diplomacy/SRAM.scala
derekpappas: OMMem is legacy and can be removed soon.

src/main/scala/diplomacy/SRAM.scala
derekpappas: Fixed

src/main/scala/diplomacy/SRAM.scala
derekpappas: OMMem is used and will be replaced in another PR

src/main/scala/rocket/ICache.scala
derekpappas: fixed

src/main/scala/util/DescribedSRAM.scala
derekpappas: removed

src/main/scala/tilelink/SRAM.scala
derekpappas: The parent is passed to the child starting from the root.

src/main/scala/diplomacy/SRAM.scala
```diff
@@ -34,22 +32,28 @@ abstract class DiplomaticSRAM(
...
+
+    val (mem, omSRAM) =  DescribedSRAM(
       name = devName.getOrElse("mem"),
       desc = devName.getOrElse("mem"),
       size = size,
       data = Vec(lanes, UInt(width = bits))
     )
     devName.foreach(n => mem.suggestName(n.split("-").last))
 
+    val id = DescribedSRAMIdAssigner.genId()
```
hcook: Is this `id` actually used anywhere?

src/main/scala/tilelink/SRAM.scala
```diff
@@ -53,7 +56,22 @@ class TLRAM(
...
+      bits = width)
+
+    parentLogicalTreeNode.map {
+      case parentLTN =>
+        def sramLogicalTreeNode = new BusMemoryLogicalTreeNode(
+          device = device,
+          omSRAMs = Seq(omSRAM),
+          busProtocol = new TL_C(None),
+          dataECC = None,
+          hasAtomics = None,
```
hcook: Why don't you set these to the actual values from the passed- in parameters? As written this will not produce correct OM information!

src/main/scala/util/DescribedSRAM.scala
```diff
@@ -26,8 +28,9 @@ object DescribedSRAM {
     name: String,
     desc: String,
     size: BigInt, // depth
-    data: T
-  ): SyncReadMem[T] = {
+    data: T,
+    uid: Int = -1
```
hcook: Why are you allowing it to be passed in here when you ignore it below?

src/main/scala/diplomacy/SRAM.scala
derekpappas: fixed

src/main/scala/tilelink/SRAM.scala
derekpappas: fixed

derekpappas: fixed

src/main/scala/util/DescribedSRAM.scala
derekpappas: fixed

src/main/scala/tilelink/SRAM.scala
hcook: I'm not seeing this as fixed, did you forget to push something?

src/main/scala/amba/axi4/SRAM.scala
```diff
@@ -115,6 +132,5 @@ object AXI4RAM
       beatBytes = beatBytes,
       devName = devName,
       errors = errors))
-    axi4ram.node
```
hcook: Was this intentional?

src/main/scala/diplomacy/SRAM.scala
```diff
@@ -5,25 +5,23 @@ package freechips.rocketchip.diplomacy
...
 import freechips.rocketchip.diplomaticobjectmodel.logicaltree.LogicalModuleTree.cache
-import freechips.rocketchip.diplomaticobjectmodel.model.{OMMemory, OMMemoryRegion, OMRTLInterface, OMRTLModule}
-import freechips.rocketchip.util.DescribedSRAM
+import freechips.rocketchip.diplomaticobjectmodel.model._
+import freechips.rocketchip.util.{DescribedSRAM, DescribedSRAMIdAssigner}
 
 abstract class DiplomaticSRAM(
     address: AddressSet,
     beatBytes: Int,
+    parentLogicalTreeNode: Option[LogicalTreeNode],
```
hcook: Why are you passing this in?

src/main/scala/diplomacy/SRAM.scala
```diff
@@ -5,25 +5,23 @@ package freechips.rocketchip.diplomacy
 import Chisel._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelAddressing
-import freechips.rocketchip.diplomaticobjectmodel.logicaltree.LogicalModuleTree
+import freechips.rocketchip.diplomaticobjectmodel.logicaltree.{BusMemoryLogicalTreeNode, LogicalModuleTree, LogicalTreeNode}
```
hcook: Unnecessary imports?

src/main/scala/tilelink/SRAM.scala
```diff
@@ -53,7 +56,22 @@ class TLRAM(
...
+      size = 1 << addrBits.size,
+      lanes = lanes,
+      bits = width)
+
+    parentLogicalTreeNode.map {
+      case parentLTN =>
+        def sramLogicalTreeNode = new BusMemoryLogicalTreeNode(
+          device = device,
+          omSRAMs = Seq(omSRAM),
+          busProtocol = new TL_C(None),
```
hcook: This should be `TL_UL`

colinschmidt: Having to explain whats already setup in diplomacy seems weird...

src/main/scala/diplomacy/SRAM.scala
derekpappas: fixed

src/main/scala/diplomacy/SRAM.scala
derekpappas: fixed

src/main/scala/tilelink/SRAM.scala
derekpappas: > This should be `TL_UL`

fixed

src/main/scala/amba/axi4/SRAM.scala
derekpappas: fixed

PR #1970, opened 2019-05-23
util: add width param to PlusArg and update it to chisel3

hcook:
- PlusArgs can now have explicitly specified widths wider than 32b.
- `PlusArg` still parses a %d decimal
- The default value is passed with a `RawParam` that sets the width (this could be turned back into an `IntParam` if FIRRTL correctly output BigInt literals)
- The default width is 32b
- Updated to use chisel3 apis and imports


hcook:
I got this to work in a flow that uses firttl's protobuf features. Seems like it fails to escape the single quote correctly going down the other path. @jackkoenig to help me resolve, probably necessitating a firrtl bump

PR #1978, opened 2019-05-26
Revert "sbus: actually use policy param"

richardxia:
Reverts freechipsproject/rocket-chip#1915

I'm temporarily reverting this PR because it's causing integration issues within SiFive's private repositories, especially because #1971 and #1976 require some code to be updated to reflect a new API for the LogicalTreeNodes/Object Model, but #1915 is caught in between those two, which makes bumping the rocket-chip pointer difficult.


richardxia:
Closing this because we've discussed a different way of working around the integration issue.

PR #1979, opened 2019-05-27
Add OMCoreComplexRTLModule and support for om signals

derekpappas:
**Related issue**: 

**Release Notes**

Add OMCoreComplexRTLModule and support for om signals

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
+++ b/src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
@@ -4,9 +4,11 @@ package freechips.rocketchip.diplomaticobjectmodel
 
 
 import freechips.rocketchip.diplomaticobjectmodel.logicaltree._
+import freechips.rocketchip.diplomaticobjectmodel.model.OMInterruptSignal
 
 trait HasLogicalTreeNode {
   def logicalTreeNode: LogicalTreeNode
   def addLogicalTreeNode(childLogicalTreeNode: LogicalTreeNode): Unit = LogicalModuleTree.add(logicalTreeNode, childLogicalTreeNode)
+  def getOMGlobalInterruptSignals(): Option[OMInterruptSignal]
 }
 
--- a/src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
+++ b/src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
@@ -9,6 +9,5 @@ import freechips.rocketchip.diplomaticobjectmodel.model.OMInterruptSignal
 trait HasLogicalTreeNode {
   def logicalTreeNode: LogicalTreeNode
   def addLogicalTreeNode(childLogicalTreeNode: LogicalTreeNode): Unit = LogicalModuleTree.add(logicalTreeNode, childLogicalTreeNode)
-  def getOMGlobalInterruptSignals(): Option[OMInterruptSignal]
 }
 
--- a/src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
+++ b/src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
@@ -4,7 +4,6 @@ package freechips.rocketchip.diplomaticobjectmodel
 
 
 import freechips.rocketchip.diplomaticobjectmodel.logicaltree._
-import freechips.rocketchip.diplomaticobjectmodel.model.OMInterruptSignal
 
 trait HasLogicalTreeNode {
   def logicalTreeNode: LogicalTreeNode
--- a/src/main/scala/diplomaticobjectmodel/model/OMCoreComplex.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMCoreComplex.scala
@@ -8,8 +8,8 @@ case class CoreComplexRTLInterface(
   resets: Seq[OMRTLReset] = Nil,
   statuses: Seq[OMStatus] = Nil,
   localInterrupts: Seq[OMInterruptSignal] = Nil,  // E.g. local_interrupts_X (all local interrupts for core X)
-  globalInterrupts: Seq[OMInterruptSignal] = Nil,  // E.g. global_interrupts
-  machineExternalInterrupts: Option[OMInterruptSignal] = None,  // E.g. meip_X
+  globalInterrupts: Option[OMInterruptSignal] = None,  // E.g. global_interrupts
+  machineExternalInterrupts: Seq[OMInterruptSignal] = Nil,  // E.g. meip_X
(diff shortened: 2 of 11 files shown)


src/main/scala/diplomaticobjectmodel/model/OMCoreComplex.scala
```diff
@@ -2,8 +2,27 @@
...
 package freechips.rocketchip.diplomaticobjectmodel.model
 
+case class CoreComplexRTLInterface(
+  clocks: List[OMClock] = Nil,
+  clockRelationships: List[OMClockRelationship] = Nil,
+  resets: List[OMRTLReset] = Nil,
+  statuses: List[OMStatus] = Nil,
+  localInterrupts: Option[OMInterruptSignal] = None,  // E.g. local_interrupts_X (all local interrupts for core X)
+  globalInterrupts: Option[OMInterruptSignal] = None,  // E.g. global_interrupts
+  machineExternalInterrupts: Option[OMInterruptSignal] = None,  // E.g. meip_X
```
richardxia: These should be Seqs, not Options. I believe the current implementation will have one instance of the local and the machine external interrupts per core. The global interrupts might be a singleton.

src/main/scala/diplomaticobjectmodel/HasLogicalTreeNode.scala
```diff
@@ -4,6 +4,7 @@ package freechips.rocketchip.diplomaticobjectmodel
 
 
 import freechips.rocketchip.diplomaticobjectmodel.logicaltree._
+import freechips.rocketchip.diplomaticobjectmodel.model.OMInterruptSignal
```
richardxia: Why is this being imported here?

src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
```diff
@@ -31,22 +35,79 @@ trait Asynchronous extends Synchronicity
...
+  synchronicity: Option[Synchronicity],
+  uid: Int,
+  _types: Seq[String] = Seq("OMReset", "OMSignal", "OMRTLComponent", "OMCompoundType")
+) extends OMSignal
+
+// Status
+trait OMStatus extends OMSignal {
+  def description: Option[String]
+  def name: String
+  uid: Int
```
richardxia: I think you mixed the syntax for case classes with traits here.

src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
```diff
@@ -31,22 +35,79 @@ trait Asynchronous extends Synchronicity
...
+}
+
+trait OMCoreStatus extends OMStatus {
+  def description: Option[String]
+  def name: String
```
richardxia: These two fields are already defined in `OMStatus`, so you don't need to repeat them here.

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -74,7 +74,27 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem with
...
+    val rtlIfc = CoreComplexRTLInterface(
+      clocks = Nil,
+      clockRelationships = Nil,
+      resets = Nil,
+      statuses = Nil,
+      localInterrupts = None, // E.g. local_interrupts_X (all local interrupts for core X)
+      globalInterrupts = None, // E.g. global_interrupts
+      machineExternalInterrupts = None, // E.g. meip_X
+      testModeSignals = Nil // E.g. debug_psd_test_mode and debug_psd_test_mode_reset
+    )
```
richardxia: What is the purpose of filling in all of these fields with `None` and `Nil`? If the whole function returns an `Option[OMCoreComplexRTLModule]`, why not just return `None` for the whole thing instead of filling out null values for each individual field?

src/main/scala/system/ExampleRocketSystem.scala
```diff
@@ -37,6 +37,8 @@ class ExampleRocketSystem(implicit p: Parameters) extends RocketSubsystem
     sbus.coupleTo("coherence_manager") { in :*= _ }
     mbus.coupleFrom("coherence_manager") { _ :=* BankBinder(mbus.blockBytes * (nBanks-1)) :*= out }
   }
+
+  override def getOMGlobalInterruptSignals() = None
```
hcook: There actually are interrupt signals, in trait `HasAsyncExtInterrupts` which is mixed in here.

src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
mwachs5: Not sure why you are calling this out here in particular... there is already quite a mix of those in this file.

src/main/scala/subsystem/BaseSubsystem.scala
mwachs5: I'm assuming this is where designer should fill in their contents?

src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
richardxia: No, I'm specifically referring to the fact that this particular trait uses both `def` and non-`def`. In every other thing in this file, all the Traits only use `def` and all the case classes just use constructor arguments. I don't actually understand what this syntax means, and I originally thought it would be a syntax error. Can you explain what this syntax means, since I haven't seen it before? Traits in Scala aren't allowed to define a constructor, so I don't really know what this specific `uid` line means.

src/main/scala/subsystem/BaseSubsystem.scala
richardxia: But there's no purpose to this object instance; it doesn't represent anything that couldn't be represented with a `None`. If someone else wanted to later fill in a value, they could easily just pattern match on the `None` and fill in their own values in the same way.

mwachs5: Ok, I think I understand what you're saying... let me rephrase it here and you sanity check?

It's basically expected that someone is going to override this `  def getCoreComplexRTLModule(): Option[OMCoreComplexRTLModule] = ` entirely. And therefore, there is no reason that this default version shouldn't just return `None`, vs bothering to fill out a bunch of useless instances.

is that what you mean?

richardxia: Yes, that's what I mean, although I don't know exactly what the original intent of this function was.

mwachs5: I assumed that Derek was just creating the skeleton, then would ask designer to come modify *this code* to fill in the appropriate values, as part of this PR. At any rate I will attempt to do *that* to see if it makes sense, otherwise I'll just return `None` as you explained.

PR #1980, opened 2019-05-28
Fix logical tree node for Rocket memories

richardxia:
**Related issue**: 

**Release Notes**

After the refactoring done in https://github.com/freechipsproject/rocket-chip/pull/1956, we accidentally introduced a regression that caused the memory addresses reported in the object model to be incorrect for the memories inside of the Rocket Tile. The cause of the bug is that the same set of resource bindings that is associated with the Rocket Core itself were getting passed in to the code that generates the Object Model for the ICache/DCache/DTIM. However, the resource bindings for the Rocket Core uses the `reg` field — which is normally used to represent a memory address — in order to represent the hart id. Hence, when passed to the ICache/DCache/DTIM, it was reading out a memory address of "0", which actually comes from it being hart 0.

The correct solution to this is that the RocketLogicalTreeNode cannot be the one that constructs the object models for the ICache or DCache; instead we need to construct separate LogicalTreeNodes for the ICache and DCache, attach them to the RocketLogicalTreeNode, and then let the `LogicalModuleTree.bind()` be responsible for traversing the tree and passing in the appropriate resource bindings to each tree node. Then, the RocketLogicalTreeNode can get access to its child OMComponents via the `components` argument to `getOMComponents()`, which we can then filter for OMICaches or OMDCaches.

I still want to do a bit more testing of this on SiFive's internal configurations, since I think I was a little too liberal with calling `.get` on options here, and I want to do a bit more testing to see how I should more safely handle these cases. Otherwise, I think this is mostly ready for a review.

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
+++ b/src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
@@ -30,8 +30,7 @@ class RocketLogicalTreeNode(
   device: SimpleDevice,
   rocketParams: RocketTileParams,
   dtim_adapter: Option[ScratchpadSlavePort],
-  XLen: Int,
-  icacheLTN: ICacheLogicalTreeNode
+  XLen: Int
 ) extends LogicalTreeNode(() => Some(device)) {
 
   def getOMDCacheFromBindings(dCacheParams: DCacheParams, resourceBindings: ResourceBindings): Option[OMDCache] = {
@@ -54,7 +53,8 @@ class RocketLogicalTreeNode(
 
     val omDCache = rocketParams.dcache.flatMap{ getOMDCacheFromBindings(_, resourceBindings)}
 
-    val omICache = icacheLTN.iCache(resourceBindings)
+    // Expect that one of the components passed in is the ICache.
+    val omICache = components.collectFirst { case x: OMICache => x }.get
 
     Seq(OMRocketCore(
       isa = OMISA.rocketISA(coreParams, XLen),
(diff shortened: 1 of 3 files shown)


richardxia:
Fixed a merge conflict. Will merged on CI pass.

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
```diff
@@ -52,9 +66,11 @@ class RocketLogicalTreeNode(
   override def getOMComponents(resourceBindings: ResourceBindings, components: Seq[OMComponent]): Seq[OMComponent] = {
     val coreParams = rocketParams.core
 
-    val omDCache = rocketParams.dcache.flatMap{ getOMDCacheFromBindings(_, resourceBindings)}
+    // Expect that one of the components passed in is the DCache/DTIM.
+    val omDCache = components.collectFirst { case x: OMDCache => x }.get
```
derekpappas: My only concern is the .get

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
```diff
@@ -52,9 +66,11 @@ class RocketLogicalTreeNode(
...
 
-    val omICache = icacheLTN.iCache(resourceBindings)
+    // Expect that one of the components passed in is the ICache.
+    val omICache = components.collectFirst { case x: OMICache => x }.get
```
derekpappas: My only concern is the .get

PR #1981, opened 2019-05-28
Add HasBlackBoxResource to EICG_wrapper so that it ends up in the .f file

jwright6323:
**Related issue**: 

**Release Notes**
This adds HasBlackBoxResource to the EICG_wrapper blackbox so that it ends up in the .f file.

### DIFF
--- a/src/main/scala/util/ClockGate.scala
+++ b/src/main/scala/util/ClockGate.scala
@@ -3,6 +3,7 @@
 package freechips.rocketchip.util
 
 import chisel3._
+import chisel3.util.HasBlackBoxResource
 import freechips.rocketchip.config.{Field, Parameters}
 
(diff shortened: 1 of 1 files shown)


aswaterman:
IIRC, the reason we didn’t do this is that it interferes with having different process-specific implementations of this module with the same module name but at different paths.

jwright6323:
Ah, ok. Couldn't you override the `ClockGateImpl` config to the process-specific wrapper to accomplish the same thing? I guess that would be an "API change", though.

aswaterman:
Yeah, that would work, but it would cause us a lot of trouble.  As it is, this way at least fails noisily: if you fail to provide an implementation, you'll get a linkage error, rather than a broken circuit.

sequencer:
closed since #2969 merged.

PR #1983, opened 2019-05-30
DTS: Allow SimpleBus to add additional 'misc' fields

mwachs5:
This adds the ability to incorporate arbitrary additional fields onto a `SimpleBus` device. 

Certainly this functionality would be generally applicable on e.g. a `SimpleDevice`, but the way the code is currently factored, it's not straightforward to pass this through.

This does not update any of the callsites for `SimpleDevice`, as the default parameter for the arbitrary additional fields is an empty list.

**Related issue**: 

**Type of change**:feature request

**Release Notes**


mwachs5:
Ok, I thought that seemed like a lot of duplicated code since the `describe` method is the vast majority of the `SimpleBus` code, and I only want to add something to it. But I can do that.

mwachs5:
Oh. I see, I can make up my own `extra` and then `super.describe`. Got it.

PR #1984, opened 2019-05-30
Switch to firrtl.stage.FirrtlMain

seldridge:
This changes the underlying FIRRTL app from `firrtl.Driver` to `firrtl.stage.FirrtlStage`. For the purposes of Rocket Chip, these should be equivalent. The only things that I think could affect Rocket Chip is the removal of `--top-name` and `--split-modules`. Top name was really a convoluted way to specify an output file which Rocket Chip never used publicly as far as I'm aware. Use explicit options (as Rocket Chip does) to do that going forward:
```
  -faf, --annotation-file <file>
                           An input annotation file
  -foaf, --output-annotation-file <file>
                           An output annotation file
  -i, --input-file <file>  An input FIRRTL file
  -o, --output-file <file>
                           The output FIRRTL file
```

For the emitters, you get a default single-file emitter derived from the compiler. If you want a one-file-per-module emitter, use an explicit `-e/--emit-modules` emitter. In summary:
```
  -E, --emit-circuit <chirrtl|high|middle|low|verilog|mverilog|sverilog>
                           Run the specified circuit emitter (all modules in one file)
  -e, --emit-modules <chirrtl|high|middle|low|verilog|mverilog|sverilog>
                           Run the specified module emitter (one file per module)
```

**Related Issue**: https://github.com/freechipsproject/firrtl/pull/1005

**Release Notes**

- Switch from firrtl.Driver to firrtl.stage.FirrtlMain


seldridge:
@debs-sifive: You may want to look at this as a follow-up to the Stage/Phase merge.

azidar:
Subsumed

PR #1985, opened 2019-06-01
added debugging and refactored the bind function

derekpappas:
**Related issue**: 

**Release Notes**

Adding requires and refactoring

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
+++ b/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
@@ -125,9 +125,10 @@ object DiplomaticObjectModelAddressing {
     )
   }
 
-  private def omAddressSets(ranges: Seq[AddressSet]): Seq[OMAddressSet] = {
+  private def omAddressSets(ranges: Seq[AddressSet], name: String): Seq[OMAddressSet] = {
     ranges.map {
       case AddressSet(base, mask) =>
+        require(mask != 0, s"""ERROR: omAddressSets: $name is mask is 0""")
         OMAddressSet(base = base, mask = mask)
     }
   }
@@ -135,8 +136,8 @@ object DiplomaticObjectModelAddressing {
   private def omMemoryRegion(name: String, description: String, value: ResourceValue, omRegMap: Option[OMRegisterMap]): OMMemoryRegion = {
     val (omRanges, permissions) = value match {
       case rm: ResourceMapping =>
-        (omAddressSets(rm.address),rm.permissions)
-      case ra: ResourceAddress => (omAddressSets(ra.address), ra.permissions)
+        (omAddressSets(rm.address, name),rm.permissions)
+      case ra: ResourceAddress => (omAddressSets(ra.address, name), ra.permissions)
       case _ => throw new IllegalArgumentException()
     }
 
--- a/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
+++ b/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
@@ -128,7 +128,7 @@ object DiplomaticObjectModelAddressing {
   private def omAddressSets(ranges: Seq[AddressSet], name: String): Seq[OMAddressSet] = {
     ranges.map {
       case AddressSet(base, mask) =>
-        require(mask != 0, s"""ERROR: omAddressSets: $name is mask is 0""")
+        require(mask != 0, s"omAddressSets: $name has invalid mask of 0")
         OMAddressSet(base = base, mask = mask)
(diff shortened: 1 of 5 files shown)


src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
```diff
@@ -125,18 +125,19 @@ object DiplomaticObjectModelAddressing {
     )
   }
 
-  private def omAddressSets(ranges: Seq[AddressSet]): Seq[OMAddressSet] = {
+  private def omAddressSets(ranges: Seq[AddressSet], name: String): Seq[OMAddressSet] = {
     ranges.map {
       case AddressSet(base, mask) =>
+        require(mask != 0, s"""ERROR: omAddressSets: $name is mask is 0""")
```
richardxia: "name is mask is 0" seems like a typo. Also, putting the `ERROR` at the beginning is not necessary because this very clearly causes a requirement failure error in Scala, and it also doesn't match our assertion message style.

```suggestion
        require(mask != 0, s"omAddressSets: $name has invalid mask of 0")
```

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
```diff
@@ -51,7 +51,8 @@ object LogicalModuleTree {
   }
 
   def resourceBindings(deviceOpt: () => Option[Device], maps: ArrayBuffer[ResourceBindingsMap]): ResourceBindings = deviceOpt() match {
-    case Some(device) => getResourceBindings(device, maps)
+    case Some(device) =>
+      getResourceBindings(device, maps)
```
richardxia: I think the previous style of having this on a single line is better and at least more consistent with the None case below.

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
```diff
@@ -16,6 +16,11 @@ import freechips.rocketchip.tile.{RocketTileParams, TileParams, XLen}
  */
 class DCacheLogicalTreeNode(deviceOpt: Option[SimpleDevice], params: DCacheParams) extends LogicalTreeNode(() => deviceOpt) {
   def getOMComponents(resourceBindings: ResourceBindings, children: Seq[OMComponent]): Seq[OMComponent] = {
+    deviceOpt match {
+      case Some(device) => require(!resourceBindings.map.isEmpty, s"""ERROR: Memory region ${device.devname} is empty""")
+      case None =>
+    }
```
richardxia: Since you don't do anything with the `None` case, this is probably clearer as:

```scala
deviceOpt.forEach {
  device => require(!resourceBindings.map.isEmpty, s"""ERROR: Memory region ${device.devname} is empty""")
}
```

Also, the error message isn't correct. "device.devname" isn't a memory region; it's a Device. The real error is that if you have a device, then we expect to have resource bindings associated with the device to be passed in, not that the device doesn't have memory regions.

src/main/scala/rocket/ScratchpadSlavePort.scala
```diff
@@ -18,11 +18,7 @@ class ScratchpadSlavePort(address: Seq[AddressSet], coreDataBytes: Int, usingAto
     this(Seq(address), coreDataBytes, usingAtomics)
   }
 
-  val device = new SimpleDevice("dtim", Seq("sifive,dtim0")) {
-    def getMemory(p: DCacheParams, resourceBindings: ResourceBindings): OMDCache = {
-      OMCaches.dcache(p, resourceBindings)
```
richardxia: Is `OMCaches.dcache()` used anywhere anymore? If not, let's delete it.

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
derekpappas: fixed

src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
derekpappas: fixed

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
derekpappas: fixed

src/main/scala/rocket/ScratchpadSlavePort.scala
derekpappas: It is used inRocketLogicalTreeNode.

PR #1998, opened 2019-06-06
Broadcast: in the case of an Acquire.NtoB, respond Grant.toT if we can

terpstra:

**Release Notes**
Improve performance of systems with a front port or more than 1 core, but no L2.


terpstra:
Just FYI, I'm waiting to hear back from a SPEC run on FPGA before I merge this. I'm a coward.

PR #1999, opened 2019-06-07
Don't accept probes while waiting for release ack

aswaterman:
Prevents propagation of stale data when C channel isn't FIFO

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -168,6 +168,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   val flushing = Reg(init=Bool(false))
   val cached_grant_wait = Reg(init=Bool(false))
   val release_ack_wait = Reg(init=Bool(false))
+  val release_ack_addr = Reg(UInt(paddrBits.W))
   val can_acquire_before_release = !release_ack_wait && release_queue_empty
   val release_state = Reg(init=s_ready)
   val any_pstore_valid = Wire(Bool())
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/DCache.scala
```diff
@@ -626,9 +626,10 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   ccover(tl_out.d.valid && !tl_out.d.ready, "BLOCK_D", "D$ D-channel blocked")
 
   // Handle an incoming TileLink Probe message
-  val block_probe = releaseInFlight || grantInProgress || blockProbeAfterGrantCount > 0 || lrscValid
-  metaArb.io.in(6).valid := tl_out.b.valid && (!block_probe || lrscBackingOff)
-  tl_out.b.ready := metaArb.io.in(6).ready && !block_probe && !s1_valid && !s2_valid
+  val block_probe_for_core_progress = blockProbeAfterGrantCount > 0 || lrscValid
+  val block_probe_for_ordering = releaseInFlight || release_ack_wait || grantInProgress
+  metaArb.io.in(6).valid := tl_out.b.valid && (!block_probe_for_core_progress || lrscBackingOff)
+  tl_out.b.ready := metaArb.io.in(6).ready && !(block_probe_for_core_progress || block_probe_for_ordering || s1_valid || s2_valid)
```
terpstra: I don't understand why the conditions for metaArg.io.in(6).valid don't match thoes for tl_outb.ready.

aswaterman: The arbiter valid signal is high more often than a probe can actually be accepted, as a means of preventing the core (which is the next-lowest priority port into the arbiter) from being able to initiate new accesses when a probe is pending.  This prevents a circular-wait condition, because probes can't be accepted while the cache pipeline is busy.

terpstra: Ahh. This makes sense. Thanks.

terpstra:
Won't this hurt performance quite badly on a multicore system? This blocks all Probes while a Release is outstanding.

terpstra:
Could we perhaps block only when the Probe's set number matches the incomplete Acquire's set number? (which must be the same as the set number of the released block)

hcook:
I would rephrase @terpstra's question as: "Could forward progress on probes still be made during`!releaseInFlight && release_ack_wait` iff the probe address's index doesn't match the index of the release whose ack is still pending?"

I'm confident that acking such probes would sound from a protocol perspective, but not confident I know how to express that progress microarchitecturally. Specifically, I couldn't convince myself that comparing `tl_out.b.address` and `probe_bits.address` was always going to be a sound thing to do once `!releaseInFlight`. Is there somewhere the pending release ack's address or index is guaranteed to be stored? Maybe from the original CPU-side request that triggered the voluntary release in the first place?

aswaterman:
Yeah, good point.  I don't think it'll make much of a difference when there's a directory, but it could sometimes make a big difference when there isn't.

Following @hcook's train of thought, I'm just going to add a new release-ack-pending set-index register.  The cost is very minor.

terpstra:
How are we going to test this? Tracegen obviously failed us last time. I suppose we need a 'TLShuffler' or something in the unit test where tracegen runs?

hcook:
[This call](https://github.com/freechipsproject/rocket-chip/blob/master/src/main/scala/groundtest/TestHarness.scala#L14) needs to make an AXI Shuffler inside itself, I think.

PR #2002, opened 2019-06-11
Genericize TestDriver.v in terms of Model

seldridge:
This adds a Verilog define holding the model name to TestDriver.v and
passes the model name (defined in the top level Makefrag) into a VCS
compilation pass.

This fixes a bug where deviating from the default model of
"TestHarness" would cause VCS to (rightly) not find the model.

**Release Notes**

- Enable vsim/vcs to work with models other than "TestHarness"

### DIFF
--- a/src/main/resources/vsrc/TestDriver.v
+++ b/src/main/resources/vsrc/TestDriver.v
@@ -3,6 +3,9 @@
 `ifndef RESET_DELAY
  `define RESET_DELAY 777.7
 `endif
+`ifndef MODEL
+ `define MODEL TestHarness
+`endif
 
 module TestDriver;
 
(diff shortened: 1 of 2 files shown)


colinschmidt:
Sorry I'm out of practice in upstream but is there not a need for a similar fix for verilator? Otherwise looks good.

seldridge:
Good thought. Verilator was working fine and I was wondering why, too...

The reason is the Verilator build maps [`$MODEL` -> `TEST_HARNESS`](https://github.com/freechipsproject/rocket-chip/blob/master/emulator/Makefrag-verilator#L65) which gets instantiated in [`src/main/resources/csrc/emulator.cc`](https://github.com/freechipsproject/rocket-chip/blob/master/src/main/resources/csrc/emulator.cc#L257).

PR #2004, opened 2019-06-12
subsystem: add 'WithNMedCores' config w. only FPU disabled

gsomlo:

**Release Notes**

Add an intermediate, "medium" config option, with only the FPU
disabled, that still provides an MMU and thus supports VM.


gsomlo:
@hcook, thanks for approving the PR! Any chance it could get pulled
soon (as I'm waiting to reference that config option from other projects)?

Thanks a ton,
--Gabriel

PR #2006, opened 2019-06-14
Equals

tampler:
Added only equals to func definitions to calm down  the compiler


tampler:
Minor refactoring. No changes to functionality

hcook:
I support updating our Scala style in this way, but I'm not convinced this is actually equivalent functionality in all cases! The `def f (...) { x }` syntax actually implies the code block returns `Unit`, regardless of what the value of the final statement x is. Your change results in functions returning the type of x instead. I think if you change your changes to be `def f (...): Unit = { x }` then it will preserve API compatibility with the original intention.

Also, can you please factor out at least the changes to the Scala/JVM settings, the new config classes, the blackbox resource update, and the annotations cleanup into separate PRs? I understand that you are not intending to change any RTL functionality with any of these, but having them in one mega PR makes it hard to assign any individual reviewer who know about all of these different packages. Smaller PRs will definitely increase review and approval throughput.

tampler:
Hi Henry
My fixes only calm down the Scala 2.12.8 compiler about your code. Chisel is still Scala DSL, so all rules apply. If you add -Xlint, -deprecated, -Xfuture  or similar checking knobs, your Rocket compilation will throw hundreds of warnings.
NOT very good for production designs, right ? So If you want to convince yourself,  simply turn on warnings and lint and think about the compiler output

tampler:
All those commits were accumulated during last 7 months. I'm not interested in maintaining those anymore. Closing the PR

PR #2009, opened 2019-06-22
subsystem: add MaskROM to RocketSubsystem

hcook:
Unify the addition of MaskROMs to all subsystem classes that inherit from `RocketSubsystem`

Factor out bus interconnections in `HasHierarchicalBusTopology`

### DIFF
--- a/src/main/scala/subsystem/BusTopology.scala
+++ b/src/main/scala/subsystem/BusTopology.scala
@@ -1,0 +1,27 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.subsystem
+
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.tilelink._
+
+trait HasHierarchicalBusTopology { this: BaseSubsystem =>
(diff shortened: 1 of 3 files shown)


src/main/scala/subsystem/BusTopology.scala
```diff
@@ -0,0 +1,27 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.subsystem
+
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.tilelink._
+
+trait HasHierarchicalBusTopology { this: BaseSubsystem =>
```
colinschmidt: Nice!

hcook: You can imagine mixing in a different set of inter-bus connections while using the same-named attachment points for all the devices...

colinschmidt: Yeah I actually already do this downstream but I had to duplicate more code than I wanted to :smile:

PR #2011, opened 2019-06-25
Changed lingering bits of dminner to use dmactive reset

ernie-sifive:
**Related issue**: 

**Release Notes**

The Debug Module had a few registers that were being reset by the default reset input (i.e. TileLink reset) rather than dmactive.  The async crossing modules from dmouter to dminner were among these and prevented DMI from accessing any registers in dminner while reset was asserted.

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -635,12 +635,14 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
     // Registers coming from 'CONTROL' in Outer
     //--------------------------------------------------------------
 
-    val selectedHartReg = RegInit(0.U(p(MaxHartIdBits).W))
+    val selectedHartReg = Reg(UInt(p(MaxHartIdBits).W))
       // hamaskFull is a vector of all selected harts including hartsel, whether or not supportHartArray is true
     val hamaskFull = Wire(init = Vec.fill(nComponents){false.B})
 
     if (nComponents > 1) {
-      when (io.innerCtrl.fire()){
+      when (~io.dmactive) {
+        selectedHartReg := 0.U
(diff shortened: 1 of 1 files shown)


src/main/scala/devices/debug/Debug.scala
```diff
@@ -1424,10 +1426,13 @@ class TLDebugModuleInnerAsync(device: Device, getNComponents: () => Int, beatByt
...
       dmInner.module.clock := gated_clock
       dmInner.module.io.dmactive := dmactive_synced
-      dmInner.module.io.innerCtrl := FromAsyncBundle(io.innerCtrl)
+      withReset (~dmactive_synced) {
+        dmInner.module.io.innerCtrl := FromAsyncBundle(io.innerCtrl)
+      }
       dmInner.module.io.debugUnavail := io.debugUnavail
       io.hgDebugInt := dmInner.module.io.hgDebugInt
       io.extTrigger.foreach { x => dmInner.module.io.extTrigger.foreach {y => x <> y}}
+      dmiXing.module.reset := false.B  // Safe AsyncQueue is reset from DMI side only
```
mwachs5: Why is this not dmactive_synced as well?

ernie-sifive: The reason is that this would prevent access to dmstatus when dmactive=0.  It's useful to be able to read the version before setting dmactive to 1.  My understanding of the Safe Async Queue is that only one side needs a reset because there is logic to reset both sides when either side's reset is asserted.

mwachs5: I see. The initial design was intended to prevent traffic to go across the crossing when periphery was in reset, but this is a conscious decision to change that design, so this makes sense. We just need to ensure that the regmapper does not need running clock or to be out of reset in order to respond properly.

PR #2012, opened 2019-06-25
PSD: Ensure that PSD is connected to dmactive-derived reset 

mwachs5:
When debug interface was APB, this was not getting connected up to the top level I/Os. Now this is connected up so that Post-Silicon-Debug controls can fully control the internal reset generated by `dmactive`.

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/devices/debug/Periphery.scala
+++ b/src/main/scala/devices/debug/Periphery.scala
@@ -96,6 +96,9 @@ trait HasPeripheryDebugModuleImp extends LazyModuleImp {
   // TODO in inheriting traits: Set this to something meaningful, e.g. "component is in reset or powered down"
   outer.debug.module.io.ctrl.debugUnavail.foreach { _ := Bool(false) }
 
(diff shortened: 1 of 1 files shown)


mwachs5:
@ernie-sifive this fails for the JTAG case with null pointer exception. If you notice anything obvious feel free to push to this branch

ernie-sifive:
Not sure I understand why, but this change seems to solve the problem ... we'll see if it passes Travis-CI.

mwachs5:
that... makes no sense to me, but 🤷‍♀ ? Maybe @jackkoenig could provide some insight into why that is the right thing to do.

ernie-sifive:
@mwachs5 I thought I was starting to understand Chisel but maybe not 😳.  What do you think - I can merge this and get it into federation if you like.

mwachs5:
If you visually inspected the verilog and the JTAG PSD is wired up all the way through, that is OK by me.

PR #1939, opened 2019-05-07
Added the HAUSER field to the ToAHB

kOstby:
This PR builds upon the work done in https://github.com/freechipsproject/rocket-chip/pull/1931 which adds add arbitrary meta-data to TileLink requests, and sends these out on the new AHB HAUSER port which has been added to the ToAHB module. 

**Related issue**: https://github.com/freechipsproject/rocket-chip/pull/1931 

**Type of change**: feature request 

**Impact**:  API modification

**Development Phase**: implementation

**Release Notes**
UserBits added to the AHB port.

PR #2014, opened 2019-06-25
APB userbits support

asmitde:
This PR adds a `pauser` optional field in APB bundles to propagate `UserBits` metadata from TileLink to APB dongles.
- Added `pauser` in `APBBundle`, and relevant support in APB port parameters
- Added support for userbits in `TLToAPB` adapter
- Added support for userbits in `TLFragmenter` node
- Verified propagation of userbits in waveform

**Related issue**:  #1931, #1939 

**Type of change**: feature add

**Impact**: no functional change | API addition (no impact on existing code)

**Release Notes**

Added `UserBits` support in APB and `TLFragmenter`.

### DIFF
--- a/src/main/scala/tilelink/Fragmenter.scala
+++ b/src/main/scala/tilelink/Fragmenter.scala
@@ -58,7 +58,15 @@ class TLFragmenter(val minSize: Int, val maxSize: Int, val alwaysMin: Boolean = 
       name        = "TLFragmenter",
       sourceId    = IdRange(0, if (minSize == maxSize) c.endSourceId else (c.endSourceId << addedBits)),
       requestFifo = true,
-      userBits    = c.clients.map( _.userBits).maxBy(_.length))))},
+      userBits    = {
+        require( c.clients.forall( _.userBits.length == c.clients(0).userBits.length ),
+          s"Length of userBits sequences of all clients must be equal.")
+        require( c.clients.forall( _.userBits.zip( c.clients(0).userBits ).forall { case (a, b) => a.width == b.width } ),
+          s"Width of corresponding userBits for all clients must match.")
+
+        c.clients(0).userBits
+      })))
+    },
     managerFn = { m => m.copy(managers = m.managers.map(mapManager)) })
 
   lazy val module = new LazyModuleImp(this) {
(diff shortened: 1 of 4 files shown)


src/main/scala/tilelink/Fragmenter.scala
```diff
@@ -57,7 +57,8 @@ class TLFragmenter(val minSize: Int, val maxSize: Int, val alwaysMin: Boolean =
     clientFn  = { c => c.copy(clients = Seq(TLClientParameters(
       name        = "TLFragmenter",
       sourceId    = IdRange(0, if (minSize == maxSize) c.endSourceId else (c.endSourceId << addedBits)),
-      requestFifo = true))) },
+      requestFifo = true,
+      userBits    = c.clients.map( _.userBits).maxBy(_.length))))},
```
hcook: Hmm, making the decision based on `length` doesn't seem necessarily correct, because you are looking at the number of different types of user bits on the link, but not the size of each type. Poking around I found:
https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/tilelink/Parameters.scala#L260
... which I think would be convenient for use with`maxBy`, but overall this raises questions for me about correct behavior when the clients dont match.

@terpstra Should mismatches between `TLClient`s' `userBits` here fail a requirement?

asmitde: `userBits` is of type `Seq[UserBits]`. So the map returns a `Seq[Seq[UserBits]]`. Shouldn't we be selecting the `Seq[UserBits]` of the maximum length from this mapped Seq?

hcook: As we discussed, the maximum bit length and maximum sequence length are not synonymous.

I talked with @terpstra offline and we agreed that for now the safe thing to do is have a requirement checking that all clients have identical userBits Seqs, and to fail if they do not match. (This strictness will require us to do more work for the target design to add user bit generation to more masters, but so be it.)

asmitde: By identical, I'm guessing you mean the bit-widths should also match and in order?

terpstra: Yes.

Also, I just spoke with Kenneth. I think that the proposal @hcook outlines will play nicely with another block we need anyway: a UserBits extractor adaptor. The idea here is that right before the TLToAPB/AHB you put down a TLUserExtract[T](x) that selects exactly one kind of user bits field and supplies a default value for any masters that had none. This has to be done on the TileLink side of the adapter, because AHB/APB have no way to identify the source of a request.

PR #2016, opened 2019-06-26
BankBinder should skip addresses with no intersecting address ranges

zhemao:
The way BankBinder is currently written, it won't work if you connect it to managers which themselves have the blocks striped across them. You would get a requirement failure because the addresses of some of the managers would be empty, as the bank and channel would be for a mutually exclusive set of blocks. This commit makes the bank skip over managers with which it has no address range in common.

**Related issue**: 

**Release Notes**


colinschmidt:
I've needed something like this before.
I'll try to see if this solves my problems but wont be able to check for a week or so.

zhemao:
Oh oops, I didn't notice I hadn't changed that. Good catch.

zhemao:
@terpstra I don't have write access. Can you perform the merge?

PR #2027, opened 2019-07-08
Add CDISCARD.D.L1 instruction

aswaterman:
Here's the spec for CFLUSH.D.L1 and CDISCARD.D.L1:

CFLUSH.D.L1
-----------------------------------

- opcode 0xFC000073, with optional rs1 field in bits 19:15.
- Only available to M-mode.
- When rs1 = x0, CFLUSH.D.L1 writes back all dirty lines and invalidates
  all lines in the L1 D$.
- When rs1 ≠ x0, CFLUSH.D.L1 writes back the line if dirty and invalidates
  the L1 D$ line containing the virtual address in integer register rs1.
  - If the effective privilege mode does not have write permissions to the
    address in rs1, then a store access or store page-fault exception is
    raised.
  - If the address in rs1 is in an uncacheable region with write
    permissions, the instruction has no effect but raises no exceptions.
  - Note that if the PMP scheme write-protects only part of a cache line,
    then using a value for rs1 in the write-protected region will cause
    an exception, whereas using a value for rs1 in the write-permitted
    region will write back the entire cache line.

CDISCARD.D.L1
---------------------------------------

- opcode 0xFC200073, with optional rs1 field in bits 19:15.
- Only available to M-mode.
- When rs1 = x0, CDISCARD.D.L1 invalidates, but does not write back,
  all lines in the L1 D$.  Dirty data within the cache is lost.
- When rs1 ≠ x0, CDISCARD.D.L1 invalidates, but does not write back,
  the L1 D$ line containing the virtual address in integer register rs1.
  Dirty data within the cache line is lost.
  - If the effective privilege mode does not have write permissions to the
    address in rs1, then a store access or store page-fault exception is
    raised.
  - If the address in rs1 is in an uncacheable region with write
    permissions, the instruction has no effect but raises no exceptions.
  - Note that if the PMP scheme write-protects only part of a cache line,
    then using a value for rs1 in the write-protected region will cause
    an exception, whereas using a value for rs1 in the write-permitted
    region will invalidate and discard the entire cache line.


aswaterman:
Motivation is to support more incoherent DMA use cases  ( :( )

aswaterman:
I agree, but I think that’s a fairly significant project so I’d like to handle it in a separate PR.

Issue #2022, opened 2019-07-01
DRET legal outside of debug mode


tzwaenn:
**Type of issue**: bug report 

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
According to the debug specification, section 4.6 "Executing dret outside of Debug Mode causes an illegal instruction exception.". However, RocketChip seems to have the illegal exception tied to M mode - so M mode without debug mode does retire the DRET instruction istead of throwing an illegal exception.
Can be observed by having the first fetch after reset return the opcode 0x7b200073 of DRET.

**Please tell us about your environment:**
This issue was found with formal verifcation tool OneSpin 360 DV

ernie-sifive:
Fixed, thanks for reporting this.

PR #2029, opened 2019-07-11
dret should be illegal outside debug mode

ernie-sifive:
**Related issue**: Issue #2022 

**Release Notes**

The dret instruction should throw an illegal instruction exception when executed outside debug mode.

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -535,6 +535,7 @@ class CSRFile(
     io_dec.system_illegal := reg_mstatus.prv < io_dec.csr(9,8) ||
       is_wfi && !allow_wfi ||
       is_ret && !allow_sret ||
+      is_ret && io.rw.addr(10) && !reg_debug ||
       is_sfence && !allow_sfence_vma
   }
 


ernie-sifive:
@aswaterman @hcook Travis test 8 seems to be timing out with this change.  Any suggestions?

aswaterman:
I don’t know what the travis tests do. Did you try restarting the travis build to see if it was a spurious failure? cc @richardxia - who’s the right person to ask?

richardxia:
I don't know much about the Travis CI tests for rocket-chip, and I'm pretty sure it was the UCB crowd that set them up. @hcook, do you know about them? In the meantime, I just hit the restart button on that job, so we can see if it was spurious.

aswaterman:
I’ll help debug tomorrow if it fails after a retry.

ernie-sifive:
@aswaterman I looked a bit at the tests and some of them run code on both the simulated hardware and on spike and compare the results.  I took a look at spike and am wondering if it has the same issue with dret.  `riscv-isa-sim/riscv/insns/dret.h` seems to only require PRV_M.  It is not obvious how to check for debug mode.  Maybe add something like this?
```
if (state.pc >= 0x1000) throw trap_illegal_instruction(0);
```
dret.h:
```
require_privilege(PRV_M);
set_pc_and_serialize(STATE.dpc);
p->set_privilege(STATE.dcsr.prv);

/* We're not in Debug Mode anymore. */
STATE.dcsr.cause = 0;

if (STATE.dcsr.step)
  STATE.single_step = STATE.STEP_STEPPING;
```

aswaterman:
Well, I retried the failing test, and it passed.  I'm going to write it off as a spurious infrastructural issue.  I'll deal with the Spike issue separately.

hcook:
@aswaterman All PRs are now failing on this test

aswaterman:
@hcook any idea why I was allowed to merge it?

ernie-sifive:
Should revert until understood. Could be a random test so it sometimes passes and sometimes fails.

aswaterman:
@ernie-sifive I'm working on replicating it locally.  If I fail to do so or can't find a quick solution, I'll revert, but I'll try to avoid doing so.

aswaterman:
It was a four-character bug I probably should've caught in code review.  D'oh.  Fixed in #2036.

PR #2030, opened 2019-07-11
TileLink: UserUniformer

hcook:
`TLUserUniformer` adapter allows a default value of a particular type of TileLink user bits to be applied to all clients that were not supplying values for those types themselves. If a client does supply a value, this is passed through the adapter. As a result, all clients appear to supply the same number of the specified type of user bits. For clients that supply their own user bits, those bits will appear at the highest indexes, and any default values will appear at lower indexes.

The adapter currently requires the widths of all UserBits of the specified type to be the same, because I didn't want to have to handle all the cases where different clients have different subsets of the same type of user bits. This restriction could probably be relaxed with sufficient effort.

Other miscellaneous helper functions also added.

re: d9035b8, https://stackoverflow.com/a/15535342/2483329

### DIFF
--- a/src/main/scala/tilelink/Fragmenter.scala
+++ b/src/main/scala/tilelink/Fragmenter.scala
@@ -318,7 +318,7 @@ object TLFragmenter
   def apply(minSize: Int, maxSize: Int, alwaysMin: Boolean = false, earlyAck: EarlyAck.T = EarlyAck.None, holdFirstDeny: Boolean = false)(implicit p: Parameters): TLNode =
   {
     val fragmenter = LazyModule(new TLFragmenter(minSize, maxSize, alwaysMin, earlyAck, holdFirstDeny))
-    fragmenter.node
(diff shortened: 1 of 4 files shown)


src/main/scala/tilelink/Fragmenter.scala
```diff
@@ -60,9 +60,9 @@ class TLFragmenter(val minSize: Int, val maxSize: Int, val alwaysMin: Boolean =
       requestFifo = true,
       userBits    = {
         require( c.clients.forall( _.userBits.length == c.clients(0).userBits.length ),
-          s"Length of userBits sequences of all clients must be equal.")
+          s"Length of userBits sequences of all clients must be equal. ${c.clients.map(x => (x.name, x.userBits.length))}")
         require( c.clients.forall( _.userBits.zip( c.clients(0).userBits ).forall { case (a, b) => a.width == b.width } ),
-          s"Width of corresponding userBits for all clients must match.")
+          s"Width of corresponding userBits for all clients must match. ${c.clients.map(x => (x.name, x.userBits))}")
```
aswaterman: This is most excellent

PR #2031, opened 2019-07-11
Debug: interface and attachment parameterization

hcook:
API changes to more consistently describe how the debug unit interfaces with external ports and buses:
- `DebugExportProtocol` exhaustively enumerates port types (`DMI|JTAG|CJTAG|APB`) rather than being a collection of one-hot boolean fields
- `BaseSubsystemBusAttachment` exhaustively enumerates tilelink bus attachment locations (`SBUS|FBUS|CBUS|MBUS|PBUS`); this will be used by other attachment methods eventually
- Debug unit can be attached to different buses based on `ExportDebug` `Field` values

(might  rebase this one #2030 goes in)

### DIFF
--- a/src/main/scala/devices/debug/Periphery.scala
+++ b/src/main/scala/devices/debug/Periphery.scala
@@ -16,24 +16,38 @@ import freechips.rocketchip.jtag._
 import freechips.rocketchip.util._
 import freechips.rocketchip.tilelink._
 
+/** Protocols used for communicating with external debugging tools */
+sealed trait DebugExportProtocol
+case object DMI extends DebugExportProtocol
+case object JTAG extends DebugExportProtocol
+case object CJTAG extends DebugExportProtocol
+case object APB extends DebugExportProtocol
+
 /** Options for possible debug interfaces */
-case object ExportDebugDMI extends Field[Boolean](true)
-case object ExportDebugJTAG extends Field[Boolean](false)
-case object ExportDebugCJTAG extends Field[Boolean](false)
-case object ExportDebugAPB extends Field[Boolean](false)
-case object ExportDisableDebug extends Field[Boolean](false)
+case class DebugInterfaceParams(
+  protocol: DebugExportProtocol = DMI,
+  externalDisable: Boolean = false,
+  attachAsTile: Boolean =  false
+) {
+  def dmi   = protocol == DMI
+  def jtag  = protocol == JTAG
+  def cjtag = protocol == CJTAG
+  def apb   = protocol == APB
+}
+
+case object ExportDebug extends Field(DebugInterfaceParams())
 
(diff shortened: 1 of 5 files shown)


src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -36,6 +36,14 @@ abstract class BareSubsystemModuleImp[+L <: BareSubsystem](_outer: L) extends La
   println(outer.dts)
 }
 
+
+sealed trait BaseSubsystemBusAttachment
+case object SBUS extends BaseSubsystemBusAttachment
+case object PBUS extends BaseSubsystemBusAttachment
+case object FBUS extends BaseSubsystemBusAttachment
+case object MBUS extends BaseSubsystemBusAttachment
+case object CBUS extends BaseSubsystemBusAttachment
```
hcook: @rmac-sifive Imagine these as `case class CBUS(idx: Int)` ...

rmac-sifive: So then somewhere in my config `Key`s for the IP I set attachment point  `= CBUS(0)` or 
`= CBUS(1)` depending on which subsystem I want the IP to attach to? Makes sense, and then we do something like:

src/main/scala/devices/debug/Periphery.scala
```diff
@@ -16,24 +16,39 @@ import freechips.rocketchip.jtag._
...
-case object ExportDebugCJTAG extends Field[Boolean](false)
-case object ExportDebugAPB extends Field[Boolean](false)
-case object ExportDisableDebug extends Field[Boolean](false)
+case class DebugAttachParams(
+  protocol: DebugExportProtocol = DMI,
+  externalDisable: Boolean = false,
+  masterWhere: BaseSubsystemBusAttachment = FBUS,
+  slaveWhere: BaseSubsystemBusAttachment = CBUS
+) {
+  def dmi   = protocol == DMI
```
mwachs5: in the future we may want to allow more than one protocol. For example JTAG AND APB. Does this refactor consider this?

mwachs5: Could we make the protocol argument to DebugAttachParams be a `Seq[DebugExportProtocol]`?

ernie-sifive: I just had an offline discussion with @hcook about this issue and this is exactly what he proposed.

src/main/scala/subsystem/BaseSubsystem.scala
hcook: Yes that is lines i was thinking along!

src/main/scala/devices/debug/Periphery.scala
hcook: I made them a `Set`

PR #2034, opened 2019-07-12
Refactor BaseSubsystem into per-trait buses

seldridge:
I have a situation where I'd like to use a somewhat different system construction that deviates from `BaseSubsystem`.  Namely, I'd like to move buses around and have the flexibility of placing certain buses inside or outside of the system. I can always define my own subsystem and go from there, but I'd like to be able to use all the existing traits defined in the subsystem package. The current construction makes this difficult as most traits are self-typed on `BaseSubsystem` when they may only use a subset of the available buses.

This refactors `BaseSubsystem` into separate traits, e.g., `trait SBus` contains a system bus and `trait MBus` contains the memory bus. Traits which previously self-typed on `BaseSubsystem` are then modified to self-type more narrowly on what buses they require.

**Release Notes**

- Breakup BaseSubsystem into separate per-bus traits

### DIFF
--- a/src/main/scala/devices/debug/Periphery.scala
+++ b/src/main/scala/devices/debug/Periphery.scala
@@ -10,6 +10,7 @@ import freechips.rocketchip.subsystem._
 import freechips.rocketchip.devices.tilelink._
 import freechips.rocketchip.amba.apb._
 import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.diplomaticobjectmodel.HasLogicalTreeNode
 import freechips.rocketchip.diplomaticobjectmodel.logicaltree.LogicalModuleTree
 import freechips.rocketchip.diplomaticobjectmodel.model.OMComponent
 import freechips.rocketchip.jtag._
@@ -41,7 +42,7 @@ class DebugIO(implicit val p: Parameters) extends ParameterizedBundle()(p) with 
   * based on a global parameter.
   */
 
-trait HasPeripheryDebug { this: BaseSubsystem =>
+trait HasPeripheryDebug { this: BareSubsystem with CBus with FBus with HasLogicalTreeNode =>
   val debug = LazyModule(new TLDebugModule(cbus.beatBytes))
 
   LogicalModuleTree.add(logicalTreeNode, debug.logicalTreeNode)
--- a/src/main/scala/devices/tilelink/BootROM.scala
+++ b/src/main/scala/devices/tilelink/BootROM.scala
@@ -4,7 +4,7 @@ package freechips.rocketchip.devices.tilelink
 
 import Chisel._
 import freechips.rocketchip.config.{Field, Parameters}
-import freechips.rocketchip.subsystem.{BaseSubsystem, HasResetVectorWire}
+import freechips.rocketchip.subsystem.{BareSubsystem, CBus, HasResetVectorWire}
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.util._
@@ -59,7 +59,7 @@ class TLROM(val base: BigInt, val size: Int, contentsDelayed: => Seq[Byte], exec
 }
 
 /** Adds a boot ROM that contains the DTB describing the system's subsystem. */
-trait HasPeripheryBootROM { this: BaseSubsystem =>
+trait HasPeripheryBootROM { this: BareSubsystem with CBus =>
   val dtb: DTB
   private val params = p(BootROMParams)
   private lazy val contents = {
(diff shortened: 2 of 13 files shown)


aswaterman:
@seldridge A regression slipped into master today and likely caused this Travis failure.  It's fixed now; try rebasing.

seldridge:
Rebased. Thanks for the quick regression fix.

hcook:
I'm not fundamentally opposed to this direction, but it is running counter trend to some functionality that I need ASAP right now (and hence have merged in a conflict, apologies). The question of how devices selectively attach to buses and whether the buses available in a particular subsystem should be reflected in the Subsystem class hierarchy is an area of active development and your input is definitely welcome.

You're correct to note that in any particular design, an individual peripheral needs to be attached to only a subset of buses. However, our recent experience has been that to which buses a peripheral should attach to is really a matter of configuration, and for most peripherals there are several suitable bus attachment points that might be used for the master or slave interfaces in different designs. Making this choice of attachment point a matter of Configuration rather than a matter of compile-time trait mix-in choice is highly desirable, and https://github.com/chipsalliance/rocket-chip/pull/2031 represents a first cut at a such a pattern (see commits related to `BaseSubsystemBusAttachment`). In contrast, I think your approach will ultimately result in a plethora of traits per peripheral (one per bus attachment point strategy), of which a particular one must be chosen to be mixed into a concrete subsystem child class (resulting in more child classes).

I would also like to point out that most of the buses in `BaseSubsystem` can be connected to each other in such a way that if no devices are connected to them, the crossbars will remove themselves. (And the functions called in `HasHierarchicalBusTopology` do provide this behavior). Thus, IMO there is no downside to depending on a BaseSubsystem that contains buses to which your particular design attaches nothing. For example, we have designs in which we define the `pbus` to exist outside the subsystem (i.e. in the parent system): in those designs no devices attach to  `subsystem.pbus` and it removes itself. So I think a maximalist approach in which `BaseSubsystem` contains as many buses as are commonly useful should not prevent you from being able to inherit from it in your own subsystem child class, and then selectively connect peripherals to additional child class buses while ignoring some buses from the parent class.

I also think some kind of mixed solution is possible, where BaseSubsystem is to some degree decomposed. For example, peripheral traits could be changed to depend only on having an abstract `attach` function defined in `BareSubsystem`.  And I don't object to adding traits that add each buses per se. For backwards compatibility reasons on our end (there are a lot of peripheral traits...), I would ask that the actual class named `BaseSubsystem` continue to have all the currently defined buses. We can add a new intermediate class hierarchy layer with only an `sbus` mixed in, if my statements from the preceding paragraph don't address your concerns about inheriting from a `BaseSubsystem` with all buses pre-defined.

Happy to chat about this more, let me know your thoughts.

seldridge:
Thanks for the detailed response @hcook. 

It didn't occur to me that unconnected buses will selectively delete themselves... Your example with `pbus` is clear to me and is basically the same thing that I'm doing.

I'm trying to move the `mbus` and associated collateral outside with the end result being a top-level TL-C port directly connected to `sbus`. Let me try the approach you outline and see if that works. My confusion was that I would need to remove the internal `mbus` for this to work properly.

And, yes, the backwards compatibility problem of redefining `BaseSubsystem` would be bad. For this PR to be approved, it would likely require a separate hierarchy.

For the time being I'll leave this open while I refactor my own stuff for the duplicate `mbus`, maximalist approach. Pending that working, I'll close this. I have [other problems with my branch working for TL-UH top, but not TL-C top](https://stackoverflow.com/questions/56549595/how-to-choose-tlmanager-parameters-widgets-for-exposed-systembus), but that's likely better geared for a SiFive forums question.

hcook:
Yeah, I will also point out that the `mbus` is actually implicitly assumed to be TL-UH; i.e. that the last-level cache is inside the subsystem and either has a CacheCork or issues Gets and Puts natively to the (presumed) DDR controller behind/outside it. I don't know of anything concrete in the mbus functionality that will break if you put more caches outside, and I know of at least one project that has done so one way or another, but just wanted to mention that it is not a well-trod path at the moment. Some implementations of the `BankedL2Key` `Field` will definitely cry about it.

I think it is an open question whether coherent managers outside the `mbus` or coherent clients inside the `fbus` should use those same attachment points too, or whether we should add some more `Xbus` names specifically for collecting those types of attachment, which might be connect in materially different ways than the non-caching variety.

jerryhethatday:
> For example, we have designs in which we define the `pbus` to exist outside the subsystem (i.e. in the parent system):

hey, Henry. Could you please clarify the termnology here: the system and the subsystem. Seems a little confusing.  In the codebase, I thought the `ExampleRocketSystem` is a system according to its name implies. The `ExampleRocketSystem` is a fully-flegded SOC, right? But what does the subsystem mean here in the context of ExampleRocketSystem? Why you have a design to `define the `pbus` to exist outside the subsystem (i.e. in the parent system)` Thanks!
@hcook

sequencer:
I think this PR has been neatly resolved by Location API.

PR #2035, opened 2019-07-12
keep valid signal low while reset is high

jerryrzhao:
**Related issue**: 

**Release Notes**

Per the tilelink specification ver. 1.7.1 section 3.2.2, valid signals should be driven low while reset is asserted. In this case, TLFuzzer is the master and should drive a.valid low during a reset.


aswaterman:
@jerryrzhao A regression slipped into master today.  I've rebased your branch on the fixed master and force-pushed it.

PR #2038, opened 2019-07-14
Debug: resethaltreq and hartreset functions

ernie-sifive:

**Release Notes**

This PR adds `resethaltreq` and `hartreset` functions to the Debug Module.  These functions are described in detail in the RISCV Debug Specification and this implementation is compliant with that.

A new DebugModuleParams configuration setting, `hasCoreResets`, is added to allow the DM to accept hart reset signals individually rather than just a single global reset.  This was added as an Option so that the DM remains backward-compatible if not selected.  When `hasCoreResets=true`, a `coreReset` input and a `coreResetReq` output are added for each hart in the system.  These are not extended into Periphery since they are intended for connections internal to the core complex.

`resethaltreq` is changed to be always available (indicated by `hasresethaltreq=1` in `dmstatus`).  If `hasCoreResets=true`, `CoreReset` inputs are used for this function, otherwise the global `reset` is used.  Logic in the DMI clock domain maintains a bitmask of which harts have `resethaltreq` enabled.  The mask is passed through the clock crossing from dmouter to dminner.  In dminner, reset inputs are monitored and when asserted, activate the corresponding `hrDebugInt` register(s) which cause a debug interrupt to the corresponding hart(s). Once reset ends, the debug interrupt is taken and the hart halts.

`hartreset` is implemented in dmouter.  Writing 1 to `dmcontrol.hartreset` asserts `coreResetReq` for all selected harts.  Writing 0 deasserts `coreResetReq`.  It is expected that these DM outputs be combined into the reset inputs of each hart.  The `ndmreset` function is still available as a system-wide reset request.


src/main/scala/devices/debug/Debug.scala
```diff
@@ -419,6 +428,32 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
       }
     }
 
+    //--------------------------------------------------------------
+    // Halt-on-reset
+    //  hrmaskReg is current set of harts that should halt-on-reset
+    //    Reset state is all zeroes
```
mwachs5: Reset state here is controlled by dmactive? or by [debug]_reset?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -419,6 +428,32 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
+    //    Bits are set by writing 1 to DMCONTROL.setresethaltreq
+    //    Bits are cleared by writing 1 to DMCONTROL.clrresethaltreq
+    //    Spec says if both are 1, then clrresethaltreq is executed
+    //--------------------------------------------------------------
```
mwachs5: I don't recall the spec for this, but basically this bit remains set even if the core DOES halt on reset, right?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -476,14 +511,19 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
...
       io.innerCtrl.bits.hasel        := Mux(DMCONTROLWrEn, DMCONTROLWrData.hasel, DMCONTROLReg.hasel)
       io.innerCtrl.bits.hamask       := hamask
     }
 
     io.ctrl.ndreset := DMCONTROLReg.ndmreset
     io.ctrl.dmactive := DMCONTROLReg.dmactive
-
+    io.coreResetReq.foreach { req =>
+      for (component <- 0 until nComponents) {
+        req(component) := DMCONTROLReg.hartreset & hartSelected(component)
```
mwachs5: I was somewhat confused by the use of `coreReset` when the spec calls this "hart" reset, and we either call them "harts" or "components" in this code. Is there any reason to use coreReset here?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -527,13 +567,15 @@ class TLDebugModuleOuterAsync(device: Device)(implicit p: Parameters) extends La
       val ctrl = new DebugCtrlBundle(nComponents)
       val innerCtrl = new AsyncBundle(new DebugInternalBundle(nComponents), AsyncQueueParams.singleton())
       val hgDebugInt = Vec(nComponents, Bool()).asInput
+      val coreResetReq = p(DebugModuleParams).hasCoreResets.option(Output(Vec(nComponents, Bool())))
```
mwachs5: why do we call this core reset instead of hart reset?

PR #2040, opened 2019-07-16
Fix TraceGen issues and allow it to use NonBlockingDCache

zhemao:
The TraceGenConfig in groundtest is pretty outdated and currently doesn't work. These changes get it working again and also allow it to check the NonBlockingDCache. 

**Related issue**: 

**Release Notes**

Fix TraceGen groundtest and allow it to use the NonBlockingDCache

### DIFF
--- a/src/main/scala/groundtest/Configs.scala
+++ b/src/main/scala/groundtest/Configs.scala
@@ -6,12 +6,13 @@ package freechips.rocketchip.groundtest
 import Chisel._
 import freechips.rocketchip.config.Config
 import freechips.rocketchip.subsystem._
+import freechips.rocketchip.system.BaseConfig
 import freechips.rocketchip.rocket.{DCacheParams}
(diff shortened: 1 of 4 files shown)


zhemao:
We did find a bug in the NonBlockingDataCache using tracegen. If the cache line reserved by an LR gets invalidated before the SC, the SC will still succeed after miss and replay. You can see this in the trace below. The second access from tile 1 is the LR/SC pair.

```
1: M[7] := 1401 @ 18170:
0: { M[7] == 1401; M[7] := 1360} @ 18228:18230
0: M[7] := 1380 @ 18355:
1: { M[7] == 1360; M[7] := 1431} @ 18383:18391
0: { M[7] == 1431; M[7] := 1396} @ 18581:
```

mn416:
@hcook @zhemao Very happy to hear this!  Thanks for letting me know.

PR #2048, opened 2019-07-22
Expand regfields on DMI interface for improved DV coverage

ernie-sifive:

**Release Notes**

The register map for the DMI interface did not describe register fields with sufficient detail to help automate verification testing.  This PR expands the regmaps to include this detail.


mwachs5:
The diff is too big for me to say whether it is bug-free, but thanks for taking the time to do this!

Issue #2024, opened 2019-07-04
Racing between ProebUnit and the MSHR


armiasalib:
I have a system that consists of multi-cores Rocket-chip, and each core has a non-blocking D$. I am considering the folloing version https://github.com/chipsalliance/rocket-chip/blob/b8baef6f26fa36c9b4e0b2a5eec12cd76ae5daf8/src/main/scala/rocket/NBDcache.scala

If the probeUnit received a probe signal on this cache line, it will NACK any new incomming memop that try to access this cacheline via s1_nack. If any memop was received previously and forwarded to the MSHR, the MSHR will stop the probeUnit from reading the tags until the tags has beed flushed, this is done via probe_rdy signal.

However, there are two conditions where the MSHR will update the tags and the ProbeUnit is allowed to read the not updated cacheline state. This happens when:
1- To set dirty bit, Line 231 in the above version.
2- upgrade permissions in Line 234. 
In these cases, te MSHR state machine jumps to s_meta_write_req and s_refill_req respectively, and the probe_rdy will continue 1, allowing the probeUnit to read non-updated state. That causes conflicts to what the MSHR is doing.

The first scenario can be easily protected by disabling the probe until the MSHR fully updated the tags; however, we cannot use this solution directly for the second since that may cause a deadlock when two cores need to update the permission for the same cacheline.

What is your opinion? Do I miss something?

Issue #2019, opened 2019-06-27
issue about lrsc_count in NBDcache


daxxyy123:
Hi，I have a question about lrsc_count in NBDcache。
The lrsc_count is used to see if there are over 16 insts between lr and sc.

// load-reserved/store-conditional 
  val lrsc_count = Reg(init=UInt(0))
  val lrsc_valid = lrsc_count > lrscBackoff //
  val lrsc_addr = Reg(UInt())
  val (s2_lr, s2_sc) = (s2_req.cmd === M_XLR, s2_req.cmd === M_XSC)
  val s2_lrsc_addr_match = lrsc_valid && lrsc_addr === (s2_req.addr >> blockOffBits)
  val s2_sc_fail = s2_sc && !s2_lrsc_addr_match
  when (lrsc_count > 0) { lrsc_count := lrsc_count - 1 }
  when (s2_valid_masked && s2_hit || s2_replay) {
    when (s2_lr) {
      lrsc_count := lrscCycles - 1 
      lrsc_addr := s2_req.addr >> blockOffBits
    }
    when (lrsc_count > 0) { 
      lrsc_count := 0
    }//why it is "lrsc_count > 0"?I think it should be "lrsc_count < 0",which mean there is too many insts between lr and sc;because there is"when (lrsc_count > 0) { lrsc_count := lrsc_count - 1 }",lrsc_count will be decrease everyclock.              
  }

Issue #495, opened 2016-12-16
wrong corrected data used in L1 NBCache's correction replay?


wsong83:
in src/main/scala/rocket/NBDcache.scala
around line 844, in the L1 NB Cache

s3_req.data := Mux(s2_data_correctable, s2_data_corrected, amoalu.io.out)

When s2_data_correctable, should it be
  s2_data_corrected >>  Cat(s2_word_idx, Bits(0,log2Up(coreDataBits)))
rather than
  s2_data_corrected
?

Since there is no real parity check in L1, this branch is never checked.
What I can see is s2_data_corrected is a flattened vector and the requested word does not always locate at index 0 unless doNarrowRead.

aswaterman:
Not surprising; the ECC support in the L1 D$ is incomplete and so might have bugs.  @terpstra can you make sure this is the right fix?

Thanks for the bug report, @wsong83!

terpstra:
I think this would be better dealt with by @hcook, who last updated the code for the TL2.

sequencer:
close since #2053 merged.

PR #2053, opened 2019-07-30
NBDCache bugfixes.

jerryz123:
**Related issue**: 

**Impact**: rtl change

**Release Notes**

Fix two bugs in the NBDCache.

- SC invalidates the reservation when it sees the reserved line as `Trunk`. This causes the replaying SC to fail always. Fix by not invalidating reservation when SC only needs to enter MSHR to set dirty bit.
- TileLink allows the ReleaseAck to arrive before the ReleaseData burst has completed. Existing MSHR assumes ReleaseAck arrives after the ReleaseData burst has finished. Fix by forcing MSHRs to wait for ReleaseAck to arrive AND for ReleaseData to finish.

Does anything still use the NBDCache? BOOM will no longer use this.

### DIFF
--- a/src/main/scala/rocket/NBDcache.scala
+++ b/src/main/scala/rocket/NBDcache.scala
@@ -187,6 +187,9 @@ class MSHR(id: Int)(implicit edge: TLEdgeOut, p: Parameters) extends L1HellaCach
   rpq.io.enq.bits := io.req_bits
   rpq.io.deq.ready := (io.replay.ready && state === s_drain_rpq) || state === s_invalid
 
+  val acked = Reg(Bool())
+  when (io.mem_grant.valid) { acked := true }
+
   when (state === s_drain_rpq && !rpq.io.deq.valid) {
     state := s_invalid
   }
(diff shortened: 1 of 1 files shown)


hcook:
SiFive does not use or test the `NBDCache`, so it is a bit neglected. I do know of other open-source contributors that are using it based on open issues:
#2019 
#2024 
#495

Also, in particular, one of your patches is to a line of code that was itself introduced by a recent bugfix:
#2040 
@zhemao Any thoughts?

jerryz123:
@zhemao 's fix was bit too aggressive in clearing reservations. LR/SC is hard :/

hcook:
Yeah totally agree, no blame assigned, just wondering whether this is a hole in the TraceGen checker that caught the issue initially, but then also approved of his bugfix while not catching this other case. Fundamental, or just not enough random fuzzing?

zhemao:
@hcook I think TraceGen doesn't report anything if an LR/SC unexpectedly fails. It just treats it as if nothing happens.

PR #2057, opened 2019-08-01
add wit dependency/git submodule check to travis

albertchen-sifive:
**Related issue**: 

**Release Notes**

add a CI script to check that git submodules match wit dependencies

### DIFF
--- a/regression/check_submodules.sh
+++ b/regression/check_submodules.sh
@@ -1,9 +1,9 @@
 #!/bin/sh
 
-filter=".[]"
-filter+="|select(.name == \\\"\$path\\\")"
-filter+="|select(.name == \\\"\$name\\\")"
-filter+="|select(.commit == \\\"\$sha1\\\")"
+filter='.[]'
+filter+='|select(.name == \"$path\")'
+filter+='|select(.name == \"$name\")'
+filter+='|select(.commit == \"$sha1\")'
 filter="\"$filter\""
 
 git submodule foreach --quiet "jq -e $filter ../wit-manifest.json" || { \
--- a/regression/check_submodules.sh
+++ b/regression/check_submodules.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+filter=".[]"
+filter+="|select(.name == \\\"\$path\\\")"
+filter+="|select(.name == \\\"\$name\\\")"
+filter+="|select(.commit == \\\"\$sha1\\\")"
+filter="\"$filter\""
+
+git submodule foreach --quiet "jq -e $filter ../wit-manifest.json" || { \
+  echo "git submodules do not match wit dependencies!" && exit 1; \
+}
--- a/regression/check_submodules.sh
+++ b/regression/check_submodules.sh
@@ -1,11 +1,7 @@
 #!/bin/sh
 
-filter='.[]'
-filter+='|select(.name == \"$path\")'
-filter+='|select(.name == \"$name\")'
-filter+='|select(.commit == \"$sha1\")'
-filter="\"$filter\""
+filter='.[]|select(.commit == "\1")|select(.name == "\2")'
 
-git submodule foreach --quiet "jq -e $filter ../wit-manifest.json" || { \
-  echo "git submodules do not match wit dependencies!" && exit 1; \
-}
+git submodule status \
+| sed -r "s/\-([a-zA-Z0-9]+)\s([a-zA-Z0-9\-]+)/$filter/g" \
+| xargs -d '\n' -I % sh -c "jq -e '%' wit-manifest.json"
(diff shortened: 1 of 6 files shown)


regression/check_submodules.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+filter='.[]'
+filter+='|select(.name == \"$path\")'
+filter+='|select(.name == \"$name\")'
```
richardxia: Why is it that we are checking that both the `$path` and the `$name` are equal to the wit submodule name? Git submodules may not always be installed at the root directory, so it's probably safer to check only the name?

Also, could you add a comment describing where these variables come from. I'm afraid that this is a pretty advanced usage of `git submodule foreach` and of `jq`, and to someone not experienced with the more advanced of both the tools, it's not exactly clear which of the two tools is setting the variables. (I eventually figured out that it's `git submodule foreach` that is defining them.)

regression/check_submodules.sh
```diff
@@ -0,0 +1,11 @@
...
+filter+='|select(.commit == \"$sha1\")'
+filter="\"$filter\""
+
+git submodule foreach --quiet "jq -e $filter ../wit-manifest.json" || { \
```
richardxia: Since it's not completely obvious what this line is doing, could you write a short comment describing what you are doing here? This file itself is just called `check_submodules.sh`, but it's not exactly clear what we are checking the submodules for. And even though I know you are checking the git submodules against the wit manifest, the algorithm you are using to check for correctness is not readily apparent without having to dig into exactly what the `git submodule foreach` command and the `jq` filter is doing.

My current understanding is that this script loops over each git submodule, and for each submodule it attempts to locate a record (JSON object) within `wit-manifest.json` which simultaneously matches the name, path, and commit hash of the git submodule.

Note that this will miss cases where the wit manifest specifies a repository that is not present as a git submodule. I am not sure if that is intentional.

regression/check_submodules.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
```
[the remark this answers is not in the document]
albertchen-sifive: yeah I agree, but would renaming this break other things? There are a bunch of existing makefiles have have the word regression in them.

regression/check_submodules.sh
albertchen-sifive: intentional. since the wit-manifest may have extra packages like `api-chisel3-sifive` that shouldn't be submodules

regression/check_submodules.sh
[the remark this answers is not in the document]
richardxia: Right, let me slightly modify my request, then. Since this specific check isn't referenced by the existing Makefile in this `regression` directory, how about we just move this one script into a new directory?

regression/check_submodules.sh
richardxia: OK, if it's intentional, then let's capture that intent in a comment, since it's not possible to infer that from just the code alone.

regression/check_submodules.sh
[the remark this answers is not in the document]
albertchen-sifive: ok

regression/check_submodules.sh
albertchen-sifive: I added `$path` as a way of enforcing that submodules are at the root, but if we don't want to check that then I'll remove this

albertchen-sifive: also I don't think this script actually works since `.travis.yml` has `submodules: false` set so this would always pass. should I just parse the `.gitmodules` file instead?

albertchen-sifive: or just do a `git submodule update` since `regression/Makefile` will do it anyway?

richardxia: There must be some other script that takes care of checking out the submodules, because otherwise the other Travis CI tests couldn't possibly be working. You could try to invoke that script before running this one? I imagine we don't tell Travis CI to clone all submodules because there was a time when rocket-chip depended on `riscv-tools`, which depended on `riscv-gnu-toolchain`, which depended on the real GCC repo, which would have pulled down a ton of unnecessary code and git history.

It looks like `git submodule status` will still work even if you haven't initialized your submodules. It's not as nice as having the commit hash and the submodule name pulled into variables, but it shouldn't be too difficult to parse. Here's what it looks like to me:

```sh
$ git submodule status 
-e1aa5f3f5c0cdeb204047c3ca50801d9f7ea25f1 chisel3
-228c9a4b7432ac52178d63b8f27fe064aec71e9c firrtl
-70c1e1dd0954bb8e9d62563b8c9730bbc39d2bf0 hardfloat
-77195ab12aefc373ca688e0a9c4d710c13191341 torture
```

richardxia: Also, the `.gitmodules` file doesn't contain the actual commit hashes, only the (initial) remote locations of the submodules.

PR #2062, opened 2019-08-06
add api-config-chipsalliance wit package/git submodule

albertchen-sifive:
**Related issue**: 

**Impact**: API modification

**Release Notes**

replace `src/main/scala/config` with a git submodule/wit package pointing at [api-config-chipsalliance](https://github.com/chipsalliance/api-config-chipsalliance) and update build rules


ucbjrl:
Unfortunately, this causes a:
```
[error] java.util.zip.ZipException: duplicate entry: config/Config.scala
```
with `sbt publishLocal`. The first file with an identical path (as far as java.util.zip.ZipOutputStream is concerned) is rocket-chip/src/main/scala/config/Config.scala.
Since these files claim to be in different packages, you might try including the package name in the file paths.

PR #2064, opened 2019-08-07
Separate ITIM DeviceTree into memory and control parts

aswaterman:
This helps introspective code know how much RAM actually lives in the ITIM.

**Impact**: API modification

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -56,7 +56,23 @@ class ICache(val icacheParams: ICacheParams, val hartId: Int)(implicit p: Parame
     name = s"Core ${hartId} ICache")))))
 
   val size = icacheParams.nSets * icacheParams.nWays * icacheParams.blockBytes
-  val device = new SimpleDevice("itim", Seq("sifive,itim0"))
+  val itim_control_offset = size - icacheParams.nSets * icacheParams.blockBytes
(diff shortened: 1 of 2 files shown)


src/main/scala/rocket/ICache.scala
```diff
@@ -56,7 +56,23 @@ class ICache(val icacheParams: ICacheParams, val hartId: Int)(implicit p: Parame
     name = s"Core ${hartId} ICache")))))
 
   val size = icacheParams.nSets * icacheParams.nWays * icacheParams.blockBytes
-  val device = new SimpleDevice("itim", Seq("sifive,itim0"))
+  val itim_control_offset = size - icacheParams.nSets * icacheParams.blockBytes
+
+  val device = new SimpleDevice("itim", Seq("sifive,itim0")) {
+    override def describe(resources: ResourceBindings): Description = {
+     val Description(name, mapping) = super.describe(resources)
+     val Seq(Binding(_, ResourceAddress(address, perms))) = resources("reg/mem")
```
terpstra: I realize you got this line from me, but doesn't it give a warning about "what if the match is refuted"?

aswaterman: I searched through the warnings and the only thing that showed up for ICache.scala is an LFSR16 deprecation notice.

PR #2066, opened 2019-08-07
omsrams

derekpappas:
**Related issue**: 

**Release Notes**

Adding OMSRAMS to caches

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
+++ b/src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
@@ -27,7 +27,7 @@ class DCacheLogicalTreeNode(memories: () => Seq[OMSRAM], deviceOpt: Option[Simpl
 }
 
 
-class ICacheLogicalTreeNode(deviceOpt: Option[SimpleDevice], icacheParams: ICacheParams, memories: () => Seq[OMSRAM]) extends LogicalTreeNode(() => deviceOpt) {
+class ICacheLogicalTreeNode(memories: () => Seq[OMSRAM], deviceOpt: Option[SimpleDevice], icacheParams: ICacheParams) extends LogicalTreeNode(() => deviceOpt) {
   def getOMICacheFromBindings(resourceBindings: ResourceBindings): OMICache = {
     getOMComponents(resourceBindings) match {
       case Seq() => throw new IllegalArgumentException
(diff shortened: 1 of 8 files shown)


src/main/scala/rocket/DCache.scala
```diff
@@ -81,6 +82,7 @@ class DCacheMetadataReq(implicit p: Parameters) extends L1HellaCacheBundle()(p)
 
 class DCache(hartid: Int, val crossing: ClockCrossingType)(implicit p: Parameters) extends HellaCache(hartid)(p) {
   override lazy val module = new DCacheModule(this)
+  override def getOMSRAMS(): Seq[OMSRAM] = Seq(module.dcacheImpl.omSRAM) ++ module.dcacheImpl.data.data_arrays.map(_._2)
```
aswaterman: `omSRAMs` instead of `omSRAMS`

src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
```diff
@@ -14,20 +14,20 @@ import freechips.rocketchip.tile.{RocketTileParams, TileParams, XLen}
  * The data memory subsystem is assumed to be a DTIM if and only if deviceOpt is
  * a Some(SimpleDevice), as a DCache would not create a Device.
  */
-class DCacheLogicalTreeNode(deviceOpt: Option[SimpleDevice], params: DCacheParams) extends LogicalTreeNode(() => deviceOpt) {
+class DCacheLogicalTreeNode(memories: () => Seq[OMSRAM], deviceOpt: Option[SimpleDevice], params: DCacheParams) extends LogicalTreeNode(() => deviceOpt) {
```
richardxia: For consistency, can you place this new `memories` parameter at the end of the parameters list? This would not only make it match the order of the `OMCaches.dcache()` function parameters, but it would also make it consistent with every other change you made to function signatures here.

src/main/scala/rocket/NBDcache.scala
```diff
@@ -669,7 +670,8 @@ class DataArray(implicit p: Parameters) extends L1HellaCacheModule()(p) {
 }
 
 class NonBlockingDCache(hartid: Int)(implicit p: Parameters) extends HellaCache(hartid)(p) {
-  override lazy val module = new NonBlockingDCacheModule(this) 
+  override lazy val module = new NonBlockingDCacheModule(this)
+  override def getOMSRAMS(): Seq[OMSRAM] = Nil
```
richardxia: Should this actually be `Nil`? Does a NonBlockingDCache really have no SRAMs within it?

aswaterman: It should not be `Nil`, of course, but in @derekpappas' defense I did suggest to him that handling the `NonBlockingDCache` in this manner was of low priority.

richardxia: Ah, I see. In that case, I'd request that we at least leave a comment here stating that this is just a dummy value and that we need to eventually fix it. Otherwise it's hard to tell the difference between "this is intentionally an empty list" from "this is being set to an empty list because we want to defer the work of setting it to the correct value".

aswaterman: Yep.

PR #2070, opened 2019-08-10
adding dummy rtl module om and the om resets

derekpappas:
**Related issue**: 

**Release Notes**

adding dummy rtl module om and the om resets
* added the OM class hierarchy for the rtl module information
* added the code to extract the reset names from the cores

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
@@ -93,7 +93,7 @@ case class InterruptSignal(
 case class OMRTLInterface(
   clocks: Seq[OMClock],
   clockRelationships: Seq[OMClockRelationship],
-  resets: Seq[OMRTLReset],
+  resets: () => Seq[OMRTLReset],
   statuses: Seq[OMStatus]
 ) extends RTLComponent
 
--- a/src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
@@ -2,6 +2,47 @@
 
 package freechips.rocketchip.diplomaticobjectmodel.model
 
+// Examples of ClockRelationships
+//  {
+//   "clock0": "core_clock_0",
+//    "clock1": "clock",
+//    "relationship": "$clock0 == m * $clock1 where m is a positive integer",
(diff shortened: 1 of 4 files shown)


src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
```diff
@@ -2,6 +2,47 @@
...
+//  {
+//   "clock0": "core_clock_0",
+//    "clock1": "clock",
+//    "relationship": "$clock0 == m * $clock1 where m is a positive integer",
+//  }
+
+//  {
+//    "clock1": "rtc_toggle",
+//    "relationship": "$clock0 >= 2 * $clock1",
+//  }
```
richardxia: Shouldn't this comment go next to the definition of `OMClockRelationship`? It's confusing to see this not next to the thing that it's actually related to.

src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
```diff
@@ -29,22 +70,34 @@ trait Synchronous extends Synchronicity
...
+  width: Int
+) extends OMSignal
 
 case class OMRTLInterface(
-  clocks: List[OMClock],
-  clockRelationships: List[OMClockRelationship],
-  resets: List[OMRTLReset]
+  clocks: Seq[OMClock],
+  clockRelationships: Seq[OMClockRelationship],
+  resets: () => Seq[OMRTLReset],
```
richardxia: Why is this a thunk that returns a `Seq[OMRTLReset]` rather than just a plain `Seq[OMRTLReset]`? All of the object model classes should just be plain data structures that don't involve function pointers.

src/main/scala/diplomaticobjectmodel/model/OMRTLModule.scala
```diff
@@ -2,6 +2,47 @@
...
+  name: String
+  description: Option[String]
+  def hartId: Int
+}
+
+case class Halt(
+  name: String,
+  description: Option[String],
+  hartId: Int
+) extends OMCoreStatus
```
richardxia: Don't you need to add in the `_types` property with the default value or else the serialized object model JSON file will lack the type information associated with each of these classes?

sequencer:
OM is purged now in #2967.

PR #2074, opened 2019-08-13
Bump chisel3 and firrtl - Async reset support

jackkoenig:
This is a big one, buckle up folks

#### Features
* Add Stage/Phase APIs
* Add "Aspect Oriented"-inspired API support
* Add native `AsyncReset` and abstract `Reset` support
* Add utility functions `unsignedBitLength` and `signedBitLength`
* Move `dontTouch`, `RawModule`, and `MultiIOModule` out of `chisel3.experiemental` to `chisel3`
* Add `reduceTree` to `Vec`
* Strip outer objects and classes from naming of Modules

#### Bugfixes
* Support `withClock` in `RawModules` with no implicit reset
* Support bulk connecting `DontCare` to Bundles with `Analog` fields
* Make LFSR support async reset
* Fix stack trimming
* Better errors for Queues with bad sizes
* Fix directions in dynamic indexing of complex Vecs

#### Features
* Dependency API
  * Support for specifying FIRRTL transforms via their dependencies on other transforms
  * Not yet used as specification for all transforms yet, but the API is there
* Constant Propagate muxes with literal inputs into Ands and Ors
* Provide SimplifyMems for lowering aggregate memories without splitting
* Add MultiTargetAnnotations
* Allow naming of blackbox `.f` file
* Add support for inferring the type of Reset
* Remove stack traces from user errors
* Improve support for nested Stage/Phase compilers
* Rename "Gender" to "Flow"
* Add basic support for `read-under-write` memories
* Improve API of `InstanceGraph`, used boxed strings

#### Bugfixes
* Improved stability of analog attachments
* Fix renaming of non-local annotation targets
* Fix TopWiringTransform
* Fix RemoveReset when init is invalid
* Fix annotation renaming in InlineInstances
* Make `TopWiringTransform` idempotent

---------

**Related issue**: 

**Release Notes**
* Use native chisel3 async reset support instead of blackboxed support

### DIFF
--- a/src/main/scala/util/AsyncResetReg.scala
+++ b/src/main/scala/util/AsyncResetReg.scala
@@ -3,8 +3,7 @@
 package freechips.rocketchip.util
 
 import Chisel._
-import chisel3.util.HasBlackBoxResource
-import chisel3.core.IntParam
+import chisel3.{withClockAndReset, withReset, RawModule}
 
 /** This black-boxes an Async Reset
   *  (or Set)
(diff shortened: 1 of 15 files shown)


jackkoenig:
Don't merge until further testing and I need to write up the summary of changes to chisel and firrtl included in this bump

src/main/scala/util/AsyncResetReg.scala
```diff
@@ -54,26 +55,15 @@ class SimpleRegIO(val w: Int) extends Bundle{
...
+  val reg = withReset(reset.asAsyncReset)(RegInit(init.U(w.W)))
+  when (io.en) {
+    reg := io.d
   }
-
-  io.q := q.asUInt
-
-  override def desiredName = s"AsyncResetRegVec_w${w}_i${init}"
-
+  io.q := reg
```
aswaterman: Omg it’s so much better

hcook:
Can you open an issue with the list of utility modules you believe could or should be refactored or removed by taking advantage of the new features, particularly AsyncReset?

src/main/scala/util/AsyncResetReg.scala
mwachs5: not that we used it much, but what about the enable signal?

PR #2076, opened 2019-08-14
Deny access to debug region for M-mode and below

aswaterman:
Per discussion with @terpstra, doing this in a direct, non-diplomatic fashion is the best near-term solution.

**Impact**: API modification


aswaterman:
:)

PR #2079, opened 2019-08-15
Split util from rocketchip package

KevinLTSiFive:
**Related issue**: 

**Release Notes**
Split util from rocketchip, sbt and wake functional. Not able to init submodules yet.


KevinLTSiFive:
init split of `util`, `utilunittest` and sources that depend on `diplomacy` and other packages are not split out. Need for further splitting.

albertchen-sifive:
why aren't the unit-tests in `library-utils-chipsalliance` also? it makes sense to me to have `utils` depend on a separate `api-unittest-chipsalliance` package as you had before and to put the unittests in the same package that the modules are defined in.

also you don't need to copy the travis and license files since rocket-chip already has them. unless you plan to split this into a submodule/wit packge

other than that this looks good to me

KevinLTSiFive:
The reason of unit-tests  of `library-utils-chipsalliance` are not in is they depend on `api-unittest-library` which depends on `api-diplomacy-chipsalliance`, which depends on `library-utils-chipsalliance`. This triggers a circle dependency that is forbidden by `wake`. The only way I think to solve this is by moving unit-tests into a  separate `library-utiltests-chipsalliance`.

KevinLTSiFive:
For now, `api-diplomacy-chipsalliance` is not split out yet. A lot of packages that depend on it should stay in. Such as `api-unittest-chipsalliance`. As for license and travis. I think the `library-utils-chipsalliance` will become a separated  sub-repo as `api-config-chipsalliance`. So I copy these files from it.

PR #2081, opened 2019-08-16
Upgrade DescribedSRAM to chisel3.

richardxia:

Being unfamiliar with the specific differences between Chisel 2 and Chisel 3, it took me a little while reading the code to realize that `SyncReadMem` and `SeqMem` are [aliases of each other](https://github.com/freechipsproject/chisel3/blob/b2a1bd7a10977d3331fee3022ec490a1aa1e0e17/src/main/scala/chisel3/compatibility.scala#L286). It doesn't seem to trigger a deprecation warning, and none of the existing chisel3 documentation appears to mention that `SyncReadMem` is the chisel3 version of `SeqMem`, despite there being independent documentation on `SyncReadMem` and `SeqMem`.

I've upgraded the `DescribedSRAM` module to chisel3 so that future readers of the code hopefully don't need to go through the same steps that I had to.

### DIFF
--- a/src/main/scala/util/DescribedSRAM.scala
+++ b/src/main/scala/util/DescribedSRAM.scala
@@ -6,8 +6,8 @@ package freechips.rocketchip.util
 import chisel3.internal.InstanceId
 import freechips.rocketchip.util.Annotated
 import freechips.rocketchip.diplomacy.DiplomaticSRAM
-import Chisel._
-import chisel3.SyncReadMem
+import chisel3.{Data, SyncReadMem, Vec}
+import chisel3.util.log2Ceil
 import freechips.rocketchip.amba.axi4.AXI4RAM
 import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelAddressing
 import freechips.rocketchip.diplomaticobjectmodel.model.OMSRAM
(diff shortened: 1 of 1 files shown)


seldridge:
@richardxia: As one possible source of information that may help with things like this, the new Chisel website has published Scaladoc for all Chisel versions through v3.1.8 (the old website did, too, but you had to manually type the URL for earlier versions).

You can get a quick rundown of all the aliases [here](https://www.chisel-lang.org/api/chisel3/latest/Chisel/index.html) as opposed to looking at the code. The latter may be just as easy to do, however.

The existing ["Migrating from Chisel2 to Chisel3" guide on the website](https://www.chisel-lang.org/chisel3/chisel3-vs-chisel2.html) (which is a copy and likely successor of the [wiki version](https://github.com/freechipsproject/chisel3/wiki/Chisel3-vs-Chisel2)) is also deficient in mentioning `SeqMem`. If you come across other things like this that you feel are unclear or could use better documentation, feel free to file issues directly on Chisel/FIRRTL or the [website repo: freechipsproject/www.chisel-lang.org](https://github.com/freechipsproject/www.chisel-lang.org).

richardxia:
Thanks, @seldridge! I've created a PR to update the Chisel website: https://github.com/freechipsproject/www.chisel-lang.org/pull/12. Let me know if that looks right, and I can port those changes over to the wiki as well.

PR #2085, opened 2019-08-19
Fix OMSRAM serialization in RocketTile memories.

richardxia:
**Related issue**: 

**Impact**: API modification

**Release Notes**
The various Rocket memories were not being correctly serialized into the Object Model JSON file, since the case class describing the `OMCache` set the type of the `memories` attribute to `() => Seq[OMSRAM]`, and it appears that the default JSON serializer serializes function types to the empty JSON object (map) type.

The main fix I made was to change the type from `() => Seq[OMSRAM]` to just `Seq[OMSRAM]`, but I also made a few other changes to how data is being passed around, partly to obviate the need to pass around thunk in the first place but also to (IMO) simplify the control flow a little bit:

- I replaced the thunk arguments of `() => Seq[OMSRAM]` with references to a LazyModule. The `getOMSRAMs()` helper functions that return the `Seq[OMSRAM]` are defined on the LazyModule to begin with, so instead of having to pass around a thunk, which defies serialization and debugging (cannot be "printed"), we can just pass around the reference to the LazyModule itself and only call the `getOMSRAMs()` method after all the Diplomacy handshakes are resolved.
- I removed the `OMCache.dcache()/icache()` static functions and just inlined them into `DCacheLogicalTreeNode` and `ICacheLogicalTreeNode`, since those were the only places where we were calling the functions, and it makes it clearer to see what is actually getting serialized into the object model without having to job to another file.

I tested this by building a few designs and confirming that the `memories` property gets serialized to an actual array of `OMSRAM` objects rather than an empty object.

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
+++ b/src/main/scala/diplomaticobjectmodel/logicaltree/RocketLogicalTreeNode.scala
@@ -3,8 +3,9 @@
 package freechips.rocketchip.diplomaticobjectmodel.logicaltree
 
 import freechips.rocketchip.diplomacy.{LazyModule, ResourceBindings, ResourceBindingsMap, SimpleDevice}
+import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelAddressing
 import freechips.rocketchip.diplomaticobjectmodel.model._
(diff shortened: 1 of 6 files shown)


richardxia:
Thanks for the review!

PR #2086, opened 2019-08-19
Added better description and tentative module name to the object mode…

john-sifive:
Add better descriptions and tentative rtl module names  to the SRAM object model. This information is currently only used for documentation. The module names are NOT guaranteed to be the final names - they are the suggested names before optimization takes place. 

**Impact**: API addition (no impact on existing code

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
+++ b/src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
@@ -166,7 +166,8 @@ object DiplomaticObjectModelAddressing {
     width: Int,
     depth: BigInt,
     granWidth: Int,
-    uid: Int
+    uid: Int,
+    rtlModule: OMRTLModule = OMRTLModule("")
   ): OMSRAM = {
     OMSRAM(
       description = desc,
@@ -174,7 +175,8 @@ object DiplomaticObjectModelAddressing {
       dataWidth = width,
       depth = depth,
       writeMaskGranularity = granWidth,
-      uid = uid
+      uid = uid,
+      rtlModule = rtlModule
     )
   }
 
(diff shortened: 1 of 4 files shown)


src/main/scala/util/DescribedSRAM.scala
```diff
@@ -34,11 +34,12 @@ object DescribedSRAM {
...
 
     val omSRAM = DiplomaticObjectModelAddressing.makeOMSRAM(
-      desc = "mem-" + uid,
+      desc = desc,
       width = data.getWidth,
       depth = size,
       granWidth = granWidth,
-      uid = uid
+      uid = uid,
+      rtlModule = OMRTLModule(name)
```
richardxia: Can you pass arguments by name to the constructor of these OM case classes? There's no real rhyme or reason to the ordering of arguments, so passing arguments positionally is less easy to read than passing arguments by name.

```suggestion
      rtlModule = OMRTLModule(moduleName = name)
```

src/main/scala/diplomaticobjectmodel/DiplomaticObjectModelUtils.scala
```diff
@@ -166,15 +166,17 @@ object DiplomaticObjectModelAddressing {
     width: Int,
     depth: BigInt,
     granWidth: Int,
-    uid: Int
+    uid: Int,
+    rtlModule: OMRTLModule = OMRTLModule("")
```
richardxia: Since you have set a default value for all the fields on the `OMRTLModule` case class, why not just call `OMRTLModule()` with no arguments?

john-sifive: Just an oversight from before the defaults. Thx for catching.

PR #2097, opened 2019-08-28
Scalafix refactoring

tampler:
Launched a single command: sbt fix

### DIFF
--- a/src/main/scala/amba/axi4/Fragmenter.scala
+++ b/src/main/scala/amba/axi4/Fragmenter.scala
@@ -7,7 +7,6 @@ import chisel3.util.IrrevocableIO
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.util.{leftOR, rightOR, UIntToOH1, OH1ToOH}
-import scala.math.{min,max}
 
 class AXI4Fragmenter()(implicit p: Parameters) extends LazyModule
 {
@@ -31,7 +30,7 @@ class AXI4Fragmenter()(implicit p: Parameters) extends LazyModule
       val beatBytes = slave.beatBytes
       val lgBytes   = log2Ceil(beatBytes)
       val master    = edgeIn.master
-      val masters   = master.masters
+      master.masters
 
       // We don't support fragmenting to sub-beat accesses
       slaves.foreach { s =>
@@ -62,7 +61,7 @@ class AXI4Fragmenter()(implicit p: Parameters) extends LazyModule
         val len  = Mux(busy, r_len,  a.bits.len)
         val addr = Mux(busy, r_addr, a.bits.addr)
 
-        val lo = if (lgBytes == 0) UInt(0) else addr(lgBytes-1, 0)
+        if (lgBytes == 0) UInt(0) else addr(lgBytes-1, 0)
         val cutoff = AXI4Parameters.lenBits + lgBytes
         val alignment = addr((a.bits.params.addrBits min cutoff)-1, lgBytes)
 
--- a/src/main/scala/tilelink/RAMModel.scala
+++ b/src/main/scala/tilelink/RAMModel.scala
@@ -34,7 +34,7 @@ class TLRAMModel(log: String = "", ignoreCorruptData: Boolean = false, ignoreDen
       val maxTransfer  = edge.manager.maxTransfer
       val beatBytes    = edge.manager.beatBytes
       val endAddressHi = (endAddress / beatBytes).intValue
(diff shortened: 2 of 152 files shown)


src/main/scala/amba/axi4/Fragmenter.scala
```diff
@@ -31,7 +30,7 @@ class AXI4Fragmenter()(implicit p: Parameters) extends LazyModule
       val beatBytes = slave.beatBytes
       val lgBytes   = log2Ceil(beatBytes)
       val master    = edgeIn.master
-      val masters   = master.masters
+      master.masters
```
hcook: Hmm, should this line just be deleted entirely?

src/main/scala/amba/axi4/Fragmenter.scala
```diff
@@ -62,7 +61,7 @@ class AXI4Fragmenter()(implicit p: Parameters) extends LazyModule
         val len  = Mux(busy, r_len,  a.bits.len)
         val addr = Mux(busy, r_addr, a.bits.addr)
 
-        val lo = if (lgBytes == 0) UInt(0) else addr(lgBytes-1, 0)
+        if (lgBytes == 0) UInt(0) else addr(lgBytes-1, 0)
```
hcook: Should this line just be deleted entirely? is it actually not used?

hcook:
Definitely more tractable to review, thanks. But I'm not a fan of the behavior where unused `val`s in particular scopes are removed and replaced by the statement itself. It's logically equivalent, but will be very confusing to read later! 

Most of those instances should probably be pruned entirely, but also I think at least one is an error (i.e. it should be used in scope but isn't). I think those will need to be singularly evaluated.

Is it possible to turn off that feature?

src/main/scala/util/Annotations.scala
```diff
@@ -261,7 +261,7 @@ object GenRegDescsAnno {
 
     val moduleName = rawModule.name
     val baseHex = s"0x${baseAddress.toInt.toHexString}"
-    val displayName = s"${moduleName}.${baseHex}"
+    s"${moduleName}.${baseHex}"
```
hcook: Seems like this one might be an actual bug, that this isn't passed to `RegistersSer.displayName` below? @derekpappas

src/main/scala/tilelink/RAMModel.scala
```diff
@@ -34,7 +34,7 @@ class TLRAMModel(log: String = "", ignoreCorruptData: Boolean = false, ignoreDen
       val maxTransfer  = edge.manager.maxTransfer
       val beatBytes    = edge.manager.beatBytes
       val endAddressHi = (endAddress / beatBytes).intValue
-      val maxLgBeats   = log2Up(maxTransfer/beatBytes)
+      log2Up(maxTransfer/beatBytes)
```
hcook: Another val removal that makes the code more confusing, though I agree the whole line could be removed

src/main/scala/amba/axi4/Fragmenter.scala
tampler: I don't understand much this "optimization". But it seems like that Scalafix removed an assignment to the local "masters" similar to the constant propagation for a Unit value.
The same name in the local scope with not functional modifications.

Actually, each "val" assumes a memory alloc on JVM, so removing extra "vals" saves on allocs

src/main/scala/amba/axi4/Fragmenter.scala
tampler: Same here. Scalafix removes "val lo" due to non-existant referrals in the local scope. However, removing the whole line is prohibited, since `addr(lgBytes-1,0)` may incur some side-effects (like db query, http, console, etc) and thus can't be eliminated. 

That's the idea of refactoring. Remove all unnecessary calls/vals/imports to simplify the code base and save on JVM allocs in case of val

src/main/scala/util/Annotations.scala
tampler: The removed `displayName` isn't referenced in the local scope

src/main/scala/tilelink/RAMModel.scala
tampler: Scalafix tracks dangling references. In this line, it (potentially) saved on a JVM alloc, but couldn't remove the same line due to a potential side effect.

So, as I see, it removes `val` decl, but keeps such calls as `Unit` references in case of side-effecting calls, which Scalafix seems not be able to discover at this moment

src/main/scala/util/Annotations.scala
hcook: I meant that it should be referenced in the local scope and it is a bug in the local scope that it is not

PR #2103, opened 2019-09-07
Fix configs with both RV32D and scratchpads

aswaterman:
A refactoring in 3133c321b7b771150168b687834fac5edd4a292a made the assumption that coreDataBits = xLen, which is not true for RV32D. Since the assumption wasn't paired with an assertion, attempting to generate such configs resulted in successful elaboration of a broken design.

**Impact**: API modification

### DIFF
--- a/src/main/scala/tile/RocketTile.scala
+++ b/src/main/scala/tile/RocketTile.scala
@@ -51,10 +51,14 @@ class RocketTile private(
   val slaveNode = TLIdentityNode()
   val masterNode = visibilityNode
 
-  val dtim_adapter = tileParams.dcache.flatMap { d => d.scratch.map(s =>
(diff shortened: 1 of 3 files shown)


src/main/scala/tile/RocketTile.scala
```diff
@@ -51,10 +51,14 @@ class RocketTile private(
   val slaveNode = TLIdentityNode()
   val masterNode = visibilityNode
 
-  val dtim_adapter = tileParams.dcache.flatMap { d => d.scratch.map(s =>
-    LazyModule(new ScratchpadSlavePort(AddressSet.misaligned(s, d.dataScratchpadBytes), xBytes, tileParams.core.useAtomics && !tileParams.core.useAtomicsOnlyForIO)))
-  }
-  dtim_adapter.foreach(lm => connectTLSlave(lm.node, xBytes))
+  val dtim_adapter = tileParams.dcache.flatMap { d => d.scratch.map { s =>
+    val coreParams = {
+      class C(implicit val p: Parameters) extends HasCoreParameters
```
davidbiancolin: Heh.

aswaterman: :(

PR #2104, opened 2019-09-07
pipe out compatname

erikdanie:
**Related issue**: 

**Release Notes**

This is a somewhat dirty way of allowing the compat string of an sram to be set, but to default to "sifive,sram0". Another option here would be to make all the parameters a case class, but I know this code is used a lot, so I would prefer not to break compatibility. If you prefer that method though, let me know and I can implement it.

### DIFF
--- a/src/main/scala/diplomacy/SRAM.scala
+++ b/src/main/scala/diplomacy/SRAM.scala
@@ -12,10 +12,10 @@ abstract class DiplomaticSRAM(
     address: AddressSet,
     beatBytes: Int,
     devName: Option[String],
-    compatName: String = "sifive,sram0")(implicit p: Parameters) extends LazyModule
+    dtsCompat: Option[Seq[String]] = None)(implicit p: Parameters) extends LazyModule
 {
   val device = devName
-    .map(new SimpleDevice(_, Seq(compatName)))
+    .map(new SimpleDevice(_, dtsCompat.getOrElse(Seq("sifive,sram0"))))
     .getOrElse(new MemoryDevice())
 
   def getOMMemRegions(resourceBindings: ResourceBindings): Seq[OMMemoryRegion] = {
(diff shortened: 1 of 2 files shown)


src/main/scala/diplomacy/SRAM.scala
```diff
@@ -11,10 +11,11 @@ import freechips.rocketchip.util.DescribedSRAM
 abstract class DiplomaticSRAM(
     address: AddressSet,
     beatBytes: Int,
-    devName: Option[String])(implicit p: Parameters) extends LazyModule
+    devName: Option[String],
+    compatName: String = "sifive,sram0")(implicit p: Parameters) extends LazyModule
```
richardxia: I hate to do this, given that you have to pipe this through so many layers and across repositories, but I think it'd be a good idea to give this parameter a name that makes it very explicitly clear that this is meant to hold a DTS `compatible` string. For example, I'd prefer a name like `dtsCompatible`, or, if you find that too verbose, at the very least the fully spelled out `compatible`, since that's at least more obviously identifiable with the equivalent property in DTS.

Also, I think you should make this a `Seq[String]`, since this would allow one to set multiple compatible strings, as is allowed by the DTS spec.

src/main/scala/tilelink/SRAM.scala
```diff
@@ -24,7 +24,8 @@ class TLRAM(
     beatBytes: Int = 4,
     ecc: ECCParams = ECCParams(),
     val devName: Option[String] = None,
-  )(implicit p: Parameters) extends DiplomaticSRAM(address, beatBytes, devName)
+    val compatName: String = "sifive,sram0"
```
richardxia: Thinking about this more, I'm wary of having the same default value set at multiple different levels, since it makes it harder to tell which level of abstraction the actual value came from. Instead, what if you made this an `Option[Seq[String]]` which defaults to `None` both here and in `DiplomaticSRAM`, and within the body of `DiplomaticSRAM`, you do a `new SimpleDevice(_, compatible.getOrElse(Seq("sfive,sram0")))`?

src/main/scala/diplomacy/SRAM.scala
erikdanie: Ah yes, that is what I should do. I was having trouble making the dafault parameter in 2 places

PR #2109, opened 2019-09-12
Remove extra (Berkeley) pom scm tag, add developer, re-indent pom.

ucbjrl:
The extra scm tag (some other piece of code is generating a SiFive scm tag) and the lack of a developer one cause pre-publish errors on Sonatype.


sequencer:
maybe the SCM should be chipsalliance now...

PR #2110, opened 2019-09-13
Adding printing of diplomacy parameters in graphml and addition of emits parameters in manager and client parameters.

brrmorre:
**Related issue**: 

**Impact**: API addition
Started adding support for clients and managers to specify exactly what kind of tile link opcodes they emit.

**Release Notes**


src/main/scala/amba/ahb/ToTL.scala
```diff
@@ -11,7 +11,17 @@ import freechips.rocketchip.util.MaskGen
...
+      TLClientParameters(name = m.name, nodePath = m.nodePath,
+        knownToEmit = Some(TLClientEmissionSizes(
+          emitsAcquireT = TransferSizes.none,
+          emitsAcquireB = TransferSizes.none,
+          emitsArithmetic = TransferSizes.none,
+          emitsLogical = TransferSizes.none,
+          emitsGet = TransferSizes(1, 128),
+          emitsPutFull = TransferSizes(1, 128),
+          emitsPutPartial = TransferSizes.none,
+          emitsHint = TransferSizes.none
```
hcook: I think for brevity this is a fine spot to rely on the default parameter values of `TransferSizes.none` and only call out the specific types that are in fact emitted.

src/main/scala/devices/debug/DMI.scala
```diff
@@ -79,7 +79,18 @@ class ClockedDMIIO(implicit val p: Parameters) extends ParameterizedBundle()(p){
...
+  val node = TLClientNode(Seq(TLClientPortParameters(Seq(TLClientParameters("debug",
+  knownToEmit = Some(TLClientEmissionSizes(
+    emitsAcquireT = TransferSizes.none,
+    emitsAcquireB = TransferSizes.none,
+    emitsArithmetic = TransferSizes.none,
+    emitsLogical = TransferSizes.none,
+    emitsGet = TransferSizes(4, 4),
+    emitsPutFull = TransferSizes(4, 4),
+    emitsPutPartial = TransferSizes.none,
+    emitsHint = TransferSizes.none
```
hcook: I think for brevity this is a fine spot to rely on the default parameter values of `TransferSizes.none` and only call out the specific types that are in fact emitted.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -10,12 +10,25 @@ import freechips.rocketchip.util.{RationalDirection,AsyncQueueParams, groupByInt
 import scala.math.max
 import scala.reflect.ClassTag
 
+case class TLManagerEmissionSizes(
+  emitsProbe:   TransferSizes   = TransferSizes.none,
```
hcook: either align all the whitespace or none #OCD

src/main/scala/tilelink/Nodes.scala
```diff
@@ -32,17 +32,52 @@ object TLImp extends NodeImp[TLClientPortParameters, TLManagerPortParameters, TL
...
+  "\nManager Name = " + currManager.name +
+  "\nManager Address = " + currManager.address +
+  "\nsupportsAcquireT = " + currManager.supportsAcquireT +
+  "\nsupportsAcquireB = " + currManager.supportsAcquireB +
+  "\nsupportsArithmetic = " + currManager.supportsArithmetic +
+  "\nsupportsLogical = " + currManager.supportsLogical +
+  "\nsupportsGet = " + currManager.supportsGet +
+  "\nsupportsPutFull = " + currManager.supportsPutFull +
+  "\nsupportsPutPartial = " + currManager.supportsPutPartial +
+  "\nsupportsHint = " + currManager.supportsHint + "\n").mkString).mkString
```
hcook: Are all these different implementations of `formatNode` actually different from each other? Can it just take some extra argument that the sub classes supply instead of copy/pasting all this code?
Pattern is something like:
```
override def formatNode() = super.formatNode("thingThatIsDifferent")
```

hcook: Oh it's the type of `def edges: Edges[List[TLEdgeIn], List[TLEdgeOut]]` that is different? hmm

hcook: can you maybe do a `private def formatAgentParams(client: Seq[TLClientParameters], managers: Seq[TLManagerParameters])` with all the string formatting and then the different node types only handle unpacking those lists from their unique Edge classes?

src/main/scala/tilelink/Nodes.scala
```diff
@@ -32,17 +32,52 @@ object TLImp extends NodeImp[TLClientPortParameters, TLManagerPortParameters, TL
...
+}
 
+case class TLClientNode(portParams: Seq[TLClientPortParameters])(implicit valName: ValName) extends SourceNode(TLImp)(portParams) with TLFormatNode
+//{
+//  override def formatNode() = "Output Edge (edges.out.head.client): \nanyEmitAcquireT = " + edges.out.head.client.anyEmitAcquireT +
+//  "\nanyEmitAcquireB = " + edges.out.head.client.anyEmitAcquireB +
+//  "\nanyEmitArithmetic = " + edges.out.head.client.anyEmitArithmetic +
+//  "\nanyEmitLogical = " + edges.out.head.client.anyEmitLogical
+//}
```
hcook: remove commented out code

src/main/scala/tilelink/Nodes.scala
```diff
@@ -32,17 +32,52 @@ object TLImp extends NodeImp[TLClientPortParameters, TLManagerPortParameters, TL
...
-case class TLClientNode(portParams: Seq[TLClientPortParameters])(implicit valName: ValName) extends SourceNode(TLImp)(portParams)
-case class TLManagerNode(portParams: Seq[TLManagerPortParameters])(implicit valName: ValName) extends SinkNode(TLImp)(portParams)
+trait TLFormatNode extends BaseNode
+{
+  def edges: Edges[List[TLEdgeIn], List[TLEdgeOut]]
+  override def formatNode() = "TileLink Node\n" + edges.out.map(currEdge =>
+  currEdge.client.clients.map(currClient => "Output Edges (edges.out.map.client.clients.map):" +
+  "\nClient Name = " + currClient.name +
+  "\nvisibility = " + currClient.visibility +
+  (if (currClient.knownToEmit == None) "\nEmits parameters are UNKNOWN\n" else
```
hcook: I think you want some tabs in the scala code here and maybe also in the output string. You might consider looking at the `"""` escape for multiline strings with `stripMargin` (https://alvinalexander.com/scala/how-to-create-multiline-strings-heredoc-in-scala-cookbook) and/or using string interpolation.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -318,6 +357,44 @@ case class TLClientPortParameters(
...
+  def emitsLogicalFast   (address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = if (clients.exists(_.knownToEmit == None)) true.B else emitHelper(false, _.knownToEmit.get.emitsLogical,    address, lgSize, range)
+  def emitsGetFast       (address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = if (clients.exists(_.knownToEmit == None)) true.B else emitHelper(false, _.knownToEmit.get.emitsGet,        address, lgSize, range)
+  def emitsPutFullFast   (address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = if (clients.exists(_.knownToEmit == None)) true.B else emitHelper(false, _.knownToEmit.get.emitsPutFull,    address, lgSize, range)
+  def emitsPutPartialFast(address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = if (clients.exists(_.knownToEmit == None)) true.B else emitHelper(false, _.knownToEmit.get.emitsPutPartial, address, lgSize, range)
+  def emitsHintFast      (address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = if (clients.exists(_.knownToEmit == None)) true.B else emitHelper(false, _.knownToEmit.get.emitsHint,       address, lgSize, range)
+
```
jackkoenig: There's a lot of duplication here, you may notice the lines around this where the code does stuff like:
```scala
val supportsProbe      = safety_helper(_.supportsProbe)      _
```
That's using currying to duplicate less, and I would suggest doing the same thing here.

eg.

Assuming we can unify `emitHelper` with `supportHelper`, we can wrap with a curried function:

```scala
  private def curriedEmitHelper(
      safe:    Boolean,
      f:  TLManagerEmissionSizes => TransferSizes)
     (address: UInt,
      lgSize:  UInt,
      range:   Option[TransferSizes] = None): Bool =
  if (clients.exists(_.knownToEmit == None)) true.B
  else emitHelper(safe, p => f(p.knownToEmit.get), address, lgSize, range)
```

Then you should be able to vastly simplify these functions:
```scala
val emitsAcquireTSafe = curriedEmitHelper(true, _.emitsAcquireT) _
...
```

brrmorre: Did you want me to change the supportsAcquireTSafe... functions as well?

jackkoenig: Ah I see you were copying that style, I'll defer to @terpstra on if those should be changed.

I think it's more important here because you have the additional `if (clients.exists(_.knownToEmit == None)) true.B else ...` boilerplate.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -318,6 +357,44 @@ case class TLClientPortParameters(
     }
   }
 
+  private def emitHelper(
```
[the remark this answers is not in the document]
brrmorre: it is very similar, but it instead looks at visibility instead of address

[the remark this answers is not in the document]
jackkoenig: That could be abstracted into a function argument. I'm concerned about copying code that looks very similar because, as an example, the existence of `groupByIntoSeq` in this method. I added that to `TLManagerPortParameters.supportHelper` because it was generating non-deterministic code. If we had had 2 functions, then I likely would've missed one. 

I'm not certain that it *should* be abstracted or what the new utility function would be called, thoughts @terpstra?

src/main/scala/devices/debug/DMI.scala
```diff
@@ -79,7 +79,13 @@ class ClockedDMIIO(implicit val p: Parameters) extends ParameterizedBundle()(p){
 
 class DMIToTL(implicit p: Parameters) extends LazyModule {
 
-  val node = TLClientNode(Seq(TLClientPortParameters(Seq(TLClientParameters("debug")))))
+  val node = TLClientNode(Seq(TLClientPortParameters(Seq(TLClientParameters("debug",
+  knownToEmit = Some(TLClientEmissionSizes(
+    emitsGet = TransferSizes(4, 4),
+    emitsPutFull = TransferSizes(4, 4),
+    emitsPutPartial = TransferSizes(4, 4)
```
hcook: can you make this whitespace look like what you did in ahb/ToTL.scala?

PR #2111, opened 2019-09-13
Wit/Wake use documentation

jackkoenig:
Squash and merge please!

This is just a start, but it's something.

**Related issue**: 

**Release Notes**

Add initial documentation for Wit and Wake flow.

### DIFF
--- a/README_WAKE.adoc
+++ b/README_WAKE.adoc
@@ -1,0 +1,50 @@
+= Rocket Chip Generator
+:toc:
+:toc-placement!:
+
+This repository contains the Rocket chip generator necessary to instantiate
+the RISC-V Rocket Core. For more information on Rocket Chip, please consult our
+http://www.eecs.berkeley.edu/Pubs/TechRpts/2016/EECS-2016-17.html[technical report].
+
+toc::[]
+
+== Dependencies
+
+This repository uses version 0.10.1 of https://github.com/sifive/wit[Wit] and version 0.15.1 of https://github.com/sifive/wake[Wake].
+Please see the respective repositories and their releases to install these dependencies.
+
+== Checkout The Code
+
+This repository uses https://github.com/sifive/wit[Wit] for dependency fetching and resolution.
+Please see the https://github.com/sifive/wit/blob/v0.10.1/share/doc/wit/tutorial.md[Wit tutorial] for more information.
+
+----
+$ wit init workspace -a git@github.com:chipsalliance/rocket-chip.git
+----
+
+This will clone rocket-chip as well as resolve and fetch all dependencies.
+
+== Compile Rocket Chip
+
+It is not necessary to compile separately from invocation, but it might be useful to understand how the different pieces work.
+
+This repository uses https://github.com/sifive/wake[Wake] for orchestrating the build.
+We strongly recommend checking out the https://github.com/sifive/wake/blob/v0.15.1/share/doc/wake/tutorial.md[Wake tutorial].
+You can find the Rocket Chip Wake rules in `build.wake` within this repository.
+
+The Rocket Chip Scala code is compiled using https://github.com/sifive/api-scala-sifive[API Scala SiFive],
+and can thus be compiled with:
+
+----
+$ wake compileScalaModule rocketchipScalaModule
+----
+
+The return value of that function is the `Path` to the compile jar.
+
+== Run Rocket Chip Generator
+
+== Tips and Tricks
+
+See the https://github.com/sifive/wake/tree/a0d99f15cf806f917d64714934d337b392d36f26/share/doc/wake[Wake `share/doc` directory] for documentation.
+Note that this link is to a newer commit than 0.15.1 because more documentation has been written since the release.
+Wake is still evolving so its hard for everything to be 100% synced, but the linked documentation should be compatible at time of writing.
--- a/README_WAKE.adoc
+++ b/README_WAKE.adoc
@@ -39,6 +39,8 @@ The return value of that function is the `Path` to the compile jar.
 
 == Run Rocket Chip Generator
 
+=== API
+
 You can run the generator with the Wake rule `runRocketChipGenerator`.
 This function accepts a single `RocketChipGeneratorOptions` as an argument.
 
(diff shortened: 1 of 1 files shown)


README_WAKE.adoc
```diff
@@ -0,0 +1,154 @@
...
+  
+global def makeRocketChipGeneratorOptions jars targetDir topModule configs =
+  RocketChipGeneratorOptions jars targetDir topModule configs Nil None
+----
+Please see the https://github.com/sifive/wake/blob/a0d99f15cf806f917d64714934d337b392d36f26/share/doc/wake/tour/tuples.adoc[Wake tuple documentation]
+for more information about Tuples.
+
+Because the constructor is private, we use `makeRocketChipGeneratorOptions` to construct `RocketChipGeneratorOptions`:
+
+* `jars` is a List of compiled Scala jars for your project, for example `rocketchipScalaModule.scalaModuleClasspath` for base Rocket Chip.
```
richardxia: Since it isn't completely obvious to me, let me double check my understanding. Are these are meant to include both the rocket-chip dependencies as well as the actual rocket-chip application? And to double check, in this flow we do not create a fat JAR, right? We may want to make that explicit here (also maybe as a comment in the actual tuple definition in the code).

Also, technical writing nit: JAR should be spelled in all caps (in your prose).
Other nit: Shouldn't `jars` be `Jars`? Wake tuple fields are spelled with an uppercase initial letter, and for people who are less familiar with Wake, you really should match the case or else it could cause confusion because it may look like `jars` is not the same thing as `Jars`.

README_WAKE.adoc
```diff
@@ -0,0 +1,154 @@
...
+* `targetDir` is the directory where the output of the generator will be placed.
+* `topModule` is the name of the top Chisel Module, eg. `"freechips.rocketchip.system.TestHarness"`
```
richardxia: Might be good to replace "name" with "fully-qualified name".

README_WAKE.adoc
```diff
@@ -0,0 +1,154 @@
...
+The return value of that function is the `Path` to the compile jar.
+
+== Run Rocket Chip Generator
+
+=== API
+
+You can run the generator with the Wake rule `runRocketChipGenerator`.
+This function accepts a single `RocketChipGeneratorOptions` as an argument.
+
+----
```
richardxia: Although GitHub doesn't have a built-in Wake syntax highlighter, if we ever produce static HTML versions of these READMEs it'll be nicer to explicitly annotate the code block with the language so that we can apply the appropriate syntax highlighting.

See how I did this in the Wake documentation that I wrote recently: https://github.com/sifive/wake/blame/2dba999205b58ace3dc7d92a1010335f201dff58/share/doc/wake/from-scala.adoc#L40

README_WAKE.adoc
```diff
@@ -0,0 +1,154 @@
...
+* Turn on verbose with `-v`. You can use this to see the type of a function, eg. `wake -v map`.
+
+* See all global variables with `-g`, this can be combined with `grep` to see available functions for a given type, try
+----
+$ wake -g | grep ScalaModule
+----
+
+* You can see the job that created a file with `wake -o <file>`, and jobs that use a file as input with `wake -i <file>`
+
+* You can combine `-o` and `-i` with `-s` to create an exectuable bash script that runs a job.
```
richardxia: Might be worth mentioning that the executable shell script includes things like setting up all the environment variables.

README_WAKE.adoc
```diff
@@ -0,0 +1,154 @@
...
+
+* By default, Wake does not maintain a stack trace, use `-d|--debug` to have Wake create one.
```
richardxia:
```suggestion
* By default, Wake does not maintain a stack trace; use `-d|--debug` to have Wake create one.
```

README_WAKE.adoc
jackkoenig: Yes it includes rocket-chip's dependencies and rocket-chip itself, everything as JARs.

We do **not** create a fat JAR, I think that does belong in the documentation of API Scala SiFIve, and yeah this tuple should mention it needs all of the necessary compiled JARs. Technically this code is agnostic as to where the JARs come from, someone could pass a fat JAR if they wanted to.

In those bullets, I'm referring to the arguments of `makeRocketChipGeneratorOptions`, I can clarify that or also include which parts of the tuple they correspond to, what do you think?

README_WAKE.adoc
jackkoenig: Agreed

README_WAKE.adoc
jackkoenig: Oh good call, I'll add that

jackkoenig: I do think we should look into generating websites sooner rather than later, easier than trying to get Github to accept Wake as a language and then we can use the more advanced features (like pulling in code rather than copy pasting it which I am only currently doing out of necessity.

README_WAKE.adoc
jackkoenig: Agreed

README_WAKE.adoc
richardxia: > We do not create a fat JAR, I think that does belong in the documentation of API Scala SiFIve, and yeah this tuple should mention it needs all of the necessary compiled JARs. Technically this code is agnostic as to where the JARs come from, someone could pass a fat JAR if they wanted to.

Even if API Scala SiFive does not create fat JARs, it wasn't completely obvious to me when reading just this documentation whether this specific function takes in just the JAR of the application or the JARs of the application and all of its dependencies.

> In those bullets, I'm referring to the arguments of makeRocketChipGeneratorOptions, I can clarify that or also include which parts of the tuple they correspond to, what do you think?


My bad, I got confused while reviewing your text and thought you were describing the Tuple itself. Maybe I was just expecting you to document the Tuple itself, since the Tuple is defined before the constructor function, and that's why I got confused.

jackkoenig: I'll try to make the prose more clear 🙂

sequencer:
wake has been removed in #2847

PR #2113, opened 2019-09-17
TileLink SRAM only check ECC errors on requested bytes

johningalls-sifive:
**Impact**: bug fix

### DIFF
--- a/src/main/scala/tilelink/SRAM.scala
+++ b/src/main/scala/tilelink/SRAM.scala
@@ -93,6 +93,7 @@ class TLRAM(
     val d_read      = Reg(Bool())
     val d_atomic    = Reg(Bool())
     val d_address   = Reg(UInt(width = addrBits.size))
+    val d_rd_mask   = MaskGen(d_address, d_size, beatBytes)
     val d_rmw_mask  = Reg(UInt(width = beatBytes))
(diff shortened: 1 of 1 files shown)


terpstra:
So, I don't like this change for two reasons:
1. It inhibits correction
2. It uses MaskGen combinationally in the readout cycle

terpstra:
I've prepared #2116 as my proposed fix. Can you rerun whatever test you used to validate your version?

johningalls-sifive:
>     1. It inhibits correction
That's right: I propose that a request should not touch lanes that it doesn't ask for.

>     2. It uses MaskGen combinationally in the readout cycle
That's right: I assume that MaskGen is faster than SRAM read + ECC decode, and cheaper to re-compute rather than add Register bits for a/d_lanes.

> I've prepared #2116 as my proposed fix. Can you rerun whatever test you used to validate your version?
Yes, I confirm that #2116 also fixes the failing test.

terpstra:
> > ```
> > 1. It inhibits correction
> > ```
> That's right: I propose that a request should not touch lanes that it doesn't ask for.

Just an FYI, if Chisel had per-lane read-enable, I would totally agree with you and have implemented it your way. But @jackkoenig doesn't care about power.

PR #2116, opened 2019-09-18
TLSRAM: only report corrupt for requested ECC blocks

terpstra:

**Release Notes**
only report corrupt for requested ECC blocks

### DIFF
--- a/src/main/scala/tilelink/SRAM.scala
+++ b/src/main/scala/tilelink/SRAM.scala
@@ -96,6 +96,7 @@ class TLRAM(
     val d_rmw_mask  = Reg(UInt(width = beatBytes))
     val d_rmw_data  = Reg(UInt(width = 8*beatBytes))
     val d_poison    = Reg(Bool())
+    val d_lanes     = Reg(UInt(width = lanes))
 
     // Decode raw unregistered SRAM output
     val d_raw_data      = Wire(Vec(lanes, Bits(width = width)))
@@ -105,7 +106,8 @@ class TLRAM(
     val d_correctable   = d_decoded.map(_.correctable)
     val d_uncorrectable = d_decoded.map(_.uncorrectable)
     val d_need_fix      = d_correctable.reduce(_ || _)
-    val d_error         = d_uncorrectable.reduce(_ || _)
+    val d_lane_error    = Cat(d_uncorrectable.reverse) & d_lanes
+    val d_error         = d_lane_error.orR
 
     notifyNode.foreach { nnode =>
       nnode.bundle.correctable.foreach { c =>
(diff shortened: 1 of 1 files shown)


terpstra:
Supersedes #2113

johningalls-sifive:
> Supersedes #2113

I confirm that this also fixes the failing test.

src/main/scala/tilelink/SRAM.scala
```diff
@@ -105,7 +106,8 @@ class TLRAM(
     val d_correctable   = d_decoded.map(_.correctable)
     val d_uncorrectable = d_decoded.map(_.uncorrectable)
     val d_need_fix      = d_correctable.reduce(_ || _)
-    val d_error         = d_uncorrectable.reduce(_ || _)
+    val d_lane_error    = Cat(d_uncorrectable.reverse) & d_lanes
+    val d_error         = d_lane_error.orR
 
     notifyNode.foreach { nnode =>
       nnode.bundle.correctable.foreach { c =>
```
johningalls-sifive: Here a request could notify a correctable ECC error using the request's address, even when the request's address+size does not include the error.

terpstra: That is intentional. Correctable ECC errors should always be corrected as soon as they are detected.

terpstra:
@hcook Does John have approval permission?

src/main/scala/tilelink/SRAM.scala
terpstra: I think it's important to also note that corrected errors are NOT reported to the user.

terpstra: Oh. I see that there is now an auxiliary 'notifyNode'. @hcook I'm inclined to leave these out-of-band notifications to repair errors as early as detected. For example, I might have had an idle check-and-scrub loop running also in the absence of requests. Do you agree?

hcook: If an error was discovered and corrected, a notification should occur. I don't see it as having anything to do semantically with a particular request, as the notification doesn't contain nor record information like a TL source id or a size. So I agree that notifications should occur for scrubbing activity. Does the address field need to be somehow more precise?

terpstra:
Re: the address, that's a good point. If you perform a bus-wide access, it does not provide information about which subset of ECC granules was broken. Performing a sub-bus-wide access ALSO does a bus-wide correction. So perhaps we should just zero the sub-beat address information to make this explicit?

terpstra:
Actually, what is being reported in "notify.bits" is NOT the physical address, but the SRAM row index. So this has already wiped out the low-order sub-beat bits. It also does not include any of the higher order bits. In the case of banked SRAMs, it doesn't even correspond to a simple bit-shift and OR.

... what was the intended report value?

If we decide the current behaviour is wrong, that is a different change. Fixing this issue is I think unrelated and does not affect the current semantics at all.

PR #2120, opened 2019-09-19
check_submodules: update regex

albertchen-sifive:
updates the regex in the `check_submodules` script. the hashes output by `git submodule status` may be prefixed with `-`, `+`, or `U` and there may be stuff from `git describe` appended to the end of each line.

**Related issue**: 

**Release Notes**

### DIFF
--- a/ci-tests/check_submodules
+++ b/ci-tests/check_submodules
@@ -9,3 +9,10 @@ filter='.[]|select(.commit == "\1")|select(.name == "\2")'
 git submodule status \
 | sed -r "s/[ -+U]([a-zA-Z0-9]+) ([a-zA-Z0-9\-]+) ?.*/$filter/g" \
 | xargs -d '\n' -I % sh -c "jq -e '%' wit-manifest.json"
+
+if [ $? -eq 0 ]; then
+  echo "Submodules are in sync with wit-manifest"
+else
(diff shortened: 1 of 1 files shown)


jackkoenig:
Could you add a print to the script to make it more obvious so that ones doesn't have to `echo $?` to see if it passed? Something like:
```diff
diff --git a/ci-tests/check_submodules b/ci-tests/check_submodules
index e2f5858..09413a8 100755
--- a/ci-tests/check_submodules
+++ b/ci-tests/check_submodules
@@ -9,3 +9,10 @@ filter='.[]|select(.commit == "\1")|select(.name == "\2")'
 git submodule status \
 | sed -r "s/[ -+U]([a-zA-Z0-9]+) ([a-zA-Z0-9\-]+) ?.*/$filter/g" \
 | xargs -d '\n' -I % sh -c "jq -e '%' wit-manifest.json"
+
+if [ $? -eq 0 ]; then
+  echo "Submodules are in sync with wit-manifest"
+else
+  echo "ERROR! Submodules are out of sync with wit-manifest!"
+  exit 1
+fi
```

albertchen-sifive:
> Could you add a print to the script to make it more obvious so that ones doesn't have to echo $? to see if it passed? Something like:

good idea

PR #2125, opened 2019-09-24
Events require hits within mask width

johningalls-sifive:

add a requirement that the number of events within a set must fit within the mask width


aswaterman:
Looks like this will break the `dump` routine defined just below?

johningalls-sifive:
> Looks like this will break the `dump` routine defined just below?

ah, right, will move the check to the evaluate function inside the EventSets class

Issue #2020, opened 2019-07-01
Both VCS and emulator stuck at WFI


yintianyu:
**Type of issue**: bug report

**Impact**: unknown

**Development Phase**: request

**Other information**
 

When simulating using vcs or emulator, it stuck at PC=0x10054, intruction WFI, Until cycle timeout.

I think it is related to fesvr.

I installed the up-to-date riscv-gnu-toolchain and riscv-isa-sim. When simulating, it says "libfesvr.so" no such file or directory. There is surely no $RISCV/lib/libfesvr.so but only $RISCV/lib/libfesvr.a exists.

I copied a libfesvr.so from old version there and it can perform simulation but stuck.

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

1. Install up-to-date riscv-gnu-toolchain
2. Install up-to-date riscv-isa-sim
3. install up-to-date riscv-tests
4. clone up-to-date rocket-chip
5. in vsim, perform "make output/qsort.riscv.out"

**What is the current behavior?**

Stuck at WFI.

**What is the expected behavior?**

Pass the test

**Please tell us about your environment:**

    - version: `bcb3d0a9c097a4d66687bcf7aded3f94048c6a2e`
    - OS: `Linux server125 5.0.2-1.el7.elrepo.x86_64 #1 SMP Wed Mar 13 18:41:45 EDT 2019 x86_64 x86_64 x86_64 GNU/Linuxx`

**What is the use case for changing the behavior?**

None

Issue #1904, opened 2019-03-27
Simulation with vsim hangs when executing WFI


hanoskoe:
**Type of issue**: bug report

**Impact**: unknown

**Development Phase**: request

**Other information**
 
Hi guys,
I am new to Rocket Chip and I could install the repo without problems. Then I tried to generate the files .out and .vcd in emulator and in vsim. What is strange for me is that I don't have any problems with emulator, but when generating the rv64ui-p-add.out file, when pc reaches the address 10054, it just hangs until it is stopped by the timeout (I once let it write until the end).

Is this a known issue or am I just doing something wrong? I am a freshman and don't understand what's going on. I didn't find any similar reported issue.

Thanks in advance

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
export RISCV= [pointing to riscv-tools]
cd RISCV
./build.sh
cd ../vsim
make output/rv64ui-p-add.out

**What is the current behavior?**
Here's the beginning of my rv64ui-p-add.out file:
`testing $random a15b114 seed 3324838479`
`C0:          0 [0] pc=[00000000fa73d4f4] W[r 0=0000000000000000][0] R[r30=638608c6638608c5] R[r 8=d14466a2d14466a2] inst=[428f4e57] DASM(428f4e57)`
`C0:          1 [0] pc=[00000000fa73d4f4] W[r 0=0000000000000000][0] R[r30=638608c6638608c5] R[r 8=d14466a2d14466a2] inst=[428f4e57] DASM(428f4e57)`
`C0:          2 [0] pc=[00000000fa73d4f4] W[r 0=0000000000000000][0] R[r30=638608c6638608c5] R[r 8=d14466a2d14466a2] inst=[428f4e57] DASM(428f4e57)`
`C0:          3 [0] pc=[00000000fa73d4f4] W[r 0=0000000000000000][0] R[r30=638608c6638608c5] R[r 8=d14466a2d14466a2] inst=[428f4e57] DASM(428f4e57)`
`C0:          4 [0] pc=[00000000fa73d4f4] W[r 0=0000000000000000][0] R[r30=638608c6638608c5] R[r 8=d14466a2d14466a2] inst=[428f4e57] DASM(428f4e57)`
`C0:          5 [0] pc=[00000000fa73d4f4] W[r 0=0000000000000000][0] R[r30=638608c6638608c5] R[r 8=d14466a2d14466a2] inst=[428f4e57] DASM(428f4e57)`
[...]

PR #2126, opened 2019-09-26
Add -static option in emulator/Makefile to link libfesvr.a correctly 

ShuyunJia:
**Related issue**: <#1904 #2020>

The emulator file size would be a bit more larger than previous since static library would be included. 

**Release Notes**

For those who have built the fesvr from a older version, both the libfesvr.so(of the older version) and libfesvr.a would be in the $RISCV/lib directory. The -static option tells the linker to use libfesvr.a explicitly.


ShuyunJia:
I found this issue disappears with the riscv-tools 072a2. Request closed.

PR #2127, opened 2019-09-27
Allow a RationalCrossingSource to remain in reset longer than sink

aswaterman:
...by preventing any enqueues that might spuriously happen during reset.

**Related issue**: 

**Release Notes**


aswaterman:
@terpstra this isn't exactly what we had talked about, but I think it does the job in a clean way.

aswaterman:
We had discussed in person moving towards a single, simpler reset strategy. This seems like a reasonable thing to do in pursuit of that. I think @hcook was thinking along the same lines?

hcook:
Seems strictly safer to me

PR #2134, opened 2019-10-02
tilelink: add TLFilter.mSubtract

hcook:
Adds a new built-in `ManagerFilter` to the `TLFilter` helper object called `mSubtract`. Factors out some common code between it and `mSelectIntersect`.

### DIFF
--- a/src/main/scala/tilelink/Filter.scala
+++ b/src/main/scala/tilelink/Filter.scala
@@ -85,7 +85,7 @@ object TLFilter
     intersectionHelper(m, filtered, alignment)
   }
 
-  // make everything except the intersecyed address sets visible
+  // make everything except the intersected address sets visible
   def mExceptIntersect(except: AddressSet): ManagerFilter = { m =>
     val filtered = m.address.flatMap(_.subtract(except))
     val alignment: BigInt = if (filtered.isEmpty) 0 else filtered.map(_.alignment).max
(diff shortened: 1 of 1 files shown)


terpstra:
I don't think this is correct. Consider a slave which has the address range [0x400, 0x800) and a 'filterExcept' of [0x600, 0x800). Presumably, you want the slave to keep [0x400, 0x600), but this proposal will remove the slave entirely.

src/main/scala/tilelink/Filter.scala
```diff
@@ -77,10 +77,23 @@ object TLFilter
...
   def mSelectIntersect(select: AddressSet): ManagerFilter = { m =>
     val filtered = m.address.map(_.intersect(select)).flatten
     val alignment = select.alignment /* alignment 0 means 'select' selected everything */
+    intersectionHelper(m, filtered, alignment)
+  }
+
+  // make everything except the intersected address sets visible
+  def mExceptIntersect(except: AddressSet): ManagerFilter = { m =>
+    val filtered = m.address.flatMap(_.subtract(except))
+    val alignment: BigInt = if (filtered.isEmpty) 0 else filtered.map(_.alignment).max
```
terpstra: I think this should be 'min', not 'max'. You can't have a transfer larger than the min alignment of any AddressSet.

PR #2135, opened 2019-10-02
Core signal monitor

jerryrzhao:

**Impact**: API addition

Added a new case class `CoreSignalMonitor` to directly access signals in the core which are not available as IOs.

### DIFF
--- a/src/main/scala/util/CoreMonitor.scala
+++ b/src/main/scala/util/CoreMonitor.scala
@@ -8,12 +8,12 @@ import Chisel._
 // this class is used to expose some internal core signals
 // to verification monitors
 case class CoreSignalMonitor(
-  val clock: Clock,
-  val reset: Reset,
-  val xLen: Int,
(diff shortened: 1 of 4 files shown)


src/main/scala/util/CoreMonitor.scala
```diff
@@ -3,17 +3,25 @@
...
+  val clock: Clock,
+  val reset: Reset,
+  val xLen: Int,
+  val cease: Bool,
+  val reg_mscratch: UInt,
+  val hartid: UInt)
 
-// this bundle is used to expose some internal core signals
-// to verification monitors which sample instruction commits
 class CoreMonitorBundle(val xLen: Int) extends Bundle with Clocked {
```
hcook: So do you actually still need this Bundle?

jerryrzhao: You're correct, this is no longer needed.

PR #2143, opened 2019-10-09
AddressAdjuster Improvements

hcook:
This PR makes the AddressAdjuster adapter more flexible in several ways:
 - It now allows the user to specify only a particular address range to be adjusted via the `adjustableRegion` optional parameter. Devices outside of this range can be attached on either the local or the remote side of the downstream ports, and traffic will be routed to them as usual. Only local devices within the `adjustableRegion` have their PMAs edited by the adapter and traffic dynamically toggled between them and the remote side device.
- It now supports flexible cardinality over its edges by defining a new CustomNode type. Specifically, it can now pass through `:=*` flexibility, allowing masters to specify the cardinality.

Also added a missing requirement and limited the scope of some existing requirements to target only the adjustable region.

### DIFF
--- a/src/main/scala/tilelink/AddressAdjuster.scala
+++ b/src/main/scala/tilelink/AddressAdjuster.scala
@@ -9,13 +9,13 @@ import freechips.rocketchip.diplomacy._
 
 // mask=0 -> passthrough
 // forceLocal -> used to ensure special devices (like debug) remain reacheable at chip_id=0
-class AddressAdjuster(mask: BigInt, forceLocal: Seq[AddressSet] = Nil)(implicit p: Parameters) extends LazyModule {
+class AddressAdjuster(mask: BigInt, adjustableRegion: AddressSet = AddressSet.everything, forceLocal: Seq[AddressSet] = Nil)(implicit p: Parameters) extends LazyModule {
   // Which bits are in the mask?
   val bits = AddressSet.enumerateBits(mask)
-  // Which idss must we route within that mask?
+  // Which ids must we route within that mask?
   val ids  = AddressSet.enumerateMask(mask)
   // forceLocal better only go one place (the low index)
-  forceLocal.foreach { as => (as.max & mask) == 0 }
+  forceLocal.foreach { as => require((as.max & mask) == 0) }
 
   val node = TLNexusNode(
     clientFn = { cp => cp(0) },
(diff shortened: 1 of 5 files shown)


src/main/scala/tilelink/AddressAdjuster.scala
```diff
@@ -7,203 +7,315 @@ import chisel3.util._
...
+
+  def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
+    require (iStars == 0, s"$context does not support :*=")
+    if (oStars > 0) {
+      require (2*iKnown >= oKnown, s"$context has $oKnown outputs and $iKnown inputs; cannot assign ${2*iKnown-oKnown} edges to resolve :=*")
+      (0, iKnown - oKnown)
+    } else {
+      require (oKnown == 2*iKnown, s"$context has $oKnown outputs and $iKnown inputs; but 2*$iKnown must be $oKnown")
+      (0, 0)
+    }
```
terpstra: I would be more strict; oStars must be either:
- 0 (your core is correct)
- 2: require oKnown==0 and output is (0, iKnown)

src/main/scala/tilelink/AddressAdjuster.scala
```diff
@@ -7,203 +7,315 @@ import chisel3.util._
...
+// forceLocal -> used to ensure special devices (like debug) remain reacheable at chip_id=0 even if in adjustableRegion
+class AddressAdjuster(mask: BigInt, adjustableRegion: AddressSet = AddressSet.everything, forceLocal: Seq[AddressSet] = Nil)(implicit p: Parameters) extends LazyModule {
   // Which bits are in the mask?
   val bits = AddressSet.enumerateBits(mask)
-  // Which idss must we route within that mask?
+  // Which ids must we route within that mask?
   val ids  = AddressSet.enumerateMask(mask)
+  // Find the intersection of the mask with some region
+  private def masked(region: Seq[AddressSet], offset: BigInt = 0): Seq[AddressSet] = {
+    region.flatMap { _.intersect(AddressSet(0 + offset, ~mask)) }
```
terpstra: + 0?

src/main/scala/tilelink/AddressAdjuster.scala
```diff
@@ -7,203 +7,315 @@ import chisel3.util._
...
-        managers   = newLocal ++ newRemote,
-        endSinkId  = local.endSinkId + remote.endSinkId,
-        minLatency = local.minLatency min remote.minLatency)
+        // Find all the holes in local routing
+        val holes = {
+          val ra = masked(adjustableRemoteManagers.flatMap(_.address))
+          val la = masked(adjustableLocalManagers .flatMap(_.address))
+          la.foldLeft(ra) { case (holes, la) => holes.flatMap(_.subtract(la)) }
+        }
+
```
terpstra: I'm concerned that we don't check that the adjustableRemoteManagers.flatMap completely cover the adjustableRegion. On the other hand, we would like adjustableRegion to be optional.

terpstra: Maybe run address unification on adjustableRemoteManagers.flatMap(_.address) and then require it to be contiguous?

terpstra: Never mind. This already suffices:
          val container = adjustableRemoteManagers.find { r => l.address.forall { la => r.address.exists(_.contains(la)) } }

src/main/scala/tilelink/AddressAdjuster.scala
```diff
@@ -7,203 +7,315 @@ import chisel3.util._
...
+      }
+
+      def containsAddress(region: Seq[AddressSet], addr: UInt): Bool =
+        region.foldLeft(false.B)(_ || _.contains(addr))
+
+      def isAdjustable(addr: UInt) = containsAddress(List(adjustableRegion), addr)
+
+      def isLocal(addr: UInt): Bool =
+        ( isAdjustable(addr) && (local_address === (addr & mask.U) || containsAddress(forceLocal, addr))) ||
+        (!isAdjustable(addr) && containsAddress(localEdge.manager.managers.flatMap(_.address), addr))
```
terpstra: Mux?

terpstra: Probably less hardware to check remoteEdge.manager.mangers (less AddressSets)

src/main/scala/tilelink/AddressAdjuster.scala
hcook: lol

PR #2148, opened 2019-10-14
Filter RegressionTestSuite based on coreParams

seldridge:
This fixes a bug where the `regression-tests` target in `freechips.rocketchip.system.$CONFIG.d` would include tests which the core couldn't support and which were not being added via `TestGeneration.addSuites`. E.g., if you build a `WithoutFPU` configuration and try to `make run-regression-tests-fast` make will error out being unable to find the `output/rv64ud-v-fcvt` test. This fixes this by filtering the possible regression tests based on what the core actually supports.

Note: this solution includes a lot of overlap with both [`isaDTS`](https://github.com/chipsalliance/rocket-chip/blob/77b4e6b84c91aa0617c7e54c7c57c8783bcd30b3/src/main/scala/tile/BaseTile.scala#L74) and [`isaString`](https://github.com/chipsalliance/rocket-chip/blob/77b4e6b84c91aa0617c7e54c7c57c8783bcd30b3/src/main/scala/rocket/CSR.scala#L427)

**Type of change**: bug report | other enhancement

**Impact**: API modification

This removes public APIs from `GeneratorApp`.

### DIFF
--- a/src/main/scala/system/Generator.scala
+++ b/src/main/scala/system/Generator.scala
@@ -11,34 +11,36 @@ import scala.collection.mutable.LinkedHashSet
 /** A Generator for platforms containing Rocket Subsystemes */
 object Generator extends GeneratorApp {
 
-  val rv64RegrTestNames = LinkedHashSet(
-        "rv64ud-v-fcvt",
-        "rv64ud-p-fdiv",
-        "rv64ud-v-fadd",
-        "rv64uf-v-fadd",
-        "rv64um-v-mul",
-        "rv64mi-p-breakpoint",
-        "rv64uc-v-rvc",
-        "rv64ud-v-structural",
-        "rv64si-p-wfi",
-        "rv64um-v-divw",
-        "rv64ua-v-lrsc",
-        "rv64ui-v-fence_i",
-        "rv64ud-v-fcvt_w",
-        "rv64uf-v-fmin",
-        "rv64ui-v-sb",
-        "rv64ua-v-amomax_d",
-        "rv64ud-v-move",
-        "rv64ud-v-fclass",
-        "rv64ua-v-amoand_d",
-        "rv64ua-v-amoxor_d",
-        "rv64si-p-sbreak",
-        "rv64ud-v-fmadd",
-        "rv64uf-v-ldst",
-        "rv64um-v-mulh",
-        "rv64si-p-dirty")
+  override def addTestSuites {
+    import DefaultTestSuites._
+    val xlen = params(XLen)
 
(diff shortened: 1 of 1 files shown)


src/main/scala/system/Generator.scala
```diff
@@ -82,7 +80,26 @@ object Generator extends GeneratorApp {
       TestGeneration.addSuites(rvi.map(_("p")))
       TestGeneration.addSuites((if (vm) List("v") else List()).flatMap(env => rvu.map(_(env))))
       TestGeneration.addSuite(benchmarks)
-      TestGeneration.addSuite(new RegressionTestSuite(if (xlen == 64) rv64RegrTestNames else rv32RegrTestNames))
+
+      /* Filter the regression tests based on what the Rocket Chip configuration supports */
+      val extensions = {
+        val fd = coreParams.fpu.map {
+          case cfg if cfg.fLen >= 64 => "d"
```
hcook: should this be "fd"?

hcook:
We have like 5 implementations of this internally too... Somehow can only add new ones and never kill the old ones. I like your style of `retain` +` case re() =>`

src/main/scala/system/Generator.scala
seldridge: Yeah, this should be... Fixed in the force-push.

seldridge:
Cool. I'm glad you like the approach. FWIW, 2.13 adds the `when` method to the `Option`object, so some of the `if (cond) Some(foo) else None` (which is a pattern I find I'm using a lot) could be cleaned up to: `Option(cond)(foo)`.

hcook:
> `Option.when(cond)(foo)`

wow yes that pattern is used in about ten thousand places!

aswaterman:
Cool, I had borrowed an implicit conversion from scalaz to handle that
pattern (you can do myBool.option(foo) if you import rocketchip util._).

seldridge:
Ah, neat! ([`implicit class BooleanToAugmentedBoolean`](https://github.com/chipsalliance/rocket-chip/blob/9efb17c9f63491ac706d4c8f81f295fd8e57faed/src/main/scala/util/package.scala#L172)) I may continue the borrowing of this into FIRRTL to help [this hot mess](https://github.com/freechipsproject/firrtl/blob/973ecf516c0ef2b222f2eb68dc8b514767db59af/src/main/scala/firrtl/ExecutionOptionsManager.scala#L315)... Or, that might be worth migrating into something common. An `import rocketchip.util._` PR for FIRRTL would be amusing, but a bridge too far.

PR #2149, opened 2019-10-15
TLB update PLRU on allocation

johningalls-sifive:
**Impact**: logic function addition


aswaterman:
Makes sense.  There's a merge conflict because of a change with the handling of the `g` bit

johningalls-sifive:
this PR is no longer necessary after other changes

PR #2153, opened 2019-10-17
Respond NtoN for all Nothing states

seldridge:
This aligns how `shrinkWrapper` handles responses for `toB` when in a `Nothing` state to be the same as responses to `toT` and `toN` when in the same state.

**Impact**: API modification

### DIFF
--- a/src/main/scala/tilelink/Metadata.scala
+++ b/src/main/scala/tilelink/Metadata.scala
@@ -126,7 +126,7 @@ class ClientMetadata extends Bundle {
       Cat(toB, Dirty)   -> (Bool(true),  TtoB, Branch),
       Cat(toB, Trunk)   -> (Bool(false), TtoB, Branch),  // Policy: Don't notify on clean downgrade
       Cat(toB, Branch)  -> (Bool(false), BtoB, Branch),
-      Cat(toB, Nothing) -> (Bool(false), BtoN, Nothing),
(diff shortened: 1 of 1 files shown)


hcook:
Uhh, seems like the error is that the transition from T to N isn't correctly reporting itself? Not sure why that is...

hcook:
Looks like it has been that way for 3 years and no one has noticed because we don't actually care about the previous state value in our extant coherence managers. I suggest making the TtoN case more accurate instead of the BtoN case less accurate.

seldridge:
Not enough info, sorry.

My (limited) understanding of the TileLink specification is that a `ProbeAck` is supposed to report the actual change that occurred in a master agent. From the 1.8.0 spec:

> `c_param` indicates the specific type of permissions change that occurred in the Master Agent as
> a result of the `Probe`. Possible transitions are selected from the **Shrink** or **Report** category of
> Table 8.3. The former indicates that permissions were decreased whereas the latter reports what
> they were and continue to be.

Currently, if a `Probe` comes in requesting a shrink to `toB` on `Nothing`, then the current response is a shrink of `BtoN` indicating that permissions changed. It seems like this should actually be a report of `NToN`. Your suggestion would seem to want to change this definition to report the requested transition and the new state (which I *think* differs from the spec)?

hcook:
Oh, hmm, you're right. I must have misread the code this morning.

PR #2155, opened 2019-10-17
Debug: Make Debug Module optional

ernie-sifive:
**Related issue**: 

**Impact**: API modification

**Release Notes**

Some have asked for a smaller debug solution. This PR makes the Debug Module optional. If there is no DM, some debug is still possible using an M-mode debugger such as GDB Stub. Another use case would be to remove debug for a production chip.


hcook:
@mwachs5 's comment isn't rendering well for me, but I think my feedback is the same:
the `DebugModuleParams` key should really be changed to be of type
```
case object DebugModuleParams extends Field[Option[DebugModuleParams]](Some(DefaultDebugModuleParams())
```

I am guessing you didn't change the type because then every lookup of the key will have to deal with the optionality. But really, they all need to be evaluated for what they should be doing when the new setting is `false/None`. Right now there are an unknown number of cases where `DebugModuleParams.hasDebug = false` is going to result in potentially undefined behavior.

For example, this line in the Rocket TLB is going to always inject the default address comparison:
```
rocket-chip/src/main/scala/rocket/TLB.scala:  val deny_access_to_debug = mpu_priv <= PRV.M && p(DebugModuleParams).address.contains(mpu_physaddr)
```

While changing the type is going to be more work up front, the Scala compiler is helping you out by at least giving you a complaint about every single case that needs to be evaluated for the new semantics. Looking in the closed source code base, I only see about 20 cases that will have to be handled, mostly Config alterations that can be changed formulaically, but also some like the one I listed above.

Finally, while we're breaking things to make them better, can we change the name of the case object `DebugModuleParams`? It is the only parameter in the whole code base that uses the same class name for the key and the value that the key is storing.

src/main/scala/devices/debug/Periphery.scala
```diff
@@ -40,7 +40,7 @@ case object ExportDebug extends Field(DebugAttachParams())
 
 class ClockedAPBBundle(params: APBBundleParameters) extends APBBundle(params) with Clocked
 
-class DebugIO(implicit val p: Parameters) extends ParameterizedBundle()(p) with CanHavePSDTestModeIO {
+class DebugIO(implicit val p: Parameters) extends ParameterizedBundle()(p) {
```
hcook: `ParameterizedBundle` is now unnecessary if all the parameters of the class are `val`s (as is the case here). So you can just extend `Bundle`

src/main/scala/devices/debug/Periphery.scala
```diff
@@ -51,77 +51,89 @@ class DebugIO(implicit val p: Parameters) extends ParameterizedBundle()(p) with
   val disableDebug = p(ExportDebug).externalDisable.option(Bool(INPUT))
 }
 
+class PSDIO(implicit val p: Parameters) extends ParameterizedBundle()(p) with CanHavePSDTestModeIO {
```
hcook: `ParameterizedBundle` is now unnecessary if all the parameters of the class are `val`s (as is the case here). So you can just extend `Bundle`

ernie-sifive: done

ipatix:
So how exactly do I remove the debug module? In one of my current designs (older rocket chip version) I disconnected the debug reset line and used tieoffDebug. However the debug module is still there (I tie its inputs to '0') and I don't really need it on my FPGA design. So I'd happily know how to just remove it using the configuration.

ernie-sifive:
case DebugModuleKey => None

ipatix:
Thank you very much!

PR #2159, opened 2019-10-22
Build AsyncResetSynchronizerShiftReg from new SynchronizerPrimitive

rmac-sifive:

**Release Notes**
`AsyncSynchronizerShiftReg` synchronizes W-bit wide signals with D stages of synchronizers and, previously, was constructed from D instantiations of a W-bit wide register. This PR changes that construction to instantiate W shift registers, each D stages deep, which is more friendly to industry CDC checking and tool flows

This PR also creates the module out of a new `SynchronizerPrimitive`, distinct from but functionally identical to `AsyncResetReg`, which makes it easier to substitute specific synchronizer cells during implementation

### DIFF
--- a/src/main/scala/util/AsyncResetReg.scala
+++ b/src/main/scala/util/AsyncResetReg.scala
@@ -4,6 +4,7 @@ package freechips.rocketchip.util
 
 import Chisel._
 import chisel3.util.HasBlackBoxResource
+import chisel3.experimental.RawModule
 import chisel3.core.IntParam
 
 /** This black-boxes an Async Reset
@@ -44,7 +45,25 @@ class AsyncResetReg(resetValue: Int = 0)
     val rst = Bool(INPUT)
   }
 
-  setResource("/vsrc/AsyncResetReg.v")
+  addResource("/vsrc/AsyncResetReg.v")
+}
+
+class SynchronizerPrimitiveReg(resetValue: Int = 0) extends RawModule {
+  val io = IO(new Bundle {
+    val d = Bool(INPUT)
+    val q = Bool(OUTPUT)
+    val en = Bool(INPUT)
+
+    val clk = Clock(INPUT)
+    val rst = Bool(INPUT)
+  })
+
+  val reg = Module(new AsyncResetReg(resetValue))
+  reg.io.d   := io.d
+  reg.io.en  := io.en
+  reg.io.clk := io.clk
+  reg.io.rst := io.rst
+  io.q := reg.io.q
 }
 
 class SimpleRegIO(val w: Int) extends Bundle{
--- a/src/main/scala/util/ShiftReg.scala
+++ b/src/main/scala/util/ShiftReg.scala
@@ -52,7 +52,6 @@ object AbstractPipelineReg {
 }
 
 class AsyncResetShiftReg(w: Int = 1, depth: Int = 1, init: Int = 0, name: String = "pipe") extends AbstractPipelineReg(w) {
-
   require(depth > 0, "Depth must be greater than 0.")
 
   override def desiredName = s"AsyncResetShiftReg_w${w}_d${depth}_i${init}"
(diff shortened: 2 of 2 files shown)


rmac-sifive:
@jackkoenig understood that the conflict exists, but I'm not sure the answer is to remove `SynchronizerPrimitiveReg`. Implementation tools may want to use a different library cell for the synchronizer register vs a regular async reset register, and the module name will tell the tool which one to use

src/main/scala/util/ShiftReg.scala
```diff
@@ -84,11 +85,41 @@ object AsyncResetShiftReg {
     apply (in, depth, init.litValue.toInt, None)
 }
 
+class SynchronizerPrimitiveShiftRegVec(depth: Int = 1, init: Int = 0) extends AbstractPipelineReg(1) {
```
mwachs5: why is this a `Vec` if it is only 1 bit wide? I think yuo should drop the `Vec` from the name

src/main/scala/util/ShiftReg.scala
```diff
@@ -84,11 +85,41 @@ object AsyncResetShiftReg {
...
+}
+
 // Note that it is important to override "name" in order to ensure that the Chisel dedup does
 // not try to merge instances of this with instances of the superclass.
-class AsyncResetSynchronizerShiftReg(w: Int = 1, sync: Int = 3, init: Int = 0) extends AsyncResetShiftReg(w, depth = sync, init, name = "sync") {
+class AsyncResetSynchronizerShiftReg(w: Int = 1, sync: Int = 3, init: Int = 0) extends AbstractPipelineReg(w) {
   require(sync > 0, "Sync must be greater than 0.")
   override def desiredName = s"AsyncResetSynchronizerShiftReg_w${w}_d${sync}_i${init}"
+
+  val output = Seq.tabulate(w) { i => SynchronizerPrimitiveShiftRegVec(io.d(i), depth = sync, init) }
```
mwachs5: this looks good here... but I think we need to do something similar for the other `SynchronizerShiftReg` variants (or get rid of them)

rmac-sifive: Sure, I changed `SyncResetSynchronizerShiftReg` to be constructed similarly. Basically WxD-sized reset synchronizers are created with W instantiations of D-deep 1-bit wide synchronizer shift regs. The other types are unchanged (constructed from D instances of a W-bit wide reg vec).

src/main/scala/util/ShiftReg.scala
```diff
@@ -84,11 +87,41 @@ object AsyncResetShiftReg {
     apply (in, depth, init.litValue.toInt, None)
 }
 
+class SynchronizerPrimitiveShiftReg(depth: Int = 1, init: Int = 0) extends AbstractPipelineReg(1) {
```
mwachs5: I think if this is primarily intended as a synchronizer primitive the default depth should be 3, or we should just remove the default and require the user to specify it.

src/main/scala/util/ShiftReg.scala
```diff
@@ -136,7 +169,8 @@ class SyncResetSynchronizerShiftReg(w: Int = 1, sync: Int = 3, init: Int = 0) ex
 
   override def desiredName = s"SyncResetSynchronizerShiftReg_w${w}_d${sync}_i${init}"
 
-  io.q := ShiftRegInit(io.d, n = sync, init = init.U, name = Some("sync"))
+  val shiftRegs = List.tabulate(w) { i => ShiftRegInit[UInt](io.d(w), n = sync, init = init.U, name = Some(s"sync_${i}")) }
```
mwachs5: but won't this not use the new Synchronizer primitive you've created? Maybe we should just delete this class entirely. I don't think we really need to make a sync-reset version of the synchronizer primitive. I'm not sure what in the code base is using this version.

The `SynchronizerShiftReg` is useful though, and should also be replaceable internally by your `SynchronizerPrimitive` but maybe should be renamed explicitly `NoResetSynchronizerShiftReg` and would just be a wrapper around the `AsyncResetSynchronizerShiftReg` that does a `withReset(false.B)`

src/main/scala/util/AsyncResetReg.scala
```diff
@@ -48,6 +48,24 @@ class AsyncResetReg(resetValue: Int = 0) extends RawModule {
   io.q := reg
 }
 
+class SynchronizerPrimitiveReg(resetValue: Int = 0) extends RawModule {
```
mwachs5: Let's just get rid of this

src/main/scala/util/ShiftReg.scala
mwachs5: Let's just build this with regular flops. And make sure it gets the proper inferred reset type.

mwachs5:
@rmac-sifive this is superceded by https://github.com/chipsalliance/rocket-chip/pull/2212, so i am going to close this one.

PR #2162, opened 2019-10-24
SBA: Fix invalid sbaccess value error case, remove spurious assert

rmac-sifive:

**Impact**: API modification

**Release Notes**

This PR fixes the error handling case where programming the Debug Module's sbaccess register to a value > 4.U would not result in the correct error handling and setting of the sberror register. Also removes a spurious assertion related to the queue size on the SBA unit's TileLink response interface

### DIFF
--- a/src/main/scala/devices/debug/SBA.scala
+++ b/src/main/scala/devices/debug/SBA.scala
@@ -180,7 +180,7 @@ object SystemBusAccessModule
                         (SBCSFieldsReg.sbaccess === 1.U) && (SBCSFieldsReg.sbaccess16  =/= 1.U) ||
                         (SBCSFieldsReg.sbaccess === 2.U) && (SBCSFieldsReg.sbaccess32  =/= 1.U) ||
                         (SBCSFieldsReg.sbaccess === 3.U) && (SBCSFieldsReg.sbaccess64  =/= 1.U) ||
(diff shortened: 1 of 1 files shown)


hcook:
Tangentially, is the SBA state machine able to deal with getting a response back on the same cycle a request is sent (d.valid and d.ready externally coupled to a.valid and a.ready`)?

If not, you need to set `minLatency = 1` in your TLClientPortParameters.

rmac-sifive:
@hcook yes, see: https://github.com/chipsalliance/rocket-chip/blob/4f0cdea85c8a2b849fd582ccc8497892001d06b0/src/main/scala/devices/debug/SBA.scala#L291

hcook:
@andrew odds of headaches are moderate

src/main/scala/devices/debug/SBA.scala
```diff
@@ -355,7 +355,6 @@ class SBToTL(implicit p: Parameters) extends LazyModule {
     tl.c.valid := false.B
     tl.e.valid := false.B
 
-    assert (!tl.d.valid || tl.d.ready, "Debug module not ready to accept TL response") // assert tl.d.valid |-> tl.d.ready
```
mwachs5: if we are going to delete this assertion can we please add a cover property instead to show that we do handle it correctly.

hcook:
Seems like the build passed but got stuck in some Travis/Github hiccup. Restarting to enable the merge in the Github UI.

PR #2176, opened 2019-11-02
Diplomatic IO Creation from Nodes

hcook:
This encapsulates the popular pattern of making IOs out of the bundles associated with SourceNodes or SinkNodes via cloning them into `HeterogeneousBags`. It is a generalization of the behavior of `makeIO()` on BundleBridge sources and sinks, and as such I named it `makeIOs()` and reimplemented the bridge methods using it. Should be totally backward compatible.

I also threw in some TL-specific utilities for printing out port metadata or monitoring port behavior.

Follow-on PR will convert the ports to use this capability.

**Related Issues**: Someone asked for this but I'm having trouble finding the issue. 

**Impact**:API addition (no impact on existing code)

**Release Notes**

Diplomacy: Source and Sink Nodes have `makeIOs` method

### DIFF
--- a/src/main/scala/diplomacy/BundleBridge.scala
+++ b/src/main/scala/diplomacy/BundleBridge.scala
@@ -22,24 +22,14 @@ case class BundleBridgeSink[T <: Data]()(implicit valName: ValName) extends Sink
 {
   def bundle: T = in(0)._1
 
-  def makeIO()(implicit valName: ValName): T = {
-    val io = IO(bundle.cloneType)
-    io.suggestName(valName.name)
-    io <> bundle
-    io
-  }
+  def makeIO()(implicit valName: ValName): T = makeIOs()(valName).head
 }
 
 case class BundleBridgeSource[T <: Data](gen: () => T)(implicit valName: ValName) extends SourceNode(new BundleBridgeImp[T])(Seq(BundleBridgeParams(gen)))
 {
   def bundle: T = out(0)._1
 
-  def makeIO()(implicit valName: ValName): T = {
-    val io = IO(Flipped(bundle.cloneType))
-    io.suggestName(valName.name)
-    bundle <> io
-    io
-  }
+  def makeIO()(implicit valName: ValName): T = makeIOs()(valName).head
 
   private var doneSink = false
   def makeSink()(implicit p: Parameters) = {
--- a/src/main/scala/diplomacy/Nodes.scala
+++ b/src/main/scala/diplomacy/Nodes.scala
@@ -3,6 +3,7 @@
 package freechips.rocketchip.diplomacy
 
 import Chisel._
+import chisel3.experimental.IO
 import chisel3.internal.sourceinfo.SourceInfo
 import freechips.rocketchip.config.{Parameters,Field}
 import freechips.rocketchip.util.HeterogeneousBag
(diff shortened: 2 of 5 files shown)


jerryhethatday:
@hcook  

> This encapsulates the popular pattern of making IOs out of the bundles associated with SourceNodes or SinkNodes via cloning them into HeterogeneousBags
I wonder what that pattern you mentioned is specificly? does that pattern have a name so that I can google it. IMHO, the makeIOs  method in the source and sink node is just creatingIOs for a source node or sink node with the coresponding bundles resolved by the diplomacy, I thought it was just help method, instead of a popular pattern. Also, why the makeIOs only exists for the source and sink node, the other type of diplomacy nodes also have IO interconnects, for a NexusNode, maybe we should have some API like `makeInputs` and `makeOutputs`, I feel confused why it just exists for the source and sink node, can you clarify this? Thanks!

hcook:
Source and Sink Nodes are were diplomacy negotiation ends. Any further extensions of the network can only have a fixed parameterization. Sometimes you want to extend the network by making a non-diplomatic, fixed IO in the parent module of a source or sink node, for example, a top-level port that match the values calculated for the node's edge. But other times you might want to connect it to a blackbox or some internal module with a fixed Chisel interface. So `makeIOs` is for the first use case, but not the second one, and by "popular" I just meant that there are often situations where the first use case occurs.

All other node types have their IOs created automatically through all module boundaries that the edge between nodes passes through. And source and sink nodes will have internal IOs created for their one inward (sink) or outward (source) edge, too. `makeIOs` affects the parent LazyModule scope and makes an additional IO there.

jerryhethatday:
> Source and Sink Nodes are were diplomacy negotiation ends. Any further extensions of the network can only have a fixed parameterization. Sometimes you want to extend the network by making a non-diplomatic, fixed IO in the parent module of a source or sink node, for example, a top-level port that match the values calculated for the node's edge. But other times you might want to connect it to a blackbox or some internal module with a fixed Chisel interface. So `makeIOs` is for the first use case, but not the second one, and by "popular" I just meant that there are often situations where the first use case occurs.
> 
> All other node types have their IOs created automatically through all module boundaries that the edge between nodes passes through. And source and sink nodes will have internal IOs created for their one inward (sink) or outward (source) edge, too. `makeIOs` affects the parent LazyModule scope and makes an additional IO there.

Sorry about the delayed reply, I got the general idea after lots of code base searching. I feel like the reason you guys introduced makeIOs for original source and sink node is because  you want to  plumbing IO from deep inside the hierarchy to
the top of the design for unit testing, etc. You firstly created a specific node type for this functionality(`BundleBridgeSink and BundleBridgeSource`) and then you guys wanted to generailize it.  I wonder what is the reason of still keeping the BundleBrideg stuff? To make it backward compatible?? Thanks for your clarification

andguilty:
I have a question why the sourcenode’s makeIOs flip the bundleout while sinknode don't flip？

jerryhethatday:
> I have a question why the sourcenode’s makeIOs flip the bundleout while sinknode don't flip？

@ddxoxpp  I posted an answer here: https://stackoverflow.com/questions/66963837/how-to-understand-the-flip-in-autobundle-and-in-makeios/67531121#67531121

PR #2178, opened 2019-11-02
Modernize Port Creation

hcook:
This builds on #2176 by using the new `makeIOs` methods to create ports in the subsystem. This 
 new methods allows the two traits used previously to be collapsed into one, which then results in some API tweaks  to how collateral is attached in the testharness.

The following traits are removed:
- `CanHaveMasterAXI4MemPortModuleImp`
- `CanHaveMasterAXI4MMIOPortModuleImp`
- `CanHaveSlaveAXI4PortModuleImp`
- `CanHaveMasterTLMMIOPortModuleImp`
- `CanHaveSlaveTLPortModuleImp`

The following APIs are changed:
- `class SimAXIMem` is moved to the `system` package
- `object SimAXIMem` has methods `connectMMIO` and `connectMem` which are used in the test harnesses by passing them the dut instead of calling them as methods of the dut.

The following soon-to-be-deprecated APIs are avoided:
- `TLBusWrapper.coupleTo` and `TLBusWrapper.coupleFrom` are used instead of the specific child class coupling methods `MemoryBus.toDRAMController` `SystemBus.toFixedWidthPort` `FrontBus.fromPort`

**Impact**: API modification

**Release Notes**

### DIFF
--- a/src/main/scala/system/SimAXIMem.scala
+++ b/src/main/scala/system/SimAXIMem.scala
@@ -1,0 +1,35 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.system // TODO this should really be in a testharness package
+
+import freechips.rocketchip.amba.axi4._
+import freechips.rocketchip.config.{Parameters}
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.subsystem.{CanHaveMasterAXI4MMIOPort, CanHaveMasterAXI4MemPort, ExtMem}
+
+/** Memory with AXI port for use in elaboratable test harnesses. */
+class SimAXIMem(edge: AXI4EdgeParameters, size: BigInt)(implicit p: Parameters) extends SimpleLazyModule {
+  val node = AXI4MasterNode(List(edge.master))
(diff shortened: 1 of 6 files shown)


src/main/scala/system/SimAXIMem.scala
```diff
@@ -0,0 +1,35 @@
...
+  srams.foreach{ s => s.node := AXI4Buffer() := AXI4Fragmenter() := xbar }
+  xbar := node
+  val io_axi4 = InModuleBody { node.makeIOs() }
+}
+
+object SimAXIMem {
+  def connectMMIO(dut: CanHaveMasterAXI4MMIOPort)(implicit p: Parameters): Unit = {
+    dut.mmio_axi4.zip(dut.mmioAXI4Node.in).foreach { case (io, (_, edge)) =>
+      // test harness size capped to 4KB (ignoring p(ExtMem).get.master.size)
+      val mmio_mem = LazyModule(new SimAXIMem(edge, size = 4096))
```
davidbiancolin: I don't know how you feel about this, but I think it might be useful if you returned the SimAxiMem instances

hcook: sgtm

Issue #2185, opened 2019-11-06
ExampleRocketSystem mem_axi -- parameterized data width?


gsomlo:
I'm using Rocket as part of a LiteX (https://github.com/enjoy-digital/litex) SoC, which, depending on the physical FPGA dev board in use, has a DRAM controller whose axi-port data width may be either 64, 128, or 256 bits.

Couple of questions:
1. If I force Rocket's mem_axi data width to 128 or 256 by explicitly setting beatBytes, am I right to expect shorter and wider bursts when cache lines (which I _think_ are 512 bits) are fetched or flushed to main RAM? Would this qualify as an unquestionably "good thing"? (Intuitively, adding external-to-Rocket RTL to convert between mem_axi's default 64bit data width to something wider on the DRAM controller side should only add waste and complexity, but I'd like a sanity check on that intuition...
2. Would it make sense to parameterize the ExtMem port beatBytes, like in this patch:
```
diff --git a/src/main/scala/subsystem/Configs.scala b/src/main/scala/subsystem/C
onfigs.scala
index ceb29e48..cc56c6c4 100644
--- a/src/main/scala/subsystem/Configs.scala
+++ b/src/main/scala/subsystem/Configs.scala
@@ -348,6 +348,15 @@ class WithDefaultMemPort extends Config((site, here, up) =>
 {
                       idBits = 4), 1))
 })
 
+class WithLitexMemPort(dw: Int) extends Config((site, here, up) => {
+  case ExtMem => Some(MemoryPortParams(MasterPortParams(
+                      base = x"8000_0000",
+                      size = x"8000_0000",
+                      // FIXME: ensure 'dw' is a power of 2 (e.g., 1, 2, 4)
+                      beatBytes = site(MemoryBusKey).beatBytes * dw,
+                      idBits = 4), 1))
+})
+
 class WithNoMemPort extends Config((site, here, up) => {
   case ExtMem => None
 })
...
diff --git a/src/main/scala/system/Configs.scala b/src/main/scala/system/Configs.scala
index 2c81ddba..47e23826 100644
--- a/src/main/scala/system/Configs.scala
+++ b/src/main/scala/system/Configs.scala
@@ -74,3 +74,28 @@ class MMIOPortOnlyConfig extends Config(
 
 class BaseFPGAConfig extends Config(new BaseConfig)
 class DefaultFPGAConfig extends Config(new WithNSmallCores(1) ++ new BaseFPGAConfig)
+
[...]

PR #2187, opened 2019-11-11
config: WithDefaultMMIOPort should use SystemBusKey not MemoryBusKey

hcook:
Per #2185 the example configuration of WithDefaultMMIOPort is not very sensible. The example MMIO port is attached to the SystemBus, not the MemoryBus, and should be configured accordingly.

**Related issue**: 

**Impact**:  API modification

### DIFF
--- a/src/main/scala/subsystem/Configs.scala
+++ b/src/main/scala/subsystem/Configs.scala
@@ -356,7 +356,7 @@ class WithDefaultMMIOPort extends Config((site, here, up) => {
   case ExtBus => Some(MasterPortParams(
                       base = x"6000_0000",
                       size = x"2000_0000",
-                      beatBytes = site(MemoryBusKey).beatBytes,
+                      beatBytes = site(SystemBusKey).beatBytes,
                       idBits = 4))
 })
 


gsomlo:
ping -- I'd be curious to know if this ends up being the way forward, so I may reference SystemBus (rather than MemoryBus) from my own alternative WithXyzMMIOPort configurations. Also, I'd like to know whether `WithEdgeDataBits` is or isn't meant to modify the MMIO port width alongside the Memory and (optional) Slave ports (personally I hope it is **not**, but would like to know for future reference)!
Thanks!

hcook:
@gsomlo The SystemBus vs MemoryBus thing is definitely more correct. I too am not sure about the intended behavior of `WithEdgeDataBits`; personally I do not use it.

@zhemao @colinschmidt Do y'all have opinions about whether the `WithEdgeDataBits` Config alteration is supposed to change the data width of all subsystem ports or only some of them?

gsomlo:
On Tue, Dec 03, 2019 at 05:46:32PM -0800, Henry Cook wrote:
> @gsomlo The SystemBus vs MemoryBus thing is definitely more correct.

Thanks, that's good to know!
 
> @zhemao @colinschmidt Do y'all have opinions about whether the
> WithEdgeDataBits Config alteration is supposed to change the
> data width of all subsystem ports or only some of them?

If it's supposed to modify the MMIO in addition to Mem and Slave, then
it needs to be updated as part of (or following) this patch (and I
won't be able to use it for my project :)

If it is NOT supposed to also touch MMIO:
1. it's been wrong all this time, and this patch will fix it
2. it doesn't need to be updated following this patch
3. I can keep using it.

Either way, a clarification would be good all around -- TIA!

gsomlo:
Ping -- Any fresh insights into this question? Thanks!

colinschmidt:
I believe that `WithEdgeDataBits` is intended to change only those subsystem ports that leave the system (i.e. external memory interfaces, and such).
I don't find myself using it often. Rather I just change the buses that I want to reconfigure explicitly.
In this definition `WithEdgeDataBits` would need to be updated to also modify `ExtBus` not just `ExtIn` and `MemoryBus`

PR #2191, opened 2019-11-13
Add optional trace stall function

ernie-sifive:
**Related issue**: 

**Release Notes**

The consumer of the TracedInstruction bundle from BaseTile.traceNode may not be able to keep up with the rate at which traced instructions appear on the interface.  This PR adds a traceStall signal in a return bundle that stalls the core when asserted, allowing the external trace unit to catch up.  If no trace consumers connect to traceNode, then traceStall is driven with false.B.


aswaterman:
What's the CDC story...  I assume like the debug interrupt this will go through a simple synchronizer?

ernie-sifive:
The stall signal must be driven synchronously to the core clock.  The clock is supplied to the external trace logic in the TracedInstruction bundle.

aswaterman:
Ack.

PR #2192, opened 2019-11-15
Monitorformaladditions

brrmorre:
**Related issue**: 

**Impact**: API modification

**Release Notes**
This adds utilities that can be used for formal.

### DIFF
--- a/src/main/scala/formal/FormalUtils.scala
+++ b/src/main/scala/formal/FormalUtils.scala
@@ -53,41 +53,28 @@ object Property {
     }
     val proposed_src = prop_type.toString + "_" + line_info
 
-//    var max_rand = 15
-//    val random_src = scala.util.Random
-//    while (!prop_name_set.add(proposed_src)) {
-//      proposed_src = src + "_" + "%x".format(random_src.nextInt(max_rand))
-//      max_rand = max_rand * 2
-//    }
     val src_wrap = s"@[${proposed_src}]"
     if (dir==MonitorDirection.Monitor) {
-//      assert(cond, s"Assert: ${prop_type.toString} ${message} ${src_wrap}")
       when(!cond) {
         printf(s"assert:${proposed_src}:${prop_type.toString} ${message}")
       }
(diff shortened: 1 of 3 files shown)


src/main/scala/formal/FormalUtils.scala
```diff
@@ -0,0 +1,247 @@
...
+      case _ => ""
+    }
+    val proposed_src = prop_type.toString + "_" + line_info
+
+//    var max_rand = 15
+//    val random_src = scala.util.Random
+//    while (!prop_name_set.add(proposed_src)) {
+//      proposed_src = src + "_" + "%x".format(random_src.nextInt(max_rand))
+//      max_rand = max_rand * 2
+//    }
```
hcook: Can you delete this commented out code? (FYI using `scala.util.Random` means your tests wont be reproducible from a known seed).

src/main/scala/formal/FormalUtils.scala
```diff
@@ -0,0 +1,247 @@
...
+            prop_type: PropertyClass)(implicit sourceInfo: SourceInfo): Unit = {
+    Property(dir, cond, crosscond, message, prop_type, "")
+  }
+  def apply(cond: Bool)(implicit sourceInfo: SourceInfo): Unit ={
+    Property(MonitorDirection.Monitor, cond, "Sanity Property", PropertyClass.LocalRTL, "")
+  }
+}
+
+object SourceGet {
+  def get_line_num(implicit sourceInfo: SourceInfo): String = {
```
hcook: camelCase is the preferred style for scala methods (as opposed to chisel wires).

brrmorre: I'm moving a file from federation to rocket chip, this cannot be changed without changing the corresponding calls (in federation).

brrmorre:
> Can you please fill out the PR message and give it a nice title. This is a public repository.
> 
> Will all the converted monitor conditions still fire in simulation as they would have when they were assertions?

The default behavior will not change. The assume calls default to normal asserts. They only change if a default parameter is changed.

PR #2194, opened 2019-11-15
Allow hartID to be any value in single-core system

ernie-sifive:
**Related issue**: 

**Release Notes**

Currently, a Scala function is used to map hartID to hart Index and vice-versa.  The hartID is written by ROM code to DM notification addresses so that the DM knows which hart is notifying it.  Being a Scala function, that means that hartID must be a constant known at compile time.  In a single core system, we already know what hart it is so don't need that information and hartID need not be a constant.

Conversely, the DM notifies a hart to request it to Go or Resume using one of the 1024 flags notification locations.  Each hart polls the flags location corresponding to its hartID.  In a single core system, there is one flags "register" that is duplicated across all 1024 locations so that the hart will see it regardless of which flags location it reads.


src/main/scala/devices/debug/Debug.scala
```diff
@@ -688,6 +687,11 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
     val nHaltGroups = if ((nComponents > 1) | (nExtTriggers > 0)) cfg.nHaltGroups
       else 0  // no halt groups possible if single hart with no external triggers
 
+    val hartSelFuncs = if (getNComponents() > 1) p(DebugModuleHartSelKey) else DebugModuleHartSelFuncs(
+      hartIdToHartSel = (x) => UInt(0),
+      hartSelToHartId = (x) => x
```
mwachs5: wouldnt this also need to be 0?

ernie-sifive: No, that would have no effect.  This code only applies to a single-core system, so hartSel is always 0.  0 => 0 in either case.

mwachs5: ok makes sense

PR #2197, opened 2019-11-18
simplify className API calling

sequencer:

```
  private def findClassName(c: Class[_]): String = {
    val n = c.getName.split('.').last
    if (n.contains('$')) findClassName(c.getSuperclass) else n
  }
```
is equal to `getSimpleName`

### DIFF
--- a/src/main/scala/diplomacy/LazyModule.scala
+++ b/src/main/scala/diplomacy/LazyModule.scala
@@ -35,12 +35,7 @@ abstract class LazyModule()(implicit val p: Parameters)
     this
   }
 
-  private def findClassName(c: Class[_]): String = {
-    val n = c.getName.split('.').last
-    if (n.contains('$')) findClassName(c.getSuperclass) else n
-  }
-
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -35,12 +35,7 @@ abstract class LazyModule()(implicit val p: Parameters)
     this
   }
 
-  private def findClassName(c: Class[_]): String = {
-    val n = c.getName.split('.').last
-    if (n.contains('$')) findClassName(c.getSuperclass) else n
-  }
-
-  lazy val className = findClassName(getClass)
+  lazy val className = getClass.getSimpleName
```
jackkoenig: `getSimpleName` has issues on Java 8 to the point that we're removing it from [FIRRTL](https://github.com/freechipsproject/firrtl/pull/1228). Furthermore, the check for `$` is a nice filter for anonymous classes where you'd rather have the superclass name than the name of the anonymous class.

Issue #1160, opened 2017-12-18
FIRRTL Bump breaks code in JTAG Tap


mwachs5:
The currently submoduled version of Chisel/FIRRTL breaks the JTAG TAP controller (which is ~~one of the few~~ the only modules in the codebase written in chisel3._ . It blows away the `irReg` due to the way it interprets something in the JtagTapController and/or DebugTransportModuleJTAG.

In the "pre-bump" Chisel, there is a line in `DebugTransportModuleJTAG` FIRRTL output:
```
JtagTapController.io is invalid
```

That line is not present in the new version. It's not SUPPOSED to have to be present. But it seems to blow things away as a result in the output Verilog. And there is no complaint. Will file a relevant bug on FIRRTL as well. If this line is manually added to the FIRRTL, then things work as expected.

Note that reverting to `Chisel._` works as a fix.

FYI @ducky64

mwachs5:
For rocket-chip this issue is fixed by using `Chisel._` instead of `chisel3._`. Since the issue is already opened in FIRRTL, I am going to close this one since it is not actually affecting rocket-chip.

mwachs5:
@jackkoenig is there reason to believe that this issue was resolved in FIRRTL so that it is no longer a problem when converting the code to chisel3?

jackkoenig:
Yes this has been fixed since https://github.com/chipsalliance/rocket-chip/commit/6bd4103c06d728707af8d4a2d7443f59b97f9228, but now with Async Reset we can both convert to `import chisel3._` and use the native async reset support!

PR #2203, opened 2019-11-21
Convert DM and JTAG to chisel3

ernie-sifive:
**Related issue**: 

**Release Notes**

This PR converts the Debug Module and JTAG unit to chisel3. This is to enable them to use the new Chisel3 reset functionality.

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -3,8 +3,8 @@
 package freechips.rocketchip.devices.debug
 
 
-import Chisel._
-import chisel3.experimental._
+import chisel3._
+import chisel3.util._
 import freechips.rocketchip.config._
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.regmapper._
@@ -203,9 +203,9 @@ class DebugInternalBundle (val nComponents: Int)(implicit val p: Parameters) ext
  */
 
 class DebugCtrlBundle (nComponents: Int)(implicit val p: Parameters) extends ParameterizedBundle()(p) {
-  val debugUnavail    = Vec(nComponents, Bool()).asInput
-  val ndreset         = Bool(OUTPUT)
-  val dmactive        = Bool(OUTPUT)
+  val debugUnavail    = Input(Vec(nComponents, Bool()))
+  val ndreset         = Output(Bool())
+  val dmactive        = Output(Bool())
 }
 
 // *****************************************
@@ -240,10 +240,10 @@ class DebugCtrlBundle (nComponents: Int)(implicit val p: Parameters) extends Par
 // Local reg mapper function : Notify when written, but give the value as well.  
 object WNotifyWire {
   def apply(n: Int, value: UInt, set: Bool, name: String, desc: String) : RegField = {
-    RegField(n, UInt(0), RegWriteFn((valid, data) => {
+    RegField(n, 0.U, RegWriteFn((valid, data) => {
       set := valid
       value := data
-      Bool(true)
+      true.B
     }), Some(RegFieldDesc(name = name, desc = desc,
       access = RegFieldAccessType.W)))
   }
@@ -253,11 +253,11 @@ object WNotifyWire {
 object RWNotify {
     def apply (n: Int, rVal: UInt, wVal: UInt, rNotify: Bool, wNotify: Bool, desc: Option[RegFieldDesc] = None): RegField = {
       RegField(n,
-        RegReadFn ((ready)       => {rNotify := ready ; (Bool(true), rVal)}),
+        RegReadFn ((ready)       => {rNotify := ready ; (true.B, rVal)}),
         RegWriteFn((valid, data) => {
           wNotify := valid
           when (valid) {wVal := data}
-          Bool(true)
+          true.B
         }
         ), desc)
     }
@@ -308,7 +308,7 @@ class TLDebugModuleOuter(device: Device)(implicit p: Parameters) extends LazyMod
     val io = IO(new Bundle {
       val ctrl = (new DebugCtrlBundle(nComponents))
       val innerCtrl = new DecoupledIO(new DebugInternalBundle(nComponents))
-      val hgDebugInt = Vec(nComponents, Bool()).asInput
+      val hgDebugInt = Input(Vec(nComponents, Bool()))
       val hartResetReq = cfg.hasHartResets.option(Output(Vec(nComponents, Bool())))
       val dmAuthenticated = cfg.hasAuthentication.option(Input(Bool()))
     })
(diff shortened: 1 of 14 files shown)


src/main/scala/devices/debug/Debug.scala
```diff
@@ -1517,10 +1517,10 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
     import CtrlState._
 
     // This is not an initialization!
-    val ctrlStateReg = Reg(CtrlState(Waiting))
+    val ctrlStateReg = RegInit(CtrlState(Waiting))
```
mwachs5: if we are turning this into a `RegInit` shouldn't we get rid of the comment above? This seems to be doing something more than the conversion to chisel3?

src/main/scala/devices/debug/DebugTransport.scala
```diff
@@ -63,63 +64,62 @@ class DTMInfo extends Bundle {
...
+  val stickyNonzeroRespReg = RegInit(false.B)
 
-  val downgradeOpReg = Reg(init = Bool(false)) // downgrade op because prev. failed.
+  val downgradeOpReg = RegInit(false.B) // downgrade op because prev. failed.
 
   val busy = Wire(Bool())
   val nonzeroResp = Wire(Bool())
 
   val busyResp    = Wire(new DMIAccessCapture(debugAddrBits))
-  val nonbusyResp = Wire(new DMIAccessCapture(debugAddrBits))
```
mwachs5: did we just get rid of this because it is unused?

src/main/scala/devices/debug/SBA.scala
```diff
@@ -100,13 +101,13 @@ object SystemBusAccessModule
     val hasSBAddr3 = (sb2tl.module.edge.bundle.addressBits >= 97)
     val hasAddr    = Seq(true, hasSBAddr1, hasSBAddr2, hasSBAddr3)
 
-    val SBADDRESSFieldsReg = Seq.fill(4)(Reg (UInt(32.W)))
+    val SBADDRESSFieldsReg = RegInit(VecInit(Seq.fill(4) {0.U(32.W)} ))
```
mwachs5: this is another place where it seems like we are adding functionality, not just changing to chisel3. Is the change to RegInit intentional?

src/main/scala/devices/debug/SBA.scala
```diff
@@ -138,14 +139,14 @@ object SystemBusAccessModule
     val hasSBData2And3 = (cfg.maxSupportedSBAccess == 128)
     val hasData        = Seq(true, hasSBData1, hasSBData2And3, hasSBData2And3)
 
-    val SBDATAFieldsReg = Seq.fill(4)(Seq.fill(4)(Reg (UInt(8.W))))
+    val SBDATAFieldsReg = RegInit(VecInit(Seq.fill(4) {VecInit(Seq.fill(4) {0.U(8.W)} )} ))
```
mwachs5: same here

src/main/scala/devices/debug/dm_registers.scala
```diff
@@ -1,6 +1,6 @@
 package freechips.rocketchip.devices.debug
 
-import Chisel._
+import chisel3._
 
 // This file was auto-generated from the repository at https://github.com/riscv/riscv-debug-spec.git,
```
mwachs5: we should fix the source script (if it is still used!) I'll make a PR for that.

mwachs5: https://github.com/riscv/riscv-debug-spec/pull/502

src/main/scala/jtag/JtagTap.scala
```diff
@@ -119,8 +114,8 @@ class JtagTapController(irLength: Int, initialInstruction: BigInt)(implicit val
     nextActiveInstruction := irChain.io.update.bits
     updateInstruction := true.B
   } .otherwise {
-    //!!! Needed when using chisel3._ (See #1160)
-    // nextActiveInstruction := DontCare
+    // Needed when using chisel3._ (See #1160)
```
mwachs5: is #1160 actually still open?

mwachs5: Can we try without this line at all since presumably this should be fixed in FIRRTL now... (@jackkoenig any reason to believe it is still needed)

mwachs5: Or, is this "proper" chisel3?

src/main/scala/jtag/JtagTap.scala
```diff
@@ -232,6 +227,8 @@ object JtagTapGenerator {
       }
     }
 
+    controllerInternal.io.dataChainIn := bypassChain.io.chainOut  // default
```
mwachs5: why do we need this now?

src/main/scala/devices/debug/DebugTransport.scala
ernie-sifive: yes - it complained about nonbusyResp not being fully initialized. True, since it was never assigned.

src/main/scala/devices/debug/SBA.scala
ernie-sifive: Yes, FIRRTL complained about these being not fully initialized.  The code below this only assigns the portion of the declared Reg that is actually used in this config, leaving the rest of it unassigned, resulting in that error.

src/main/scala/devices/debug/dm_registers.scala
ernie-sifive: thanks, I hadn't gotten to that yet :-)

src/main/scala/jtag/JtagTap.scala
ernie-sifive: Without this line, FIRRTL complains about nextActiveInstruction being not fully initialized -- meaning that in some cases the Wire is never assigned a value.  You need either this or to initialize it with WireInit, for example, to ensure that the Wire is always assigned to something.

src/main/scala/jtag/JtagTap.scala
ernie-sifive: Same thing - FIRRTL reports "not fully initialized" -- this ensures there is a value.

src/main/scala/devices/debug/Debug.scala
ernie-sifive: I couldn't figure out how to get the type needed for Reg( ).  I have now and changed it back to the original functionality, i.e. no initialization.

src/main/scala/jtag/JtagTap.scala
ernie-sifive: Same thing about "not fully initialized" due to not being assigned in some cases.  This ensures there is a value assigned in all cases.

src/main/scala/devices/debug/SBA.scala
mwachs5: So are we Ok with this being reset by the module reset? This is a change in functionality.

src/main/scala/devices/debug/SBA.scala
mwachs5: So are we Ok with this being reset by the module reset? This is a change in functionality.

src/main/scala/devices/debug/Debug.scala
mwachs5: ok sounds good

src/main/scala/devices/debug/SBA.scala
ernie-sifive: You're right, this did change functionality.  I changed to a different method that should maintain existing function.

src/main/scala/devices/debug/SBA.scala
ernie-sifive: You're right, this did change functionality.  I changed to a different method that should maintain existing function.

PR #2210, opened 2019-11-27
Addition of RV32E and RV64E to rocket core object json

mhtwn:
**Related issue**: 

RV32E and RV64E feature is added to object json.

**Release Notes**

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/model/OMISA.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMISA.scala
@@ -63,7 +63,7 @@ object OMISA {
     val baseISAVersion = baseInstructionSet match {
       case RV32E => "1.9"
       case RV32I => "2.0"
-      case RV64E => "2.0"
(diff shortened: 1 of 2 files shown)


aswaterman:
While you are at it, can you please create an `RV64E` case object and do the same thing for RV64E/RV64I?

mhtwn:
> While you are at it, can you please create an `RV64E` case object and do the same thing for RV64E/RV64I?

Sure!

src/main/scala/diplomaticobjectmodel/model/OMISA.scala
```diff
@@ -62,6 +63,7 @@ object OMISA {
     val baseISAVersion = baseInstructionSet match {
       case RV32E => "1.9"
       case RV32I => "2.0"
+      case RV64E => "2.0"
```
aswaterman: I think this should also be version 1.9 for the time being.

PR #2212, opened 2019-11-30
Synchronizer Primitive Updates

mwachs5:
**Related issue**: 

This builds on top of #2159 

**Impact**: API modification

**Release Notes**

* Change to the internals of `*SynchronizerShiftReg` primitives to build upon an easily replaceable (by integrators) metastable-hardened synchronizer cell, `SynchronizerPrimitiveShiftReg`
* Backwards-incompatible change that `*SynchronizerShiftReg` must now have synchronizer depth > 1
* Introduction of `ClockCrossingReg` for places where `*SynchronizerShiftReg` were being used to indicate a clock crossing but where depth was == 1.
* Changes to `AsyncCrossingSource` and `AsyncCrossingSink` in light of the above changes which have functional (additional delay out of reset) changes
* Deprecation of `IntSyncCrossingSink` API which used the number of synchronizer stages to distinguish between Async, Rational, and Sync crossing types. Instead, use specific IntSync*CrossingSink types to distinguish the different types.
* Deprecation of `SyncResetSynchronizerShiftReg` as it is better inferred from the reset type. Use `ResetSynchronizerShiftReg` instead.

### DIFF
--- a/src/main/scala/interrupts/Crossing.scala
+++ b/src/main/scala/interrupts/Crossing.scala
@@ -45,6 +45,17 @@ class IntSyncCrossingSource(alreadyRegistered: Boolean = false)(implicit p: Para
 }
 
 
+object IntSyncCrossingSink
+{
+  @deprecated("IntSyncCrossingSink which used the `sync` parameter to determine crossing type is deprecated. Use IntSyncAsyncCrossingSink, IntSyncRationalCrossingSink, or IntSyncSyncCrossingSink instead for > 1, 1, and 0 sync values respectively", "rocket-chip 1.2")
+  def apply(sync: Int = 3)(implicit p: Parameters) =
+  {
+    val intsink = LazyModule(new IntSyncAsyncCrossingSink(sync))
+    intsink.node
+  }
+}
+
+
 class IntSyncAsyncCrossingSink(sync: Int = 3)(implicit p: Parameters) extends LazyModule
 {
   val node = IntSyncSinkNode(sync)
--- a/src/main/scala/interrupts/Crossing.scala
+++ b/src/main/scala/interrupts/Crossing.scala
@@ -45,7 +45,7 @@ class IntSyncCrossingSource(alreadyRegistered: Boolean = false)(implicit p: Para
 }
 
 
-class IntSyncCrossingSink(sync: Int = 3)(implicit p: Parameters) extends LazyModule
+class IntSyncAsyncCrossingSink(sync: Int = 3)(implicit p: Parameters) extends LazyModule
 {
   val node = IntSyncSinkNode(sync)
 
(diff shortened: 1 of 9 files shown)


hcook:
I don't feel qualified to opine on the internal implementation changes, but I support the following API changes:
> Deprecation of IntSyncCrossingSink API which used the number of synchronizer stages to distinguish between Async, Rational, and Sync crossing types. Instead, use specific IntSync*CrossingSink types to distinguish the different types.
> Introduction of ClockCrossingReg for places where *SynchronizerShiftReg were being used to indicate a clock crossing but where depth was == 1.

ernie-sifive:
@mwachs5 Should we add this check to the other apply methods -- SyncResetSynchronizerShiftReg and ResetSynchronizerShiftReg?

mwachs5:
"@mwachs5 Should we add this check to the other apply methods -- SyncResetSynchronizerShiftReg and ResetSynchronizerShiftReg?"
I am a little bit loathe to do so, because it's not quite correct --- what does it mean to take an initial value for something that you don't actually initialize (because it's just a wire)?
Are you running into cases where you need the API to allow sync=0 for the others?

mwachs5:
@ernie-sifive I added the same to the SyncResetSynchronizerShiftReg -- does this work?

ernie-sifive:
@mwachs5 Looks good and the builds that were failing before are now working!

src/main/scala/interrupts/Crossing.scala
```diff
@@ -56,11 +67,51 @@ class IntSyncCrossingSink(sync: Int = 3)(implicit p: Parameters) extends LazyMod
...
+  }
+}
+
+class IntSyncRationalCrossingSink()(implicit p: Parameters) extends LazyModule
+{
+  val node = IntSyncSinkNode(1)
+
+  lazy val module = new LazyModuleImp(this) {
+    (node.in zip node.out) foreach { case ((in, edgeIn), (out, edgeOut)) =>
+      out := RegNext(in.sync)
```
rmac-sifive: This never needs to be more than 1 register stage deep? I would've expected this to be implemented similarly to the above `class IntSyncAsyncCrossingSink(sync: Int = 3)` with configurable depth

src/main/scala/util/SynchronizerReg.scala
```diff
@@ -0,0 +1,202 @@
...
+  def apply [T <: Chisel.Data](in: T, sync: Int, init: T): T =
+    apply (in, sync, init.litValue.toInt, None)
+}
+
+class ResetSynchronizerShiftReg(w: Int = 1, sync: Int, init: Int) extends AbstractPipelineReg(w) {
+  require(sync > 1, s"Sync must be greater than 1, not ${sync}.")
+  override def desiredName = s"ResetSynchronizerShiftReg_w${w}_d${sync}_i${init}"
+  val output = Seq.tabulate(w) { i =>
+    val initBit = ((init >> i) & 1) > 0
+    SynchronizerPrimitiveShiftReg(io.d(i), sync, initBit, SynchronizerResetType.Inferred)
```
rmac-sifive: It may've been cleaner to expose the reset type as a constructor argument to the class to avoid DRY on each of `SyncresetSynchronizerShiftReg`, `ResetSynchronzierShiftReg` since you do essentially the same thing in each of these classes `val output = Seq.tabulate(w) { ... }`, but that would've required a massive code refactor everywhere we currently instantiate these registers so I think this is ok for now

PR #2215, opened 2019-12-05
AXIS: add diplomatic AXI-Stream support

terpstra:
**Release Notes**
This adds support for AXI-Stream connections diplomatically. It also prototypes several innovations we should port back to the other diplomatic busses:
1. BundleMap allows us to systematically add 'user' fields to busses with associated meta-data and categorization which helps adapters operate on fields they don't know anything about
2. beatBytes as an Option. This lets the bus width be controlled by common attachment points instead of slaves.
3. all Parameters are classes with private destructor, without unapply, and with versioned copy and constructors -> this allows us to upgrade the parameter API in a backwards compatible manner
4. all Parameters contain their duals in Master/Slave parameters

### DIFF
--- a/src/main/scala/amba/axis/Parameters.scala
+++ b/src/main/scala/amba/axis/Parameters.scala
@@ -1,0 +1,307 @@
+// See LICENSE.SiFive for license details.
+package freechips.rocketchip.amba.axis
+
+import chisel3._
+import chisel3.util._
+import chisel3.internal.sourceinfo.SourceInfo
+import freechips.rocketchip.config.Parameters
+import freechips.rocketchip.util.BundleField
+import freechips.rocketchip.diplomacy._
+
+class AXISSlaveParameters private (
+  val name:          String,
+  val supportsSizes: TransferSizes,
+  val destinationId: Int,
+  val resources:     Seq[Resource],
+  val nodePath:      Seq[BaseNode])
+{
+  require (!supportsSizes.none)
+  require (destinationId >= 0)
+
+  def v1copy(
+      name:          String        = name,
+      supportsSizes: TransferSizes = supportsSizes,
+      destinationId: Int           = destinationId,
+      resources:     Seq[Resource] = resources,
+      nodePath:      Seq[BaseNode] = nodePath) =
+  {
+    new AXISSlaveParameters(
+      name         = name,
+      supportsSizes = supportsSizes,
+      destinationId = destinationId,
+      resources     = resources,
+      nodePath      = nodePath)
+  }
+}
+
+object AXISSlaveParameters {
+  def v1(
+      name:          String,
+      supportsSizes: TransferSizes,
+      destinationId: Int           = 0,
+      resources:     Seq[Resource] = Nil,
+      nodePath:      Seq[BaseNode] = Nil) =
+  {
+    new AXISSlaveParameters(
+      name         = name,
+      supportsSizes = supportsSizes,
+      destinationId = destinationId,
+      resources     = resources,
+      nodePath      = nodePath)
+  }
+}
+
+class AXISSlavePortParameters private (
+  val slaves:        Seq[AXISSlaveParameters],
+  val reqAligned:    Boolean, /* 'Position byte's are unsupported */
+  val reqContinuous: Boolean, /* 'Null byte's inside transfers unsupported */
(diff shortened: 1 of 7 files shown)


src/main/scala/amba/axis/Parameters.scala
```diff
@@ -0,0 +1,93 @@
...
+case class AXISEdgeParameters(
+  master:     AXISMasterPortParameters,
+  slave:      AXISSlavePortParameters,
+  params:     Parameters,
+  sourceInfo: SourceInfo)
+{
+  require (slave.beatBytes.isEmpty && master.beatBytes.isEmpty,
+    s"Neither master nor slave port specify a bus width (insert an AXISBusBinder between them?) at ${sourceInfo}")
+  require (!slave.beatBytes.isEmpty && !master.beatBytes.isEmpty && slave.beatBytes != master.beatBytes,
+    s"Master and slave ports specify incompatible bus widths (insert an AXISWidthWidget between them?) at ${sourceInfo}")
```
davidbiancolin: Are the BusBinders and WidthWidgets coming in later PRs?

terpstra: Yes

grebe:
This overall looks pretty similar to [my diplomatic AXI4-Stream implementation](https://github.com/ucb-bar/dsptools/tree/master/rocket/src/main/scala/amba/axi4stream).

One thing you may find useful in your parameters is having the ability to specify if a slave interface is always ready or if a master interface is always valid (I only did this for the slave interface). Some blocks (e.g. a block that duplicates and splits a stream) are a lot simpler if you know this. Perhaps that's too niche to consider at this point.

If it's of interest to you, I'd be happy to port some of my existing blocks to this node implementation and upstream them- a "stream Mux", width adapter, and AXI MM <-> stream DMA may be useful to have in here. I haven't done it yet, but ideally the DMA would be abstracted to support whatever memory interface you want. Let me know if you're interested and maybe we should chat offline.

terpstra:
Yeah, I considered adding an 'alwaysReady'. I need this as well. The only reason I didn't is that then I would not be able to use the DecoupledIO from chisel. :-/

terpstra:
I did not know about your existing AXIS implementation; sorry for the duplicated effort! However, this was a good vehicle for show-casing the style we would like to migrate to our other busses. If you want to port your width adapter and PR it, I would be happy to accept it.

terpstra:
FYI, the 'AXISXbar' I implemented is already both a stream mux and demux.

grebe:
OK, cool. I'll put together a PR for something small and see what it looks like.

Re: `DecoupledIO`, yeah it's kind of a problem. I debating trying to do something *very* hacky by overriding the bool to be a UInt(0) and overriding `fire()`, but that seems ill advised. I think putting something in the parameters system that says always ready, adds an assertion to check that that's actually true, and then lets the generator deal with setting the field may be useful. I guess it's a compromise- rather than making the signal truly optional, you can get a lot of the benefits by letting the generator treat it as optional even if it is still present.

Another option I debated an never got around to is emitting a firrtl annotation that triggers a pass to do some checking and remove/replace the ready signals.

jerryhethatday:
> OK, cool. I'll put together a PR for something small and see what it looks like.
> 
> Re: `DecoupledIO`, yeah it's kind of a problem. I debating trying to do something _very_ hacky by overriding the bool to be a UInt(0) and overriding `fire()`, but that seems ill advised. I think putting something in the parameters system that says always ready, adds an assertion to check that that's actually true, and then lets the generator deal with setting the field may be useful. I guess it's a compromise- rather than making the signal truly optional, you can get a lot of the benefits by letting the generator treat it as optional even if it is still present.
> 
> Another option I debated and never got around to is emitting a firrtl annotation that triggers a pass to do some checking and remove/replace the ready signals.

Hello, @grebe  Sorry to bother.   I felt still confused about the rationale of  creating this BundleMap abstraction. What is it really for?
`BundleMap allows us to systematically add 'user' fields to busses with associated meta-data and categorization which helps adapters operate on fields they don't know anything about`  I think this claim is a little vague. Is this mainly for parameterization to dynamic configure how many and what items user fields could have?
BundleMap extends from Record, and is just a collection of chisel Data with corresponding BundleKey.  Why introduce this key<-> value stuff?

PR #2225, opened 2019-12-13
Fix Makefile rules for api-configs-chipsalliance

jackkoenig:
api-configs-chipsalliance does not use the Maven/SBT standard
src/main/scala, instead it puts main sources in src/

**Related issue**: 

**Release Notes**

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -37,8 +37,10 @@ src_path := src/main/scala
 resources := $(base_dir)/src/main/resources
 csrc := $(resources)/csrc
 vsrc := $(resources)/vsrc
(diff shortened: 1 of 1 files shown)


colinschmidt:
Why doesn't it use the standard path?

jackkoenig:
In building the Wake Scala flow, I looked around at a bunch of other ways of doing things and not all tools follow that standard (Mill for example). We decided to go with stuff like `craft/src` for Scala code (Craft being a name for Chisel/FIRRTL/Diplomacy and related tools), and then Verilog sources would be in `rtl/src` etc.

PR #2226, opened 2019-12-14
tilelink: address adjuster supports reporting a default region id offset

hcook:

**Impact**: API addition 

**Release Notes**
Address Adjuster junction now supports having a default region id. This is only used for reporting the default location of adjustable devices to DTS and OM, change has no effect on the circuit behavior of the adapter.

### DIFF
--- a/src/main/scala/tilelink/AddressAdjuster.scala
+++ b/src/main/scala/tilelink/AddressAdjuster.scala
@@ -7,14 +7,20 @@ import chisel3.util._
 import freechips.rocketchip.config._
 import freechips.rocketchip.diplomacy._
 
-// mask=0 -> passthrough
-// adjustableRegion -> only devices in this regions get adjusted
-// forceLocal -> used to ensure special devices (like debug) remain reacheable at chip_id=0 even if in adjustableRegion
-class AddressAdjuster(mask: BigInt, adjustableRegion: Option[AddressSet] = Some(AddressSet.everything), forceLocal: Seq[AddressSet] = Nil)(implicit p: Parameters) extends LazyModule {
+class AddressAdjuster(
+    mask: BigInt, // 0 -> passthrough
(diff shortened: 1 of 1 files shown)


src/main/scala/tilelink/AddressAdjuster.scala
```diff
@@ -7,14 +7,20 @@ import chisel3.util._
...
 
-// mask=0 -> passthrough
-// adjustableRegion -> only devices in this regions get adjusted
-// forceLocal -> used to ensure special devices (like debug) remain reacheable at chip_id=0 even if in adjustableRegion
-class AddressAdjuster(mask: BigInt, adjustableRegion: Option[AddressSet] = Some(AddressSet.everything), forceLocal: Seq[AddressSet] = Nil)(implicit p: Parameters) extends LazyModule {
+class AddressAdjuster(
+    mask: BigInt, // 0 -> passthrough
+    adjustableRegion: Option[AddressSet] = Some(AddressSet.everything), // only devices in this region get adjusted
+    forceLocal: Seq[AddressSet] = Nil, // ensure special devices (e.g. debug) remain reacheable at id=0 even if also asjusted
+    defaultId: Option[Int] = None) // default base address to use for reporting manager address metadata
```
mwachs5: is this an address or a integer multiple for the reloc region? if its an addres shouldn't it be a `BigInt`?

hcook: The address adjuster works mostly in terms of "region ids", which are basically the multiples of the relocatable local region. (This encoding ambiguity is because sometimes the dynamic input is driven from a chip_id/mode_select device instead of from a top-level wire encoded as an address; I do think this implementation is more general).

So the argument is selecting which id / multiple to use as the default. I see now that I wrote the comment poorly...

PR #2228, opened 2019-12-16
AXIS fixes

terpstra:

**Release Notes**
AXIS is irrevocable and some fields appear/disappear based on adapters.

### DIFF
--- a/src/main/scala/util/BundleMap.scala
+++ b/src/main/scala/util/BundleMap.scala
@@ -13,6 +13,12 @@ case class BundleField(key: BundleKeyBase, data: Data)
     key.setDefault(out)
     out
   }
+  def isOutput = DataMirror.specifiedDirectionOf(data) match {
+    case SpecifiedDirection.Unspecified => true
+    case SpecifiedDirection.Output      => true
+    case SpecifiedDirection.Input       => false
+    case SpecifiedDirection.Flip        => false
+  }
 }
 
 sealed trait BundleKeyBase {
@@ -38,7 +44,26 @@ class BundleMap(val fields: Seq[BundleField]) extends Record {
   require(fields.map(_.key.name).distinct.size == fields.size)
   override def cloneType: this.type = (new BundleMap(fields)).asInstanceOf[this.type]
   val elements = ListMap(fields.map { bf => bf.key.name -> chisel3.experimental.DataMirror.internal.chiselTypeClone(bf.data) } :_*)
-  def apply[T <: Data](key: BundleKey[T]) = elements(key.name).asInstanceOf[T]
+
+  def apply[T <: Data](key: BundleKey[T]): T         = elements(key.name).asInstanceOf[T]
+  def lift [T <: Data](key: BundleKey[T]): Option[T] = elements.lift(key.name).map(_.asInstanceOf[T])
+
+  def apply(key: BundleKeyBase): Data         = elements(key.name)
+  def lift (key: BundleKeyBase): Option[Data] = elements.lift(key.name)
+
+  // For every key where this.output or that.input
+  def assignL(that: BundleMap): Seq[BundleKeyBase] = {
+    val a = this.fields.filter( _.isOutput).map(_.key)
+    val b = that.fields.filter(!_.isOutput).map(_.key)
+    (a ++ b).toList.distinct
+  }
+
+  // For every key where this.input or that.output
+  def assignR(that: BundleMap): Seq[BundleKeyBase] = {
+    val a = this.fields.filter(!_.isOutput).map(_.key)
+    val b = that.fields.filter( _.isOutput).map(_.key)
+    (a ++ b).toList.distinct
+  }
 }
 
 // Implement :<= :=> and :<>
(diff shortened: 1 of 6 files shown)


terpstra:
@jackkoenig I'd be happy if `:<=` `:=>` and `:<>` were in chisel3 itself, but I'm not going to wait.

src/main/scala/util/BundleMap.scala
```diff
@@ -37,6 +44,81 @@ class BundleMap(val fields: Seq[BundleField]) extends Record {
...
+  private def descendR(x: Data, y: Data): Unit = {
+    DataMirror.specifiedDirectionOf(y) match {
+      case SpecifiedDirection.Unspecified => assignR(x, y)
+      case SpecifiedDirection.Output      => ()
+      case SpecifiedDirection.Input       => y := x
+      case SpecifiedDirection.Flip        => assignL(y, x)
+    }
+  }
+
+  def assignL(x: Data, y: Data): Unit = {
```
seldridge: If this PR is going to `FixChisel3`, did you consider returning `x` here instead of `Unit`?

That has seemed to be one of the nice behaviors of the diplomatic connectors that might be worth exploring here.

terpstra: That's an interesting idea. I'm not sure it's that useful in this context, though. In diplomacy, when you go `a := b := c`, generally `b` does something between the two connections. In chisel, this sort of expression would only be legal if 'b' were a wire, and thus it's a no-op.

seldridge: I think I've wanted this for constructs like the following where you're iterating through something connecting things as you go along. I realize this effectively saves one line, but it's just always seemed obnoxious that we return `Unit` when I'm not sure there's a good reason for it. Granted `def foo { val x = 3 }` is of type `Unit`.

```scala
class ShiftRegister(a: Int) extends MultiIOModule {

  val in: Bool = IO(Input(Bool()))
  val out = IO(Output(Bool()))

  val state: Seq[Bool] = Seq.fill(a)(Reg(Bool()))

  out := state.foldLeft(in){ case (right, left) => left := right; left }

}
```

src/main/scala/util/BundleMap.scala
```diff
@@ -37,6 +44,81 @@ class BundleMap(val fields: Seq[BundleField]) extends Record {
...
+  // For every key where this.input or that.output
+  def assignR(that: BundleMap): Seq[BundleKeyBase] = {
+    val a = this.fields.filter(!_.isOutput).map(_.key)
+    val b = that.fields.filter( _.isOutput).map(_.key)
+    (a ++ b).toList.distinct
+  }
 }
 
+// Implement :<= :=> and :<>
+object FixChisel3 {
```
seldridge: This would probably be better as `private[freechips]` if you are proposing something you want to get into Chisel3.

src/main/scala/util/BundleMap.scala
terpstra: Ah. So your point is it's not just legal for wires, but also registers. I still think that until/unless chisel3 supports connecting circuits with ':=' instead of just Data, this is going to have very limited utility. Basically, it's only good for shift registers.

src/main/scala/util/BundleMap.scala
terpstra: Could you elaborate? I'm using these operators in other repositories, too. Are you proposing some form of upgrade path?

seldridge: My understanding of this is that this PR is a proposed API addition to Chisel3, but has immediate need and is too high priority to go into Chisel3 first. So, the upgrade path should look like this:

1) `object FixChisel3` is added to Rocket Chip
2) A PR adding `:<=`, `:=>`, and `:<>` is filed on Chisel3 and gets into the merge window for 3.3 (closing in ~January)
3) Rocket bumps to 3.3 at the end of January (or sooner if comfortable with master) and `object FixChisel3` is deprecated and removed.

The best situation is to limit how far `FixChisel3` gets in terms of usage to limit the deprecation/removal of (3). One way to do that is to limit this to `private[freechips]`. However, if you need this in `sifive` packages, that won't work.

Alternatively, this is not intended to ever go into Chisel3 and/or not something SiFive wants to contribute, then there's no need to limit the scope.

terpstra: I am using these operators in other packages, yes. I can't imagine removing this object in the same PR that bumps chisel would be a huge problem, would it?

seldridge: It's an API breakage. I was hoping there might be a way to avoid that. Sounds like there isn't. :( No reason to hold this up.

jackkoenig: I was thinking about this and I think it's actually perfectly safe to add methods that are previously defined using implicits. The only catch is the methods must be added to the base class rather than using an implicit conversion, but any new connection operators should be added in such a fashion. See this Scastie worksheet for an example: https://scastie.scala-lang.org/dGdCMA2NS3GwyIyDNzI7uQ

PR #2230, opened 2019-12-19
Add LazyModule.with* methods for clock/reset scope

jackkoenig:
Adds LazyModule.withClock, LazyModule.withReset, and
LazyModule.withClockAndReset to support setting custom clocks and resets
for LazyModuleImps

This enables using Chisels `with*` APIs in `LazyModules`. Currently, child `LazyModuleImps` will get the clock and reset from their parent (defaulting to false in `LazyRawModuleImps`), with the expectation that the user will override via direct connections. this doesn't work for changing the reset type because FIRRTL requires that the type of all reset connections are the same. Arguably, FIRRTL should be changed to allow last-connect semantics to override inferred reset type, but regardless, it seems useful to enable using the `with*` APIs on `LazyModuleImps`.

Example:
```scala
class MyLazyModule extends LazyModule {
  val child = LazyModule.withReset(module.syncReset)(new MyChildLazyModule)

    val syncReset = MyAsyncToSyncConverter(reset)
     ...
  }
}
```

I haven't quite figured out how to test this, so I'm opening this draft PR to show the proposed API.

**Related issue**: 

**Release Notes**

Adds LazyModule.withClock, LazyModule.withReset, and
LazyModule.withClockAndReset to support setting custom clocks and resets
for LazyModuleImps

### DIFF
--- a/src/main/scala/diplomacy/LazyModule.scala
+++ b/src/main/scala/diplomacy/LazyModule.scala
@@ -16,6 +16,9 @@ abstract class LazyModule()(implicit val p: Parameters)
   protected[diplomacy] var info: SourceInfo = UnlocatableSourceInfo
   protected[diplomacy] val parent = LazyModule.scope
 
+  // Wrapper function used for with(Clock|Reset) scoping, see LazyModule.withClock
+  private[diplomacy] var wrap: LazyModuleImpLike => LazyModuleImpLike = identity
+
   // code snippets from 'InModuleBody' injection
   protected[diplomacy] var inModuleBody = List[() => Unit]()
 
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -136,6 +139,28 @@ object LazyModule
...
+    res
+  }
+
+  def withReset[T <: LazyModule](reset: => Reset)(bc: T)(implicit valName: ValName, sourceInfo: SourceInfo): T = {
+    val res = apply(bc)
+    res.wrap = (x: LazyModuleImpLike) => chisel3.withReset(reset)(x)
+    res
+  }
+
+  def withAndClock[T <: LazyModule](clock: => Clock, reset: => Reset)
```
ernie-sifive: Did you intend this to be def withClockAndReset ... ?

jackkoenig: I did indeed! Whoops

jackkoenig:
We decided against this approach. Instead we enhanced FIRRTL, see #2239

PR #2232, opened 2019-12-20
Add chicken bit to disable speculative I$ refill

aswaterman:
This is mostly intended as a post-silicon debug feature to eliminate a common source of spurious events.

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -76,7 +76,8 @@ class RocketCustomCSRs(implicit p: Parameters) extends CustomCSRs with HasRocket
     val mask = BigInt(
       tileParams.dcache.get.clockGate.toInt << 0 |
       rocketParams.clockGate.toInt << 1 |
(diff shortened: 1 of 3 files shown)


src/main/scala/rocket/RocketCore.scala
```diff
@@ -76,7 +76,8 @@ class RocketCustomCSRs(implicit p: Parameters) extends CustomCSRs with HasRocket
     val mask = BigInt(
       tileParams.dcache.get.clockGate.toInt << 0 |
       rocketParams.clockGate.toInt << 1 |
-      rocketParams.clockGate.toInt << 2
+      rocketParams.clockGate.toInt << 2 |
+      1 << 3 // disableSpeculativeICacheRefill
```
mwachs5: do we want to hardcode this?

aswaterman: I can add a knob for the presence of this chicken bit if you prefer, but my inclination is to make the knob default to having the chicken bit.

PR #2240, opened 2020-01-03
Adding case object for looking up base tile params

kenzhang82:
**Related issue**: 

**Release Notes**

Add the case object for creating a sequence of parameters in the base tiles, so its parameters can be looked up from `subsystem/Configs.scala`. For instance, one can specify different clock crossings to different tiles like:
```
... extends Config((site, here, up) => {
  case BuildBaseTile => site(TileKey).hartId match {
    case 0: List(RocketCrossingParams(
      crossingType = SynchronousCrossing()
    case 1: List(RocketCrossingParams(
      crossingType = AsynchronousCrossing()
    ...
})

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -12,6 +12,8 @@ import freechips.rocketchip.rocket._
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.util._
 
+case object BuildBaseTile extends Field[Seq[Parameters => BaseTile]](Nil)
+
 case object TileVisibilityNodeKey extends Field[TLEphemeralNode]
 case object TileKey extends Field[TileParams]
 case object ResetVectorBits extends Field[Int]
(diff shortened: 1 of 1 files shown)


kenzhang82:
@jackkoenig @hcook could you please review this?

hcook:
I don't understand how this Field is supposed to be used; you have added a Field that is not actually used to look up any information in any public code. Furthermore, the example you provide in your PR does not match the type signature of the Field in the code. Does looking up `BuildBaseTile` return an instance of `BaseTile` or of `RocketCrossingParams`? `RocketCrossingKey` already allows for different tiles to have different cross types, why is this additional field necessary?

kenzhang82:
Thanks Henry. This is supposed to return an instance of BaseTile with its params.

I looked at the example in Config.scala, and I couldn’t figure out how we can reference HartId (which we want to match the tile it is on) from RocketCrossingKey without going through baseTile hierarchy.

Can you provide an example how to have different tiles with different crossings with the existing field? 

Thanks!

kenzhang82:
Never mind @hcook, I managed to figure it out. Closing this.

PR #2241, opened 2020-01-07
Tile connectMasterPortsToSBus TLFIFOFixer

johningalls-sifive:
**Impact**: addition

add TLFIFOFixer between the Tile port and the SBus to support requestFifo from the core to non-fifo memory regions

### DIFF
--- a/src/main/scala/subsystem/HasTiles.scala
+++ b/src/main/scala/subsystem/HasTiles.scala
@@ -47,8 +47,8 @@ trait HasTiles extends HasCoreMonitorBundles { this: BaseSubsystem =>
     sbus.fromTile(tile.tileParams.name, crossing.master.buffers) {
         crossing.master.cork
           .map { u => TLCacheCork(unsafe = u) }
-          .map { _ :=* tile.crossMasterPort() }
-          .getOrElse { tile.crossMasterPort() }
+          .map { _ :=* TLFIFOFixer() :=* tile.crossMasterPort() }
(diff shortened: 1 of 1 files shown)


hcook:
Not sure unconditionally adding this adapter to every single tile ever produced is a good idea...

johningalls-sifive:
> Not sure unconditionally adding this adapter to every single tile ever produced is a good idea...

Will the FIFOFixer logic be optimized away if not needed, similar to WidthWidget?

terpstra:
Yes. But people complain about empty/useless modules.

johningalls-sifive:
@terpstra said:

> Actually, I think we did this because of performance on MAIN MEMORY (not the L2LIM). The goal was to exclude the L2$ from the FIFOFixer, as otherwise you could only do one operation at a time.
> So the gist of it seems to be: we strongly order MMIO to all volatile devices.
> All cacheable devices (TRACKED/CACHED) follow the weakest RISC-V memory model.
> Devices that fall into the IDEMPOTENT/UNCACHED camp each get to be in their own ordering domain, and those devices count as MMIO, but you need to put fences if you need order between domains.
> Soo.... we won't be merging this PR to add more FIFOFixers. Instead, we will accept a change to requireFIFO that accepts a TLFIFOFixer.Policy
> And you should call that method on TLFIFOFixer.allVolatile because we only guarantee FIFO order without fences for VOLATILE devices.

PR #2248, opened 2020-01-10
Fix offsets of DMOuter registers in Object Model.

richardxia:
The previous Object Model implementation did not account for the relative offset of the dmOuter register address block. This commit adds some logic to translate the offsets into absolute addresses.

**Related issue**: 

**Release Notes**

There was a bug in the generated Object Model for the Debug Module, where the register field bit offsets were incorrect. This has now been corrected.


src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTrees.scala
```diff
@@ -38,11 +38,44 @@ class DebugLogicalTreeNode(
...
+    regMap.copy(
+      registerFields = regMap.registerFields.map(
+        field => field.copy(
+          bitRange = field.bitRange.copy(
+            base = field.bitRange.base + baseAddressBytes * 8
+          )
+        )
+      )
+    )
+  }
```
richardxia: @jackkoenig, @azidar, are we using any lens library in rocket-chip? It'd be pretty convenient for expressing code like the above.

richardxia:
Thanks for taking a look, @ernie-sifive!

PR #2250, opened 2020-01-10
Rocket Core DCache TileLink: dont require Fifo just yet

johningalls-sifive:
**Related issue**: https://github.com/chipsalliance/rocket-chip/pull/2246 causes problems in downstream repositories


terpstra:
Could you please fix the requireFIFO also?

johningalls-sifive:
> Could you please fix the requireFIFO also?

Will do in a separate PR after we sort out the issues between this change and the downstream repository.

johningalls-sifive:
> > Could you please fix the requireFIFO also?
> 
> Will do in a separate PR after we sort out the issues between this change and the downstream repository.

done in https://github.com/chipsalliance/rocket-chip/pull/2252

PR #2251, opened 2020-01-10
Use ungated clock in TracedInstruction bundle

ernie-sifive:
**Related issue**: 

**Release Notes**


aswaterman:
What's the problem with using the gated clock, given that t.valid should only toggle when the gated clock is running?

PR #2256, opened 2020-01-14
supporting D$ way size larger than 4KB

ShuyunJia:
Signed-off-by: shuyun <shuyunjia@outlook.com>

**Related issue**: 

**Release Notes**
For a VIPT D$ with way size larger than 4KB, multiply synonyms with the same physical address would reside inside. By probing back to the master initiates the line fill request, the synonyms would be cleaned automatically. For an incoming probe transaction in physical address, multiple tag lines should be checked since it may appears in several possible virtual indexes. 

This proposal is not fully tested and there would be extra cycles' cost on line fill. Besides, the timing would become a bit worse since there would be more check for multiply synonyms. So I requested a draft pr so that we can disccuss about it.

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -37,39 +37,48 @@ class DCacheDataReq(implicit p: Parameters) extends L1HellaCacheBundle()(p) {
   val wordMask = UInt(width = rowBytes / wordBytes)
   val eccMask = UInt(width = wordBytes / eccBytes)
   val way_en = Bits(width = nWays)
+  val slice_en = Vec(nSlices, Bool())
 }
 
 class DCacheDataArray(implicit p: Parameters) extends L1HellaCacheModule()(p) {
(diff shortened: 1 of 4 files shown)


aswaterman:
I certainly agree with the goal of circumventing the VIPT constraint, but doing it this way hurts PPA.  It can also be done in a way that's less invasive to the codebase.  I will see if I can get permission to open-source the technique eventually.  Until then, you might want to just maintain this on your own branch if you need it.

aswaterman:
We recently added a parameter that allows you to override the function that constructs the data cache, provided it adheres to the same interface: https://github.com/chipsalliance/rocket-chip/blob/27120ee42ff4ff41908981025efcf7d59f879894/src/main/scala/rocket/HellaCache.scala#L232

If you use rocket-chip as a library in your broader project, you can maintain your alternate data cache design in your project's repo, and override the parameter to instantiate your cache.

PR #2262, opened 2020-01-21
Remove unused type parameter on RegisterRouter.

richardxia:
**Related issue**: 

**Impact**: API modification (I guess this is technically backwards incompatible)

**Release Notes**

`RegisterRouter` no longer takes a type parameter, as it was not being used.

---

I was trying to read the definition of `RegisterRouter`, and I was having difficulty understanding what its type parameter `T <: Data` was being used for. As far as I can tell, it is not being used by anything, and simply removing it and making it a non-generic type seems to work fine. If it is in fact actually being used for something, I'd love to learn more about it, since at my current level of understanding of Scala, I can't figure out how it is being used, since none of its members or methods depend on it.


hcook:
Yeah, I see the IO subclass having a `[T <: Data]` parameter which is its out port bundle type, but not what application that has in the base class. @terpstra any thoughts?

PR #2264, opened 2020-01-21
Enable Tile Link Printing as part of monitors [WIP]

mwachs5:
**Related issue**: 

One request has been more "printf" style logging similar to what is output for processor trace. This is a a strawman example of how printfs could be selectively added as part of diplomatic monitors. 

The idea is that using the same scoping mechanism by which monitors are included at all, we can add a Prefix option which when set would make the monitors add a `fprintf`.

The PR as currently done adds the following to the monitor on the TL interface to the PLIC, which only happens if both ready and valid are set.

```
    `ifndef SYNTHESIS
    `ifdef PRINTF_COND
      if (`PRINTF_COND) begin
    `endif
        if (_T_702 & _T_112) begin
          $fwrite(32'h80000002,"PLIC_TL_IFC DecoupledIO(ready -> %d, valid -> %d, bits -> TLBundleA(opcode -> %d, param -> %d, size -> %d, source -> %d, address -> %d, mask -> %d, data -> %d, corrupt -> %d))",io_in_a_ready,io_in_a_valid,io_in_a_bits_opcode,io_in_a_bits_param,io_in_a_bits_size,io_in_a_bits_source,io_in_a_bits_address,io_in_a_bits_mask,io_in_a_bits_data,io_in_a_bits_corrupt); // @[Monitor.scala 573:56:freechips.rocketchip.system.DefaultConfig.fir@63461.8]
        end
    `ifdef PRINTF_COND
      end
    `endif
    `endif // SYNTHESIS
```

Potential issues with this approach:
  * logging/prefixing is controlled by designer, not by user of the code, which may be not the level of control desired.
  * prints only happening right now when ready & valid asserted - do we want to show if ready | valid asserted?
  * This may get noisy fast. Maybe want to add a `PlusArgReader` to control the logging at a finer grain?

**Release Notes**

### DIFF
--- a/src/main/scala/devices/tilelink/Plic.scala
+++ b/src/main/scala/devices/tilelink/Plic.scala
@@ -349,7 +349,8 @@ trait CanHavePeripheryPLIC { this: BaseSubsystem =>
   val plicOpt  = p(PLICKey).map { params =>
     val tlbus = attach(p(PLICAttachKey).slaveWhere)
     val plic = LazyModule(new TLPLIC(params, tlbus.beatBytes))
-    plic.node := tlbus.coupleTo("plic") { TLFragmenter(tlbus) := _ }
+    val tmp = tlbus.coupleTo("plic") { TLFragmenter(tlbus) := _ }
+    EnableMonitorPrint("PLIC_TL_IFC", implicit p => plic.node := tmp)
     plic.intnode :=* ibus.toPLIC
     plic
   }
--- a/src/main/scala/diplomacy/Nodes.scala
+++ b/src/main/scala/diplomacy/Nodes.scala
@@ -12,6 +12,7 @@ import scala.util.matching._
 
 case object MonitorsEnabled extends Field[Boolean](true)
 case object RenderFlipped extends Field[Boolean](false)
+case object MonitorPrintPrefix extends Field[Option[String]](None)
 
 case class RenderedEdge(
   colour:  String,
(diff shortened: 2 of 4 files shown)


mwachs5:
Also, I tried to make the output "prettier" with this but it did not compile:

```
+++ b/src/main/scala/tilelink/Bundles.scala
@@ -183,9 +183,23 @@ final class TLBundleA(params: TLBundleParameters)
   val mask    = UInt(width = params.dataBits/8)
   val data    = UInt(width = params.dataBits)
   val corrupt = Bool() // only applies to *Data messages
+
+  // override def toPrintable: Printable = {
+  //   p"A:\t" +
+  //   p"opcode[${opcode}]\t" +
+  //   p"param[${param}]\t" +
+  //   p"size[${size}]\t" +
+  //   p"source[${source}]\t" +
+  //   p"address[${address}]\t" +
+  //   p"user[${user}]\t" +
+  //   p"mask[${mask}]\t" +
+  //   p"data[${data}]\t" +
+  //   p"corrupt[${corrupt}]\n"
+  // }
 }
+
```

terpstra:
I agree with all of your trade-off analysis. I think the most relevant is the "must be done by the designer". We could consider using a per-monitor plus arg? Not sure.

I would change your EnableMonitorPrint to:
```scala
   def EnableMonitorPrint[T](prefix: String)(body: Parameters => T)(implicit p: Parameters) = body(p.alterPartial
```

ie: I would curry it. That way you can use it like this:
```scala
  EnableMonitorPrint("hax") { implicit p =>
    ... more complicated code that needs curly braces
  }
```

PR #2267, opened 2020-01-23
Add a new set of deprecation warnings

hcook:
**Related issue**: 

**Release Notes**


src/main/scala/tilelink/RegisterRouter.scala
```diff
@@ -138,8 +138,10 @@ class TLRegBundleBase(arg: TLRegBundleArg) extends Bundle
   implicit val p = arg.p
 }
 
+@deprecated("Use HasTLControlRegMap trait in place of TLRegisterRouter+TLRegBundle+TLRegModule", "rocket-chip 1.4")
```
mwachs5: what does "HasTLControlRegMap" go . with? a LazyModule? LazyModuleImp? Module..?

mwachs5: also what is a "ControlRegMap"? The name seems a bit specific. Why isn't it just "HasTLRegMap"? Are there other kinds of TL____RegMap?

hcook: if you scroll down in the same file you will see:
```
/** Mix this trait into a RegisterRouter to be able to attach its register map to a TL bus */
trait HasTLControlRegMap { this: RegisterRouter[_] =>
```
and if you check the defintion of `RegisterRouter` you will find it `extends LazyModule`

hcook: A control reg map is presumed to consist of MMIO control registers. This trait adds a node called `controlNode`. It doesn't cover use cases like SPIFlash that have another node for "data"; users can always add extra nodes for additional, non-control maps. Having an BundleBridge IO node is likewise handled by mixing in an additional trait on top of this one.

hcook: `Why isn't it just "HasTLRegMap"? ` ...because what if you have more than one map?

PR #2271, opened 2020-01-30
Tl vip2

brrmorre:
**Related issue**: 

**Impact**: API modification

**Release Notes**

Updating the TL Monitors to add more support for formal verification and additional constraints. This branch is copied from https://github.com/dgupta-oski/rocket-chip/tree/TL-VIP

Similar PRs here:
https://github.com/chipsalliance/rocket-chip/pull/2269
https://github.com/chipsalliance/rocket-chip/pull/2138

### DIFF
--- a/src/main/scala/formal/FormalUtils.scala
+++ b/src/main/scala/formal/FormalUtils.scala
@@ -5,7 +5,7 @@ import Chisel._
 import chisel3.{VecInit}
 import chisel3.util.Cat
 import chisel3.internal.sourceinfo.{SourceInfo, SourceLine}
-import freechips.rocketchip.config._
+import freechips.rocketchip.config.Field
 
 sealed abstract class MonitorDirection(name: String) {
   override def toString: String = name
--- a/src/main/scala/formal/FormalUtils.scala
+++ b/src/main/scala/formal/FormalUtils.scala
@@ -5,6 +5,7 @@ import Chisel._
 import chisel3.{VecInit}
 import chisel3.util.Cat
 import chisel3.internal.sourceinfo.{SourceInfo, SourceLine}
+import freechips.rocketchip.config._
 
 sealed abstract class MonitorDirection(name: String) {
   override def toString: String = name
@@ -21,6 +22,10 @@ object MonitorDirection {
 
   object Cover extends MonitorDirection("Cover") { override def flip: MonitorDirection = Cover }
 }
+
+case class TLMonitorStrict(on: Boolean = true)
+
+case object TLMonitorStrictMode extends Field[TLMonitorStrict](TLMonitorStrict(true))
 
 sealed abstract class PropertyClass(name: String) {
   override def toString: String = name
--- a/src/main/scala/formal/FormalUtils.scala
+++ b/src/main/scala/formal/FormalUtils.scala
@@ -23,9 +23,7 @@ object MonitorDirection {
   object Cover extends MonitorDirection("Cover") { override def flip: MonitorDirection = Cover }
 }
 
-case class TLMonitorStrict(on: Boolean = true)
-
-case object TLMonitorStrictMode extends Field[TLMonitorStrict](TLMonitorStrict(true))
+case object TLMonitorStrictMode extends Field[Boolean](true)
 
 sealed abstract class PropertyClass(name: String) {
   override def toString: String = name
--- a/src/main/scala/formal/FormalUtils.scala
+++ b/src/main/scala/formal/FormalUtils.scala
@@ -25,6 +25,15 @@ object MonitorDirection {
 
 case object TLMonitorStrictMode extends Field[Boolean](true)
 
+case class TestImplType(
+  simulation: Boolean = true,
+  formal: Boolean = false,
+  fpga: Boolean = false
+)
+
+// Determine if test should be generated for formal and/or simulation
+case object TestplanTestType extends Field[TestImplType]
+
 sealed abstract class PropertyClass(name: String) {
   override def toString: String = name
 }
--- a/src/main/scala/formal/FormalUtils.scala
+++ b/src/main/scala/formal/FormalUtils.scala
@@ -32,7 +32,7 @@ case class TestImplType(
 )
 
 // Determine if test should be generated for formal and/or simulation
-case object TestplanTestType extends Field[TestImplType]
+case object TestplanTestType extends Field[TestImplType](TestImplType())
 
 sealed abstract class PropertyClass(name: String) {
   override def toString: String = name
--- a/src/main/scala/tilelink/Monitor.scala
+++ b/src/main/scala/tilelink/Monitor.scala
@@ -575,7 +575,7 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
     }
   }
 
-  def legalizeADSource(bundle: TLBundle, edge: TLEdge) {
+  def legalizeADSource(bundle: TLBundle, edge: TLEdge) = if (TLMonitorStrictMode == true) {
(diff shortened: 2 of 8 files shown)


src/main/scala/tilelink/Monitor.scala
```diff
@@ -484,34 +574,74 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
...
       d_clr := UIntToOH(bundle.d.bits.source)
-      assume((a_set | inflight)(bundle.d.bits.source), "'D' channel acknowledged for nothing inflight" + extra)
+      d_opcodes_clr := size_to_numfullbits(1.U << log_a_opcode_bus_size.U) << (bundle.d.bits.source << log_a_opcode_bus_size.U)
+      d_sizes_clr := size_to_numfullbits(1.U << log_a_size_bus_size.U) << (bundle.d.bits.source << log_a_size_bus_size.U)
+    }
+    when (bundle.d.valid && d_first && edge.isResponse(bundle.d.bits) && !d_release_ack) {
+      assert(((inflight)(bundle.d.bits.source)) || (bundle.a.valid && (bundle.a.bits.source === bundle.d.bits.source)), "'D' channel acknowledged for nothing inflight" + extra)
+      // TODO: Commenting out due to error - not enough arguments for method apply: (n: Int, gen: T)(implicit sourceInfo: chisel3.internal.sourceinfo.SourceInfo,
+      //       implicit compileOptions: chisel3.CompileOptions)chisel3.Vec[T] in trait VecFactory
+      // assert(((bundle.d.bits.opcode === Vec(responseMap)(a_opcode_lookup)) || (bundle.d.bits.opcode === Vec(responseMapSecondOption)(a_opcode_lookup)))
```
brrmorre: This assertion is commented out for some reason...

src/main/scala/tilelink/Monitor.scala
```diff
@@ -484,34 +574,74 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
   }
 
   def legalizeADSource(bundle: TLBundle, edge: TLEdge) {
+    val a_size_bus_size = edge.bundle.sizeBits
```
brrmorre: Theres no additional left shifting. Unless theres an additional flag that I'm missing?

brrmorre: Compare with: https://github.com/chipsalliance/rocket-chip/pull/2138/files

src/main/scala/tilelink/Monitor.scala
brrmorre: I fixed this

src/main/scala/tilelink/Monitor.scala
brrmorre: I fixed this

src/main/scala/tilelink/Monitor.scala
```diff
@@ -31,18 +33,26 @@ object TLMonitor {
...
+@chiselName
 class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirection.Monitor) extends TLMonitorBase(args)
 {
+  require (args.edge.params(TLMonitorStrictMode) || (! args.edge.params(TestplanTestType).formal))
+
   val cover_prop_class = PropertyClass.Default
   val desc_text = "Placeholder"
 
-  /*def assert(cond: Bool, message: String): Unit = {
+  def monAssert(cond: Bool, message: String): Unit = if (monitorDir == MonitorDirection.Monitor)
```
brrmorre: Need to move monAssert and assume to FormalUtils...

jackkoenig: This is a nit but stylistically I'd write this as:
```scala
def monAssert(cond: Bool, message: String): Unit =
  if (monitorDir == MonitorDirection.Monitor) {
    assert(cond, message)
  } else {
    Property(monitorDir, cond, message, PropertyClass.Default, desc_text)
  }
```

In particular, Scala-style dictates that curly-braces go on the same line as the declaration they represent (https://docs.scala-lang.org/style/nested-blocks.html#curly-braces). It's usually harmless to put it on the next line, but apparently it can cause issues with the parser.

I would also add a comment explaining why you have `monAssert` instead of just `assert` (making it "flippable") just for future people looking at this code.

src/main/scala/formal/FormalUtils.scala
```diff
@@ -5,6 +5,7 @@ import Chisel._
 import chisel3.{VecInit}
 import chisel3.util.Cat
 import chisel3.internal.sourceinfo.{SourceInfo, SourceLine}
+import freechips.rocketchip.config._
```
jackkoenig:
```suggestion
import freechips.rocketchip.config.Field
```
Better to not wildcard import

src/main/scala/tilelink/Monitor.scala
```diff
@@ -517,6 +610,98 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
     when (bundle.a.fire() || bundle.d.fire()) { watchdog := 0.U }
   }
 
+  def legalizeADSourceStrict(bundle: TLBundle, edge: TLEdge) {
+    val a_size_bus_size = edge.bundle.sizeBits + 1 //add one so that 0 is not mapped to anything (size 0 -> size 1 in map, size 0 in map means unset)
+    val a_opcode_bus_size = 3 + 1 //opcode size is 3, but add so that 0 is not mapped to anything
+    val log_a_opcode_bus_size = log2Ceil(a_opcode_bus_size)
+    val log_a_size_bus_size = log2Ceil(a_size_bus_size)
+    def size_to_numfullbits(x: UInt): UInt = (1.U << x) - 1.U //convert a number to that many full bits
+
```
jackkoenig: Since you're using `@chiselName`, can't you get rid of all of these `suggestNames`?

brrmorre: For some reason @chiselName wasn't working for me

src/main/scala/tilelink/Monitor.scala
jchang0: I also suggested that this definition should go somewhere more common (maybe formal.utils) instead of making it special to Monitor.  I also would think there should be a better way to be more future compatible, but haven't think too hard about it yet.  Not sure if @jackkoenig  have better suggestions.

brrmorre: Jacob's suggested change is gonna go into a future PR

brrmorre: I implemented Jack's suggested changes.

src/main/scala/tilelink/Monitor.scala
```diff
@@ -31,30 +33,27 @@ object TLMonitor {
...
-    Property(monitorDir,
-        cond,
-        message,
-        PropertyClass.Default,
-        desc_text)
-  }*/
-
-  def assume(cond: Bool, message: String): Unit = if (monitorDir == MonitorDirection.Monitor)
-  {
+  val desc_text = "TLMonitor Assertion"
```
jchang0: desc_text is supposed to be for message, which should be set at each property instantiation.

src/main/scala/tilelink/Monitor.scala
```diff
@@ -517,6 +599,98 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
     when (bundle.a.fire() || bundle.d.fire()) { watchdog := 0.U }
   }
 
+  def legalizeADSourceStrict(bundle: TLBundle, edge: TLEdge) {
```
jchang0: rename to legalizeADSource (since everybody should be using this one)

add depricated to the current legalizeADSource and rename it to something else that shows it's only a workaround for almond

jackkoenig:
Please don't merge broken commits, either Squash-and-Merge or cleanup the history before using a merge commit.

PR #2272, opened 2020-01-30
Port some selected modules to chisel3 for properly inferring resets

mwachs5:
This ports some selected modules to `chisel3` so that resets can be properly inferred in those modules.

- [ ] The results have been LEC-ed for consistency.

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/Multiplier.scala
+++ b/src/main/scala/rocket/Multiplier.scala
@@ -3,29 +3,30 @@
 
 package freechips.rocketchip.rocket
 
-import Chisel._
+import chisel3._
+import chisel3.util.{Cat, log2Up, log2Ceil, log2Floor, Log2, Decoupled, Enum, Fill, Valid, Pipe}
 import Chisel.ImplicitConversions._
 import freechips.rocketchip.util._
 import ALU._
 
 class MultiplierReq(dataBits: Int, tagBits: Int) extends Bundle {
-  val fn = Bits(width = SZ_ALU_FN)
-  val dw = Bits(width = SZ_DW)
-  val in1 = Bits(width = dataBits)
-  val in2 = Bits(width = dataBits)
-  val tag = UInt(width = tagBits)
+  val fn = Bits(SZ_ALU_FN.W)
+  val dw = Bits(SZ_DW.W)
+  val in1 = Bits(dataBits.W)
+  val in2 = Bits(dataBits.W)
+  val tag = UInt(tagBits.W)
   override def cloneType = new MultiplierReq(dataBits, tagBits).asInstanceOf[this.type]
 }
 
 class MultiplierResp(dataBits: Int, tagBits: Int) extends Bundle {
-  val data = Bits(width = dataBits)
-  val tag = UInt(width = tagBits)
+  val data = Bits(dataBits.W)
+  val tag = UInt(tagBits.W)
   override def cloneType = new MultiplierResp(dataBits, tagBits).asInstanceOf[this.type]
 }
 
 class MultiplierIO(dataBits: Int, tagBits: Int) extends Bundle {
-  val req = Decoupled(new MultiplierReq(dataBits, tagBits)).flip
-  val kill = Bool(INPUT)
+  val req = Flipped(Decoupled(new MultiplierReq(dataBits, tagBits)))
+  val kill = Input(Bool())
   val resp = Decoupled(new MultiplierResp(dataBits, tagBits))
 }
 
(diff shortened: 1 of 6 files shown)


src/main/scala/rocket/Multiplier.scala
```diff
@@ -42,23 +43,23 @@ class MulDiv(cfg: MulDivParams, width: Int, nXpr: Int = 32) extends Module {
...
   val fastMulW = if (cfg.mulUnroll == 0) false else w/2 > cfg.mulUnroll && w % (2*cfg.mulUnroll) == 0
  
-  val s_ready :: s_neg_inputs :: s_mul :: s_div :: s_dummy :: s_neg_output :: s_done_mul :: s_done_div :: Nil = Enum(UInt(), 8)
-  val state = Reg(init=s_ready)
+  val s_ready :: s_neg_inputs :: s_mul :: s_div :: s_dummy :: s_neg_output :: s_done_mul :: s_done_div :: Nil = Enum(8)
+  val state = RegInit(s_ready)
  
-  val req = Reg(io.req.bits)
-  val count = Reg(UInt(width = log2Ceil(
+  val req = Reg(chiselTypeOf(io.req.bits))
```
mwachs5: @aswaterman this is the sort of thing I always get nervous about. The old `Reg(io.req.bits)` was just taking the type of that parameter, correct? It's not like a `RegNext` or `RegInit`?

src/main/scala/rocket/PMP.scala
```diff
@@ -23,15 +24,16 @@ object PMP {
 
   def apply(reg: PMPReg): PMP = {
     val pmp = Wire(new PMP()(reg.p))
-    pmp := reg
+    pmp.cfg := reg.cfg
```
mwachs5: Seems like I can't do this because chisel3 doesn't like it if names don't match exactly. Happy to take any suggestion other way to do this other than looking at the bundles and seeing what I need to attach one-by-one...

src/main/scala/tilelink/WidthWidget.scala
```diff
@@ -28,27 +29,27 @@ class TLWidthWidget(innerBeatBytes: Int)(implicit p: Parameters) extends LazyMod
...
-          corrupt_reg := Bool(false)
+          count := 0.U
+          corrupt_reg := false.B
         }
       }
 
       def helper(idata: UInt): UInt = {
-        val odata = Seq.fill(ratio) { idata }
-        val rdata = Reg(Vec(ratio-1, idata))
+        val odata = Seq.fill(ratio) { WireInit(idata) }
```
mwachs5: @aswaterman , @hcook this is another place where I would like a sanity check on my translation... is this equivalent?

src/main/scala/tilelink/WidthWidget.scala
```diff
@@ -109,14 +110,16 @@ class TLWidthWidget(innerBeatBytes: Int)(implicit p: Parameters) extends LazyMod
 
       val index  = sel | count
       def helper(idata: UInt, width: Int): UInt = {
-        val mux = Vec.tabulate(ratio) { i => idata((i+1)*outBytes*width-1, i*outBytes*width) }
+        val mux = VecInit.tabulate(ratio) { i => idata((i+1)*outBytes*width-1, i*outBytes*width) }
         mux(index)
       }
 
-      out <> in
+      out.bits := in.bits
```
mwachs5: 😭 I think out and in are both wires so I had to do this...

src/main/scala/tilelink/WidthWidget.scala
```diff
@@ -133,12 +136,14 @@ class TLWidthWidget(innerBeatBytes: Int)(implicit p: Parameters) extends LazyMod
     def splice[T <: TLDataChannel](edgeIn: TLEdge, in: DecoupledIO[T], edgeOut: TLEdge, out: DecoupledIO[T], sourceMap: UInt => UInt) = {
       if (edgeIn.manager.beatBytes == edgeOut.manager.beatBytes) {
         // nothing to do; pass it through
-        out <> in
+        out.bits := in.bits
```
mwachs5: another sad chisel3 conversion

src/main/scala/util/Repeater.scala
```diff
@@ -24,16 +25,18 @@ class Repeater[T <: Data](gen: T) extends Module
...
 }
 
 object Repeater
 {
   def apply[T <: Data](enq: DecoupledIO[T], repeat: Bool): DecoupledIO[T] = {
-    val repeater = Module(new Repeater(enq.bits))
+    val repeater = Module(new Repeater(chiselTypeOf(enq.bits)))
     repeater.io.repeat := repeat
-    repeater.io.enq := enq
+    repeater.io.enq.bits := enq.bits
```
mwachs5: one more where it seemed I had to do this manually instead of using `<>`

mwachs5: actually, I never tried `<>`... let me give that a shot

mwachs5: yay `<>` did work.

src/main/scala/rocket/Multiplier.scala
aswaterman: You got it right.

src/main/scala/rocket/PMP.scala
aswaterman: I think maybe you can use @terpstra's FixChisel3 operators, but for cases like this with 2-or-so fields, making it explicit seems OK.

seldridge: And I would prefer to get a discussion going about migrating those operators into Chisel 3 before mass proliferation in Rocket Chip.

src/main/scala/tilelink/WidthWidget.scala
aswaterman: I don't think this one should be necessary, given that odata is never assigned to.  If you back out this particular change, what goes wrong?

src/main/scala/tilelink/WidthWidget.scala
aswaterman: This is another place where @terpstra's FixChisel3 might help, but OK for now.

src/main/scala/rocket/PMP.scala
aswaterman: @seldridge I would also like to avoid the proliferation of custom-but-actually-generally-useful operators.

src/main/scala/tilelink/WidthWidget.scala
mwachs5: it complains that `idata` needs to be hardware. Let me back it out and double check

src/main/scala/rocket/PMP.scala
mwachs5: I opened a Chisel issue:https://github.com/freechipsproject/chisel3/issues/1316

src/main/scala/tilelink/WidthWidget.scala
aswaterman: Yeah, I see it.  But I think this is actually masking a different bug.  I'm looking into it.

mwachs5: sorry fighting merge conflicts will make this PR against master as well, performing the LEC checks there. Haven't gotten to try backing this out yet

src/main/scala/tilelink/WidthWidget.scala
terpstra: Try: out.bits :<> in.bits

src/main/scala/util/Repeater.scala
```diff
@@ -2,38 +2,41 @@
...
 // When repeat is asserted, the Repeater copies the input and repeats it next cycle.
 class Repeater[T <: Data](gen: T) extends Module
 {
-  val io = new Bundle {
-    val repeat = Bool(INPUT)
-    val full = Bool(OUTPUT)
-    val enq = Decoupled(gen).flip
-    val deq = Decoupled(gen)
-  }
+  val typ = chiselTypeOf(gen)
```
terpstra: I think you should have just let it using 'gen'. That will make it the caller's job to chiselTypeOf() on the constructor argument, which is probably a better API.

aswaterman: While I agree your thinking is more in line with the chisel3 philosophy, I thought it was more practical to maintain API compatibility, rather than having to change all the call sites (some of which have nothing to do with this PR).

mwachs5:
I have merged this to master and will rebase multiple-reset-schemes branch instead of merging this PR.

PR #2273, opened 2020-01-31
Convert some modules to Chisel3

mwachs5:
Convert some modules to chisel3 to allow properly infer resets. Same as to #2272 but targeting master

**Related issue**: 

**Release Notes**


aswaterman:
Thanks again for doing this.

PR #2275, opened 2020-02-04
Add feature-disable bit to suppress corrupt on GrantData

aswaterman:
Temporarily suppressing corrupt GrantDatas facilitates using normal loads and stores to wipe down uncorrectable errors.

### DIFF
--- a/src/main/scala/tile/CustomCSRs.scala
+++ b/src/main/scala/tile/CustomCSRs.scala
@@ -34,11 +34,11 @@ class CustomCSRs(implicit p: Parameters) extends CoreBundle {
 
   def flushBTB = getOrElse(bpmCSR, _.wen, false.B)
   def bpmStatic = getOrElse(bpmCSR, _.value(0), false.B)
-  def disableDCacheClockGate = getOrElse(chickenCSR, _.value(0), true.B)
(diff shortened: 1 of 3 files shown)


src/main/scala/tile/CustomCSRs.scala
```diff
@@ -38,6 +38,7 @@ class CustomCSRs(implicit p: Parameters) extends CoreBundle {
   def disableICacheClockGate = getOrElse(chickenCSR, _.value(1), true.B)
   def disableCoreClockGate = getOrElse(chickenCSR, _.value(2), true.B)
   def disableSpeculativeICacheRefill = getOrElse(chickenCSR, _.value(3), true.B)
+  def suppressCorruptOnGrantData = getOrElse(chickenCSR, _.value(9), true.B)
```
mwachs5: i just realized that if we DON'T have the chickenCSR, that means that all these bits are hard-coded to true? I guess I would have expected the inverse.

mwachs5: but otherwise given that this seems fine to me

aswaterman: Uh, I guess you're right.  It hasn't been an issue since the chickenCSR is always present in Rocket cores.  Might as well fix this for clarity.

PR #2280, opened 2020-02-07
bump FIRRTL for async reset register fixes

mwachs5:
**Related issue**: 
See discussion on https://github.com/freechipsproject/firrtl/pull/1355

^---- this template question makes no sense for a PR

**Impact**: API modification

**Release Notes**

Async registers which are only reset, never assigned, will be emitted properly in a way that prevents lint tools from complaining about a latch.

### DIFF
--- a/wit-manifest.json
+++ b/wit-manifest.json
@@ -15,7 +15,7 @@
         "source": "git@github.com:freechipsproject/chisel3.git"
     },
     {
-        "commit": "4b1f71e0c8a979ca9248a2ab62256ada12a683df",
+        "commit": "61bf6373bc752d1194f5fdc97db10297f82a0bb4",
         "name": "firrtl",
         "source": "git@github.com:freechipsproject/firrtl.git"
     },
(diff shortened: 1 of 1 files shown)


ernie-sifive:
@jackkoenig Does this FIRRTL issue ring any bells?
`%Warning-WIDTH: /home/travis/build/chipsalliance/rocket-chip/emulator/generated-src/freechips.rocketchip.system.WithJtagDTMSystem_DefaultConfig.v:184060: Operator NEGATE expects 14 bits on the LHS, but LHS's CONST '13'sh700' generates 13 bits.`

jackkoenig:
This one looks new, @azidar any ideas? I suspect one (or a combination of) the Verilog improvements are to blame.

azidar:
Yes it is. We can either revert some backported Verilog fixups to 1.2.x, or try to fix this with another verilog improvement transform.

ucbjrl:
There's some evidence pointing to freechipsproject/firrtl#1264

mwachs5:
Just to be clear I bumped this to what looked like the FIRRTL release branch, not FIRRTL master... was that the right thing to do?

Issue #867, opened 2017-07-17
rocket-chip release tag


yuz835:
Hi,

The rocket-chip github master branch is under development,
is there any plan to make stable release tags?

xfguo:
Agree! That's really important for commercial use.

ianperryman:
Yes this would be very helpful for evaluating the RISC-V architecture.  Releases will make it easier to discuss feature content and issues.

minmossi:
Any update?

jrpetrus:
I agree. It's really hard to evaluate the RocketChip master branch. The various submodules do not compiling without considerable hacking. I'm looking at you riscv-tools and chisel3.

bonannojj:
A related question (not sure if this justifies a separate issue): Is there any documentation about what versions of the RISC-V spec, and which extensions, the master branch implements?

sequencer:
The reason not releasing RC is because RC is so huge to main the API compatibility.
We are considering splitting RC into more small repos #2961 to maintain the API compatibility.
Also there is other needs to clean up dead code and documentation.
We will reach to a concrete plan to releasing RC, but it might not be the current RC, it might be diplomacy, rocket-core, tilelink, AMBA4, and other small components.

Issue #1289, opened 2018-03-16
Keep track for bug fixes?


joisty:
Hi,

While I was studying Rocket cores, I come up to following question.
As the Rocket is an open-source project, it is good that various useful patches
are being done daily, and Rocket is becoming more reliable core. However,
as Rocket is not a SW but a HW, at some point it will become a chip and it will
not be able to apply any further patches.
Basically, for most minor updates and patches that just clean-up codes or enhances
performance slightly, it does not really need to be  a big concern. However, from time
to time, there seems to be some patches related to critical functional bugs. 
These are different story.
If you already have your sillicon, of course, you cannot apply the patches even if it
will fix a critical bug. However, you should at least be easy to  keep tracks of which 
bugs will be appearing in your version of sillicon, to apply any kind of SW turn-arounds,
or even at least be aware of. 
This "issues" board possibly could be doing this role, but I do not think this board 
neccessarily holds all and only the "HW bugs".
I think it'd be the best if there already is some kind of bug-tracking platforms (e.g. errata),
that I just don't know yet. Otherwise it'be better if we can start making one.

Thanks.
Jamie

michael-etzkorn:
100% agree. WG has started a discussion about organizing tickets.

michael-etzkorn:
Closed by #3141 

RC-WG aims to keep a changelog as a resolution to this issue.

PR #2284, opened 2020-02-09
Create CHANGELOG.md and VERSIONING.md document

mwachs5:
**Related issue**: https://github.com/chipsalliance/rocket-chip/issues/867

**Impact**:  API modification

**Release Notes**

Suggest that we switch from **Release Notes** in the PR which seems to go into the ether to an actual Changelog a la

https://keepachangelog.com/en/1.0.0/

To my knowledge there isn't an actual "release" we can start from, but we could start from whenever this PR gets merged...

This also requires us to adopt an actual versioning strategy, captured in VERSIONING.md

### DIFF
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -44,7 +44,7 @@ Roadmap
 * Upcoming Versions:
 
   - Version ~2020.04.04.0: Released approximately before the breaking changes discussed for version NEXT:
-  - Version NEXT YY.MM.DD.0: To be released after:
+  - Version NEXT YY.0M.0D.0: To be released after:
     - Chisel is bumped to 3.3
     - FIRRTL is bumped to X.XX
     - Command line interface updated for staged generator
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -14,7 +14,6 @@ Version Naming
   * API Additions
   * Bug Fixes (which can result in changes in the emitted hardware considered bugs)
 * MINOR changes do NOT include:
-  * Changes in top-level I/Os for < level of hierarchy >
   * Backwards incompatible changes in the API (see below)
 
 API
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -36,6 +36,8 @@ The following is NOT part of the API:
   * Internals of scripts like vlsi-mem-gen, the contents of such scripts
   * Internals of any verilog black boxes
   * C Code ...?
+  * Makefiles and Makefile fragments
+  
       
 Roadmap
 -------
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,4 +1,4 @@
-Rocket Chip Versioning Strategy ** All Strawman to be discussed**
+Rocket Chip Versioning Strategy
 ===============================
 
 Version Naming
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -59,7 +59,18 @@ Development occurs for the next YYYY.0M.0D.0 release on the `master` branch.
 When the YYYY.0M.0D.0 is made, a protected branch is created with the name: `YYYY.0M.0D.X`.
 The MINOR releases will be made as tagged commits on `YYYY.0M.0D.X branch`.
 
+Naming Your Own Releases
+---------------------------
 
+
+If you would like to fork rocket-chip and release your own derivative versions, we suggest:
+  * If you fork and cut a specific release off an existing commit off this repo:
+    * Name it as <release>-SNAPSHOT-<git commit hash>
+    * For example if you want to make a specific release off commit hash `1234beef` which is a commit after the 2020.04.04.2 release, call it `2020.04.04.2-SNAPSHOT-1234beef`
+  * If you fork and modify a release branch and want to make a release:
+    * Name it as <release>-<identifier>-<arbitrary number with meaning decided by the entity doing the release>
+  * For example, if GitHub user `@mwachs5` did this: `2020.04.04.2-mwachs5-0`
+ * This document suggests no requirements/restrictions on the changes made in such forks
 
 Previous Versions
 -----------------
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -8,7 +8,7 @@ Version Naming
 * The format is YYYY.MM.DD.MINOR
 * MINOR changes include:
   * API Additions
-  * Bug Fixes (which can result in changes in the emitted hardware considered buges)
+  * Bug Fixes (which can result in changes in the emitted hardware considered bugs)
 * MINOR changes do NOT include:
   * Changes in top-level I/Os for < level of hierarchy >
   * Backwards incompatible changes in the API (see below)
@@ -52,4 +52,3 @@ Previous Versions
 -----------------
 
 Many 1.2.x versions have been released previously as needed.
-
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -25,10 +25,11 @@ The Rocket-Chip API is:
   * The version(s) of Chisel, FIRRTL, rocket-tools pointed to by rocket-chip.
   * Any code that generates IO(s) of any Module
   * The cmd line interface to scripts like vlsi-mem-gen
+  * `public val`s within `LazyModule`s which translate to diplomatic nodes (?? -- this could go either way)
 
 The following is NOT part of the API:
-  * public vals within `Module`s which translate to Wires, Regs, module instance names
-    * We note that this may cause issues for people using Cross-Module references
+  * `public val`s within `Module`s which translate to Wires, Regs, module instances
+    * (We note that this may cause issues for people using Cross-Module references)
   * Internals of scripts like vlsi-mem-gen, the contents of such scripts
   * Internals of any verilog black boxes
   * C Code ...?
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,0 +1,50 @@
+= Rocket Chip Versioning Strategy
+
+== Version Naming
+
+* Rocket Chip uses a Calendar Versioning release strategy (https://calver.org/).
+* The format is YYYY.MM.DD.MINOR
+* MINOR changes include:
+  * API Additions
+  * Bug Fixes (which can result in changes in the emitted hardware considered buges)
+* MINOR changes do NOT include:
+  * Changes in top-level I/Os for < level of hierarchy >
+  * Backwards incompatible changes in the API (see below)
+
+== API **Strawman -- to be discussed **
+
+The Rocket-Chip API is: 
+  * All public Scala methods and classes
+  * All public Scala methods and classes of included Scala libraries (hardfloat, CDE, etc)
+  * The command line used to invoke the generator
+  * All Verilog interfaces (I/Os and parameterization) for verilog files
+  * C code...?
+  * The version(s) of Chisel, FIRRTL, rocket-tools pointed to by rocket-chip.
+  * Any code that generates IO(s) of any Module
+  * The cmd line interface to scripts like vlsi-mem-gen
+
+The following is NOT part of the API:
+  * public vals within `Module`s which translate to Wires, Regs, module instance names
+    * We note that this may cause issues for people using Cross-Module references
+  * Internals of scripts like vlsi-mem-gen, the contents of such scripts
+  * Internals of any verilog black boxes
+  * C Code ...?
+      
+== Roadmap
+
+* Upcoming Versions:
+
+  - Version ~2020.04.04.0: Released approximately before the breaking changes discussed for version NEXT:
+  - Version NEXT YY.MM.DD.0: To be released after:
+    - Chisel is bumped to 3.3
+    - FIRRTL is bumped to X.XX
+    - Command line interface updated for staged generator
+    
+* Following versions released quarterly
+
+* We attempt to deprecate APIs gracefully over at least 2 quarterly releases
+
+== Previous Versions
+
+Many 1.2.x versions have been released previously as needed.
+
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -5,7 +5,11 @@ Version Naming
 --------------
 
 * Rocket Chip uses a Calendar Versioning release strategy (https://calver.org/).
-* The format is YYYY.MM.DD.MINOR
+* The format is YYYY.0M.0D.MINOR
+* Example:
+  * 2020.04.04.0 is the release which occurred on April 4, 2020.
+  * 2020.04.04.1 is a release that could occur at any time, but introduces only MINOR changes over the 2020.04.04.0 release
+  * 2020.06.01.0 is a release that occurs on June 1, 2020, with MAJOR (API breaking) changes over the 2020.04.04.X release.
 * MINOR changes include:
   * API Additions
   * Bug Fixes (which can result in changes in the emitted hardware considered bugs)
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -43,7 +43,7 @@ Roadmap
 * Upcoming Versions:
 
   - Version ~2020.04.04.0: Released approximately before the breaking changes discussed for version NEXT:
-  - Version NEXT YY.0M.0D.0: To be released after:
+  - Version NEXT YYYY.0M.0D.0: To be released after:
     - Chisel is bumped to 3.3
     - FIRRTL is bumped to X.XX
     - Command line interface updated for staged generator
@@ -51,6 +51,15 @@ Roadmap
 * Following versions released quarterly
 
 * We attempt to deprecate APIs gracefully over at least 2 quarterly releases, but do not guarantee this.
+
+Version Branches
+-----------------
+
+Development occurs for the next YYYY.0M.0D.0 release on the `master` branch.
+When the YYYY.0M.0D.0 is made, a protected branch is created with the name: `YYYY.0M.0D.X`.
+The MINOR releases will be made as tagged commits on `YYYY.0M.0D.X branch`.
+
+
 
 Previous Versions
 -----------------
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -1,6 +1,8 @@
-= Rocket Chip Versioning Strategy
+Rocket Chip Versioning Strategy ** All Strawman to be discussed**
+===============================
 
-== Version Naming
+Version Naming
+--------------
 
 * Rocket Chip uses a Calendar Versioning release strategy (https://calver.org/).
 * The format is YYYY.MM.DD.MINOR
@@ -11,7 +13,8 @@
   * Changes in top-level I/Os for < level of hierarchy >
   * Backwards incompatible changes in the API (see below)
 
-== API **Strawman -- to be discussed **
+API
+---
 
 The Rocket-Chip API is: 
   * All public Scala methods and classes
@@ -30,7 +33,8 @@ The following is NOT part of the API:
   * Internals of any verilog black boxes
   * C Code ...?
       
-== Roadmap
+Roadmap
+-------
 
 * Upcoming Versions:
 
@@ -44,7 +48,8 @@ The following is NOT part of the API:
 
 * We attempt to deprecate APIs gracefully over at least 2 quarterly releases
 
-== Previous Versions
+Previous Versions
+-----------------
 
 Many 1.2.x versions have been released previously as needed.
 
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -23,7 +23,7 @@ The Rocket-Chip API is:
   * All public Scala methods and classes
   * All public Scala methods and classes of included Scala libraries (hardfloat, CDE, etc)
   * The command line used to invoke the generator
-  * All Verilog interfaces (I/Os and parameterization) for verilog files
+  * All Verilog interfaces (I/Os and parameterization) for verilog resources
   * C code...?
   * The version(s) of Chisel, FIRRTL, rocket-tools pointed to by rocket-chip.
   * Any code that generates IO(s) of any Module
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -47,7 +47,7 @@ Roadmap
     
 * Following versions released quarterly
 
-* We attempt to deprecate APIs gracefully over at least 2 quarterly releases
+* We attempt to deprecate APIs gracefully over at least 2 quarterly releases, but do not guarantee this.
 
 Previous Versions
 -----------------
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -38,7 +38,14 @@ The following is NOT part of the API:
   * C Code ...?
   * Makefiles and Makefile fragments
   
-      
+API Dependencies on included libraries/dependencies:
+------------------------------------------------------
+
+* A new MINOR release may update `firrtl`, `chisel`, and `rocket-tools` to a newer minor release of those dependencies.
+* A new MAJOR release may update `firrtl`, `chisel`, and `rocket-tools` to a newer major release of those dependencies.
+
+* A new MINOR release may update `wit` dependencies listed in `wit-manifest.json` to newer minor release of those dependencies.
+* A new MAJOR release may update `wit` dependencies listed in `wit-manifest.json` to newer major release of those dependencies.
 Roadmap
 -------
 
--- a/VERSIONING.md
+++ b/VERSIONING.md
@@ -5,11 +5,11 @@ Version Naming
 --------------
 
 * Rocket Chip uses a Calendar Versioning release strategy (https://calver.org/).
-* The format is YYYY.0M.0D.MINOR
+* The format is YYYY.0M.MINOR
 * Example:
-  * 2020.04.04.0 is the release which occurred on April 4, 2020.
-  * 2020.04.04.1 is a release that could occur at any time, but introduces only MINOR changes over the 2020.04.04.0 release
-  * 2020.06.01.0 is a release that occurs on June 1, 2020, with MAJOR (API breaking) changes over the 2020.04.04.X release.
+  * 2020.04.0 is the release which occurred in April, 2020.
+  * 2020.04.1 is a release that could occur at any time, but introduces only MINOR changes over the 2020.04.0 release
+  * 2020.06.0 is a release that occurs in June 2020, with MAJOR (API breaking) changes over the 2020.04.X release(s).
 * MINOR changes include:
   * API Additions
   * Bug Fixes (which can result in changes in the emitted hardware considered bugs)
@@ -42,8 +42,8 @@ Roadmap
 
 * Upcoming Versions:
 
-  - Version ~2020.04.04.0: Released approximately before the breaking changes discussed for version NEXT:
-  - Version NEXT YYYY.0M.0D.0: To be released after:
+  - Version 2020.04.0: To be released just before the breaking changes discussed for version NEXT:
+  - Version NEXT YYYY.0M.0: To be released after:
     - Chisel is bumped to 3.3
     - FIRRTL is bumped to X.XX
     - Command line interface updated for staged generator
@@ -55,9 +55,9 @@ Roadmap
 Version Branches
 -----------------
 
-Development occurs for the next YYYY.0M.0D.0 release on the `master` branch.
-When the YYYY.0M.0D.0 is made, a protected branch is created with the name: `YYYY.0M.0D.X`.
-The MINOR releases will be made as tagged commits on `YYYY.0M.0D.X branch`.
+Development occurs for the next major YYYY.0M.0 release on the `master` branch.
+When the YYYY.0M.0 release is made, a protected (no force-pushing) branch is created with the name: `YYYY.0M.X`.
+The first MAJOR and subsequent MINOR releases will be made as tagged commits on `YYYY.0M.X branch`.
 
 Naming Your Own Releases
 ---------------------------
@@ -66,10 +66,10 @@ Naming Your Own Releases
 If you would like to fork rocket-chip and release your own derivative versions, we suggest:
   * If you fork and cut a specific release off an existing commit off this repo:
     * Name it as <release>-SNAPSHOT-<git commit hash>
-    * For example if you want to make a specific release off commit hash `1234beef` which is a commit after the 2020.04.04.2 release, call it `2020.04.04.2-SNAPSHOT-1234beef`
+    * For example if you want to make a specific release off commit hash `1234beef` which is a commit after the 2020.04.2 release, call it `2020.04.2-SNAPSHOT-1234beef`
   * If you fork and modify a release branch and want to make a release:
     * Name it as <release>-<identifier>-<arbitrary number with meaning decided by the entity doing the release>
-  * For example, if GitHub user `@mwachs5` did this: `2020.04.04.2-mwachs5-0`
+    * For example, if GitHub user `@mwachs5` did this: `2020.04.2-mwachs5-0`
  * This document suggests no requirements/restrictions on the changes made in such forks
 
 Previous Versions
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,5 +1,3 @@
-https://github.com/freechipsproject/firrtl/blob/master/.github/PULL_REQUEST_TEMPLATE.md
-
 ### Contributor Checklist
 
 <!-- - [ ] Did you add Scaladoc to every public function/method? -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,18 +1,54 @@
-<!--
-Please select the item best describing the pull request in each category and delete the other items.
--->
-**Related issue**: <!-- if applicable -->
+https://github.com/freechipsproject/firrtl/blob/master/.github/PULL_REQUEST_TEMPLATE.md
 
-<!-- choose one -->
-**Type of change**: bug report | feature request | other enhancement
+### Contributor Checklist
 
-<!-- choose one -->
-**Impact**: no functional change | API addition (no impact on existing code) | API modification
+<!-- - [ ] Did you add Scaladoc to every public function/method? -->
+<!-- - [ ] Did you add at least one test demonstrating the PR? -->
+- [ ] Did you delete any extraneous printlns/debugging code?
+- [ ] Did you link to any relevant issues / discussions?
+- [ ] Did you specify the type of improvement?
+- [ ] Did you state the API impact?
+- [ ] Did you specify the code generation impact?
+- [ ] Did you request a desired merge strategy?
+- [ ] Did you add text to be included in the Release Notes for this change in CHANGELOG.md?
 
-<!-- choose one -->
-**Development Phase**: proposal |  implementation
+#### Relevant Issues/Discussion 
 
-**Release Notes**
-<!--
-Text from here to the end of the body will be considered for inclusion in the release notes for the version containing this pull request.
--->
+<!-- Link to any issues/external discussions related to this PR -->
+
+#### Description of What was Changed
+
+#### Type of Improvement
+
+<!-- Choose one or more from the following: -->
+<!--   - bug fix                            -->
+<!--   - performance improvement            -->
+<!--   - documentation                      -->
+<!--   - code refactoring                   -->
+<!--   - code cleanup                       -->
+<!--   - backend code generation            -->
+<!--   - new feature/API                    -->
+
+#### API Impact
+
+<!-- How would this affect the current API? Does this add, extend, deprecate, remove, or break any existing API? -->
+
+#### Backend Code Generation Impact
+
+<!-- Does this change any generated Verilog?  -->
+<!-- How does it change it or in what circumstances would it?  -->
+
+#### Desired Merge Strategy
+
+<!-- If approved, how should this PR be merged? -->
+<!-- Options are: -->
+<!--   - Squash: The PR will be squashed and merged (choose this if you have no preference). -->
+<!--   - Rebase: You will rebase the PR onto master and it will be merged with a merge commit. -->
+
+
+### Reviewer Checklist (only modified by reviewer)
+<!-- - [ ] Did you add the appropriate labels? -->
+<!-- - [ ] Did you mark the proper milestone (1.2.x, 1.3.0, 1.4.0) ? -->
+- [ ] Did you review the changes?
+- [ ] Did you read the `CHANGELOG.md` description of the changes?
+<!-- - [ ] Did you mark as `Please Merge`? -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,7 +9,7 @@ https://github.com/freechipsproject/firrtl/blob/master/.github/PULL_REQUEST_TEMP
 - [ ] Did you specify the type of improvement?
 - [ ] Did you state the API impact?
 - [ ] Did you specify the code generation impact?
-- [ ] Did you request a desired merge strategy?
+<!-- - [ ] Did you request a desired merge strategy? -->
 - [ ] Did you add text to be included in the Release Notes for this change in CHANGELOG.md?
 
 #### Relevant Issues/Discussion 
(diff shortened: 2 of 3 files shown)


richardxia:
👍 on doing this, but we probably need to talk about what our release process/cadence/methodology is, or else this is just going to become an unbounded list of changes that is no different than the output of `git log`.

For a strawman, how about we try to do quarterly releases, maintain backwards compatibility for APIs for at least one release? This means holding off from removing any deprecated APIs until there has at least been one release with an official deprecation notice. I know this is very idealized, and that in practice, we may need to break backwards compatibility sooner, but perhaps it could at least be a goal we strive towards.

mwachs5:
> For a strawman, how about we try to do quarterly releases, maintain backwards compatibility for APIs for at least one release? This means holding off from removing any deprecated APIs until there has at least been one release with an official deprecation notice.

I think a quarterly release makes sense, but think 6 months is probably a more optimistic goal for keeping deprecated APIs. which would be a 2-quarter/release thing.

Another strawman to add to yours... we use Semantic Versioning ([https://semver.org/]), so the "quarterly release" would just have "whatever number we got up to at the end of that quarter"...

I guess we are assuming we are versioning the whole thing, vs separately versioning any components like diplomacy, tilelink, etc.

richardxia:
> I think a quarterly release makes sense, but think 6 months is probably a more optimistic goal for keeping deprecated APIs. which would be a 2-quarter/release thing.

Just to clarify what I meant, I was thinking of something like this:

Assuming releases are done on months that are divisible by 3:

- In March (3), we have some stable release
- In April, we decide that we want to get rid of some API
- In the June release (6), that API still exists but is _deprecated_, and we could perhaps even use Scala deprecation warnings to indicate that
- Only in the September (9) release do we actually remove that API.

> Another strawman to add to yours... we use Semantic Versioning ([https://semver.org/]), so the "quarterly release" would just have "whatever number we got up to at the end of that quarter"...

There is a little bit of interplay between SemVer and our deprecation policy. In SemVer, [marking an API as deprecated will increase the middle number](https://semver.org/#spec-item-7), and actually removing said API will increase the increase the first number. I think our deprecation policy would basically require us to hold off from removing deprecated APIs for at least 3 months, but it could take 6 months for it to affect an actual release.

richardxia:
> I guess we are assuming we are versioning the whole thing, vs separately versioning any components like diplomacy, tilelink, etc.

Yes, I agree with that.

mwachs5:
I wrote up some strawman proposals and added a VERSIONING.md. Greatly interested in others feedback on whether this would be possible / useful.

VERSIONING.md
```diff
@@ -0,0 +1,55 @@
...
+===============================
+
+Version Naming
+--------------
+
+* Rocket Chip uses a Calendar Versioning release strategy (https://calver.org/).
+* The format is YYYY.MM.DD.MINOR
+* MINOR changes include:
+  * API Additions
+  * Bug Fixes (which can result in changes in the emitted hardware considered buges)
```
mwachs5, line 11:
```suggestion
  * Bug Fixes (which can result in changes in the emitted hardware considered bugs)
```

VERSIONING.md
```diff
@@ -0,0 +1,50 @@
+= Rocket Chip Versioning Strategy
+
+== Version Naming
+
+* Rocket Chip uses a Calendar Versioning release strategy (https://calver.org/).
+* The format is YYYY.MM.DD.MINOR
```
ucbjrl, line 8: Just to be clear, the major release designation is the `YYYY.MM.DD` component. Minor releases will retain this designation (regardless of their release date), and simply bump the `MINOR` component. Perhaps a couple of examples would help clarify this.

mwachs5: ah yes good point

VERSIONING.md
```diff
@@ -0,0 +1,55 @@
...
+
+  - Version ~2020.04.04.0: Released approximately before the breaking changes discussed for version NEXT:
+  - Version NEXT YY.MM.DD.0: To be released after:
+    - Chisel is bumped to 3.3
+    - FIRRTL is bumped to X.XX
+    - Command line interface updated for staged generator
+    
+* Following versions released quarterly
+
+* We attempt to deprecate APIs gracefully over at least 2 quarterly releases
```
mwachs5, line 50:
```suggestion
* We attempt to deprecate APIs gracefully over at least 2 quarterly releases, but do not guarantee this.
```

tmagik:
6 months may not long enough given how long it's taken Berkeley to look at upgrading rocket-chip in https://github.com/ucb-bar/chipyard/tree/master/generators

However, if we had a set of projects that use rocket-chip and run them all through some sort of CI on every rocket-chip release we might be able to keep the 6 month cycle.

VERSIONING.md
```diff
@@ -0,0 +1,55 @@
+Rocket Chip Versioning Strategy ** All Strawman to be discussed**
+===============================
+
+Version Naming
+--------------
+
+* Rocket Chip uses a Calendar Versioning release strategy (https://calver.org/).
+* The format is YYYY.MM.DD.MINOR
```
mwachs5, line 8:
```suggestion
* The format is YYYY.MM.DD.MINOR
  * For example, 2020.04.04.0 is the release which occurred on April 4, 2020.
  * 2020.04.04.1 is a release that could occur at any time, but introduces only MINOR changes over the 2020.04.04 release
  * 2020.06.01.0 is a release that occurs on June 1, 2020, with major API breaking changes over the 2020.04.04.X release.
```

VERSIONING.md
```diff
@@ -0,0 +1,55 @@
...
+* MINOR changes do NOT include:
+  * Changes in top-level I/Os for < level of hierarchy >
+  * Backwards incompatible changes in the API (see below)
+
+API
+---
+
+The Rocket-Chip API is: 
+  * All public Scala methods and classes
+  * All public Scala methods and classes of included Scala libraries (hardfloat, CDE, etc)
```
grebe, line 21: It isn't clear to me if the definition of the Scala API implies either source or binary compatibility for minor version bumps (the discussion below about `public val`s in `Module`s seems to imply neither, which is a pragmatic choice but differs from how most scala projects work). I think it would be good to be specific as to whether one of (or any of) binary and source compatibility are intended to be preserved.

I see that mima is in plugins.sbt- is the intention for CI to report binary compatibility for PRs? That would be awesome, even if it can't always be preserved.

VERSIONING.md
```diff
@@ -0,0 +1,55 @@
...
+  * The command line used to invoke the generator
+  * All Verilog interfaces (I/Os and parameterization) for verilog files
+  * C code...?
+  * The version(s) of Chisel, FIRRTL, rocket-tools pointed to by rocket-chip.
```
grebe, line 25: Is the intention here to have an `=` or a `>=` relationship with the Chisel/FIRRTL/etc. versions?

mwachs5: i think my intention is if you bump the major version of CHisel/FIRRTL, that would count as a major (not MINOR) change. But bumping the Chisel/FIRRTL version to one that is also backwards compatible would be considered a MINOR change.

Does that answer your question? How can I make it more clear?

grebe: Ah, I see. That makes sense. I think for sbt this means an `=` relationship, i.e. a newer version of the dep does not satisfy the requirement. It also makes sense that major releases should happen together.

My main reason for asking about this is that sometimes different libraries could want different deps, for example my radio baseband might want a newer version of chisel than rocket, but I want to use them together. Because of the way binary compatibility works with chisel stuff, it would be hard to do the normal scala thing and have a `>=` dependency constraint. It's probably better to explicitly say you are not using a standard release (which may come with a recommendation that you re-run tests) if you are bumping deps.

Perhaps it should have subsections that say something like:

* A new minor release of rocket may update firrtl, chisel, and rocket-tools to a newer minor release of those dependencies. 
* A new major release of rocket may update firrtl, chisel, and rocket tools to a newer major release of those dependencies.

richardxia:
> 6 months may not long enough given how long it's taken Berkeley to look at upgrading rocket-chip in https://github.com/ucb-bar/chipyard/tree/master/generators

One could make the argument that the reason it takes a long time to upgrade rocket-chip is because we haven't been doing any versioning at all and not paying attention to or defining APIs. I think by going through this exercise, it'll make it much easier to upgrade between versions of rocket-chip. Once we have explicit, public APIs, it'll become easier for us to maintain backwards compatibility. Today, it's impossible to do that because it's not obvious to us who is even relying on which interfaces, especially since we in general have not developed a good habit of marking certain interfaces as `private` or `protected`.

VERSIONING.md
richardxia: I just want to mention that this violates two different parts of CalVer, so we probably shouldn't call this CalVer:

- Major, Minor, and Micro are used to describe the ordinal position of the version component, with Minor specifically being used to refer to the second position.
- They discourage the use of 4-component versions, and there is nothing that matches the component that you are indicating with the "MINOR" designation above.

VERSIONING.md
richardxia: This does not match the CalVer formatting conventions. Specifically:

> - MM - Short month - 1, 2 ... 11, 12
> - 0M - Zero-padded month - 01, 02 ... 11, 12

You should either update the string to say `YYYY.0M.0D` or update the examples to say `2020.4.4`.

VERSIONING.md
```diff
@@ -0,0 +1,55 @@
...
+  * Backwards incompatible changes in the API (see below)
+
+API
+---
+
+The Rocket-Chip API is: 
+  * All public Scala methods and classes
+  * All public Scala methods and classes of included Scala libraries (hardfloat, CDE, etc)
+  * The command line used to invoke the generator
+  * All Verilog interfaces (I/Os and parameterization) for verilog files
```
richardxia, line 23: Is this just of the top-level module or all modules?

VERSIONING.md
```diff
@@ -0,0 +1,55 @@
...
+  * C code...?
+  * The version(s) of Chisel, FIRRTL, rocket-tools pointed to by rocket-chip.
+  * Any code that generates IO(s) of any Module
+  * The cmd line interface to scripts like vlsi-mem-gen
+  * `public val`s within `LazyModule`s which translate to diplomatic nodes (?? -- this could go either way)
```
richardxia, line 28: How about the Makefiles/Makefrags?

Also, what about the Wake files and wit-manifest.json?

VERSIONING.md
```diff
@@ -0,0 +1,55 @@
...
+
+* Upcoming Versions:
+
+  - Version ~2020.04.04.0: Released approximately before the breaking changes discussed for version NEXT:
+  - Version NEXT YY.MM.DD.0: To be released after:
+    - Chisel is bumped to 3.3
+    - FIRRTL is bumped to X.XX
+    - Command line interface updated for staged generator
+    
+* Following versions released quarterly
```
richardxia, line 48: If we're planning to do quarterly releases, is it necessary to include the day of the month in the version string? I have doubts to our ability to predict the exact day on which we will cut a release, and I'm not sure if that level of granularity warrants inclusion in the version string.

VERSIONING.md
mwachs5: "Major, Minor, and Micro are used to describe the ordinal position of the version component, with Minor specifically being used to refer to the second position.
"
Hm, that didn't match my read of the calver link posted above, which called out several examples of projects using various combinations of these things, but perhaps i should read it more closely to see if they still considered those "calver" or "calver-inspired"

mwachs5: Ok, I see what you are saying. The only example on their site which violates the "dot-position of MAJOR/MINOR/MICRO defines what it is" is Teradata, which they explicitly call out a having a "useful combination of semvar and calver":
Screen Shot 2020-04-04 at 2 59 32 AM

mwachs5: There is also this one that seems to use an `_` to get around the "MICRO is in the wrong spot" but they don't give any further explanation on whether that is encouraged/discouraged:
Screen Shot 2020-04-04 at 3 03 47 AM

mwachs5: @richardxia do you think that we "shouldn't call this calver" or should change it until it matches calver (which presumably is a good thing)? For example I can't tell if using something like `YYYY_0M_0D.MINOR` would address your points since there aren't a lot of "underscore" examples.

VERSIONING.md
mwachs5: > It isn't clear to me if the definition of the Scala API implies either source or binary compatibility for minor version bumps (the discussion below about public vals in Modules seems to imply neither, which is a pragmatic choice but differs from how most scala projects work).

Yeah, this is a big question in my mind. On the one hand, just using "normal" Scala API makes a lot of sense, since in actuality every `val foo = Wire(...)` you create really is an API point for someone else even in scala-land (there are utilities to leverage that to generate cross-module references, etc).

On the other hand, the thing that often causes/cures pain for Chisel consumers is things like:
`val foo = Reg(...).suggestName("a_better_name")`
because it gives an API to the *output* of the generated code.
"a_better_name"  becomes a pretty important part of the API too, but will be harder to detect with binary compatibility.

I'll be honest that I am just getting familiarized with this topic myself, and would be interested in learning more how this binary compatability stuff should be done.

VERSIONING.md
mwachs5: There is no "top-level module" in rocket-chip which is a library of module generators. So, really I'd have to say "all modules". But yeah not sure how this could be enforced especially with diplomatic punching of things other than the previous requirement that all classes (and `Bundle`s are classes) stay the same, so maybe this is captured in that.

Actually most `val io = IO ( new Bundle {...})` are actually anonymous classes. Not sure how an anonymous class is considered in a "normal" scala API. Presumably if I added/removed a var or changed its type inside that `Bundle` that would be considered a change in the type of `val io` so would be considered a breaking change by my above rules?

In thinking about this more it's probably best to just use Scala standard definition of the API as it will probably capture *most* of the hardware use cases we think about.

VERSIONING.md
mwachs5: > How about the Makefiles/Makefrags?
> Also, what about the Wake files

Good point. Do you have a suggestion? How is this dealt with in other projects?

> and wit-manifest.json?

To me this is like asking "what about the submodules" which I tried to capture as well. Is there something special about wit-manifest.json that should be captured differently than submodules? In particular, shoudl we be explicit about changes to the version of wit/wake needed to deal with this repo (for example, if we make a change to this repo that requires you to bump to a new major version of wit/wake, in my mind that should be a MAJOR change).

Not sure the right classification if we make a change that requires you to bump to a new minor version of wit/wake. I think this is sort of what @grebe was discussing above RE chisel/firrtl, but it's a little inverted because rocket-chip does not *provide* wit/wake wheras rocket-chip is often used as the *provider* of chisel/firrtl to projects which use rocket-chip as a library.

VERSIONING.md
```diff
@@ -0,0 +1,59 @@
...
+  * All Verilog interfaces (I/Os and parameterization) for verilog files
+  * C code...?
+  * The version(s) of Chisel, FIRRTL, rocket-tools pointed to by rocket-chip.
+  * Any code that generates IO(s) of any Module
+  * The cmd line interface to scripts like vlsi-mem-gen
+  * `public val`s within `LazyModule`s which translate to diplomatic nodes (?? -- this could go either way)
+
+The following is NOT part of the API:
+  * `public val`s within `Module`s which translate to Wires, Regs, module instances
+    * (We note that this may cause issues for people using Cross-Module references)
```
mwachs5, line 36: I'm thinking i should remove this relaxation so that we can use more automated tooling to find changes

VERSIONING.md
```diff
@@ -0,0 +1,59 @@
...
+  * Internals of any verilog black boxes
+  * C Code ...?
+      
+Roadmap
+-------
+
+* Upcoming Versions:
+
+  - Version ~2020.04.04.0: Released approximately before the breaking changes discussed for version NEXT:
+  - Version NEXT YY.MM.DD.0: To be released after:
```
mwachs5, line 47:
```suggestion
  - Version NEXT YY.0M.0D.0: To be released after:
```

VERSIONING.md
mwachs5: I have zero intention of "predicting" the exact day we will cut a release. I think we should tag it when it happens
However I also think it's quite likely we will cut a release multiple times in a month (hoping that the features listed above happen before the end of April, as a very real example).

So maybe the bug in this section is really "following versions released quarterly" if the releases are going to really be feature-driven

VERSIONING.md
```diff
@@ -0,0 +1,59 @@
...
+* The format is YYYY.0M.0D.MINOR
+* Example:
+  * 2020.04.04.0 is the release which occurred on April 4, 2020.
+  * 2020.04.04.1 is a release that could occur at any time, but introduces only MINOR changes over the 2020.04.04.0 release
+  * 2020.06.01.0 is a release that occurs on June 1, 2020, with MAJOR (API breaking) changes over the 2020.04.04.X release.
+* MINOR changes include:
+  * API Additions
+  * Bug Fixes (which can result in changes in the emitted hardware considered bugs)
+* MINOR changes do NOT include:
+  * Changes in top-level I/Os for < level of hierarchy >
```
mwachs5, line 17:
```suggestion
  * Changes in top-level I/Os for < level of hierarchy >
```

deleting this because it's captured (or not) in the API below

VERSIONING.md
```diff
@@ -0,0 +1,58 @@
...
+* Example:
+  * 2020.04.04.0 is the release which occurred on April 4, 2020.
+  * 2020.04.04.1 is a release that could occur at any time, but introduces only MINOR changes over the 2020.04.04.0 release
+  * 2020.06.01.0 is a release that occurs on June 1, 2020, with MAJOR (API breaking) changes over the 2020.04.04.X release.
+* MINOR changes include:
+  * API Additions
+  * Bug Fixes (which can result in changes in the emitted hardware considered bugs)
+* MINOR changes do NOT include:
+  * Backwards incompatible changes in the API (see below)
+
```
mwachs5, line 18:
```suggestion

If you would like to fork rocket-chip and release your own derivative versions, we suggest:
  * If you fork and cut a specific release off an existing commit off this repo:
    * Name it as <release>-SNAPSHOT-<git commit hash>
    * For example if you want to make a specific release off commit hash `1234beef` which is the next commit after the 2020.04.04.2 release, call it `2020.04.04.2-SNAPSHOT-1234beef`
  * If you fork and modify a release branch and want to make a release:
    * Name it as <release>-<identifier>-<arbitrary number with meaning decided by the entity doing the release>
  * For example, if GitHub user `@mwachs5` did this: 2020.04.04.2-mwachs5-0
  * For example, a project like SiFive Freedom did this: 2020.04.04.2-sifivefreedom-0
  * This document imposes no requirements/restrictions on the changes made in such forks, those can be managed by the projects/users as they see fit.

```

VERSIONING.md
```diff
@@ -0,0 +1,67 @@
...
+* We attempt to deprecate APIs gracefully over at least 2 quarterly releases, but do not guarantee this.
+
+Version Branches
+-----------------
+
+Development occurs for the next YYYY.0M.0D.0 release on the `master` branch.
+When the YYYY.0M.0D.0 is made, a protected branch is created with the name: `YYYY.0M.0D.X`.
+The MINOR releases will be made as tagged commits on `YYYY.0M.0D.X branch`.
+
+
```
mwachs5, line 62:
```suggestion
Naming Your Own Releases
---------------------------


If you would like to fork rocket-chip and release your own derivative versions, we suggest:
  * If you fork and cut a specific release off an existing commit off this repo:
    * Name it as <release>-SNAPSHOT-<git commit hash>
    * For example if you want to make a specific release off commit hash `1234beef` which is a commit after the 2020.04.04.2 release, call it `2020.04.04.2-SNAPSHOT-1234beef`
  * If you fork and modify a release branch and want to make a release:
    * Name it as <release>-<identifier>-<arbitrary number with meaning decided by the entity doing the release>
  * For example, if GitHub user `@mwachs5` did this: `2020.04.04.2-mwachs5-0`
 * This document suggests no requirements/restrictions on the changes made in such forks

File: VERSIONING.md (hunk 13 above)
Comment (mwachs5) [reply VERSIONING.md:18]: I moved this further down

File: VERSIONING.md (hunk 2 above)
Comment (richardxia) [reply VERSIONING.md]: I think we should pop up a level to discuss what kinds of compatibility guarantees we want to provide and what our branching strategy and bugfix policy will be. Despite using dates in the format string, you seem to want to provide some notion of semantic versioning by adopting a definition of "minor" that is equivalent to SemVer's. If this is important to us, then we shouldn't obfuscate it using CalVer.

You can see that in the examples on the calver.org website, besides the `MICRO` version components, the projects their either fall into the category of "every non-micro release is guaranteed to break compatibility and we are not trying to address that at all" (e.g. Ubuntu) or "my library is basically permanently backwards compatible, but I am dependent on stuff that changes over time, like time zones or SSL certificate bundles". Most of the libraries on that page which have a `MICRO` version only really use it for bugfixes or security patches.

Note that you _are_ allowed to use SemVer while trying to stick to a date-based release schedule; you can fully adopt SemVer's versioning schemes, and instead allowing changes to dictate SemVer version bumps, you can let your release schedule dictate what kinds of changes are allowed to be made.

File: VERSIONING.md (hunk 7 above)
Comment (richardxia) [reply VERSIONING.md:26]: > There is no "top-level module" in rocket-chip which is a library of module generators.

Let me reword my question, since this is a bit more subtle than I had realized.

For a given "top" module (the actual module that the Chisel generation flow starts with or perhaps also any markDUT module), are the Verilog interfaces of the Verilog modules below it in that particular generation allowed to change?

For example, suppose I had the two Chisel modules (and please excuse the syntax I am making up because I don't actually know Chisel):

```scala
class OuterModule extends Module {
  val io = Bundle { val out = Output(Bool), val in1 = Input(Bool), val in2 = Input(Bool) }
  val inner_module = InnerModule()
  inner_module.io.in1 := io.in1
  inner_module.io.in2 := io.in2
  io.out := inner_module.io.out
}

class InnerModule extends Module {
  val io = Bundle { val out = Output(Bool), val in1 = Input(Bool), val in2 = Input(Bool) }
  io.out := io.in1 && io.in2
}
```

Now, suppose in a future version, we have the following change to `OuterModule`:

```diff
-  inner_module.io.in2 := io.in2
+  inner_module.io.in2 := false.asBool
```

Due to constant propagation, `InnerModule`s interface will likely change due to optimizing away the constant signal.

Now, if you instead chose `InnerModule` as your top-level module for a particular run of the generators, the IOs would _not_ change (I think?).

VERSIONING.md
richardxia: > > How about the Makefiles/Makefrags?
> > Also, what about the Wake files
> 
> Good point. Do you have a suggestion? How is this dealt with in other projects?

In my experience, most other software projects do not attempt to rely on the build scripts of their dependencies at all, and they are not part of the "interface". However, I was under the impression that some UCB folks or other open source users may be relying on them.

> Is there something special about wit-manifest.json that should be captured differently than submodules / the repos it points to?

I agree that it's mostly the same as the Git submodules. However, one big technical difference is that the `wit-manifest.json` describes a `>=` relationship, not an `==` relationship, so if in the `wit-manifest.json` you say that you depend on version A of a library, you are really saying that you depend on A or any version that has A as an ancestor. This does make it a little bit easier on versioning since users are allowed to use any version of the dependency that is newer.

VERSIONING.md
richardxia: > However at the current rate of change it's quite likely we will cut a release multiple times in a month (hoping that the features listed above happen before the end of April, as a very real example, though perhaps would stabilize after that?)

That was not the impression that I was getting from the document, since the only release cadence that was explicitly described was quarterly. Maybe the disconnect is that I was under the impression that SiFive internally would still be using SNAPSHOT/bleeding-edge releases, but that anyone (externally or internally) would have the option of using a more stable release created quarterly (and perhaps with bugfixes). There's always going to be a bit of a tradeoff between stability and new features, and that's why many large projects end up having different channels to cater to different needs.

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+  * Any code that generates IO(s) of any Module
+  * The cmd line interface to scripts like vlsi-mem-gen
+  * `public val`s within `LazyModule`s which translate to diplomatic nodes (?? -- this could go either way)
+
+The following is NOT part of the API:
+  * `public val`s within `Module`s which translate to Wires, Regs, module instances
+    * (We note that this may cause issues for people using Cross-Module references)
+  * Internals of scripts like vlsi-mem-gen, the contents of such scripts
+  * Internals of any verilog black boxes
+  * C Code ...?
```
mwachs5, line 38:
```suggestion
  * C Code ...?
  * Makefiles and Makefile fragments
  
```

VERSIONING.md
mwachs5: I'm coming around to agree it's pointless to put the day in there. 

Also, i think the concept of a "stable branch" is useful even if releases aren't getting cut from it per se.

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
+Rocket Chip Versioning Strategy ** All Strawman to be discussed**
+===============================
+
+Version Naming
+--------------
+
+* Rocket Chip uses a Calendar Versioning release strategy (https://calver.org/).
+* The format is YYYY.0M.0D.MINOR
```
mwachs5, line 8:
```suggestion
* The format is YYYY.0M.MINOR
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+* Example:
+  * 2020.04.04.0 is the release which occurred on April 4, 2020.
+  * 2020.04.04.1 is a release that could occur at any time, but introduces only MINOR changes over the 2020.04.04.0 release
+  * 2020.06.01.0 is a release that occurs on June 1, 2020, with MAJOR (API breaking) changes over the 2020.04.04.X release.
```
mwachs5, line 12:
```suggestion
* Example:
  * 2020.04.0 is the release which occurred in April, 2020.
  * 2020.04.1 is a release that could occur at any time, but introduces only MINOR changes over the 2020.04.0 release
  * 2020.06.0 is a release that occurs in June 2020, with MAJOR (API breaking) changes over the 2020.04.X release(s).
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+  * Backwards incompatible changes in the API (see below)
+
+API
+---
+
+The Rocket-Chip API is: 
+  * All public Scala methods and classes
+  * All public Scala methods and classes of included Scala libraries (hardfloat, CDE, etc)
+  * The command line used to invoke the generator
+  * All Verilog interfaces (I/Os and parameterization) for verilog files
```
mwachs5, line 26:
```suggestion
  * All Verilog interfaces (I/Os and parameterization) for verilog resources
```

VERSIONING.md
mwachs5: I realized that this comment was in particular on a line i meant to target the verilog resources in this file (black boxes), not generated verilog.

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+  * The cmd line interface to scripts like vlsi-mem-gen
+  * `public val`s within `LazyModule`s which translate to diplomatic nodes (?? -- this could go either way)
+
+The following is NOT part of the API:
+  * `public val`s within `Module`s which translate to Wires, Regs, module instances
+    * (We note that this may cause issues for people using Cross-Module references)
+  * Internals of scripts like vlsi-mem-gen, the contents of such scripts
+  * Internals of any verilog black boxes
+  * C Code ...?
+
```
mwachs5, line 39:
```suggestion
API Dependencies on included libraries/dependencies:
------------------------------------------------------

* A new MINOR release may update `firrtl`, `chisel`, and `rocket-tools` to a newer minor release of those dependencies.
* A new MAJOR release may update `firrtl`, `chisel`, and `rocket-tools` to a newer major release of those dependencies.

* A new MINOR release may update `wit` dependencies listed in `wit-manifest.json` to newer minor release of those dependencies.
* A new MAJOR release may update `wit` dependencies listed in `wit-manifest.json` to newer major release of those dependencies.
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+  * Internals of scripts like vlsi-mem-gen, the contents of such scripts
+  * Internals of any verilog black boxes
+  * C Code ...?
+      
+Roadmap
+-------
+
+* Upcoming Versions:
+
+  - Version ~2020.04.04.0: Released approximately before the breaking changes discussed for version NEXT:
```
mwachs5, line 45:
```suggestion
  - Version 2020.04.0: To be released just before the breaking changes discussed for version NEXT:
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+  - Version NEXT YYYY.0M.0D.0: To be released after:
```
mwachs5, line 46:
```suggestion
  - Version NEXT YYYY.0M.0: To be released after:
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+    - Command line interface updated for staged generator
+    
+* Following versions released quarterly
+
+* We attempt to deprecate APIs gracefully over at least 2 quarterly releases, but do not guarantee this.
+
+Version Branches
+-----------------
+
+Development occurs for the next YYYY.0M.0D.0 release on the `master` branch.
```
mwachs5, line 58:
```suggestion
Development occurs for the next major YYYY.0M.0 release on the `master` branch.
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+When the YYYY.0M.0D.0 is made, a protected branch is created with the name: `YYYY.0M.0D.X`.
```
mwachs5, line 59:
```suggestion
When the YYYY.0M.0 release is made, a protected (no force-pushing) branch is created with the name: `YYYY.0M.X`.
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+The MINOR releases will be made as tagged commits on `YYYY.0M.0D.X branch`.
```
mwachs5, line 60:
```suggestion
The first MAJOR and subsequent MINOR releases will be made as tagged commits on `YYYY.0M.X branch`.
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+
+Naming Your Own Releases
+---------------------------
+
+
+If you would like to fork rocket-chip and release your own derivative versions, we suggest:
+  * If you fork and cut a specific release off an existing commit off this repo:
+    * Name it as <release>-SNAPSHOT-<git commit hash>
+    * For example if you want to make a specific release off commit hash `1234beef` which is a commit after the 2020.04.04.2 release, call it `2020.04.04.2-SNAPSHOT-1234beef`
```
mwachs5, line 69:
```suggestion
    * For example if you want to make a specific release off commit hash `1234beef` which is a commit after the 2020.04.2 release, call it `2020.04.2-SNAPSHOT-1234beef`
```

VERSIONING.md
```diff
@@ -0,0 +1,78 @@
...
+  * If you fork and modify a release branch and want to make a release:
+    * Name it as <release>-<identifier>-<arbitrary number with meaning decided by the entity doing the release>
+  * For example, if GitHub user `@mwachs5` did this: `2020.04.04.2-mwachs5-0`
+ * This document suggests no requirements/restrictions on the changes made in such forks
```
mwachs5, line 73:
```suggestion
  * If you fork and modify a release branch and want to make a release:
    * Name it as <release>-<identifier>-<arbitrary number with meaning decided by the entity doing the release>
    * For example, if GitHub user `@mwachs5` did this: `2020.04.2-mwachs5-0`
 * This document suggests no requirements/restrictions on the changes made in such forks
```

VERSIONING.md
davidbiancolin: FWIW, we (chipyard & firesim) do not use Rocket Chip's checked-in makefrags but do use the generated ones.

VERSIONING.md
davidbiancolin: What do you mean here?

davidbiancolin:
> > 6 months may not long enough given how long it's taken Berkeley to look at upgrading rocket-chip in https://github.com/ucb-bar/chipyard/tree/master/generators
> 
> One could make the argument that the reason it takes a long time to upgrade rocket-chip is because we haven't been doing any versioning at all and not paying attention to or defining APIs. I think by going through this exercise, it'll make it much easier to upgrade between versions of rocket-chip. Once we have explicit, public APIs, it'll become easier for us to maintain backwards compatibility. Today, it's impossible to do that because it's not obvious to us who is even relying on which interfaces, especially since we in general have not developed a good habit of marking certain interfaces as `private` or `protected`.



> > 6 months may not long enough given how long it's taken Berkeley to look at upgrading rocket-chip in https://github.com/ucb-bar/chipyard/tree/master/generators
> 
> One could make the argument that the reason it takes a long time to upgrade rocket-chip is because we haven't been doing any versioning at all and not paying attention to or defining APIs. I think by going through this exercise, it'll make it much easier to upgrade between versions of rocket-chip. Once we have explicit, public APIs, it'll become easier for us to maintain backwards compatibility. Today, it's impossible to do that because it's not obvious to us who is even relying on which interfaces, especially since we in general have not developed a good habit of marking certain interfaces as `private` or `protected`.

This is more or less my position on the matter. 

Would the proposal also be to have RC developers use stricter access modifiers when writing new Chisel code?

davidbiancolin:
FireSim developer here; all-in-all this is a really cool proposal and i'm excited to see and use a versioned rocket-chip. My only concern pertains to the restriction on public `val`s. 

To be clear, i think the only sane thing to do is to mandate that these are not part of the RC API in general, however there are cases where such members should probably be included. We use cross module references (i.e., `instA.instB.member`) in Chisel to annotate key structures like memories (e.g., one backing a regfile) or to inject debug features on important signals without modifying RC sources. 

I don't think there needs to be many instances of this, but there should probably be a means to indicate certain public vals _are_ part of the API, perhaps simply a source comment, and will be subject to the standard deprecation schedule.

Somewhat related, it seems to me that @azidar 's AOP work is pretty reliant on having access to these aforementioned `val`s; am I mistaken?

davidbiancolin:
This is probably already captured but it's probably worthwhile to state explicitly that default values for all `config.Field`s constitute part of the API.

VERSIONING.md
mwachs5: I am editing the text to say "verilog resources", which may not really resolve this conversation but my intention was the files which we are explicitly including as verilog here.

VERSIONING.md
mwachs5: I *think* i am agreeing with your larger point below... people use the Scala API to do cross-module references, so we should NOT allow changes to `public vals` within Modules. This also lets us use normal binary compatability testing frameworks to detect these sort of changes.

None of this helps the person doing a verilog CMR, where things like annotations applied, .suggestName(s) applied, configuration settings might change, but this seems more straightforward to implement and audit to move this bullet up to the "part of the API" section above.

VERSIONING.md
mwachs5: i removed the day from the propsoed date string

VERSIONING.md
mwachs5: I added Makefiles/Makefrags to the "NOT part of the API" section, and then tried to clarify the wit-manifest dependencies behavior. Can you take a look @richardxia to see if what I wrote in https://github.com/chipsalliance/rocket-chip/pull/2284/files#diff-baa362176b63e6912fc6721235a68502R47
helps answer your question?

VERSIONING.md
mwachs5: @grebe take a look at https://github.com/chipsalliance/rocket-chip/pull/2284/files#diff-baa362176b63e6912fc6721235a68502R41 and let me know if this is more clear to you

mwachs5:
"This is probably already captured but it's probably worthwhile to state explicitly that default values for all config.Fields constitute part of the API."

-- are default values detectable by binary compatibility  tools (@jackkoenig, @davidbiancolin ?)

mwachs5:
@davidbiancolin i am leaning towards saying public vals of any type are part of the API. For CMR, and the AOP work you mentioned, and just general naming. If you were fixing a bug you could always deprecate the public val gracefully to by giving a pointer back to whatever you renamed the bug. Also, note this proposal isn't very strict about NOT making backwards incompatible changes except on release branches... just hoping to capture and track them better

davidbiancolin:
> @davidbiancolin i am leaning towards saying public vals of any type are part of the API. For CMR, and the AOP work you mentioned, and just general naming.

Well the compatibility tools will back you up on that decision. :)

However, worry that if you include public vals in the API you are increasing your API surface area so tremendously it'll just be too difficult to manage. I think in an ideal world you'd just "change" the default access modifier to private in all new Chisel modules added to RC, and then you'd just publicize certain members on an as needed basis.  It's a pity it's just so painful to write chisel like that. 

Chisel has in the past been able to name nodes assigned to protected and private vals, so that final concern can be fixed.

> Also, note this proposal isn't very strict about NOT making backwards incompatible changes except on release branches... just hoping to capture and track them better

Roger that.

VERSIONING.md
richardxia: That answers my question, but it sounds like chipyard and firesim use the generated Makefiles.

FWIW, SiFive does not use any of rocket-chip Makefiles/Makefrags/generated Makefiles internally, either.

richardxia:
Just had a thought pop in my mind. Suppose that we've adopted the versioning strategy described in this document, and that we have an existing 2020.04.0 release. If before the 2020.07.0 release we merge in a new API to master, but then later make a backwards-incompatible change to that API in master, we do _not_ need to document the backwards-incompatible change in 2020.07.0's release notes, correct?

Another way of interpreting the scenario I described above is that `master` is inherently unstable, and it is the act of actually creating a versioned release that "blesses" the API. We are still free to break that API, but the cost of doing so is now having to document it in the CHANGELOG?

mwachs5:
>  We are still free to break that API, but the cost of doing so is now having to document it in the CHANGELOG?

yes, on master I think that is the right way to think about it. Master is unstable.

I imagine that part of (in your example) actually making the 2020.07.0 release is that we would go through the CHANGELOG.md and essentially "squash" it. So if on master branch we had a CHANGELOG.md that was (as we went along under the "next release" section) we saw "add amazing new feature (#1234)" then "revert add amazing new feature (#1235)", in the 2020.07.0 CHANGELOG.md that shows up on the branch it would omit both those entries
Though that's obviously additional work for release manager, but auditing the release notes is going to have to be part of the release process anyway.

One thing I wanted to add to this PR was a more rigorous PR checklist that would show that I'd expect the CHANGELOG.md edits to be part of the PR process, not all piled on at the end in the release process.

richardxia:
> we would go through the CHANGELOG.md and essentially "squash" it.

Yeah, that makes a lot of sense.

mwachs5:
I added the updated PR template to this PR. I’m worried that every PR will have a conflict in the CHANGLOG.md, looking into if there is a way to hint to git that all conflicts are additions...

mwachs5:
Unfortunately, the whole "How to properly version control your CHANGELOG.md" is apparently a well-known problem:

https://github.com/olivierlacan/keep-a-changelog/issues/253#issuecomment-413014151

So may need to update the pull request template to capture the release notes then as part of the release we gather them into the CHANGELOG.md manually later :/

mwachs5:
I am looking at the commit history and it seems like 2cee947 should be the 2020.04.0 release, because after that were the major breaking changes intended for 2020.05.0 (which we could at this point make at any time).

( I was looking at this to try poking around at a release notes flow)

mwachs5:
This is a promising approach to the changelog maintenance... https://about.gitlab.com/blog/2018/07/03/solving-gitlabs-changelog-conflict-crisis/

michael-etzkorn:
Superseded by #3141 

For simplicity of maintenance, our versioning policy is an immature, ad-hoc take on `major.minor.patch` with no official VERSIONING.md for policy. The minor versions are not guaranteed to be API compatible, but the versioning and PR ideas of this PR could be reintroduced as we mature the WGs versioning and PR policies.

jackkoenig:
Immature and ad-hoc versioning is better than no versioning 🙂. I am glad to see rocket-chip moving forward with releases!

michael-etzkorn:
I'll add a viewpoint to the discussion here that while certainly the software side of rocket-chip (e.g. APIs) is important to consider as a versioning policy is matured, a main motivator for implementing versioning and a changelog was consideration of the hardware. Any verification policies are best implemented over frozen code. Any mask of a particular version of the code will retain whatever bugs are discovered and may only be patchable through a metal fix. Because of the hardware-impact, a well-matured versioning scheme will require something more nuanced involving hardware components rather than solely considering user APIs. 


https://github.com/chipsalliance/rocket-chip/issues/1289

PR #2285, opened 2020-02-10
Custommessage

brrmorre:
**Related issue**: 

**Release Notes**

Support for custom messages in Property class.

### DIFF
--- a/src/main/scala/formal/FormalUtils.scala
+++ b/src/main/scala/formal/FormalUtils.scala
@@ -46,12 +46,13 @@ object Property {
 
   def apply(dir: MonitorDirection, cond: Bool, message: String,
             prop_type: PropertyClass=PropertyClass.Default,
-            idx: String = "")(implicit sourceInfo: SourceInfo): Unit = {
+            idx: String = "",
+            custom_name: String = "")(implicit sourceInfo: SourceInfo): Unit = {
     val line_info = sourceInfo match {
       case SourceLine(filename, line, col) => s"${filename}_L${line}_C${col}_I${idx}".replace('.', '_')
       case _ => ""
     }
-    val proposed_src = prop_type.toString + "_" + line_info
+    val proposed_src = if (custom_name == "") prop_type.toString + "_" + line_info else custom_name
 
     val src_wrap = s"@[${proposed_src}]"
     if (dir==MonitorDirection.Monitor) {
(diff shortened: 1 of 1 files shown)


mwachs5:
what is an example of a custom message?

brrmorre:
> what is an example of a custom message?
Basically you can name assertions using the Property class. This is used on this branch here that I also want to merge.
https://github.com/sifive/federation/compare/custommessage

src/main/scala/formal/FormalUtils.scala
```diff
@@ -46,12 +46,13 @@ object Property {
...
             prop_type: PropertyClass=PropertyClass.Default,
-            idx: String = "")(implicit sourceInfo: SourceInfo): Unit = {
+            idx: String = "",
+            custom_name: String = "")(implicit sourceInfo: SourceInfo): Unit = {
     val line_info = sourceInfo match {
       case SourceLine(filename, line, col) => s"${filename}_L${line}_C${col}_I${idx}".replace('.', '_')
       case _ => ""
     }
-    val proposed_src = prop_type.toString + "_" + line_info
+    val proposed_src = if (custom_name == "") prop_type.toString + "_" + line_info else custom_name
```
jchang0: Line info seems to be helpful.  Can you still keep that somehow? (Maybe don't need it as part of a label, but add it to message or something).

src/main/scala/formal/FormalUtils.scala
```diff
@@ -46,12 +46,13 @@ object Property {
 
   def apply(dir: MonitorDirection, cond: Bool, message: String,
             prop_type: PropertyClass=PropertyClass.Default,
-            idx: String = "")(implicit sourceInfo: SourceInfo): Unit = {
+            idx: String = "",
```
jchang0: We should take the chance to actually git rid of this.  It used to prevent it from generating the same label, but firrtl have  a built in mechanism for that now.

src/main/scala/formal/FormalUtils.scala
brrmorre: Why/where would you want the line info if the person who wrote the assertion specified a specific name and a specific message? Can't you just search for that specific name in the codebase?

src/main/scala/formal/FormalUtils.scala
jchang0: As discussed, we should make this change in a separate PR.

PR #2287, opened 2020-02-11
Put timebase-frequency in the correct place

tmagik:
This is WIP to move the timebase-frequency DTS line from the
/cpus/cpu@X/ location to a single instance at /cpus/.

A final version for upstream should have a single entry, as
the current code generates something like:

timebase-frequency = <1000000 1000000>

as there seem to be multiple objects, which should all be checked
to be the same value with something like 'require'

**Related issue**: 

**Release Notes**


terpstra:
We used to have it where you propose to move it. It was changed. I would like more explanation before I sign off on this.

tmagik:
@terpstra @sequencer the linux kernel DTS standard (and code) require that the timebase-frequency item be in the top-level 'cpus' node. Can you point to the commit that moved it?

There may be good reasons to do it the way it's currently done in rocket-chip, however those reasons have never been publicly debated on any linux-kernel list I know of. It's also my understanding that anything that does not have the same timebase-frequency for all cpus nodes is not compliant with the RiscV spec.

PR #2294, opened 2020-02-13
Support for versioning in diplomacy

brrmorre:
**Related issue**: 

**Impact**: API modification

**Release Notes**

Support for versioning in the parameters in diplomacy


aswaterman:
what's the deal with the "removed comments" commit?

brrmorre:
> what's the deal with the "removed comments" commit?

I removed duplicate comments.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -10,32 +10,59 @@ import freechips.rocketchip.util.{RationalDirection,AsyncQueueParams, groupByInt
...
-  supportsAcquireB:   TransferSizes = TransferSizes.none,
-  supportsArithmetic: TransferSizes = TransferSizes.none,
-  supportsLogical:    TransferSizes = TransferSizes.none,
-  supportsGet:        TransferSizes = TransferSizes.none,
-  supportsPutFull:    TransferSizes = TransferSizes.none,
-  supportsPutPartial: TransferSizes = TransferSizes.none,
-  supportsHint:       TransferSizes = TransferSizes.none,
-  userBits:           Seq[UserBits] = Nil,
+  supportedSizesP:      TLSupportedSizes = TLSupportedSizes(),
+  emissionSizesP:       TLSupportedSizes = TLSupportedSizes(),
```
terpstra: Why not call these: supports and emits?

Then the API looks like: client.supports.arithmetic instead of client.supportsSizesP.arithmetic

... and supports should to none (like it does), but emits should default to TransferSizes(1, 4096) ... the safe default

PR #2295, opened 2020-02-14
Add vector extension to OMISA

aswaterman:


aswaterman:
Closes #2292, which was approved.  Had to re-PR to work around Travis failure.

PR #2298, opened 2020-02-19
 assertion for speculative fetch chicken bit

surajvellengar:

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/Frontend.scala
+++ b/src/main/scala/rocket/Frontend.scala
@@ -173,6 +173,7 @@ class FrontendModule(outer: Frontend) extends LazyModuleImp(outer)
   fq.io.enq.bits.btb := s2_btb_resp_bits
   fq.io.enq.bits.btb.taken := s2_btb_taken
   fq.io.enq.bits.xcpt := s2_tlb_resp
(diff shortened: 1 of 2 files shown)


src/main/scala/rocket/ICache.scala
```diff
@@ -400,6 +400,7 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
       }
   }
 
+  assert (!(io.s2_kill && tl_out.a.ready && tl_out.a.valid))
```
aswaterman: I don't think this is right.  tl_out.a.ready & valid can legitimately happen when s2_kill is high, for the purposes of sending a prefetch hint.

src/main/scala/rocket/Frontend.scala
```diff
@@ -173,6 +173,7 @@ class FrontendModule(outer: Frontend) extends LazyModuleImp(outer)
   fq.io.enq.bits.btb := s2_btb_resp_bits
   fq.io.enq.bits.btb.taken := s2_btb_taken
   fq.io.enq.bits.xcpt := s2_tlb_resp
+  assert(!(s2_speculative && io.ptw.customCSRs.asInstanceOf[RocketCustomCSRs].disableSpeculativeICacheRefill && !icache.io.s2_kill))
```
aswaterman: You can show statically that this assertion can never be true, so do we really need it?

PR #2305, opened 2020-02-25
Add option to change cache clean downgrade policy

### LINKED ISSUE #2119: DCache ProbeAckData (Exclusive Caching?)
**Type of issue**: feature request

**Development Phase**: request

**What is the current behavior?**

Rocket's DCache is configured such that on a `Probe` the choice of response being `ProbeAck` or `ProbeAckData` is a [function of the data being dirty or not](https://github.com/chipsalliance/rocket-chip/blob/b05fb5e68f96efa3243e7c28381dcbbfebca74a7/src/main/scala/tilelink/Metadata.scala#L116). (Technically, this is hardcoded into how any `ClientMetadata` responds to probes.)

This works fine for inclusive caching, but an exclusive cache where the L2 *does not* include a copy of the data would result in additional memory accesses. E.g., a clean downgrade from `Trunk` to `None` where ownership is transferring from one core's data cache to another core's data cache would (I think) require the L2 to fetch a new copy of the data from memory.

Alternatively, I'm missing something here or there's some clever approach to handle this with the L2.

**What is the expected behavior?**

It seems like the choice of how the DCache responds (either returning data with `ProbeAckData` or not returning data with `ProbeAck`) should be easily configurable. However, there's no option for this due to assumptions in `ClientMetadata`.

A top level DCache parameter seems initially attractive. However, this is likely better handled through diplomacy to make DCache behavior a function of the memory system it's connected to.

**What is the use case for changing the behavior?**

Better support for alternative memory hierarchies.

seldridge:
Adds a configuration option to rocket data caches to configure whether or not data is written back on a clean downgrade. This, for TileLink, means that the policy can be changed from the hard-coded "always `ProbeAck`" to "always `ProbeAckData`". 

This is implemented as an additional data cache parameter as well as a parameter used to construct `ClientMetadata` objects.

A subsystem level configuration partial function is included for mixing in this configuration.

**Related issue**: Fixes #2119 

**Release Notes**

- Add ability to configure DCache clean downgrade policy


aswaterman:
Usually this kind of parameterization would affect the cache behavior by way of the protocol configuration--i.e. the cache configuration object would not be aware of this detail.  @hcook ?

seldridge:
I think that's reasonable... Just so that I understand, this would be implemented as an additional member in `TLClientParameters`?

seldridge:
@hcook: (I missed your comment...) Thanks for the software architecture feedback. The structure you suggest makes sense and removes the unscalable parameter passing. Let me hack this up and rebase the PR.

seldridge:
@hcook: I updated and rebased based on the first four of your suggestions.

I introduced a small type hierarchy around `ClientMetadata` that tries to decouple the interface, `ClientMetadataLike`, from helpers to implement the interface. It's a combination of what you suggest.

For the fifth item, I haven't touched this and left the cache policy configuration as part of `DCacheParams`. A global policy object could work, too. That wouldn't be that hard to write. I do think that this specific writeback policy is a function of the outer memory hierarchy. That could, hyperbolically, be different on a per-cache basis. I'm not doing this, so either solution would fit my needs. What do you think?

seldridge:
Friendly ping @hcook.

seldridge:
Friendly ping @hcook. Any thoughts on the revised version? (I'm attempting to prevent divergence and the best way to do that is upstreaming early and often...)

seldridge:
The failure on Travis is just a remote url resolution problem that would pass with a restart. 

However, this is also failing locally when I'm testing the new configuration. (I'll need a little more time.)

seldridge:
This is rebased to actually build. The handling of the `ClientMetadata` factory needs to make sure that `Mux(sel, a, b)` works such that `a` and `b` have the same type as opposed to just being a subtype of `ClientMetadata`.

This then hits an intermittent TL monitor failure (`'C' channel carries ReleaseData type unsupported by manager (connected at HellaCache.scala:280:21)`) when always responding with `ProbeAckData`. The included `FooConfig` will fail in verilator with `--seed=0` and pass with `--seed=1` for `rv64ui-v-fence_i`.

I'm attempting to debug, but someone else may be able to pick out the problem more directly.

```
./emulator-freechips.rocketchip.system-FooConfig \
  --verbose \
  --cycle-count \
  --seed 0 \
  rv64ui-v-fence_i 2>&1 | spike-dasm | tail -n10
C0:     188813 [1] pc=[0000000000002bac] W[r14=0000000000000000][1] R[r14=0000000000000001] R[r31=00000000000014d7] inst=[fff70713] addi    a4, a4, -1
C0:     188814 [1] pc=[0000000000002bb0] W[r29=0000000000000000][0] R[r14=0000000000000000] R[r 0=0000000000000000] inst=[fe071ee3] bnez    a4, pc - 4
C0:     188818 [1] pc=[0000000000002bb4] W[r 5=0000000000002bb4][1] R[r 0=0000000000000000] R[r 0=0000000000000000] inst=[00000297] auipc   t0, 0x0
Assertion failed: 'C' channel carries ReleaseData type unsupported by manager (connected at CrossingHelper.scala:30:80)
    at Monitor.scala:44 assert(cond, message)
Assertion failed: 'C' channel carries ReleaseData type unsupported by manager (connected at HellaCache.scala:280:21)
    at Monitor.scala:44 assert(cond, message)
[189258] %Error: freechips.rocketchip.system.FooConfig.v:201431: Assertion failed in TOP.TestHarness.ldut.tile.buffer.TLMonitor
%Error: /home/schuyler/repos/github.com/freechipsproject/rocket-chip/emulator/generated-src/freechips.rocketchip.system.FooConfig.v:201431: Verilog $stop
Aborting...
```

```
./emulator-freechips.rocketchip.system-FooConfig \
  --cycle-count \
  --seed 1 \
  rv64ui-v-fence_i 2>&1 | spike-dasm
This emulator compiled with JTAG Remote Bitbang client. To enable, use +jtag_rbb_enable=1.
Listening on port 41701
*** PASSED *** Completed after 215936 cycles
```

src/main/scala/system/Configs.scala
```diff
@@ -24,6 +24,8 @@ class BaseConfig extends Config(
 
 class DefaultConfig extends Config(new WithNBigCores(1) ++ new BaseConfig)
 
+class FooConfig extends Config(new WithL1DCacheExclusiveL2 ++ new DefaultConfig)
```
hcook, line 27: Foo ;)

hcook:
I like the revised version, makes sense to parameterize the underlying helper and using the trait to override the version used plus copy method seems slick.

I'm not sure about your error, I think the problem might actually be in the TLManagerParameters offered by your L2. Are you claiming your manager supportsAcquireB?

seldridge:
@hcook wrote:

> I like the revised version, makes sense to parameterize the underlying helper and using the trait to override the version used plus copy method seems slick.

Nice. Thanks!

Note: it turns out I now may not need this so urgently, though I still think it makes sense to (eventually) upstream.

@hcook wrote:
> I'm not sure about your error, I think the problem might actually be in the TLManagerParameters offered by your L2. Are you claiming your manager supportsAcquireB?

For the error, this is running with the `DefaultConfig` coherency fabric (the default implementation of `BankedL2Params.coherenceManager`). So, the manager is supposed to support `acquireB`. However, I'm at a loss as to why this is failing. The address it's trying to release looks unexpected: `0x9d965600`. I think this is beyond main memory? And this is only failing for certain seeds.

wat

As I didn't remember this failing before, I bisected on Friday and found that, oddly, reversing the bit order of the LFSR used in the replacement policy (#2355) was the PR that seemed to introduce intermittency. However, this is obviously a red herring...

hcook:
Yeah, `DefaultConfig` main memory address range is `80000000 - 90000000  RWXC memory@80000000` and the monitor knows which transactions are allowed on which address ranges; you can't release to an address that doesn't exist. Since you acquired the block successfully in the first place, I would assume you are not releasing to the same address you acquired from. Why that is I am not sure, but it has to be inside the L1...

Revisiting your problem statement, you said you are attempting to "configure whether or not data is written back on a clean downgrade" ... but there are actually two separate concerns which I don't think the current code factors well irregardless of your changes:
1) Whether upon a clean downgrade a message should be sent at all (i.e. send Release[Data] vs silent eviction of the block)
2) Whether upon a clean downgrade the data should be sent along with the block even though it is clean (i.e. send ReleaseData instead of Release)
I think you want to change 2, but I believe you are in practice actually changing 1, except even more state machine changes are apparently needed since you are now sending a message in a case where no message would have been sent previously, and thus end up using some random address in your message. The current code definitely assumes that only ReleaseData's will ever be sent, as indicated by this line, which unconditionally uses [the edge method that includes a data payload](https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/rocket/DCache.scala#L782).

While it sounds like you too only want to ever send ReleaseDatas, I think this  dichotomy means the conditionality you are looking for is not adequately encapsulated within the signals driven by `shrinkHelper`. Put another way, `s2_victim_dirty` being driven high is apparently not sufficient to [guarantee `probe_bits.address` will be the address you want](https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/rocket/DCache.scala#L740)... Unfortunately the release state machine logic overall is beastly, since it is both very tightly coupled with the cache pipeline and also tries to minimize muxing logic on the outbound path. It's not clear to me whether `s2_victim_dirty` just needs to be tied into the logic somewhere else, or whether `shrinkHelper` needs to be expanded to have an addition return `Bool` to disambiguate sending a message from attaching data to that message...

PR #2306, opened 2020-02-26
bug fix for DeviceRegName

sequencer:

**Release Notes**
bug fix to comparing `Option[String]` with `String`
```
DiplomacyUtils.regName(x._1) == "control"
```
will compare between `Option[String]` and `String`, which never equal.
If comparing this equivalency, it should be
```
DiplomacyUtils.regName(x._1).contains("control")
```

### DIFF
--- a/src/main/scala/diplomacy/Resources.scala
+++ b/src/main/scala/diplomacy/Resources.scala
@@ -150,7 +150,7 @@ trait DeviceRegName
       devname
     } else {
       val (named, bulk) = reg.partition { case (k, v) => DiplomacyUtils.regName(k).isDefined }
-      val mainreg = reg.find(x => DiplomacyUtils.regName(x._1) == "control").getOrElse(reg.head)._2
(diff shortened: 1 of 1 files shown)


richardxia:
I agree that this is correct now, but it concerns me that it has had this behavior for three years, ever since that comparison was first added in https://github.com/chipsalliance/rocket-chip/commit/0bf46edb6c4cb4fcc81ee3b17c022f23f8dc876e. I am a little bit worried that making this change could cause a change in behavior for existing designs. I would almost prefer just removing the part of the expression that checks for the string "control", since it has never worked in the past. What do you all think?

By the way, is there a way to avoid this kind of problem in Scala?

sequencer:
> By the way, is there a way to avoid this kind of problem in Scala?

I'm using Intellij IDEA for code linting. It can also be achieved by Scala compiler plugin.
In `sbt`/`mill`, we can pass option to `scalaCompilerPlugin` to add these plugins.
I think bug like this can be analysed by https://github.com/sksamuel/scapegoat too.

PR #2308, opened 2020-02-26
WIP: adding scaladoc comments for Diplomacy (Nodes.scala).

richardxia:
**Related issue**: 

This is not really ready for a review, but I kind of wanted this to be an FYI to the both of you on this. I had a lot of spare cycles on my flight, and I just tried to dive right in to the Diplomacy codebase. This is speaking from the perspective of someone who doesn't really know any of our TileLink code, who doesn't really work on hardware, and who is approaching this from a much more graph theoretical/abstract functional programming types perspective.

I have to say, a lot of this code and its design is elegant in how simple it is (I am not using that sarcastically). I think the main stumbling blocks for people are probably due to how generic they are (everything is super type parameterized) and not understanding the mnemonics and metaphors that guide the naming conventions in this code.

I wanted to jot down more of the latter to help others with reading this.

Also, pulling in @jackkoenig in case if he has thoughts on how I should be representing that giant blob of comments at the beginning of the file. It doesn't exactly feel appropriate as a Scaladoc comment, since it's not about documenting individual classes or methods.

I am definitely planning to spend more time working on this.

**Release Notes**

### DIFF
--- a/src/main/scala/diplomacy/Nodes.scala
+++ b/src/main/scala/diplomacy/Nodes.scala
@@ -10,6 +10,90 @@ import freechips.rocketchip.util.HeterogeneousBag
 import scala.collection.mutable.ListBuffer
 import scala.util.matching._
 
+/**
+ * Concepts, metaphors, and mnemonics to help with understanding Diplomacy code.
+ *
+ *
+ * # Abstract types
+ *
+ * Diplomacy is a set of abstractions for describing directed, acyclic graphs
+ * where parameters may be negotiated between nodes. These abstractions are
+ * expressed in the form of abstract classes, traits, and type parameters, which
+ * comprise nearly all of the types defined in this file.
+ *
+ * The NodeImp ("node implementation") is the main abstract type that determines
+ * the type parameters of all other abstract types. Defining a concrete
+ * implementation of NodeImp will therefore determine concrete types for all
+ * type parameters. For example, passing in a concrete instance of NodeImp to a
+ * SourceNode will fully determine concrete types for all of a SourceNode's type
+ * parameters.
+ *
+ * Specific applications of Diplomacy are expected to either extend these types
+ * or to specify concrete types for the type parameters. This allows for
+ * creating application-specific node, edge, parameter, and bundle types.
+ *
+ * Is an "edge" and a binding the same thing?
+ *
+ *
+ * # Inward/Outward vs. Upward/Downward
+ *
+ * Diplomacy defines two dimensions: inward/outward and upward/downward.
+ *
+ * Inward/outward refer to the direction of the directed graph itself. For a
+ * given node:
+ * - Inward refers to edges that point into a node.
+ * - Outward refers to edges that point out of a node.
+ *
+ * A useful mnemonic for distinguishing between inward and outward is to always
+ * view it from the perspective of a particular node. Because Diplomacy
+ * describes directed, acyclic graphs, this direction will always be consistent
+ * across the entire graph.
+ *
+ * Upward/downward refer to the direction of the parameter negotiation, where
+ * the direction is relative to the inward/outward direction. For a given edge:
+ * - Upward refers to a flow of parameters in the outward direction.
+ * - Downward refers to a flow of parameters in the inward direction.
+ *
+ * A useful mnemonic for distinguishing between inward and outward is to imagine
+ * the diplomatic graph as a literal network of rivers where upward refers to
+ * parameters that move in the upstream direction while downward refers to
+ * parameters that move in the downstream direction.
+ *
+ *
+ * # Handles
+ *
+ * Two Diplomatic nodes can be bound together using the := operator or one of
+ * its sibling operators. Binding is asymmetric, and the binding operation will
+ * connect the outer side of one node to the inner side of the other.
+ *
+ * For example, the expression a := b will connect the outer side of b to the
+ * inner side of a.
+ *
+ * We would like the := operator to have additional properties which make it
+ * intuitive to use:
+ *
+ * 1. It should be chainable, so that a := b := c will have the intuitive effect
+ *    of binding c to b and b to a. This requires that return type of := be the
+ *    same as its arguments, because the result of one := operation must be
+ *    valid as an argument to other := operation.
+ *
+ * 2. It should be associative, so that (a := b) := c is equivalent to a := (b
+ *    := c). This means that the order in which the bind operations execute does
+ *    not matter, even if split across multiple files.
+ *
+ * 3. a := b should only be allowed if and only if b allows outward edges and a
+ *    allows inward edges. This should be preserved even when chaining
+ *    operations, and it should ideally be enforced at compile-time.
+ *
+ * Handles are a way of satisfying all of these properties. A Handle represents
+ * the aggregation of a chain of Nodes, and it preserves information about
+ * whether connectability of the innermost and the outermost sides of the chain.
+ *
+ * If b supports inward edges, then a := b returns a Handle that supports inward
+ * edges that go into b. If a supports outward edges, then a := b returns a
+ * Handle that supports outward edges coming out of a.
+ */
+
 case object MonitorsEnabled extends Field[Boolean](true)
 case object RenderFlipped extends Field[Boolean](false)
 
@@ -49,10 +133,21 @@ trait OutwardNodeImp[DO, UO, EO, BO <: Data]
   def getI(pd: DO): Option[BaseNode] = None // most-inward common node
 }
 
+/**
+ * A node implementation.
+ *
+ * This class has no members and is solely used for holding type information.
+ * Applications of Diplomacy should extend NodeImp with a case object that sets
+ * concrete type arguments.
+ */
 abstract class NodeImp[D, U, EO, EI, B <: Data]
   extends Object with InwardNodeImp[D, U, EI, B] with OutwardNodeImp[D, U, EO, B]
 
-// If your edges have the same direction, using this saves you some typing
+/**
+ * A NodeImp where the inward and outward edges are of the same type.
+ *
+ * If your edges have the same direction, using this saves you some typing.
+ */
 abstract class SimpleNodeImp[D, U, E, B <: Data]
   extends NodeImp[D, U, E, E, B]
 {
@@ -64,6 +159,9 @@ abstract class SimpleNodeImp[D, U, E, B <: Data]
   def bundleI(e: E) = bundle(e)
 }
 
+/**
+ * The base Node type at the top of the Node type hierarchy.
+ */
 abstract class BaseNode(implicit val valName: ValName)
 {
   val scope = LazyModule.scope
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomacy/Nodes.scala
```diff
@@ -49,10 +133,21 @@ trait OutwardNodeImp[DO, UO, EO, BO <: Data]
   def getI(pd: DO): Option[BaseNode] = None // most-inward common node
 }
 
+/**
+ * A node implementation.
+ *
+ * This class has no members and is solely used for holding type information.
+ * Applications of Diplomacy should extend NodeImp with a case object that sets
+ * concrete type arguments.
+ */
```
jackkoenig:
```suggestion
/**
  * A node implementation.
  *
  * This class has no members and is solely used for holding type information.
  * Applications of Diplomacy should extend NodeImp with a case object that sets
  * concrete type arguments.
  */
```
This applies to every one of these, if the following starts line up with the left `*` it is JavaDoc, you line up with the right `*` for ScalaDoc

richardxia: Looks like I need to go figure out how to change this in my vim settings. It was automatically inserting the `*` on each new line, aligned to the left one.

...And it looks like it's right there at the [bottom of the `vim-scala` README](https://github.com/derekwyatt/vim-scala#scaladoc-comment-indentation). Kind of annoying that it's not on by default.

jackkoenig:
> Also, pulling in @jackkoenig in case if he has thoughts on how I should be representing that giant blob of comments at the beginning of the file. It doesn't exactly feel appropriate as a Scaladoc comment, since it's not about documenting individual classes or methods.

For overarching documentation, the standard style seems to be either wholly separate documentation, or to put it on the package, eg. `scala.collection`: https://www.scala-lang.org/api/2.12.10/scala/collection/index.html. It is also pretty normal to put usage notes and examples that don't apply to a particular function on the relevant [often abstract] class, eg. `List`: https://www.scala-lang.org/api/2.12.10/scala/collection/immutable/List.html.

Glancing at the big comment you wrote, I think it would be reasonable to put it on `package diplomacy` with links to more information (and examples) on classes like `NodeImp`.

src/main/scala/diplomacy/Nodes.scala
richardxia: Fixed my editor settings in https://github.com/richardxia/dotfiles/commit/55518fbd1166ca7fde2266fabfdfcd14e831bfdf.

I'll remember to fix up all of these later.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -10,6 +10,90 @@ import freechips.rocketchip.util.HeterogeneousBag
...
+ * The NodeImp ("node implementation") is the main abstract type that determines
+ * the type parameters of all other abstract types. Defining a concrete
+ * implementation of NodeImp will therefore determine concrete types for all
+ * type parameters. For example, passing in a concrete instance of NodeImp to a
+ * SourceNode will fully determine concrete types for all of a SourceNode's type
+ * parameters.
+ *
+ * Specific applications of Diplomacy are expected to either extend these types
+ * or to specify concrete types for the type parameters. This allows for
+ * creating application-specific node, edge, parameter, and bundle types.
```
mwachs5:
```suggestion
 * creating and associating application-specific node, edge, parameter, and bundle types.
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -10,6 +10,90 @@ import freechips.rocketchip.util.HeterogeneousBag
...
+ * view it from the perspective of a particular node. Because Diplomacy
+ * describes directed, acyclic graphs, this direction will always be consistent
+ * across the entire graph.
+ *
+ * Upward/downward refer to the direction of the parameter negotiation, where
+ * the direction is relative to the inward/outward direction. For a given edge:
+ * - Upward refers to a flow of parameters in the outward direction.
+ * - Downward refers to a flow of parameters in the inward direction.
+ *
+ * A useful mnemonic for distinguishing between inward and outward is to imagine
```
mwachs5:
```suggestion
 * A useful mnemonic for distinguishing between upward and downward is to imagine
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -10,6 +10,90 @@ import freechips.rocketchip.util.HeterogeneousBag
...
+ * or to specify concrete types for the type parameters. This allows for
+ * creating application-specific node, edge, parameter, and bundle types.
+ *
+ * Is an "edge" and a binding the same thing?
+ *
+ *
+ * # Inward/Outward vs. Upward/Downward
+ *
+ * Diplomacy defines two dimensions: inward/outward and upward/downward.
+ *
```
mwachs5: So, is it correct to say that inward/outward refer to the way edges move, and upward/downward refer to the way parameters move? Is this a helpful summary?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -10,6 +10,90 @@ import freechips.rocketchip.util.HeterogeneousBag
...
+ * For example, the expression a := b will connect the outer side of b to the
+ * inner side of a.
+ *
+ * We would like the := operator to have additional properties which make it
+ * intuitive to use:
+ *
+ * 1. It should be chainable, so that a := b := c will have the intuitive effect
+ *    of binding c to b and b to a. This requires that return type of := be the
+ *    same as its arguments, because the result of one := operation must be
+ *    valid as an argument to other := operation.
```
mwachs5:
```suggestion
 *    valid as an argument to the other := operation.
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -10,6 +10,90 @@ import freechips.rocketchip.util.HeterogeneousBag
...
+ *
+ * 2. It should be associative, so that (a := b) := c is equivalent to a := (b
+ *    := c). This means that the order in which the bind operations execute does
+ *    not matter, even if split across multiple files.
+ *
+ * 3. a := b should only be allowed if and only if b allows outward edges and a
```
mwachs5:
```suggestion
 * 3. a := b should be allowed if and only if b allows outward edges and a
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -10,6 +10,90 @@ import freechips.rocketchip.util.HeterogeneousBag
...
+ *    allows inward edges. This should be preserved even when chaining
+ *    operations, and it should ideally be enforced at compile-time.
+ *
+ * Handles are a way of satisfying all of these properties. A Handle represents
+ * the aggregation of a chain of Nodes, and it preserves information about
+ * whether connectability of the innermost and the outermost sides of the chain.
```
mwachs5:
```suggestion
 * the connectability of the innermost and the outermost sides of the chain.
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -49,10 +133,21 @@ trait OutwardNodeImp[DO, UO, EO, BO <: Data]
...
+ * concrete type arguments.
+ */
 abstract class NodeImp[D, U, EO, EI, B <: Data]
   extends Object with InwardNodeImp[D, U, EI, B] with OutwardNodeImp[D, U, EO, B]
 
-// If your edges have the same direction, using this saves you some typing
+/**
+ * A NodeImp where the inward and outward edges are of the same type.
+ *
+ * If your edges have the same direction, using this saves you some typing.
```
mwachs5:
```suggestion
 * If your edges have the same type in both directions, using this saves you some typing.
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -125,18 +223,48 @@ trait FormatNode[I <: FormatEdge, O <: FormatEdge] extends BaseNode {
...
+ *
+ * left side is outer, right side is inner
+ *
+ * Two Handles can be bound if their adjacent ends are both <.
+ */
 trait NoHandle
 case object NoHandleObject extends NoHandle
 
+/**
+ * A Handle that can be on both sides of a bind operator.
```
mwachs5:
```suggestion
 * A Handle that can be used on either side of a bind operator.
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -10,6 +10,90 @@ import freechips.rocketchip.util.HeterogeneousBag
...
+ * A useful mnemonic for distinguishing between inward and outward is to imagine
+ * the diplomatic graph as a literal network of rivers where upward refers to
+ * parameters that move in the upstream direction while downward refers to
+ * parameters that move in the downstream direction.
+ *
+ *
+ * # Handles
+ *
+ * Two Diplomatic nodes can be bound together using the := operator or one of
+ * its sibling operators. Binding is asymmetric, and the binding operation will
```
mwachs5: Why did you decide to put so much here about the `:=` operator, but not go into `:=*` and friends? Did you just run out of time or was that a concious division? Later we talk about "ostars" without any intro

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -248,6 +415,13 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   val inward = this
   val outward = this
 
+  /**
+   * Given counts of known inward and outward binding and inward and outward
+   * star bindings, return the resolved inward stars and outward stars.
```
mwachs5: what is a "star binding"? We never talked about it. Not sure if that was intentional vs running out of time to type :)

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -560,7 +747,9 @@ class IdentityNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])()(imp
   } 
 }
 
-// EphemeralNodes disappear from the final node graph
+/**
+ * EphemeralNodes disappear from the final node graph
```
mwachs5: i never understood what this meant. So you put them beween two nodes and then it's like they were never there...?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -447,10 +624,18 @@ abstract class MixedCustomNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
+ * A Node that may be extended with custom behavior.
+ *
+ * Different from a MixedNode in that the inner and outer NodeImps are the same.
+ */
 abstract class CustomNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(
   implicit valName: ValName)
   extends MixedCustomNode(imp, imp)
 
+/**
+ *
```
mwachs5:
```suggestion
 * A node that can connect multiple inward edges of one NodeImp to multiple outward edges of a different NodeImp
```

src/main/scala/diplomacy/Nodes.scala
richardxia: I was sort of thinking of mentioning the other operators, but I think that probably goes into a different section that describes the operators in more detail. My primary goal for writing this section was to understand what a `Handle` is, why it exists, and something to help with why it's specifically named a "handle". I'd also want to run everything I've written here by @terpstra, since everything I wrote is just me trying to "create a theory of handles" from reading the code, and I'm not sure if the design principles I post hoc derived are actually what @terpstra's goals were.

This Handle stuff is only really a concept that you'd need to understand if you were trying to deeply understand the code here, but it's not necessarily something you'd need to know as a user of a library implemented in Diplomacy. For example, anyone using the Diplomatic TileLink framework needs to understand `:=`, `:=*`, etc., but they wouldn't need to know anything about a Handle.

For what it's worth, the last part of the code that I got to on the plane while trying to was the four different `NodeBinding` types and their associated operators, but I had not gotten to the point of actually comprehending the code yet, since the code is recursive, meaning that in order to understand `resolveStar()`, you need to understand how `resolveStar()` works on the neighboring nodes. This is a bit challenging when you don't have internet or any prior knowledge on using any of the operators besides the plain "bind-once" `:=` operator.

src/main/scala/diplomacy/Nodes.scala
richardxia: This was a running out of time (or really, the passing out due to being awake for 20 hours) thing. This is where I got to when I hit the "Understanding NodeBindings and running into a recursive block" described above.

src/main/scala/diplomacy/Nodes.scala
richardxia: Note that I didn't touch this at all besides changing the plain comment to a Scaladoc comment. I have not yet gotten to a point where I understand what an EphemeralNode is or what it should be used for.

richardxia:
> Thank you for doing this! Do you mind if i push directly to this branch for the Nodes which you are missing?

That works for me, but I think @sequencer has a more complete list of Scaladoc comments for the classes here, so I wonder if we should work on getting his work integrated in first?

sequencer:
Sure! I'll PR these documentation tonight.

sequencer:
comments of `LazyModule` are added, I'll work on `Nodes.scala` tomorrow.

src/main/scala/diplomacy/Nodes.scala
sequencer: > My primary goal for writing this section was to understand what a Handle is, why it exists, and something to help with why it's specifically named a "handle".

I think handle is implemented as its name.
```
nodeA := nodeB
```
will return a new handle with `nodeA` as `outwardNode` and `nodeB` as `inwardNode`.
If `nodeB` doesn't have a `inwardNode`, `bind` will just return a `OutwardNodeHandle` which only have a `outwardNode` `nodeA`.

sequencer:
closed since #2560.

PR #2312, opened 2020-02-27
Correct VCD dump-start

colinschmidt:

$dumpvars starts dumping so in order to delay the start we need to turn off dumping before later enabling it

Relevant portion of verilog manual:
```
18.1.3 Stopping and resuming the dump ($dumpoff/$dumpon)
Executing the $dumpvars task causes the value change dumping to start at the end of the current simulation time unit
```


colinschmidt:
I've tested this down stream it behaves as expected.

PR #2318, opened 2020-03-04
BundleMap: improved API for user bits

terpstra:
- Field unification can be customized and defaults to equality required
- Bulk assignment of BundleMaps now uses defaults for missing fields
- You must now define both a Field and Key class

**Impact**: API modification

**Release Notes**
User bits API composability improved


terpstra:
@jackkoenig In case you were planning on including these operators in chisel3; note the two improvements over the old version:
- scala type equality is required, but you can reverse which type is needed with a trailing :
- you can provide overload the bulk assignment operators for your type

PR #2325, opened 2020-03-07
Fix D$ miss performance counter

haoozi:
**Type of change**: Performance counter enhancement

**Release Notes**

D$ miss event is originally linked to outgoing requests from D$. This assumption is not correct when a D$ prefetcher exists, memory request caused by prefetcher should not be considered as D$ miss.


johningalls-sifive:
@haoozi your question could be rephrased as "should prefetch cache misses be counted as cache misses?"

I agree with @aswaterman that we should leave the current event as-is, and create other events to sub-divide "cache miss" due to load/store/ptw/prefetch/etc.

haoozi:
That would also be fine. Adding more event may take some time to talk, should I close this PR?

PR #2326, opened 2020-03-10
User bits overhaul

terpstra:
Completely replace user bits in every rocket-chip diplomatic bus. One application is that we can now carry AMBA-style protection bits between AMBA adapters over a TileLink-based SoC.

**Impact**: API modification


hcook:
Looks good to me, modulo what we discussed about intersection vs union of the master and slave sides.

`BundleMap` is really powerful.

Issue #2078, opened 2019-08-15
What is the meaning and function of fbus(front bus)?


clx782782:
I don’t know well what is the definition of “front bus” and what is the function of it.

And what is the difference between the “fbus” and the other buses?

Any suggestions will be appreciated.

michael-etzkorn:
FrontBus is simply a TL crossbar for external TL requests. This reduces contention to the system bus with the rocket cores. 

A hierarchical design for the buses in the SoC is a useful categorization that allows for making decisions for each category on things like power and bandwidth maintenance. 

https://chipyard.readthedocs.io/en/latest/Generators/Rocket-Chip.html shows an example of how they're all connected in the default coherent topology.

Closing this issue. Further follow-up can be made by opening a question on StackOverflow.

Issue #2025, opened 2019-07-06
Difference between crossToBus and crossFromBus


Tong-Lin:
I changed cross function between cbus and sbus like follow:
```
trait HasHierarchicalBusTopology { this: BaseSubsystem =>
  // The sbus masters the cbus; here we convert TL-UH -> TL-UL
  //sbus.crossToBus(cbus, NoCrossing)
  cbus.crossFromBus(sbus, SynchronousCrossing(FastToSlow))
```
It because that I want adapt modules like 'TLWidthWidget' and 'TLFragmenter' been instanced in cbus which is slower side; And then got following error:
> [error] (run-main-0) java.lang.reflect.InvocationTargetException
> [error] java.lang.reflect.InvocationTargetException
> [error]         at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
> [error]         at java.base/jdk.internal.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
> [error]         at java.base/jdk.internal.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
> [error]         at java.base/java.lang.reflect.Constructor.newInstance(Constructor.java:490)
> [error]         at freechips.rocketchip.util.HasGeneratorUtilities.$anonfun$elaborate$1(GeneratorUtils.scala:54)
> [error]         at chisel3.core.Module$.do_apply(Module.scala:52)
> [error]         at chisel3.Driver$.$anonfun$elaborate$1(Driver.scala:95)
> [error]         at chisel3.internal.Builder$.$anonfun$build$2(Builder.scala:343)
> [error]         at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
> [error]         at chisel3.internal.Builder$.$anonfun$build$1(Builder.scala:341)
> [error]         at scala.util.DynamicVariable.withValue(DynamicVariable.scala:58)
> [error]         at chisel3.internal.Builder$.build(Builder.scala:341)
> [error]         at chisel3.Driver$.elaborate(Driver.scala:95)
> [error]         at freechips.rocketchip.util.HasGeneratorUtilities.elaborate(GeneratorUtils.scala:59)
[...]

PR #2327, opened 2020-03-10
Make TLBusWrappers' topology a configurable Field

hcook:
### Overview

This PR changes the selection of the topology of TLBusWrappers that are included in the `BaseSubsystem` layer of a Rocket-Chip-based design from being a static `trait` that is mixed in at compile time to a child class of `BaseSubystem` to being a  dynamic choice driven at runtime from a configurable `Field`. This change means that the choice of internal network-on-chip topology no longer affects the Scala class hierarchy of Subsystems. Instead, a `Field` is supplied by which users can create a topology on a per-`Config` basis. The description of the topology is intended to be readable from a serializable format, although the current implementation maintains some programmatic escape hatches. Any number of `TLBusWrappers` in any connectivity network shape is possible, though the requirements of diplomatic TileLink must still be satisfied (no cycles, single apparent path to each address from each agent, etc).

A topological `Config` alteration is named `WithCoherentBusTopology` is supplied to recreate the extant, default, 5-hierarchical-crossbars topology from `DefaultConfig`+`BaseSubsystem with HasHierarchicalBusTopology`. The `HasHierarchicalBusTopology` trait itself is removed, and any child class that extends `BaseSubsystem` only to supply a different topology from that one should now be redundant (can just extend `BaseSubsystem` and supply a different value for the `Field` `TLNetworkTopologyLocated(where: HierarchicalLocation)`. Three included example sub-topologies are included, which build on each other additively in order to recreate the original 5-xbar design. The three sub-topologies are `JustOneBusTopologyParams` (`sbus`), `HierarchicalBusTopologyParams` (`fbus`, `cbus`, `pbus`) and `CoherentBusTopologyParams` (adds a coherence manager and `mbus`).

### API Additions and Changes

What is changed:
- Use recently-added classes `Location` and `LocationMap`: These dynamic maps can store arbitrary things indexable by "location" (just a string name for now). For example, `TLBusWrapper` references are stored in named locations in `HasTileLinkLocations .tlBusWrapperLocationMap`.
- Added `trait HasTileLinkLocations`: intended to be an even more generic version of the extant `Attachable` trait, which now extends it.
- `HasTileLinkLocations` provides `def locateTLBusWrapper` function for child devices to locate a `TLBusWrapper` instance inside a parent that extends this trait based on a location label.
- `TLBusWrapperTopology` is the new intended serializable entry point for describing bus topologies, see below.
- `CoherenceManagerWrapper` is a new wrapper for all collateral related to attaching coherence managers within a topology, allowing use of the `TLBusWrapper` connectivity and clock crossing APIs. This will change the instance hierarchy of the coherence manager in existing designs by adding another module hierarchy layer wrapping it and its associated adapters together.
- `HasHierarchicalBusTopology` trait is removed, but configuration equivalents `WithCoherentBusTopology` and `WithIncoherentBusTopology` are provided to be applied as alterations to `DefaultConfig` instead.
- A new class of runtime error becomes possible, where a device or wrapper with a configurable attachment or connection location tries to use a location for an instance that actually does not exist in a given configured topology. So it goes with dynamic configurability...

What is not changed:
- Accessor `val` members of `BaseSubsystem` named `sbus`, `fbus`, `cbus` etc. still exist; I would like to deprecate them in the near future in favor of using only the `locateTLBusWrapper` method. 
- @rmac-sifive's recent changes to `sifive-blocks` are compatible with this PR
- Users can still make `BaseSubsystem` subclasses with statically defined subnetwork graphs and connect them to the dynamically-created topology. The dynamic topology can have nothing in it (though such a configuration will fail to build e.g. `DefaultConfig` in this repo). Set `TLNetworkTopologyLocated("InSubsystem") => up(TLNetworkTopologyLocated("InSubsystem")).copy(connections = Nil)` in your `Config` class to wire the standard bus locations together by hand.

### Intended Usage

`TLBusWrapperTopology` is a serializable representation of a TileLink-based interconnect. A topology comprises instances of `TLBusWrapper`, a (previously existing) class which is an abstraction around a piece of interconnect fabric that provides some kind of connectivity between TileLink endpoints. There are currently several subtypes of `TLBusWrapper`, which each contain a particular crossbar, set of crossbars, chain of TL adapters, and/or internal devices. (In the future, some of these subtypes may just become parameterizations of a single base type.) `TLBusWrappers` have helper functions for attaching themselves to endpoint devices or to each other with configurable clock crossings. 

The new `TLBusWrapperTopology` representation has two components, `instances` and `connections`. 

Each `instance` is tuple that contains a `Location` and a `TLBusWrapperInstantiationLike`, where the former is a named reference to the latter, which can later be used when making connections between instances (or when attaching other things to the instance). A `TLBusWrapperInstantiationLike` is any class that knows how to instantiate a `TLBusWrapper`. The four previously-existing `[Name]BusParams` case classes all inherit from it and instantiate some particular `TLBusWrapper` subtype; the parameters are things like bus data width and parameters for internal devices and adapters' control interfaces.

Each `connection` is a triple that contains a pair of `Locations` and a `TLBusWrapperConnectionLike`, where the former pair identifies two previously instantiated `TLBusWrappers`, and the latter provides a method for actually programmatically connecting the instances. There is currently a single case class for describing parameterized connection between `TLBusWrappers`, named `TLBusWrapperConnection`.

A `TLBusWrapperConnection` contains the following serializable parameters:
- `xType: ClockCrossingType`: what type of TL clock crossing adapter to insert between the buses. The appropriate half of the crossing adapter ends up inside each bus.
- `driveClockFromMaster: Option[Boolean]`: if `None`, don't bind the bus's diplomatic `clockGroupNode`, otherwise have either the master or the slave bus bind the other one's `clockGroupNode`, assuming the inserted crossing type is not asynchronous.
- `nodeBinding: NodeBinding`: fine-grained control of multi-edge cardinality resolution for diplomatic bindings within the connection.
- `flipRendering: Boolean`: fine-grained control of the graphML rendering of the connection.
There are also three programmatic escape hatches allowing users to inject custom node chains into the connection, or control the specific point within the `TLBusWrapper` that is connected to or from on either side.

Three example sub-topologies are defined, `JustOneBusTopologyParams` (`sbus`), `HierarchicalBusTopologyParams` (`fbus`, `cbus`, `pbus`) and `CoherentBusTopologyParams` (adds a coherence manager and `mbus`). These can be combined additively.

At the moment, the `BaseSubsystem` class looks up the `TLNetworkTopologyLocated("InSubsystem")` `Field` to discover its configurable topology. This call into the `Parameters` object could be replaced with reading in the topology from some serialized format. Given a topology, the subsystem first instantiates all instances, then connects all connections, passing in itself as the instantiating or connecting scope:
```
topology.foreach(_.instantiate(this))
topology.foreach(_.connect(this))
```

**Related issues**: #2034 #2025 #2078 

**Impact**:  API modification (but largely backward compatible)


src/main/scala/subsystem/BusTopology.scala
```diff
@@ -2,29 +2,61 @@
...
 
+import freechips.rocketchip.config.{Field, Parameters}
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.tilelink._
+import freechips.rocketchip.util.Location
 
-trait HasHierarchicalBusTopology { this: BaseSubsystem =>
-  sbus.clockGroupNode := asyncClockGroupsNode
+// These fields control parameters of the five traditional tilelink bus wrappers
+case object SystemBusKey extends Field[SystemBusParams]
```
mwachs5: so is the idea that these will be deprecated in the future?

hcook: These particular ones I am fine with leaving in for backward compatibility for awhile, they're pretty harmless. The analogy here is: we still allow `Config` users to set `RocketTilesKey` manually, even though there are also ways of setting that `Field` based on reading other configuration formats...

src/main/scala/subsystem/BusTopology.scala
```diff
@@ -2,29 +2,79 @@
...
+) extends TLBusWrapperTopology(
+  instantiations = List(
+    (PBUS, pbus),
+    (FBUS, fbus),
+    (CBUS, cbus)),
+  connections = List(
+    (SBUS, CBUS, TLBusWrapperCrossToConnection  (xTypes.sbusToCbusXType)()),
+    (CBUS, PBUS, TLBusWrapperCrossToConnection  (xTypes.cbusToPbusXType)()),
+    (SBUS, FBUS, TLBusWrapperCrossFromConnection(xTypes.fbusToSbusXType)()))
+)
```
rmac-sifive: 🎉👍

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -40,65 +32,39 @@ abstract class BareSubsystemModuleImp[+L <: BareSubsystem](_outer: L) extends La
...
     p(SubsystemDriveAsyncClockGroupsKey)
       .map(_.drive(asyncClockGroupsNode))
       .getOrElse(InModuleBody { HeterogeneousBag[ClockGroupBundle](Nil) })
 
+  // Find the topology configuration for the TL buses located in this subsystem.
+  // Calling these functions populates tlBusWrapperLocationMap and connects the locations to each other.
+  val location = HierarchicalLocation("InSubsystem")
+  val topology = p(TLNetworkTopologyLocated(location.name))
+  topology.foreach(_.instantiate(this))
+  topology.foreach(_.connect(this))
```
rmac-sifive: Similar to the stuff I'm moving to `sifive-blocks`, maybe this should live in a trait?

hcook: good idea, i did this

mwachs5:
@hcook can you take a look at https://github.com/chipsalliance/rocket-chip/pull/2410 and see if that approach is advised given that this change is coming? Would like to get those OM changes in before this change if there is huge incompatibility. Since that PR is basically assuming `BaseSubsystem` has 4 `bus` variables wondering how it will interact with this

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -45,72 +37,59 @@ abstract class BareSubsystemModuleImp[+L <: BareSubsystem](_outer: L) extends La
...
+
+  // TODO must there really always be an "sbus"?
+  val sbus = tlBusWrapperLocationMap(SBUS)
+  tlBusWrapperLocationMap.lift(SBUS).map { _.clockGroupNode := asyncClockGroupsNode }
+
+  // TODO deprecate these public members to see where users are manually hardcoding a particular bus that might actually not exist in a certain dynamic topology
+  val pbus = tlBusWrapperLocationMap.lift(PBUS).getOrElse(sbus)
+  val fbus = tlBusWrapperLocationMap.lift(FBUS).getOrElse(sbus)
+  val mbus = tlBusWrapperLocationMap.lift(MBUS).getOrElse(sbus)
+  val cbus = tlBusWrapperLocationMap.lift(CBUS).getOrElse(sbus)
```
mwachs5, line 90: is there a reason there is no `sbus` in this list?

mwachs5: oh NVM i see it is above

jerryhethatday:
is there some more detailed infomation I can get about the TLBusWrapper class?

PR #2329, opened 2020-03-10
Revert staged generator PRs

debs-sifive:
**Related issue**: #2322 #2309 #2274

**Impact**: API modification - use prior Makefile conventions

**Release Notes**

This is causing backwards compatibility issues. Will be re-PRed in the future.


mwachs5:
I think this will also resolve this SO ticket:

https://stackoverflow.com/questions/60479390/sbt-trapexitsecurityexception-error-message-in-from-chisel-rocket-chip/60626145#60626145

debs-sifive:
It should, but as Wes diagnosed, we will want to fix this in ChiselStage in the future, since we want to un-revert these changes ASAP as well.

PR #2332, opened 2020-03-12
Add hardware-perf-monitor-counters to DTS

mwachs5:
**Related issue**: 

**Type of change**:  feature request

**Release Notes**

DTS now contains additional property `hardware-perf-monitor-counters` in the `cpu` node

Example: 
```
L15: cpus {
#address-cells = <1>;
#size-cells = <0>;
L7: cpu@0 {
...
hardware-perf-monitor-counters = <2>;
...
};
};

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -214,7 +214,8 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
       "clock-frequency"      -> tileParams.core.bootFreqHz.asProperty,
       "riscv,isa"            -> isaDTS.asProperty,
       "timebase-frequency"   -> p(DTSTimebase).asProperty,
-      "hardware-exec-breakpoint-count" -> tileParams.core.nBreakpoints.asProperty
+      "hardware-exec-breakpoint-count" -> tileParams.core.nBreakpoints.asProperty,
+      "hardware-perf-monitor-counters" -> tileParams.core.nPerfCounters.asProperty
   )
 
   // The boundary buffering needed to cut feed-through paths is

### CODE REVIEW

**Comment by aswaterman** [comment]
I’m somewhat opposed to this sort of thing on principle. I have been telling SiFive’s software team that using precise exceptions to detect feature presence is the way to go. We architected things this way on purpose, so that software can readily discover features in this manner. So we might as well rely upon that, rather than adding redundant info to DT.

Tangentially, adding SiFive-specific extensions to DT is politically fraught. We really should be clearing our DT extensions with the Linux kernel folks, rather than acting unilaterally.

I have been work with @bsousi5 to adopt my test-by-trapping approach, and he seems to think it’s workable. So far, we have had success in using it for similar things, and I think we should follow that pattern here, too.

**Comment by bsousi5** [comment]
Software can test for the number of counters available, but it seems odd that software has to do this since it is a user/design configurable option. This make sense for Linux since Device Tree in Linux is used to describe non-discoverable hardware. However, in Embedded where memory is limited and code size is an important factor, have additional code to test for hardware may not be best. Given the incompatible requirements, SiFive software team would need to not use the DTS from the rocket-chip generator, but derive its own DTS or some other representation for Embedded software.

**Comment by aswaterman** [comment]
It is not odd, IMO. Software tests for all sorts of stuff about the hardware, of both coarse and fine grain. (Software designed for other systems asks questions like “do I have an NVIDIA GPU connected over PCIe?” or “does the CLZ instruction return the correct result with a zero input?”)

We all agree that self-discovery of PCIe devices is totally normal. And that’s not just out of necessity: self-discovery of PCIe devices is used even in systems where all the bus connections are hardwired.

I find it hard to accept the argument that processor features are somehow different. Processor features are the easy case in this respect! The self-discovery mechanism is already defined, and is lightweight. (Certainly lighter-weight than defining new and redundant APIs.)

Code size is a legitimate argument, but I don’t think it’s a winner in this case. The extra code to search the DT for these entries is large relative to the code to probe a CSR.

**Comment by aswaterman** [comment]
BTW, here is an efficient routine to count the number of counters during the M-mode boot process.

```
  .globl num_mhpmcounters
num_mhpmcounters:

  la t0, 1f
  csrrw t2, mtvec, t0
  auipc t0, 0

  csrr x0, mhpmcounter3
  csrr x0, mhpmcounter4
  csrr x0, mhpmcounter5
  csrr x0, mhpmcounter6
  csrr x0, mhpmcounter7
  csrr x0, mhpmcounter8
  csrr x0, mhpmcounter9
  csrr x0, mhpmcounter10
  csrr x0, mhpmcounter11
  csrr x0, mhpmcounter12
  csrr x0, mhpmcounter13
  csrr x0, mhpmcounter14
  csrr x0, mhpmcounter15
  csrr x0, mhpmcounter16
  csrr x0, mhpmcounter17
  csrr x0, mhpmcounter18
  csrr x0, mhpmcounter19
  csrr x0, mhpmcounter20
  csrr x0, mhpmcounter21
  csrr x0, mhpmcounter22
  csrr x0, mhpmcounter23
  csrr x0, mhpmcounter24
  csrr x0, mhpmcounter25
  csrr x0, mhpmcounter26
  csrr x0, mhpmcounter27
  csrr x0, mhpmcounter28
  csrr x0, mhpmcounter29
  csrr x0, mhpmcounter30
  csrr x0, mhpmcounter31

  li a0, 31
  j 2f

  .align 2
1:
  csrr a0, mepc
  sub a0, a0, t0
  srl a0, a0, 2
  addi a0, a0, -1
2:
  csrw mtvec, t2
  ret
```

aswaterman:
BTW, there is an efficient way to check how many counters exist - write -1 to `mcounteren` and read it back to see what the index of the highest bit set is.

bsousi5:
Nice. This is more elegant. Thanks.

mwachs5:
@bsousi5 are you OK if i close this PR then?

bsousi5:
> @bsousi5 are you OK if i close this PR then?

Yes, @mwachs5  you can close the PR.

mwachs5:
closed since there is SW alternatives to determining this vs putting it into DTS

PR #2338, opened 2020-03-14
Use inferred resets for JTAG and DebugTransport modules

ernie-sifive:
**Related issue**: 

**Impact**: API modification

**Release Notes**

This PR changes the DebugTransport and JTAG modules to use abstract reset.  The jtag_reset signal is driven with an AsyncReset signal in tieoffDebug.  Old-style explicit instantiations of neg-edge registers and async-reset registers have been replaced by standard registers wrapped with the appropriate withClockAndReset.

### DIFF
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -644,15 +644,12 @@ class TLDebugModuleOuterAsync(device: Device)(implicit p: Parameters) extends La
       val hartResetReq = p(DebugModuleKey).get.hasHartResets.option(Output(Vec(nComponents, Bool())))
       val dmAuthenticated = p(DebugModuleKey).get.hasAuthentication.option(Input(Bool()))
     })
-    val rf_reset = IO(Input(Reset()))
+    val rf_reset = IO(Input(Reset()))    // RF transform
 
     childClock := io.dmi_clock
     childReset := io.dmi_reset
 
     withClockAndReset(childClock, childReset) {
-      dmOuter.module.clock := io.dmi_clock
-      dmOuter.module.reset := io.dmi_reset
-
       dmi2tlOpt.foreach { _.module.io.dmi <> io.dmi.get }
 
       dmiBypass.module.io.bypass := ~io.ctrl.dmactive | ~AsyncResetSynchronizerShiftReg(in=io.ctrl.dmactiveAck, sync=3, name=Some("dmactiveAckSync"))
@@ -1710,7 +1707,7 @@ class TLDebugModuleInnerAsync(device: Device, getNComponents: () => Int, beatByt
       val hartIsInReset = Input(Vec(getNComponents(), Bool()))
       val auth = p(DebugModuleKey).get.hasAuthentication.option(new DebugAuthenticationIO())
     })
-    val rf_reset = IO(Input(Reset()))
+    val rf_reset = IO(Input(Reset()))    // RF transform
 
     childClock := io.debug_clock
     childReset := io.debug_reset
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -629,11 +629,13 @@ class TLDebugModuleOuterAsync(device: Device)(implicit p: Parameters) extends La
   val dmiInnerNode = TLAsyncCrossingSource() := dmiBypass.node := dmiXbar.node
   dmOuter.dmiNode := dmiXbar.node
   
-  lazy val module = new LazyModuleImp(this) {
+  lazy val module = new LazyRawModuleImp(this) {
 
     val nComponents = dmOuter.intnode.edges.out.size
 
     val io = IO(new Bundle {
+      val dmi_clock = Input(Clock())
+      val dmi_reset = Input(Reset())
       val dmi   = (!p(ExportDebug).apb).option(Flipped(new DMIIO()(p)))
       // Optional APB Interface is fully diplomatic so is not listed here.
       val ctrl = new DebugCtrlBundle(nComponents)
@@ -644,16 +646,23 @@ class TLDebugModuleOuterAsync(device: Device)(implicit p: Parameters) extends La
     })
     val rf_reset = IO(Input(Reset()))
 
-    dmi2tlOpt.foreach { _.module.io.dmi <> io.dmi.get }
+    childClock := io.dmi_clock
+    childReset := io.dmi_reset
 
-    dmiBypass.module.io.bypass := ~io.ctrl.dmactive | ~AsyncResetSynchronizerShiftReg(in=io.ctrl.dmactiveAck, sync=3, name=Some("dmactiveAckSync"))
+    withClockAndReset(childClock, childReset) {
+      dmOuter.module.clock := io.dmi_clock
+      dmOuter.module.reset := io.dmi_reset
 
-    io.ctrl <> dmOuter.module.io.ctrl
-    io.innerCtrl <> ToAsyncBundle(dmOuter.module.io.innerCtrl, AsyncQueueParams.singleton(safe=cfg.crossingHasSafeReset))
-    dmOuter.module.io.hgDebugInt := io.hgDebugInt
-    io.hartResetReq.foreach { x => dmOuter.module.io.hartResetReq.foreach {y => x := y}}
-    io.dmAuthenticated.foreach { x => dmOuter.module.io.dmAuthenticated.foreach { y => y := x}}
+      dmi2tlOpt.foreach { _.module.io.dmi <> io.dmi.get }
 
+      dmiBypass.module.io.bypass := ~io.ctrl.dmactive | ~AsyncResetSynchronizerShiftReg(in=io.ctrl.dmactiveAck, sync=3, name=Some("dmactiveAckSync"))
+
+      io.ctrl <> dmOuter.module.io.ctrl
+      io.innerCtrl <> ToAsyncBundle(dmOuter.module.io.innerCtrl, AsyncQueueParams.singleton(safe=cfg.crossingHasSafeReset))
+      dmOuter.module.io.hgDebugInt := io.hgDebugInt
+      io.hartResetReq.foreach { x => dmOuter.module.io.hartResetReq.foreach {y => x := y}}
+      io.dmAuthenticated.foreach { x => dmOuter.module.io.dmAuthenticated.foreach { y => y := x}}
+    }
   }
 }
 
@@ -1784,14 +1793,14 @@ class TLDebugModule(beatBytes: Int)(implicit p: Parameters) extends LazyModule {
 
     dmOuter.module.io.dmi.foreach { dmOuterDMI =>
       dmOuterDMI <> io.dmi.get.dmi
-      dmOuter.module.reset := io.dmi.get.dmiReset
-      dmOuter.module.clock := io.dmi.get.dmiClock
+      dmOuter.module.io.dmi_reset := io.dmi.get.dmiReset
+      dmOuter.module.io.dmi_clock := io.dmi.get.dmiClock
       dmOuter.module.rf_reset := io.dmi.get.dmiReset
     }
 
     (io.apb_clock zip io.apb_reset)  foreach { case (c, r) =>
-      dmOuter.module.reset := r
-      dmOuter.module.clock := c
+      dmOuter.module.io.dmi_reset := r
+      dmOuter.module.io.dmi_clock := c
       dmOuter.module.rf_reset := r
     }
 
--- a/src/main/scala/devices/debug/DebugTransport.scala
+++ b/src/main/scala/devices/debug/DebugTransport.scala
@@ -65,24 +65,29 @@ class DTMInfo extends Bundle {
 /** A wrapper around JTAG providing a reset signal and manufacturer id. */
 class SystemJTAGIO extends Bundle {
   val jtag = Flipped(new JTAGIO(hasTRSTn = false))
-  val reset = Input(Bool())
+  val reset = Input(Reset())
   val mfr_id = Input(UInt(11.W))
   val part_number = Input(UInt(16.W))
   val version = Input(UInt(4.W))
 }
 
 class DebugTransportModuleJTAG(debugAddrBits: Int, c: JtagDTMConfig)
-  (implicit val p: Parameters) extends Module  {
+  (implicit val p: Parameters) extends RawModule  {
 
   val io = IO(new Bundle {
+    val jtag_clock = Input(Clock())
+    val jtag_reset = Input(Reset())
+    val tapReset = Input(Bool())
     val dmi = new DMIIO()(p)
     val jtag = Flipped(new JTAGIO(hasTRSTn = false)) // TODO: re-use SystemJTAGIO here?
-    val jtag_reset = Input(Bool())
     val jtag_mfr_id = Input(UInt(11.W))
     val jtag_part_number = Input(UInt(16.W))
     val jtag_version = Input(UInt(4.W))
     val fsmReset = Output(Bool())
   })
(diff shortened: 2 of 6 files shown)


src/main/scala/devices/debug/Debug.scala
```diff
@@ -644,16 +646,23 @@ class TLDebugModuleOuterAsync(device: Device)(implicit p: Parameters) extends La
...
     val rf_reset = IO(Input(Reset()))
 
-    dmi2tlOpt.foreach { _.module.io.dmi <> io.dmi.get }
+    childClock := io.dmi_clock
+    childReset := io.dmi_reset
 
-    dmiBypass.module.io.bypass := ~io.ctrl.dmactive | ~AsyncResetSynchronizerShiftReg(in=io.ctrl.dmactiveAck, sync=3, name=Some("dmactiveAckSync"))
+    withClockAndReset(childClock, childReset) {
+      dmOuter.module.clock := io.dmi_clock
+      dmOuter.module.reset := io.dmi_reset
```
mwachs5:
```suggestion
```

I *think* that since you are basically assigning these to `childClock/childReset`, you don't need to say that. That is the rationale for having `childClock` and `childReset`, all Lazy submodules should get that clock/reset automatically. Did you try without this and have any issues?

src/main/scala/devices/debug/Debug.scala
```diff
@@ -644,16 +646,23 @@ class TLDebugModuleOuterAsync(device: Device)(implicit p: Parameters) extends La
...
 
-    io.ctrl <> dmOuter.module.io.ctrl
-    io.innerCtrl <> ToAsyncBundle(dmOuter.module.io.innerCtrl, AsyncQueueParams.singleton(safe=cfg.crossingHasSafeReset))
-    dmOuter.module.io.hgDebugInt := io.hgDebugInt
-    io.hartResetReq.foreach { x => dmOuter.module.io.hartResetReq.foreach {y => x := y}}
-    io.dmAuthenticated.foreach { x => dmOuter.module.io.dmAuthenticated.foreach { y => y := x}}
+      dmi2tlOpt.foreach { _.module.io.dmi <> io.dmi.get }
+
+      dmiBypass.module.io.bypass := ~io.ctrl.dmactive | ~AsyncResetSynchronizerShiftReg(in=io.ctrl.dmactiveAck, sync=3, name=Some("dmactiveAckSync"))
```
mwachs5: but I think you are right you need to put these non-lazy modules in explicit withClockAndReset blocks, but actually I am not sure. @jackkoenig , @terpstra do you know if this would automatically get the `childClock` and `childReset` domain without a withClockAndReset(childClock, childReset)?

src/main/scala/devices/debug/DebugTransport.scala
```diff
@@ -65,24 +65,29 @@ class DTMInfo extends Bundle {
 /** A wrapper around JTAG providing a reset signal and manufacturer id. */
 class SystemJTAGIO extends Bundle {
   val jtag = Flipped(new JTAGIO(hasTRSTn = false))
-  val reset = Input(Bool())
+  val reset = Input(Reset())
```
mwachs5: is there ever a case where we wouldn't expect this to be `AsyncReset`? Should we just set it that way...?

src/main/scala/devices/debug/DebugTransport.scala
```diff
@@ -65,24 +65,29 @@ class DTMInfo extends Bundle {
...
   val version = Input(UInt(4.W))
 }
 
 class DebugTransportModuleJTAG(debugAddrBits: Int, c: JtagDTMConfig)
-  (implicit val p: Parameters) extends Module  {
+  (implicit val p: Parameters) extends RawModule  {
 
   val io = IO(new Bundle {
+    val jtag_clock = Input(Clock())
+    val jtag_reset = Input(Reset())
```
mwachs5: again, do we expect this to ever be anything but `AsyncReset`?

src/main/scala/devices/debug/DebugTransport.scala
```diff
@@ -65,24 +65,29 @@ class DTMInfo extends Bundle {
...
+    val tapReset = Input(Bool())
```
mwachs5: I am curious why we kept this one as `Bool`. I actually forget if there was something in the JTAG spec that implied something should be sync reset, or it was general laziness in the fact that we had to put down `AsyncResetRegVec` that kept much of the DTM as synchronous reset.

src/main/scala/devices/debug/DebugTransport.scala
```diff
@@ -65,24 +65,29 @@ class DTMInfo extends Bundle {
...
     val dmi = new DMIIO()(p)
     val jtag = Flipped(new JTAGIO(hasTRSTn = false)) // TODO: re-use SystemJTAGIO here?
-    val jtag_reset = Input(Bool())
     val jtag_mfr_id = Input(UInt(11.W))
     val jtag_part_number = Input(UInt(16.W))
     val jtag_version = Input(UInt(4.W))
     val fsmReset = Output(Bool())
   })
+  val rf_reset = IO(Input(Reset()))
```
mwachs5: I think this one deserves a comment / should there be a `DebugTransport` parameter for this for whether or not it's present?

src/main/scala/devices/debug/Periphery.scala
```diff
@@ -159,17 +159,18 @@ trait HasPeripheryDebugModuleImp extends LazyModuleImp {
...
     val psdio = psd.psd.getOrElse(WireInit(0.U.asTypeOf(new PSDTestMode)))
 
-    dtm.clock          := sj.jtag.TCK
+    dtm.io.jtag_clock  := sj.jtag.TCK
     dtm.io.jtag_reset  := sj.reset
     dtm.io.jtag_mfr_id := sj.mfr_id
     dtm.io.jtag_part_number := sj.part_number
     dtm.io.jtag_version := sj.version
-    dtm.reset          := dtm.io.fsmReset
+    dtm.io.tapReset     := dtm.io.fsmReset
```
mwachs5: since we aren't actually using the module implicit reset and we can do `withClockAndReset`, do we actually need to bring this back out as an IO? Can we just keep this logic internnal to the DTM?

src/main/scala/jtag/JtagStateMachine.scala
```diff
@@ -66,7 +66,7 @@ object JtagState {
   * Usage notes:
   * - 6.1.1.1b state transitions occur on TCK rising edge
   * - 6.1.1.1c actions can occur on the following TCK falling or rising edge
-  *
+  * Implicit reset is AsyncReset
```
mwachs5: there is now a `RequireAsyncReset` mixin we could use to enforce this below (though we would need to get up to date on chisel):

`class JtagStateMachine ... extends Module with RequireAsyncReset`

src/main/scala/jtag/JtagStateMachine.scala
```diff
@@ -76,13 +76,10 @@ class JtagStateMachine(implicit val p: Parameters) extends Module() {
   }
   val io = IO(new StateMachineIO)
 
-  val nextState = WireInit(JtagState.TestLogicReset.U)
-  val currStateReg = Module (new AsyncResetRegVec(w = JtagState.State.width,
-    init = JtagState.State.toInt(JtagState.TestLogicReset)))
-  currStateReg.io.en := true.B
-  currStateReg.io.d  := nextState
-  val currState = currStateReg.io.q
+  val nextState = Wire(UInt(JtagState.State.width.W))
```
mwachs5: why did you get rid of the `WireInit` here in favor of the two-line version for `nextState`?

src/main/scala/jtag/JtagTap.scala
```diff
@@ -31,7 +31,8 @@ class JtagOutput(irLength: Int) extends Bundle {
 }
 
 class JtagControl extends Bundle {
-  val jtag_reset = Input(Bool())
+  val jtag_reset = Input(Reset())
```
mwachs5: still curious why this is not `AsyncReset` and why the `tapReset` is `Bool` vs `AsyncReset`

src/main/scala/jtag/JtagTap.scala
```diff
@@ -79,15 +82,19 @@ class JtagTapController(irLength: Int, initialInstruction: BigInt)(implicit val
   // combined with any POR, and it should also be
   // synchronized to TCK.
   require(!io.jtag.TRSTn.isDefined, "TRSTn should be absorbed into jtckPOReset outside of JtagTapController.")
-  withReset(io.control.jtag_reset) {
+  withReset(io.control.jtag_reset.asAsyncReset) {
```
mwachs5: we could get rid of the cast here if we know that it's gotta be AsyncReset already

src/main/scala/jtag/Utils.scala
```diff
@@ -13,35 +12,6 @@ class Tristate extends Bundle {
   val driven = Bool()  // active high, pin is hi-Z when driven is low
 }
 
-/** Generates a register that updates on the falling edge of the input clock signal.
-  */
-object NegEdgeReg {
```
mwachs5: 🎉 on deleting this chunk of code

src/main/scala/devices/debug/Debug.scala
ernie-sifive: You are correct - fixed.

src/main/scala/devices/debug/Debug.scala
ernie-sifive: Yes, without the withClockAndReset, there is a FIRRTL exception.

src/main/scala/devices/debug/DebugTransport.scala
ernie-sifive: Done.

src/main/scala/devices/debug/DebugTransport.scala
ernie-sifive: Done

src/main/scala/devices/debug/DebugTransport.scala
ernie-sifive: I rearranged this and renamed it to `tapIsInTestLogicReset` to make it clearer that it's not a "reset".  JTAG spec says that the IDCODE IR is loaded by traversing the TAP through the Test-Logic-Reset state.  This is done synchronously to avoid creating an async reset within the design.

src/main/scala/devices/debug/DebugTransport.scala
ernie-sifive: Comment added.  The signal does not appear in Verilog if the RF transform is not applied.

src/main/scala/devices/debug/Periphery.scala
ernie-sifive: Moved to internal logic.

src/main/scala/jtag/JtagStateMachine.scala
ernie-sifive: Will add when available.

src/main/scala/jtag/JtagStateMachine.scala
ernie-sifive: Fixed

src/main/scala/jtag/JtagTap.scala
ernie-sifive: Done.

src/main/scala/jtag/JtagTap.scala
ernie-sifive: Done.

src/main/scala/devices/debug/DebugTransport.scala
```diff
@@ -260,8 +263,14 @@ class DebugTransportModuleJTAG(debugAddrBits: Int, c: JtagDTMConfig)
   tapIO.control.jtag_reset := io.jtag_reset
 
   //--------------------------------------------------------
-  // Reset Generation (this is fed back to us by the instantiating module,
-  // and is used to reset the debug registers).
+  // TAP Test-Logic-Reset state resets the debug registers.
```
mwachs5:
```suggestion
  // TAP Test-Logic-Reset state synchronously resets the debug registers.
```

src/main/scala/jtag/JtagStateMachine.scala
```diff
@@ -76,12 +74,8 @@ class JtagStateMachine(implicit val p: Parameters) extends Module() {
   }
   val io = IO(new StateMachineIO)
 
-  val nextState = WireInit(JtagState.TestLogicReset.U)
-  val currStateReg = Module (new AsyncResetRegVec(w = JtagState.State.width,
-    init = JtagState.State.toInt(JtagState.TestLogicReset)))
-  currStateReg.io.en := true.B
-  currStateReg.io.d  := nextState
-  val currState = currStateReg.io.q
+  val nextState = WireInit(JtagState.TestLogicReset.U)  // (JtagState.State.width.W))
```
mwachs5: Are you concerned the width may not be properly inferred?

mwachs5: It should be : https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/jtag/JtagStateMachine.scala#L13

src/main/scala/devices/debug/DebugTransport.scala
```diff
@@ -260,8 +263,14 @@ class DebugTransportModuleJTAG(debugAddrBits: Int, c: JtagDTMConfig)
...
 
-  io.fsmReset := tapIO.output.reset
-}
+  when (tapIO.output.reset) {
```
mwachs5: so is this the same as `tapIsInTestLogicReset`...?

src/main/scala/jtag/JtagTap.scala
```diff
@@ -104,23 +112,18 @@ class JtagTapController(irLength: Int, initialInstruction: BigInt)(implicit val
...
-  when (currState === JtagState.UpdateIR.U) {
-    nextActiveInstruction := irChain.io.update.bits
-    updateInstruction := true.B
-  } .otherwise {
-    // Needed when using chisel3._ (See #1160)
-    nextActiveInstruction := DontCare
-    updateInstruction := false.B
+  withClockAndReset(clock_falling, io.control.jtag_reset) {
+    val activeInstruction = RegInit(initialInstruction.U(irLength.W))
+    when (tapIsInTestLogicReset) {
```
mwachs5: thanks, this is much easier to read & understand

src/main/scala/jtag/JtagTap.scala
```diff
@@ -104,23 +112,18 @@ class JtagTapController(irLength: Int, initialInstruction: BigInt)(implicit val
...
+    }.elsewhen (currState === JtagState.UpdateIR.U) {
+      activeInstruction := irChain.io.update.bits
+    }
+    io.output.instruction := activeInstruction
   }
-  io.output.instruction := activeInstruction
 
-  io.output.reset := currState === JtagState.TestLogicReset.U
+  tapIsInTestLogicReset := currState === JtagState.TestLogicReset.U
+  io.output.reset := tapIsInTestLogicReset
```
mwachs5: can we rename this output as well to `tapIsInTestLogicReset` for clarity?

src/main/scala/devices/debug/DebugTransport.scala
ernie-sifive: done

src/main/scala/jtag/JtagStateMachine.scala
ernie-sifive: unnecessary comment removed

src/main/scala/jtag/JtagTap.scala
ernie-sifive: done

src/main/scala/devices/debug/DebugTransport.scala
ernie-sifive: name changed.

PR #2339, opened 2020-03-15
FixChisel3: Add some commentary to scaladoc

mwachs5:
**Related issue**: 

**Release Notes**

Added some scaladoc commentary to the "FixChisel3" operators :<>, :<=, :=> to explain what they do and the rationale for their creation.


src/main/scala/util/package.scala
```diff
@@ -234,23 +234,55 @@ package object util {
     map.view.map({ case (k, vs) => k -> vs.toList }).toList
   }
 
+/** provides operators useful for working with bidirectional [[Bundle]]s
+  * 
+  * In terms of [[Flipped]] with a producer 'p' and 'consumer' c:
+  * c :<= p // means drive all unflipped fields of 'c' from 'p' (e.g.: valid/bits)  
+  * c :=> p // means drive all flipped fields of 'c' from 'p' (e.g.: ready)  
+  * p :<> c // do both of the above
+  * c :<> p // do both of the above, but you'll probably get a Gender error later.
```
terpstra: Pretty sure you meant:

  * c :<> p // do both of the above
  * p :<> c // do both of the above, but you'll probably get a Gender error later.
Write

mwachs5:
```suggestion
  * c :<> p // do both of the above
  * p :<> c // do both of the above, but you'll probably get a Gender error later.
```

src/main/scala/util/package.scala
```diff
@@ -234,23 +234,55 @@ package object util {
...
+  * c :<> p // do both of the above
+  * p :<> c // do both of the above, but you'll probably get a Gender error later.
+  * 
+  * This utility class is needed because in [[chisel3]]:
+  * c := p // only works if there are no directions on fields.  
+  * c <> p // only works if one of those is an [[IO]] (not a [[Wire]]).
+  * 
+  * Compared with [[chisel3]] operators:
+  * c <> p   is an 'actual-direction'-inferred 'c :<> p' or 'p :<> c'
+  * c := p   is equivalent to 'c :<= p' + 'p :=> c'
```
mwachs5:
```suggestion
  * c := p   is equivalent to 'c :<= p' + 'c :=> p'
```

@terpstra (and/or @leikou01 )  can you double check-this statement? Given the first two lines in this it seems like this change is what it should be. Otherwise I don't really understand this sentence.

src/main/scala/util/package.scala
```diff
@@ -234,23 +234,55 @@ package object util {
...
+  * This is how Chisel._ (compatability mode) and firrtl work.
+  * Some find that  ':<>' has superior readability (even if the direction can be inferred from an IO),
+  * because it clearly states the intended producer/consumer relationship. 
+  * You will get an appropriate error if you connected it the wrong way
+  * (usually because you got the IO direction wrong) instead of silently succeeding.
+  * 
+  * What if you want to connect all of the signals (e.g. ready/valid/bits)
+  * from producer 'p' to a monitor 'm'?
+  * For example in order to tap the connection to monitor traffic on an existing connection.
+  * In that case you can do 'm :<= p' and 'm :=> p'.
```
mwachs5:
```suggestion
  * In that case you can do 'm :<= p' and 'p :=> m'.
```

@terpstra can you double-check this one too? Just intuitively I feel like this is how the sentence should read... all the normal P signals go to M, and then also all other P signals go to M.

src/main/scala/util/package.scala
leikou01: just tested it. the latter is correct.
`* c := p   is equivalent to 'c :<= p' + 'p :=> c' or 'c:<>= p'

Consumer always on LHS.

mwachs5: ok, i will resolve this and leave it as-is.

src/main/scala/util/package.scala
```diff
@@ -234,23 +234,55 @@ package object util {
     map.view.map({ case (k, vs) => k -> vs.toList }).toList
   }
 
+/** provides operators useful for working with bidirectional [[Bundle]]s
+  * 
+  * In terms of [[Flipped]] with a producer 'p' and 'consumer' c:
+  * c :<= p // means drive all unflipped fields of 'c' from 'p' (e.g.: valid/bits)  
+  * c :=> p // means drive all flipped fields of 'c' from 'p' (e.g.: ready)
```
leikou01, line 241: this one actually means
` * c :=> p // means drive all flipped fields of 'p' from 'c' (e.g.: ready) ` , e.g.: p.ready := c.ready ?

src/main/scala/util/package.scala
leikou01: sorry, even the one sentence from myself is not consistent. corrected it in the above.

src/main/scala/util/package.scala
mwachs5: so should this say p :=> c for a more consistent example?

src/main/scala/util/package.scala
```diff
@@ -234,23 +234,55 @@ package object util {
     map.view.map({ case (k, vs) => k -> vs.toList }).toList
   }
 
+/** provides operators useful for working with bidirectional [[Bundle]]s
+  * 
+  * In terms of [[Flipped]] with a producer 'p' and 'consumer' c:
+  * c :<= p // means drive all unflipped fields of 'c' from 'p' (e.g.: valid/bits)  
+  * c :=> p // means drive all flipped fields of 'c' from 'p' (e.g.: ready)  
+  * c :<> p // do both of the above
+  * p :<> c // do both of the above, but you'll probably get a Gender error later.
```
jackkoenig, line 243: `Gender` has been replaced by `Flow`. Also if this results in flow errors in FIRRTL, I think it would be better to just catch them here instead.

src/main/scala/util/package.scala
leikou01: My understanding is "no". 

Rule #1 for this set of operator is "consumer always on LHS". (It is sort of directional operator from the producer-consumer's viewpoint. not sure of the best way to express this).

p:=> c means that the consumer's ready (input) is driven by the producer's ready (output), which is wrong.

src/main/scala/util/package.scala
leikou01: based on my current understanding, the former is wrong. the latter might be correct (m.ready = p.ready) for a monitor.

src/main/scala/util/package.scala
```diff
@@ -234,23 +234,55 @@ package object util {
     map.view.map({ case (k, vs) => k -> vs.toList }).toList
   }
 
+/** provides operators useful for working with bidirectional [[Bundle]]s
+  * 
+  * In terms of [[Flipped]] with a producer 'p' and 'consumer' c:
+  * c :<= p // means drive all unflipped fields of 'c' from 'p' (e.g.: valid/bits)
```
mwachs5, line 240:
```suggestion
  * c :<= p // means drive all unflipped fields of 'c' from 'p' (e.g.: c.valid := p.valid)
```

src/main/scala/util/package.scala
mwachs5, line 241:
```suggestion
  * c :=> p // means drive all flipped fields of 'c' from 'p' (e.g.: `p.ready := c.ready`)
```

src/main/scala/util/package.scala
mwachs5, line 243:
```suggestion
  * p :<> c // do both of the above, but you'll probably get a Flow error later.
```

src/main/scala/util/package.scala
mwachs5, line 251:
```suggestion
  * c := p is equivalent to 'c :<= p' + 'c :=> p' or 'c:<> p'
```

```

File: src/main/scala/util/package.scala (hunk 4 above)
Comment (terpstra) [reply src/main/scala/util/package.scala:241]: You are correct. Consumer is always on the left-hand-side. The comment should have been:
```
c :=> p // means drive all flipped fields of 'p' from 'c' (e.g.: p.ready := c.ready)
```

File: src/main/scala/util/package.scala (hunk 5 above)
Comment (terpstra) [reply src/main/scala/util/package.scala:243]: You propose to try and introspect the actual direction? And if it cannot be determined (ie: is a wire), just trust the user? That is fine by me.

File: src/main/scala/util/package.scala [hunk 7]
```diff
@@ -234,23 +234,55 @@ package object util {
...
+  * c :<> p // do both of the above
+  * p :<> c // do both of the above, but you'll probably get a Gender error later.
+  * 
+  * This utility class is needed because in [[chisel3]]:
+  * c := p // only works if there are no directions on fields.  
+  * c <> p // only works if one of those is an [[IO]] (not a [[Wire]]).
+  * 
+  * Compared with [[chisel3]] operators:
+  * c <> p   is an 'actual-direction'-inferred 'c :<> p' or 'p :<> c'
+  * c := p is equivalent to 'c :<= p' + 'c :=> p' or 'c:<> p'
```
terpstra, line 251: This is wrong.
c := p is equivalent to 'c :<= p' + 'p :=> c'.

'c :<> p' is not related to 'c := p' at all.

terpstra: In other words, 'c := p' says drive ALL fields of 'c' from 'p' regardless of their direction.

(which is pretty much only useful when you want to put a VIP monitor down that watches the wires of an existing producer-consumer pair)

src/main/scala/util/package.scala
```diff
@@ -234,23 +234,55 @@ package object util {
...
+  * c :<> p // do both of the above
+  * p :<> c // do both of the above, but you'll probably get a Flow error later.
+  * 
+  * This utility class is needed because in [[chisel3]]:
+  * c := p // only works if there are no directions on fields.  
+  * c <> p // only works if one of those is an [[IO]] (not a [[Wire]]).
+  * 
+  * Compared with [[chisel3]] operators:
+  * c <> p   is an 'actual-direction'-inferred 'c :<> p' or 'p :<> c'
+  * c := p is equivalent to 'c :<= p' + 'c :=> p' or 'c:<> p'
```
mwachs5, line 251:
```suggestion
  * c := p is equivalent to 'c :<= p' + 'p :=> c'. In other words, drive ALL fields of 'c' from 'p' regardless of their direction.
```

src/main/scala/util/package.scala
```diff
@@ -234,23 +234,55 @@ package object util {
...
     }
-    // x <> y   is an 'actual-direction'-inferred 'x :<> y' or 'y :<> x'
-    // x := y   is equivalent to 'x :<= y' + 'y :=> x'
+
 
     // Versions of the operators that use the type from the RHS
     // y :<=: x  ->  x.:<=:(y)  ->  y :<= x  ->  FixChisel3.assignL(y, x)
+    /** version of the :<= operator that uses the type from the RHS */
     def :<=: (y: T): Unit = { FixChisel3.assignL(y, x) }
+    /** version of the :>= operator that uses the type from the RHS */
```
mwachs5, line 283:
```suggestion
    /** version of the :=> operator that uses the type from the RHS */
```

src/main/scala/util/package.scala
jerryhethatday: @jackkoenig  do you have any plan of supporting these operators in chisel?

PR #2344, opened 2020-03-18
Add SeparateButSynchronousCrossingType

mwachs5:
**Related issue**: None

**Release Notes**

Add a new `ClockCrossingType` : `SeparateButSynchronousCrossingType`. This represents an interface which (for whatever reason, perhaps power gating) has a separate reset and clock, but it should be synchronous to the other reset and clock. 

This isn't something that makes a ton of sense generally, so this is only supported in this PR far for interrupt crossings and tile link crossings. Anything else in this codebase would give a `MatchError` if it is attempted.


mwachs5:
yes, that was intentional.... i dont think it makes sense for the general case to use this unusual crossing type so I wanted people to get an error if they tried

mwachs5:
but maybe i should reconsider PRCI

mwachs5:
I don't think I want to go this route at this time. May reopen this PR later

PR #2346, opened 2020-03-19
Subsystem: Add HierarchicalLocation

rmac-sifive:
**Related issue**: 

**Release Notes**
Adding `HierarchicalLocation` which config keys may use to control hierarchy of instantiation


hcook:
by which i mean
```
 case class HierarchicalLocation(override val name: String) extends Location[LazyScope](name)
 case object InTile extends HierarchicalLocation("InTile")
 case object InSubsystem extends HierarchicalLocation("InSubsystem")
 case object InSystem extends HierarchicalLocation("InSystem")
```

hcook:
Sorry, might be a bug in my PR, but here is my code proposal:
```
 class HierarchicalLocation(override val name: String) extends Location[LazyScope](name)    
 case object InTile extends HierarchicalLocation("InTile")                                  
 case object InSubsystem extends HierarchicalLocation("InSubsystem")                        
 case object InSystem extends HierarchicalLocation("InSystem")
```

hcook:
Also I'll note that doing this requires juggling some internal imports, because a few files import `sifive.devices._` and `rocketchip.subsystem._`
Looks like only two or three files are affected though...

PR #2353, opened 2020-03-22
Chisel compile warning: TileLink Parameters v1

johningalls-sifive:
**Impact**: (should be) no functional change

```
method apply in object TLClientParameters is deprecated: Use TLMasterParameters.v1 instead of TLClientParameters
method apply in object TLClientPortParameters is deprecated: Use TLMasterPortParameters.v1 instead of TLClientPortParameters
method apply in object TLManagerParameters is deprecated: Use TLSlaveParameters.v1 instead of TLManagerParameters
method apply in object TLManagerPortParameters is deprecated: Use TLSlavePortParameters.v1 instead of TLManagerPortParameters
method copy in class TLMasterParameters is deprecated: Use v1copy instead of copy
method copy in class TLMasterPortParameters is deprecated: Use v1copy instead of copy
method copy in class TLSlaveParameters is deprecated: Use v1copy instead of copy
```

### DIFF
--- a/src/main/scala/amba/apb/ToTL.scala
+++ b/src/main/scala/amba/apb/ToTL.scala
@@ -11,9 +11,9 @@ import freechips.rocketchip.util._
 
 case class APBToTLNode()(implicit valName: ValName) extends MixedAdapterNode(APBImp, TLImp)(
   dFn = { mp =>
-    TLClientPortParameters(
+    TLMasterPortParameters.v1(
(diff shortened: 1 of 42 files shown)


terpstra:
We should really finish the job by adding the '.v2' constructors where managers->slaves and clients->masters. And then rename the TL node appropriately. (I'm not asking you to do this)

PR #2355, opened 2020-03-22
Chisel compile warning: random LFSR16

johningalls-sifive:

```
[warn] /rocket-chip/src/main/scala/rocket/ICache.scala:191:14: method apply in object LFSR16 is deprecated (since 3.2): Use chisel3.util.random.LFSR(16) for a 16-bit LFSR
[warn]     val v0 = LFSR16(refill_fire)(log2Up(nWays)-1,0)
[warn]              ^
[warn] /rocket-chip/src/main/scala/tilelink/Arbiter.scala:100:14: method apply in object LFSR16 is deprecated (since 3.2): Use chisel3.util.random.LFSR(16) for a 16-bit LFSR
[warn]   val lfsr = LFSR16(Bool(true))
[warn]              ^
[warn] /rocket-chip/src/main/scala/util/ECC.scala:186:14: method apply in object LFSR16 is deprecated (since 3.2): Use chisel3.util.random.LFSR(16) for a 16-bit LFSR
[warn]     UIntToOH(LFSR16()(log2Up(width)+f-1,0))(width-1,0)
[warn]              ^
[warn] /rocket-chip/src/main/scala/util/Misc.scala:157:28: method apply in object LFSR16 is deprecated (since 3.2): Use chisel3.util.random.LFSR(16) for a 16-bit LFSR
[warn]   private def randomizer = LFSR16()
[warn]                            ^
[warn] /rocket-chip/src/main/scala/util/Replacement.scala:17:14: method apply in object LFSR16 is deprecated (since 3.2): Use chisel3.util.random.LFSR(16) for a 16-bit LFSR
[warn]   val lfsr = LFSR16(replace)
[warn]              ^
```

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -13,6 +13,7 @@ import freechips.rocketchip.util.{DescribedSRAM, _}
 import freechips.rocketchip.util.property._
 import chisel3.internal.sourceinfo.SourceInfo
 import chisel3.dontTouch
+import chisel3.util.random.LFSR
 import freechips.rocketchip.diplomaticobjectmodel.DiplomaticObjectModelAddressing
 import freechips.rocketchip.diplomaticobjectmodel.model._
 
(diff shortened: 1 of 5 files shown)


seldridge:
Note: that the get the *exact* behavior of `LFSR16` you need to use a special construction:

```scala
FibonacciLFSR
  .maxPeriod(16, increment, seed = Some(BigInt(1) << 15))
  .asBools
  .reverse
```

The new LFSR works in reverse of the old one.

johningalls-sifive:
> Note: that the get the _exact_ behavior of `LFSR16` you need to use a special construction:
> The new LFSR works in reverse of the old one.

Noted, and thanks, I didn't know that difference.
For all of these purposes, the new behavior should be fine too: we aren't trying for predictable behavior from the pseudo-random generators.  :-)

aswaterman:
Yeah, the semantic difference is fine in this case

PR #2363, opened 2020-03-25
Using AOP to query Diplomacy

azidar:
**Related issue**: 

**Release Notes**


sequencer:
Is possible to use AOP modifying `LazyModule`s?
for example:
`nodeA := nodeB`
->
`nodeA := Monitor := nodeB`

diplomacy.view
```diff
@@ -0,0 +1,6 @@
+//graph     dot
+//// graph  graphML
+//
+//BaseNode  all
+//BaseNode  byGroup
```
azidar: Delete this file?

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -16,6 +17,10 @@ abstract class LazyModule()(implicit val p: Parameters)
   protected[diplomacy] var info: SourceInfo = UnlocatableSourceInfo
   protected[diplomacy] val parent = LazyModule.scope
 
+
+  var childrenAop = ()=>children
+  var nodesAop = ()=>nodes
```
azidar: This pattern of a `var` and `() => ...` is strange. Should you use a `lazy val` instead? Why does it need to be mutable? Also, please add a comment to describe the reason for this change.

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -117,11 +122,14 @@ abstract class LazyModule()(implicit val p: Parameters)
     children.foreach( _.nodeIterator(iterfunc) )
   }
 
-  def getChildren = children
+  def getChildren() = children
+//  def getNodes() = nodes
```
azidar: What's the reason for this change? Delete extraneous comments.

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -117,11 +122,14 @@ abstract class LazyModule()(implicit val p: Parameters)
...
+
 }
 
 object LazyModule
 {
+  val impToLazyModuleMap = new mutable.HashMap[String, (LazyModuleImp, LazyModule)]
```
azidar: All `LazyModuleImpLike`'s have a `wrapper` field. Why can't we reconstruct this `impToLazyModuleMap` from the pointer to the top-level module's wrapper?

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -184,6 +192,11 @@ sealed trait LazyModuleImpLike extends RawModule
 
 class LazyModuleImp(val wrapper: LazyModule) extends MultiIOModule with LazyModuleImpLike {
   val (auto, dangles) = instantiate()
+
+  if(! s"${this.getClass.getName}".contains("anon")){
```
azidar: Why delete anonymous LazyModuleImps? Why use the className, instead of the module name (which is guaranteed to be unique)?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -389,7 +389,7 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   private var bundlesSafeNow = false
   // Accessors to the result of negotiation to be used in LazyModuleImp:
   def out: Seq[(BO, EO)] = {
-    require(bundlesSafeNow, s"${name}.out should only be called from the context of its module implementation")
+//    require(bundlesSafeNow, s"${name}.out should only be called from the context of its module implementation")
```
azidar: We need to figure out a better way to do this. Let me know if you want to chat about it.

src/main/scala/monitors/TLMonitorInjToDcache.scala
```diff
@@ -0,0 +1,89 @@
...
+  *     val identityNodes = SelectDiplomacy.identityNodes()
+  *     val ephemeralNodes = SelectDiplomacy.ephemeralNodes()
+  *     val mixedNexusNodes = SelectDiplomacy.mixedNexusNodes()
+  *     val nexusNodes = SelectDiplomacy.nexusNodes()
+  *     val sourceNodes = SelectDiplomacy.sourceNodes()
+  *     val sinkNodes = SelectDiplomacy.sinkNodes()
+  *     val mixedTestNodes = SelectDiplomacy.mixedTestNodes()
+  * 
+  *     val csNode = getSrcNodes[DCache]()
+  *     val cmNode = getManagementNodes[DCache()
```
azidar: These access functions are fine, but I'm not certain they are powerful enough. We want the user to be able to express queries like:

1. Select the link between CPU and L1, and inject a monitor into the CPU
1. Select the link between CPU and L1, and inject a monitor into the L1
1. Select all sink nodes, and instantiate a TLMonitor on each (don't care which module it is injected to)

Can you show concrete examples demonstrating each of the above queries?

src/main/scala/util/SelectDiplomacy.scala
```diff
@@ -0,0 +1,407 @@
...
+
+  val digraph = new MutableDiGraph[RTreeData]
+
+  def findNode(name:String): Any = {
+    None
+  }
+
+  /** Collects all Lazy Module Imp as a List.
+    */
+  def collectImpModules(): List[LazyModuleImp] = {
```
azidar: Many of your collect functions share similar structure, but do not directly reuse code. They all iterate of teh edges, match on the parent.item, and then return something. Can you capture that pattern in a separate function which is called by these collection functions?

src/main/scala/util/SelectDiplomacy.scala
```diff
@@ -0,0 +1,407 @@
...
+        case x: LazyModule => lst = x :: lst
+        case _ => None
+      }
+    }
+    lst
+  }
+
+  /** Create a file with all Lazy Module names.
+    */
+  def collectLazyModules(targetDir: String, fname: String): String = {
```
azidar: Can this function reuse `def collectLazyModules(): List[LazyModule]` and then serialize/write the output?

emulator/load-modules
```diff
@@ -0,0 +1,10 @@
+module load synopsys/vcs/P-2019.06-SP2
```
mwachs5: seems like we don't need this in rocket-chip

emulator/riscv-tools.bashrc
```diff
@@ -0,0 +1,12 @@
+# Remove any old tools from PATH:
+export PATH="$(perl -e 'print join(":", grep { ! m{^/work/tools/riscv-tools/[0-9a-f]+/bin$} } split(/:/, $ENV{PATH}))')"
+
+# Add latest tools to RISCV and PATH:
+
+export FSGCC=$(cat $(dirname $(readlink -f ${BASH_SOURCE[0]}))/../../freedom-studio-toolchain.path)
```
mwachs5: remove this from rocket-chip

freedom-studio-toolchain.path
```diff
@@ -0,0 +1 @@
+/sifive/tools/riscv-tools/riscv64-unknown-elf-gcc-8.3.0-2019.08.0-x86_64-linux-ubuntu14
```
mwachs5: this doesn't belong in rocket-chip

src/main/scala/diplomacy/LazyModule.scala
mwachs5: suggest we call this "childrenAOP" if AOP is an acronym.

src/main/scala/diplomacy/LazyModule.scala
mwachs5: can you add a comment for this? What is the `String` used to lookup here?

@azidar are you saying `wrapper` already gives a mapping from any LazyModuleImpLink back to its instantiating `LazyModule`?

src/main/scala/diplomacy/LazyModule.scala
mwachs5: how will this interact with FIRRTL transforms that modify module /instance names after chisel elaboration?

src/main/scala/monitors/TLMonitorInjToDcache.scala
```diff
@@ -0,0 +1,89 @@
...
+import freechips.rocketchip.diplomacy._
+import scala.reflect.ClassTag
+
+/** API to access Diplomacy Node data:
+  *  {{{
+  *     val baseNodes = SelectDiplomacy.collectBaseNodes()
+  *     val impMod = SelectDiplomacy.collectImpModules()
+  *     val lazyMod = SelectDiplomacy.collectLazyModules()
+  * 
+  *     val mixedNodes = SelectDiplomacy.mixedNodes()
```
mwachs5: whats a use case for all these? Why would i want to query diplomatic nodes like this? I can imagine wanting to grab for example all "TLSinkNodes" or "AXI4MasterNodes", but why would I abstractly want to query all `mixedNexusNode`?

src/main/scala/monitors/TLMonitorInjToDcache.scala
```diff
@@ -0,0 +1,89 @@
...
+  *     val sourceNodes = SelectDiplomacy.sourceNodes()
+  *     val sinkNodes = SelectDiplomacy.sinkNodes()
+  *     val mixedTestNodes = SelectDiplomacy.mixedTestNodes()
+  * 
+  *     val csNode = getSrcNodes[DCache]()
+  *     val cmNode = getManagementNodes[DCache()
+  *  }}}
+ */
+
+case class AopTestModule (wide: Int) extends Module {
```
mwachs5: suggest we use AOP not Aop if AOP is an acronym

src/main/scala/monitors/TLMonitorInjToDcache.scala
```diff
@@ -58,6 +58,28 @@ case class AopTestModule (wide: Int) extends Module {
   io.random := 0.U
 }
 
+case object MonitorPrintPrefix extends Field[Option[String]](None)
+
+object BundlePrintTest {
+  def print_bundle(bundle: TLBundle, edge: TLEdge) {
+    edge.params(MonitorPrintPrefix).foreach { prefix =>
```
mwachs5, line 65: is `MonitorPrintPrefix` actually set here? i think you should just take `prefix` as a parameter to `print_bundle`

PR #2365, opened 2020-03-26
Added v2 constructors and supportHelper to Edges

brrmorre:
**Related issue**: 

**Release Notes**


src/main/scala/tilelink/Parameters.scala
```diff
@@ -276,13 +276,42 @@ class TLSlaveParameters private(
...
+    setName:       String,
+    address:       Seq[AddressSet],
+    regionType:    RegionType.T,
+    executable:    Boolean,
+    fifoId:        Option[Int],
+    supports:      TLMasterToSlaveTransferSizes,
+    emits:         TLSlaveToMasterTransferSizes,
+    alwaysGrantsT: Boolean,
+    mayDenyGet:    Boolean,
+    mayDenyPut:    Boolean) =
```
terpstra: You need to take default arguments from the current values of the class.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -276,13 +276,42 @@ class TLSlaveParameters private(
...
       mayDenyGet    = mayDenyGet,
       mayDenyPut    = mayDenyPut,
       alwaysGrantsT = alwaysGrantsT,
       fifoId        = fifoId)
   }
 
+  def v2copy(
+    nodePath:      Seq[BaseNode],
+    resources:     Seq[Resource],
+    setName:       String,
```
terpstra: The argument should probably just be called 'name' (but still default to the value of setName).

src/main/scala/tilelink/Parameters.scala
```diff
@@ -366,6 +395,35 @@ object TLSlaveParameters {
       alwaysGrantsT = alwaysGrantsT,
       fifoId        = fifoId)
   }
+
+  def v2(
+    nodePath:      Seq[BaseNode]                = Seq(),
+    resources:     Seq[Resource]                = Seq(),
+    setName:       String,
```
terpstra: Again, this should probably be called name, and default to "".

src/main/scala/tilelink/Parameters.scala
```diff
@@ -366,6 +395,35 @@ object TLSlaveParameters {
...
+    address:       Seq[AddressSet],
```
terpstra: I am a bit worried that we've made the non-default argument not the first argument.

The plus side: it will force people to write 'address = ' in the constructor. That's also the downside.

@hcook ?

src/main/scala/tilelink/Parameters.scala
```diff
@@ -819,6 +904,33 @@ object TLMasterParameters {
       neverReleasesData = false,
       sourceId          = sourceId)
   }
+
+  def v2(
+    nodePath:          Seq[BaseNode]                = Seq(),
+    resources:         Seq[Resource]                = Nil,
+    name:              String,
```
terpstra: Again, the one non-default value is not the first.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1087,6 +1199,64 @@ case class TLEdgeParameters(
...
+    val supportCases = groupByIntoSeq(client.clients)(m => trim(member(m))).map { case (k, vs) =>
+      k -> vs.flatMap(_.visibility)
+    }
+    val mask = if (safe) ~BigInt(0) else AddressDecoder(supportCases.map(_._2))
+    val simplified = supportCases.map { case (k, seq) => k -> AddressSet.unify(seq.map(_.widen(~mask)).distinct) }
+    simplified.map { case (s, a) =>
+      (Bool(Some(s) == range) || s.containsLg(lgSize)) &&
+      a.map(_.contains(address)).reduce(_||_)
+    }.foldLeft(Bool(false))(_||_)
+  }
```
terpstra: This seems wrong. Shouldn't emitHelper be checking against the source ID, not the address?

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1087,6 +1199,64 @@ case class TLEdgeParameters(
...
+    val supportCases = groupByIntoSeq(manager.managers)(m => trim(member(m))).map { case (k, vs) =>
+      k -> vs.flatMap(_.address)
+    }
+    val mask = if (safe) ~BigInt(0) else AddressDecoder(supportCases.map(_._2))
+    val simplified = supportCases.map { case (k, seq) => k -> AddressSet.unify(seq.map(_.widen(~mask)).distinct) }
+    simplified.map { case (s, a) =>
+      (Bool(Some(s) == range) || s.containsLg(lgSize)) &&
+      a.map(_.contains(address)).reduce(_||_)
+    }.foldLeft(Bool(false))(_||_)
+  }
```
terpstra: This function already exists on TLSlavePortParameters. Why is it being duplicated here?

src/main/scala/tilelink/Parameters.scala
terpstra: Also, this method should be defined on TLMasterPortParameters.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1087,6 +1199,64 @@ case class TLEdgeParameters(
...
+  def supportsLogicalFast   (address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = supportHelper(false, _.supports.logical,    address, lgSize, range) && emitHelper(false, _.emits.logical,    address, lgSize, range)
+  def supportsGetFast       (address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = supportHelper(false, _.supports.get,        address, lgSize, range) && emitHelper(false, _.emits.get,        address, lgSize, range)
+  def supportsPutFullFast   (address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = supportHelper(false, _.supports.putFull,    address, lgSize, range) && emitHelper(false, _.emits.putFull,    address, lgSize, range)
+  def supportsPutPartialFast(address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = supportHelper(false, _.supports.putPartial, address, lgSize, range) && emitHelper(false, _.emits.putPartial, address, lgSize, range)
+  def supportsHintFast      (address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = supportHelper(false, _.supports.hint,       address, lgSize, range) && emitHelper(false, _.emits.hint,       address, lgSize, range)
```
terpstra: I think these should be called something else, that suggests it requires both supports and emits. Maybe expectsGetFast? Or something else @hcook ?

src/main/scala/tilelink/Parameters.scala
hcook: arguments with no default value should always go first IMO

src/main/scala/tilelink/Parameters.scala
hcook: i like expects

src/main/scala/tilelink/Parameters.scala
seldridge: What about `Option[String] = None`?

hcook: Not using an Option for this cause bug #2428 IMO

PR #2367, opened 2020-03-26
Fix D$ elaboration with < 4 MiB of physical address space

aswaterman:
Yes, apparently people care about this case still.


aswaterman:
This temporarily reverts #2352 to untangle a pull-request Uroboros

Issue #2366, opened 2020-03-26
Should clockGate be a member or CoreParams


davidmlw:
Two reasons for this:
1. clockGate is an attribute of every core, or every tile.
2. for rocket core, FrontendModule depends on HasRocketCoreParameters to get clockGate attribute, making FrontedModule a specific core dependent design. if clockGate is a global attribute of CoreParams, then FrontenModule will be a generic design to any core.
https://github.com/chipsalliance/rocket-chip/blob/2f6c74990dceb484ec032d8c31ff86fec8ec5a6e/src/main/scala/rocket/Frontend.scala#L91

aswaterman:
FrontendModule is not especially generic. It’s microarchitecturally coupled
to the Rocket design. In my experience, trying to make these large core
modules generic has been unproductive—see the FPU, for example. Making
smaller modules more generic has been somewhat more successful.

I’m fine with making clockGate more generic.

davidmlw:
I agree your insight. Chisel is very cool to make the small module generic and even make them in language itself.

Back to this specific rocket.FrontendModule, it is make some sense to make clockGate in CoreParams, so FrontendModule can be decoupled with rocket.

I will file a PR to master to make clockGate more generic.

PR #2369, opened 2020-03-27
mv clockGate from RocketCoreParams to CoreParams

davidmlw:
**Related issue**: 
https://github.com/chipsalliance/rocket-chip/issues/2366

**Impact**: API modification

**Release Notes**

clockGate is more a Core Params than a specific Rocket Core Params. With this change, FrontendModule is decouple from RocketCore in Scala code level.

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -39,7 +39,7 @@ case class RocketCoreParams(
   fastLoadWord: Boolean = true,
   fastLoadByte: Boolean = false,
   branchPredictionModeCSR: Boolean = false,
-  //clockGate: Boolean = false,
+  clockGate: Boolean = false,
   mvendorid: Int = 0, // 0 means non-commercial implementation
   mimpid: Int = 0x20181004, // release date in BCD
   mulDiv: Option[MulDivParams] = Some(MulDivParams()),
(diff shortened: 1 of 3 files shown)


mwachs5:
The PR description is wrong. This is an API-breaking change. Please update and add release notes

src/main/scala/tile/Core.scala
```diff
@@ -41,6 +41,7 @@ trait CoreParams {
   val nL2TLBEntries: Int
   val mtvecInit: Option[BigInt]
   val mtvecWritable: Boolean
+  val clockGate: Boolean = false
```
hcook: Don't put a default value in a trait, then all users will have to `override` it. Here it should be abstract, and sub-classes of this trait should fill in the value.

src/main/scala/rocket/RocketCore.scala
```diff
@@ -39,7 +39,7 @@ case class RocketCoreParams(
   fastLoadWord: Boolean = true,
   fastLoadByte: Boolean = false,
   branchPredictionModeCSR: Boolean = false,
-  clockGate: Boolean = false,
+  //clockGate: Boolean = false,
```
hcook: But this is still something that RocketCores can configure right? Why are you remove the ability to configure it?

davidmlw: Yes, RocketCores can configure it. I just move this clockGate from rocketParams to CoreParams, so any module extends HasCoreParams can see this value.

src/main/scala/tile/Core.scala
davidmlw: Oh, I didn't know this rule. will remove default value.

sequencer:
This should be merged a long time ago.

PR #2370, opened 2020-03-27
Use Verilator binary from SiFive.

richardxia:
**Related issue**: 

**Release Notes**

It has recently come to my attention that rocket-chip's Travis CI builds compiles Verilator from source, albeit with some caching logic. I think I can save us some time by using the prebuilt [Verilator binaries](https://github.com/sifive/verilator/releases) (for Ubuntu 16.04) that I've been building on behalf of SiFive for the past couple of years.

Note that this specific change will bump Verilator from 4.008 to 4.028. I suppose I could build a 4.008 if we really wanted to, but hopefully this doesn't cause any issues.


.travis.yml
```diff
@@ -68,11 +71,11 @@ jobs:
...
       before_cache:
         - ls -t regression/install | tail -n+2 | sed s@^@regression/install/@ | xargs rm -rf
-    - stage: prepare cache-verilator
+    - stage: set up verilator
       script:
-        - travis_wait 120 make verilator -C regression SUITE=none
-      before_install:
-        - export CXX=g++-4.8 CC=gcc-4.8
+        # This installs into /usr/local/bin/verilator
+        - curl -L -O verilator.deb https://github.com/sifive/verilator/releases/download/4.028-0sifive1/verilator_4.028-0sifive1_amd64.deb
```
mwachs5: presumably the "make verilator" target was picking the version somehow? Should we make those two versions line up (the one this curl uses and the one you'd get if you decided to run "make verilator"?

richardxia: That sounds reasonable, but I can't think of a good way of doing this besides just manually making sure that these version numbers are updated in both places. Do you have any suggestions?

richardxia:
I apparently don't know how to run `curl`. Let me figure out why this isn't working... I swear I've been testing this locally.

.travis.yml
mwachs5: could we put the version number in a shared location (like "verilator-version.txt") and use that whenever we need it? Similar to the rocket tools hash file at the top level?

richardxia: That will rely on me/SiFive not changing the package naming convention for the SiFive hosted .deb package, which I guess will at least be true until we start building packages for other .deb-based distros, like Debian or other versions of Ubuntu. At least if it changes, it'll error out.

I'll make this change (if I can get the stupid build to work at all).

richardxia:
```
$ apt install -y verilator.deb
E: Could not open lock file /var/lib/dpkg/lock-frontend - open (13: Permission denied)
E: Unable to acquire the dpkg frontend lock (/var/lib/dpkg/lock-frontend), are you root?
The command "apt install -y verilator.deb" exited with 100.
```


sudo make me a sandwich

.travis.yml
```diff
@@ -68,11 +71,11 @@ jobs:
...
         - ls -t regression/install | tail -n+2 | sed s@^@regression/install/@ | xargs rm -rf
-    - stage: prepare cache-verilator
+    - stage: set up verilator
       script:
-        - travis_wait 120 make verilator -C regression SUITE=none
-      before_install:
-        - export CXX=g++-4.8 CC=gcc-4.8
+        # This installs into /usr/local/bin/verilator
+        - curl -L -o verilator.deb https://github.com/sifive/verilator/releases/download/4.028-0sifive1/verilator_4.028-0sifive1_amd64.deb
+        - apt install -y verilator.deb
```
mwachs5: can you do this with the travis package thing above instead of doing it directly here?

mwachs5: by "package thing" i am referring to this list: https://github.com/chipsalliance/rocket-chip/blob/8af2fad8e5dcfc66d75c175d7999728d043adb20/.travis.yml#L22

richardxia: No, since this is an unofficial package. You can if you create your own apt repository, but I think that's kind of a lot of work.

I guess I could sign up for a [Launchpad](https://launchpad.net/ubuntu/+ppas) account, but that might be more trouble than necessary, since I'd have to learn how the Ubuntu package building actually works.

richardxia:
I'm getting kind of frustrated with trying to make this work, and at this point I'm willing to give up, since I have no idea how these Makefiles actually work nor can I figure out how Travis CI actually handles package setup with multiple jobs and stages.

I think I'm going to just close this out as a wontfix, since it doesn't seem to be worth the effort.

PR #2373, opened 2020-03-27
BEU <-> HART link added in DTS

mhtwn:
**Related issue**: 

**Release Notes**
The CPU section in the generated DTS file will have link to the associated BEU. This should help downstream software flows.

**Example of the emitted DTS (updated):**
   ```
{
...
	model = "SiFive,FU500";
....
	L34: cpus {
		#address-cells = <1>;
		#size-cells = <0>;
		L9: cpu@0 {
			clock-frequency = <0>;
			...
                         ...
                         ...
			device_type = "cpu";
			....
			sifive,buserror = <&L8>;
			....
		};
	};
	L33: soc {
                ....
                ....
		L8: bus-error-unit@1700000 {
			compatible = "sifive,buserror0";
			interrupt-parent = <&L3>;
			interrupts = <303>;
			reg = <0x1700000 0x1000>;
			reg-names = "control";
		};
                ....
                ....
     }

```

### DIFF
--- a/src/main/scala/tile/RocketTile.scala
+++ b/src/main/scala/tile/RocketTile.scala
@@ -88,11 +88,15 @@ class RocketTile private(
 
   val itimProperty = frontend.icache.itimProperty.toSeq.flatMap(p => Map("sifive,itim" -> p))
 
+  val beuProperty = bus_error_unit.map(d => Map(
+          "sifive,buserror0" -> d.device.asProperty)).getOrElse(Nil)
+
   val cpuDevice: SimpleDevice = new SimpleDevice("cpu", Seq("sifive,rocket0", "riscv")) {
     override def parent = Some(ResourceAnchors.cpus)
     override def describe(resources: ResourceBindings): Description = {
       val Description(name, mapping) = super.describe(resources)
-      Description(name, mapping ++ cpuProperties ++ nextLevelCacheProperty ++ tileProperties ++ dtimProperty ++ itimProperty)
+      Description(name, mapping ++ cpuProperties ++ nextLevelCacheProperty
+                  ++ tileProperties ++ dtimProperty ++ itimProperty ++ beuProperty)
     }
   }
 
--- a/src/main/scala/tile/RocketTile.scala
+++ b/src/main/scala/tile/RocketTile.scala
@@ -89,7 +89,7 @@ class RocketTile private(
   val itimProperty = frontend.icache.itimProperty.toSeq.flatMap(p => Map("sifive,itim" -> p))
 
   val beuProperty = bus_error_unit.map(d => Map(
-          "sifive,buserror0" -> d.device.asProperty)).getOrElse(Nil)
+          "sifive,buserror" -> d.device.asProperty)).getOrElse(Nil)
 
   val cpuDevice: SimpleDevice = new SimpleDevice("cpu", Seq("sifive,rocket0", "riscv")) {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by mwachs5** [comment]
Can you put an example of the emitted DTS in the PR notes?

**Comment by mhtwn** [comment]
> Can you put an example of the emitted DTS in the PR notes?

Done. (kept only relevant-to-the-PR content)

**Comment by mwachs5** [comment]
Thank you! I assume in that example `<L3>` points to the PLIC, for example?

**Comment by mhtwn** [comment]
> Thank you! I assume in that example `<L3>` points to the PLIC, for example?

Yes, that is right.

File: src/main/scala/tile/RocketTile.scala [hunk 1]
```diff
@@ -88,11 +88,15 @@ class RocketTile private(
 
   val itimProperty = frontend.icache.itimProperty.toSeq.flatMap(p => Map("sifive,itim" -> p))
 
+  val beuProperty = bus_error_unit.map(d => Map(
+          "sifive,buserror0" -> d.device.asProperty)).getOrElse(Nil)
```
richardxia: I think it would be more idiomatic and readable to call `.toSeq` instead of `.getOrElse(Nil)`.

Also, do we actually want to call this `sifive,buserror0` or should it just be `sifive,buserror`?

richardxia:
cc @bsousi5 @nategraff-sifive, who should comment on the actual produced DTS.

src/main/scala/tile/RocketTile.scala
bsousi5: It should be not be "sifive,buserror0", but simply be "sifive,buserror" as the version as described in the compatible string.

mwachs5:
```suggestion
          "sifive,buserror" -> d.device.asProperty)).getOrElse(Nil)
```

mwachs5: ok, i suggested the change, but would like @mhtwn to show what the outputted DTS looks like with this change and get @bsousi5 's final approval before approving the PR

mhtwn: @mwachs5 its a plain string - it will show in the dts as is.

mwachs5: well i see that string in the DTS above in 2 places. I thought that @bsousi5 was saying in the "compat" he expected it to say sifive,buserror0 but the device type to be sifive,buserror

mhtwn: ok, just to not create any confusion, the dts in the release notes (updated above) is how the generated dts looks:
1.  In the cpu component listing: `buserror0` is now -> `buserror`
2. Compatible field in the 'bus_error_unit' component still lists `sifive_buserror0`

mwachs5: thanks! @bsousi5 if that looks good to you can you please approve the PR? If that's not enough to make it mergable i will approve it

bsousi5: I am good with what @mhtwn commented and shown in the dts.

bsousi5: It seems I don't have the permission to hit the approve button. So @mwachs5, you need to approve.

daveparry: I do. So I’ve approved based on your, @richardxia and @mwachs5’s reviews.

PR #2374, opened 2020-03-28
Add CSRMonitorBundle to monitor CSR values

jchang0:
**Related issue**: 

Adding monitor bundle to monitor CSR registers (similar to coreBundle)

 feature request

**Release Notes**

### DIFF
--- a/src/main/scala/subsystem/RocketSubsystem.scala
+++ b/src/main/scala/subsystem/RocketSubsystem.scala
@@ -55,7 +55,8 @@ trait HasRocketTiles extends HasTiles
   }
 
   def coreMonitorBundles = (rocketTiles map { t =>
-    t.module.core.rocketImpl.coreMonitorBundle
+    (t.module.core.rocketImpl.coreMonitorBundle,
+      t.module.core.rocketImpl.csr.csrMonitorBundle)
   }).toList
 }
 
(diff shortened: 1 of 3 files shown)


src/main/scala/util/CoreMonitor.scala
```diff
@@ -22,7 +22,12 @@ class CoreMonitorBundle(val xLen: Int) extends Bundle with Clocked {
   val inst = UInt(width = 32.W)
 }
 
+class CSRMonitorBundle(val xLen: Int) extends Bundle with Clocked {
+  val mie = UInt(width = xLen.W)
```
mwachs5: is there any reason you've picked  these two? Why not mstatus, etc?

aswaterman:
You should probably add mstatus to the bundle, too, since that contains the global interrupt enable.

src/main/scala/util/CoreMonitor.scala
jchang0: Those two are the one that @surajvellengar wanted

src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -55,7 +55,8 @@ trait HasRocketTiles extends HasTiles
   }
 
   def coreMonitorBundles = (rocketTiles map { t =>
-    t.module.core.rocketImpl.coreMonitorBundle
+    (t.module.core.rocketImpl.coreMonitorBundle,
+      t.module.core.rocketImpl.csr.csrMonitorBundle)
```
johningalls-sifive: I foresee problems mixing types here: **csr**MonitorBundle in **core**MonitorBundles

jchang0:
Okay. I  did it from within federation only.  Going to hand that work off to Suraj

PR #2375, opened 2020-03-28
Support for Multiple Reset Schemes (for some parts of the codebase)

mwachs5:
~Currently this PR includes parts of https://github.com/chipsalliance/rocket-chip/pull/2205, which I would like to be merged first and then I will rebase this on top of that~ (note, this has been done). ~This PR will be rebased often before merging if necessary and commit history will be cleaned up.~ (this has also been done)

This PR includes/replaces #2237 and #2217 

**Related issue**: 

**Impact**:  API modification

**Release Notes**

Adds support for multiple reset schemes for much of the code base (some parts of the code base would still need to be converted to chisel3 or use the `DefaultCompileOptions.NonStrictInferReset` capability to infer resets, as previously this code base only allowed Bool resets. Notably, async reset of the Rocket Core/Tile is *not* supported.

Important changes to note:
**Adds** parameterization scheme for reset types (`SubsystemResetSchemeKey`)
~**Adds** LazyModule `withClock`, `withClockAndReset`, `withReset`~
**Changes** significantly the exported I/Os for debug related logic
**Removes** internal synchronization logic for debug reset-- this must be done externally now

More detailed release notes from #2237:

This PR converts the debug module to the new Configurable Reset Scheme using chisel3 abstract reset.

* There is now a 1:1 relationship required between resets and clocks (except for rational crossing clock/reset pairs). dmOuter uses dmi_clock and dmi_reset and dmInner uses new inputs debug_clock and debug_reset. A portion of SBA is in the TileLink clock / reset domain.
debug_clock must be synchronous to clock. The clock gate formerly in dmInnerAsync now resides outside the debug module in customer logic. User logic can call connectDebugClockAndReset to achieve the same functionality as before, in a different level of the hierarchy.
* A new input dmactiveAck is returned from customer logic and is used to indicate when dmInner is able to accept DMI transactions (i.e. debug_clock is running and debug_reset is negated). When dmactiveAck is negated, a bus blocker returns denied for transactions to dmInner. The current version of OpenOCD is tolerant of this blocker, but other software may need to be updated.
* dmi_reset must be asynchronous. debug_reset may be either synchronous or asynchronous but its deassertion must always occur either when debug_clock is stopped or synchronously to debug_clock.
* The APB and JTAG interface conforms to the above and now require reset synchronization.
* hart-reset and halt-on-reset functions have been changed to simple I/O and require customer logic involvement. The two signals are grouped in a new ResetCtrlIO bundle. hartIsInReset is a logical equivalent of core_reset and is synchronous to core_clock. For debug module halt-on-reset function, this signal must remain asserted for at least 4 debug_clock cycles for the DM to assert debug interrupt plus 3 core_clock cycles for the debug interrupt to reach the core prior to the core_reset signal being deasserted. Halt-on-reset will not work if either clock or core_clock is not running.
* The clock and reset signals formerly in TracedInstruction have been removed since a trace module attached would not be allowed to use them because of the 1:1 rule.


src/main/scala/tilelink/Arbiter.scala
```diff
@@ -18,12 +18,12 @@ object TLArbiter
   val roundRobin: Policy = (width, valids, select) => if (width == 1) UInt(1, width=1) else {
     val valid = valids(width-1, 0)
     assert (valid === valids)
-    val mask = RegInit(~UInt(0, width=width))
-    val filter = Cat(valid & ~mask, valid)
-    val unready = (rightOR(filter, width*2, width) >> 1) | (mask << width)
+    val maskn = RegInit(UInt(0, width=width)) // val mask = RegInit(~UInt(0, width=width)) fails with AsyncReset
```
mwachs5: @jackkoenig @azidar this "fails with AsyncReset comment" is presumably something where it can't tell that ~UInt(const) is also a const. I assume that is still an issue and I should not try to undo this change?

aswaterman: In the mean time, why don't we do `val mask = RegInit(((BigInt(1) << width)-1).U(width.W))`, to avoid the other RTL changes while keeping the constant constant?

src/main/scala/util/AsyncQueue.scala
```diff
@@ -78,17 +78,17 @@ class AsyncQueueSource[T <: Data](gen: T, params: AsyncQueueParams = AsyncQueueP
   val bits = params.bits
   val sink_ready = WireInit(true.B)
   val mem = Reg(Vec(params.depth, gen)) // This does NOT need to be reset at all.
-  val widx = GrayCounter(bits+1, io.enq.fire(), !sink_ready, "widx_bin")
+  val widx = withReset(reset.asAsyncReset)(GrayCounter(bits+1, io.enq.fire(), !sink_ready, "widx_bin"))
```
mwachs5: if we have to do this .asAsyncReset anyway, is there any reason not to just keep the `AsyncResetReg` in `GrayCounter` (and not change this bit of code)?

src/main/scala/util/AsyncQueue.scala
```diff
@@ -78,17 +78,17 @@ class AsyncQueueSource[T <: Data](gen: T, params: AsyncQueueParams = AsyncQueueP
...
   val ridx = AsyncResetSynchronizerShiftReg(io.async.ridx, params.sync, Some("ridx_gray"))
   val ready = sink_ready && widx =/= (ridx ^ (params.depth | params.depth >> 1).U)
 
   val index = if (bits == 0) 0.U else io.async.widx(bits-1, 0) ^ (io.async.widx(bits, bits) << (bits-1))
   when (io.enq.fire()) { mem(index) := io.enq.bits }
 
-  val ready_reg = AsyncResetReg(ready.asUInt, "ready_reg")(0)
+  val ready_reg = withReset(reset.asAsyncReset)(RegNext(next=ready, init=false.B).suggestName("ready_reg"))
```
mwachs5: same here... AsyncResetReg basically just does what this line of code is doing anyway

src/main/scala/util/AsyncQueue.scala
```diff
@@ -158,10 +158,10 @@ class AsyncQueueSink[T <: Data](gen: T, params: AsyncQueueParams = AsyncQueuePar
   val deq_bits_nxt = io.async.mem(if (params.narrow) 0.U else index)
   io.deq.bits := ClockCrossingReg(deq_bits_nxt, en = valid, doInit = false, name = Some("deq_bits_reg"))
 
-  val valid_reg = AsyncResetReg(valid.asUInt, "valid_reg")(0)
+  val valid_reg = withReset(reset.asAsyncReset)(RegNext(next=valid, init=false.B).suggestName("valid_reg"))
   io.deq.valid := valid_reg && source_ready
 
-  val ridx_reg = AsyncResetReg(ridx, "ridx_gray")
+  val ridx_reg = withReset(reset.asAsyncReset)(RegNext(next=ridx, init=0.U).suggestName("ridx_gray"))
```
mwachs5: if these changes are all functionally equivalent in all cases (I think it is) I am inclined to just keep this file unchanged

src/main/scala/util/AsyncQueue.scala
ernie-sifive: I thought one of the goals was to remove all the AsyncResetReg instantiations and replace them with standard chisel Reg driven with the appropriate type of reset.

mwachs5: I think that is correct that was a goal, but when I look at this all I see is the same thing that `AsyncResetReg` is doing internally anyway. If we have to cast the reset (as in, we are explicitly putting down AsyncResetReg even if the reset coming in is not Async reset otherwise) then I dont know if this has an advantage over `AsyncResetReg`, which is no longer a black-box instantiator anyway.

I could be convinced otherwise, just gunshy of modifications to these primitives

ernie-sifive: I think there is also a problem with reset type reassignment.  Trying this just now in an RF system, I get:
Exception in thread "main" firrtl.passes.PassExceptions: 
firrtl.passes.CheckTypes$InvalidConnect: : [module AsyncQueueSource]  Type mismatch in 'widx_bin.reset <= reset'.
  widx_bin.reset: UInt<1>
  reset: AsyncReset

mwachs5: ah, i see, you can't actually can't use the `AsyncResetReg.apply()` without casing the reset yourself to `asAsyncReset` because:

https://github.com/chipsalliance/rocket-chip/blob/4f0cdea85c8a2b849fd582ccc8497892001d06b0/src/main/scala/util/AsyncResetReg.scala#L75

Ok, I am fine with this then

mwachs5: but now that I look again at what `AsyncResetReg` is actually doing, yeah let's just use the native RegInit stuff :)

src/main/scala/tilelink/Arbiter.scala
```diff
@@ -18,12 +18,12 @@ object TLArbiter
...
+    val filter = Cat(valid & maskn, valid)
+    val unready = (rightOR(filter, width*2, width) >> 1) | (~maskn << width)
```
mwachs5:
```suggestion
    mask = RegInit(((BigInt(1) << width)-1).U(width.W))))
    val filter = Cat(valid & ~mask, valid)
    val unready = (rightOR(filter, width*2, width) >> 1) | (mask << width)
```

src/main/scala/tilelink/Arbiter.scala
```diff
@@ -18,12 +18,12 @@ object TLArbiter
...
     val readys = ~((unready >> width) & unready(width-1, 0))
     when (select && valid.orR) {
-      mask := leftOR(readys & valid, width)
+      maskn := ~leftOR(readys & valid, width)
```
mwachs5:
```suggestion
      mask := leftOR(readys & valid, width)
```

mwachs5:
This is failing because fesvr writes dmactive =1, then immediately tries to do another operation and gets an error response. This is due to the synchronization on dmActiveAck which adds a delay before the blocker is released. Another rocket-tools bump to fesvr 😭 ? or other ideas @ernie-sifive ?

mwachs5:
@ernie-sifive I filed FESVR issue (https://github.com/riscv/riscv-isa-sim/issues/435) and added a workaround. But I am confused why the dmactiveAck needs to be synchronized in the testbench anyway. Isn't it already synchronized here:

https://github.com/chipsalliance/rocket-chip/pull/2375/files#diff-579e415dc2dfe635c9d572021aeeccf8R657

ernie-sifive:
@mwachs5 dmactive is generated in dmi_clock domain, get synced to clock in the testbench (where dmactiveAck is used to control the clock gate), then dmactiveAck returns to dmOuter where it is synced back to dmi_clock.  If you remove the sync to clock then won't the clock gate misbehave?

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -198,7 +199,7 @@ class LazyRawModuleImp(val wrapper: LazyModule) extends RawModule with LazyModul
   // It is recommended to drive these even if you manually shove most of your children
   // Otherwise, anonymous children (Monitors for example) will not be clocked
   val childClock = Wire(Clock())
-  val childReset = Wire(Bool())
+  val childReset = Wire(Reset())
```
mwachs5: FYI @hcook

src/main/scala/devices/debug/Periphery.scala
```diff
@@ -246,16 +273,52 @@ object Debug {
...
+      debug_reset := ~AsyncResetSynchronizerShiftReg(in=true.B, sync=3, name=Some("debug_reset_sync"))
+    }
+    // Need to clock DM during debug_reset because of synchronous reset, so keep
+    // the clock alive for one cycle after debug_reset asserts to action this behavior.
+    // The unit should also be clocked when dmactive is high.
+    withClockAndReset(c, debug_reset.asAsyncReset) {
+      // This "if" is really a workaround for FESVR code which does not wait or handle errors after writing dmactive=1
+      // to actually see that the debug module is ready to to be accessed.
+      // See https://github.com/riscv/riscv-isa-sim/issues/435
+      val dmactiveAck = if (dmactiveSyncNeeded) ResetSynchronizerShiftReg(in=debug.dmactive, sync=3, name=Some("dmactiveAck")) else debug.dmactive
```
mwachs5:
```suggestion
      val dmactiveAck = ResetSynchronizerShiftReg(in=debug.dmactive, sync=3, name=Some("dmactiveAck"))
```

mwachs5:
Am going to do a cleanup rebase now

mwachs5:
closing to reopen because travis seems borked

PR #2379, opened 2020-03-30
fix Chisel compile warning: Diplomacy AddressRange match may not be exhaustive

johningalls-sifive:

fix this Chisel compile Scala warning:
```
[warn] /rocket-chip/src/main/scala/diplomacy/AddressRange.scala:51:44: match may not be exhaustive.
[warn] It would fail on the following input: (_, _)
[warn]     ranges.tail.foldLeft(Seq(ranges.head)) { case (head :: tail, x) =>
[warn]                                            ^
```

### DIFF
--- a/src/main/scala/diplomacy/AddressRange.scala
+++ b/src/main/scala/diplomacy/AddressRange.scala
@@ -53,6 +53,7 @@ object AddressRange
         case Some(z) => z :: tail
         case None => x :: head :: tail
       }
+      case (head, tail) => head ++ Seq(tail)
     }.reverse
   }
   // Set subtraction... O(n*n) b/c I am lazy
--- a/src/main/scala/diplomacy/AddressRange.scala
+++ b/src/main/scala/diplomacy/AddressRange.scala
@@ -48,12 +48,13 @@ object AddressRange
   def unify(seq: Seq[AddressRange]): Seq[AddressRange] = {
     if (seq.isEmpty) return Nil
     val ranges = seq.sorted
-    ranges.tail.foldLeft(Seq(ranges.head)) { case (head :: tail, x) =>
-      head.union(x) match {
-        case Some(z) => z :: tail
-        case None => x :: head :: tail
-      }
-      case (head, tail) => head ++ Seq(tail)
+    ranges.tail.foldLeft(Seq(ranges.head)) {
+      case (head :: tail, x) =>
+        head.union(x) match {
+          case Some(z) => z :: tail
+          case None => x :: head :: tail
+        }
+      case _ => Unreachable()
     }.reverse
   }
   // Set subtraction... O(n*n) b/c I am lazy
--- a/src/main/scala/diplomacy/Unreachable.scala
+++ b/src/main/scala/diplomacy/Unreachable.scala
@@ -1,0 +1,7 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.diplomacy
+
+case object Unreachable {
+  def apply(): Nothing = throw new AssertionError("unreachable code")
+}


aswaterman:
I believe in this case, the alternate case is impossible because of the `return` statement above.  Adding the impossible-to-match case makes the code a bit harder to understand.  Scala provides an escape hatch for this case: instead of `ranges.tail.foldLeft(Seq(ranges.head)) { case (head :: tail, x) =>` you can write `(ranges.tail.foldLeft(Seq(ranges.head)): @unchecked) { case (head :: tail, x) =>`.  Might be clearer to first assign `ranges.tail.foldLeft(Seq(ranges.head))` to a `val`, though.

johningalls-sifive:
> `(ranges.tail.foldLeft(Seq(ranges.head)): @unchecked) { case (head :: tail, x) =>`

This gives me an error `@unchecked does not take parameters`.
I've tried a few different places trying to get `@unchecked` to apply to the list internal to the `foldLeft`, but I haven't gotten it to stick anywhere.  :-/

aswaterman:
Even if you first assign `ranges.tail.foldLeft(Seq(ranges.head)` to `val foo` and then do `(foo: @unchecked) match { ...` ?

aswaterman:
Here's the doc btw. https://www.scala-lang.org/api/2.12.3/scala/unchecked.html

johningalls-sifive:
> Even if you first assign `ranges.tail.foldLeft(Seq(ranges.head)` to `val foo` and then do `(foo: @unchecked) match { ...` ?

That gives me error `missing argument list for method foldLeft`.
If I don't create a val and instead add `match` after `@unchecked` then I get error `@unchecked does not take parameters`.

aswaterman:
Uhh.. sorry, I misread the code.  Sorry for wasting your time on that.

aswaterman:
Scala experts will know how better to express this sort of thing, but you can also replace your `case (head, tail) => head ++ Seq(tail)` with `case _ => ???` to explicitly throw an exception in the event of a match error.  I guess the author of this code should decide.

johningalls-sifive:
no problem @aswaterman, thank you for jumping in to help!

src/main/scala/diplomacy/AddressRange.scala
```diff
@@ -53,6 +53,7 @@ object AddressRange
         case Some(z) => z :: tail
         case None => x :: head :: tail
       }
+      case (head, tail) => head ++ Seq(tail)
```
terpstra, line 56: Pretty sure this should just be: `case (Nil, _) => require(false)`

terpstra: What I don't like is your PR seems to suggest this case is possible and that there is something to be done.

johningalls-sifive: > Pretty sure this should just be: `case (Nil, _) => require(false)`

How about this, to get rid of the warning?
```
case (_, _) => require(false); Nil
```

terpstra: So I just spent some time googling for this and playing around with scala, because I have often wondered how best to mark something unreachable. The problem with 'require(false)' is that it does not have polymorphic return type. That's because, usually, a require can succeed. This is why you added '; Nil'.

There are some ways to get a polymorphic return type in scala that I found:
- you can `throw` something, perhaps `AssertionError("unreachable")`
- you can put `???` which results in [error] Caused by: scala.NotImplementedError: an implementation is missing

Having to write `throw AssertionError("unreachable") ` is more typing than `??? // unreachable`, but it is more descriptive.

Frankly, I would like to `def unreachable: Empty = throw AssertionError("unreachable")` somewhere in rocket-chip and then we can just put `unreachable` in these places.

terpstra:
I just force-pushed this branch with my version of the fix rebased onto current master.

terpstra:
@johningalls-sifive If you approve of my version, feel free to merge.

terpstra:
Actually, do either of @hcook @jackkoenig know of an existing 'Unreachable()' equivalent in scala? I could not find one. That's why I rolled my own. `???` is not the same, as it implies you haven't implemented that code yet, but it could be reachable.

jackkoenig:
I don't know of one and I don't see anything like that in Predef. I think your `Unreachable()` is good and clear. It could alternatively be fun to make a def named `!!!` as an analogy to `???`.

terpstra:
I can't believe I'm the one arguing against using an operator but I think !!! less readable.

PR #2381, opened 2020-03-30
Only checkout the latest 50 commits of rocket-tools submodules.

richardxia:
**Related issue**: 

**Release Notes**

I think this should speed up the rocket-tools checkout step by not cloning the entire history of the whole `gcc` Git repo. I chose 50 since it's what Travis CI defaults to and because it seems to be a good balance of keeping the checkouts small while also pulling more history if it's useful.

I have not tested this and plan to rely on Travis CI to test it for me.

### DIFF
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -150,13 +150,15 @@ stamps/other-submodules.stamp:
 	git -C $(abspath $(TOP)) submodule update --init --recursive $(submodule_names)
 	date > $@
 
+ROCKET_TOOLS_CHECKOUT_DEPTH ?= 50
+
 stamps/rocket-tools_checkout.stamp:
 	mkdir -p $(dir $@)
 	git -C $(abspath $(TOP)) clone -n https://github.com/freechipsproject/rocket-tools.git
 	git -C $(RISCV_TOOLS) checkout $(TOOLS_HASH)
-	git -C $(RISCV_TOOLS) submodule update --init --recursive riscv-isa-sim  riscv-tests riscv-pk riscv-openocd fsf-binutils-gdb
-	git -C $(RISCV_TOOLS) submodule update --init riscv-gnu-toolchain
-	git -C $(RISCV_TOOLS)/riscv-gnu-toolchain submodule update --init riscv-binutils-gdb riscv-gcc riscv-newlib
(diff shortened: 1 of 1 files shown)


aswaterman:
How does depth work for nested submodules?  Does it check out the last 50 commits from each submodule, or does it check out the commits corresponding to the last 50 commits of its parent?

richardxia:
> How does depth work for nested submodules? Does it check out the last 50 commits from each submodule, or does it check out the commits corresponding to the last 50 commits of its parent?

The specifics of this are apparently very complicated and depend on the specific version of Git you are running: https://stackoverflow.com/questions/2144406/how-to-make-shallow-git-submodules

I think it usually checks out the last 50 commits from each submodule. However, on newer-ish versions of Git, if the requested submodule commit hash was not fetched, it will attempt to fetch that commit directly from the remote server. Unfortunately, GitHub does not support directly fetching a commit by commit hash...

I got the following error trying to run this on Git 2.20.1:

```sh
error: Server does not allow request for unadvertised object ffea1427a7872b8b1237a3400dadf2e34da8f793
Fetched in submodule path 'fsf-binutils-gdb', but it did not contain ffea1427a7872b8b1237a3400dadf2e34da8f793. Direct fetching of that commit failed.
Makefile:156: recipe for target 'stamps/rocket-tools_checkout.stamp' failed
```

I got the following error trying to run this on Git 2.7.4 (default on Ubuntu 16.04):

```sh
Unable to checkout 'ffea1427a7872b8b1237a3400dadf2e34da8f793' in submodule path 'fsf-binutils-gdb'
```

I believe Git 2.7.4 does not exhibit the "fetch sha1 directly if it was not fetched originally" behavior, since it was introduced in this commit which only came in Git 2.8: https://github.com/git/git/commit/fb43e31f2b43076e7a30c9cd00d0241cb8cf97eb

In any case, it seems like the strategy of doing a shallow clone won't actually work since even if we use a newer Git, GitHub won't let you fetch SHAs directly. Because the version of GDB we want is way older than 50 commits than master, this just won't work.

---

It seems like none of my attempts at improving the build on rocket-chip are really working. At least these PRs can serve as a record of what was attempted and why they failed.

aswaterman:
Ugh. Thanks for the explanation.

PR #2383, opened 2020-03-30
Drive protection user bits on TL

jsmithsf:
**Related issue**: 

Drive some relevant AMBA protection field information from Rocket.

**Impact**: API modification
When Rocket is configured to not use virtual memory, it will drive extra information on the TL user bits that can be converted into the AMBA protection fields.

**Release Notes**

The instruction vs data, cacheable vs. non-cacheable, and privileged vs. non-privileged information should be driven now.

Now we will mark an access as privileged if the core is in machine mode, or if the access is to a cacheable region.


jsmithsf:
The changes are pretty minor, just looking for an approval from someone

src/main/scala/rocket/DCache.scala
```diff
@@ -551,6 +552,20 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
     Mux(s2_req.cmd === M_PWR, putpartial,
     Mux(!s2_read, put, atomics))))
 
+  // Drive APROT Bits
+  tl_out_a.bits.user.lift(AMBAProt).foreach { x =>
+    val user_bit_cacheable = edge.manager.supportsAcquireTFast(access_address, a_size)
+
+    x.privileged  := s2_req.dprv === PRV.M || user_bit_cacheable
```
mwachs5: so all cacheable accesses are also by definition privileged?

src/main/scala/rocket/ICache.scala
```diff
@@ -436,6 +443,19 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
     ccover(!refill_valid && (tl_out.d.fire() && !refill_one_beat), "PREFETCH_D_AFTER_MISS_D", "I$ prefetch resolves after miss")
     ccover(tl_out.a.fire() && hint_outstanding, "PREFETCH_D_AFTER_MISS_A", "I$ prefetch resolves after second miss")
   }
+  // Drive APROT information
+  tl_out.a.bits.user.lift(AMBAProt).foreach { x =>
+    val user_bit_cacheable = edge_out.manager.supportsAcquireTFast(refill_paddr, lgCacheBlockBytes.U)
+
+    x.privileged  := io.privileged || user_bit_cacheable    // privileged if machine mode or memory port
```
mwachs5: oh i see the comment here. so we are assuming everythign cacheable is going to memory port? If not this comment is a bit misleading

src/main/scala/rocket/DCache.scala
jsmithsf: The relevant point from the AMBA spec is that if a master cannot generate accurate protection information, it is recommended that we mark the transaction as privileged.  Because caches like the L2 can merge requests from multiple different privilege modes, we can't accurately provide the protection information.

src/main/scala/rocket/ICache.scala
jsmithsf: Yes, currently we are assuming any cacheable thing is going to the memory port.  During a short discussion that I had with @hcook he mentioned that it's possible this may change in the future if we add other types of ports.

solomatnikov: It seems it should be the opposite for safety, i.e. if you cannot figure out the exact privilege level, you should assume the lowest level of privilege.

jsmithsf: I don't understand ARM's reasoning behind their decision, but we are making this change to be more compatible with ARM behavior.  So I think for that reason we should follow the same thing.

mwachs5:
If @daveparry can review i can rubber stamp (assuming his review is not strong enough)

src/main/scala/rocket/DCache.scala
```diff
@@ -551,6 +552,20 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
...
+    x.cacheable   := user_bit_cacheable
+
+    // Following are always tied off
+    x.fetch       := false.B
+    x.secure      := true.B
+    x.bufferable  := false.B
+    x.modifiable  := false.B
```
terpstra: I wonder whether bufferable/modifiable should be !user_bit_cacheable ?

src/main/scala/rocket/ICache.scala
terpstra: I am confused. Since we cache everything on fetch, why is privileged not just tied to true?

jsmithsf: @terpstra, @aswaterman posed a similar question but was unsure.

jerryhethatday:
@jsmithsf Hi, Thanks for your work. I am rookie to rocketchip, and want to understand why AMBAProtField is needed when virtual memory is disabled.
```
  val masterNode = TLClientNode(Seq(TLMasterPortParameters.v1(
    clients = Seq(TLMasterParameters.v1(
      sourceId = IdRange(0, 1 + icacheParams.prefetch.toInt), // 0=refill, 1=hint
      name = s"Core ${staticIdForMetadataUseOnly} ICache")),
    requestFields = useVM.option(Seq()).getOrElse(Seq(AMBAProtField())))))
```
I am working on a chipyard configuration(rv32imac without os support), and when I port the deign to vcu118 fpga prototyping, the firrtl will complain some amba signal is not initialized. 
I do some debugging, and found the problem is icache masterNode will have amba related userbits, through diplomacy negotiation, these  `requestFields` will go downsteam and reach an end of fpga signal while there is no corresponding amba signal. So, I want to know why you add this in the first place. Thanks

PR #2386, opened 2020-03-31
amba: combine modifiable+cacheable and add {read,write}alloc

terpstra:
I misunderstood the AMBA specifications:
  HPROT[3] (cacheable)
and
  ARPROT[1] (modifiable)
are supposed to be the same, despite using different names in AHB vs. AXI4.

This mismatch then led me to incorrectly map ARPROT[2]+ARPROT[3] to
the AHB cacheable bit.

Add explicit readalloc+writealloc bits to fully cover the AXI4 bits.

Also, in most places, where we say 'cacheable', we mean set all of the bits:
  bufferable (writeback ok)
  modifiable (write-combining / fetch whole blocks ok)
  readalloc  (use outer caches for reads)
  writealloc (use outer caches for writes)

... I updated rocket accordingly.

### DIFF
--- a/src/main/scala/amba/apb/ToTL.scala
+++ b/src/main/scala/amba/apb/ToTL.scala
@@ -92,8 +92,9 @@ class APBToTL()(implicit p: Parameters) extends LazyModule
         prot.secure     := !in.pprot(1)
         prot.fetch      :=  in.pprot(2)
         prot.bufferable :=  true.B
-        prot.cacheable  :=  true.B
         prot.modifiable :=  true.B
+        prot.readalloc  :=  true.B
+        prot.writealloc :=  true.B
       }
       when (out.a.fire()) {
         assert(in.paddr === out.a.bits.address, "Do not expect to have to perform alignment in APB2TL Conversion")
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -554,7 +554,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
 
   // Drive APROT Bits
   tl_out_a.bits.user.lift(AMBAProt).foreach { x =>
-    val user_bit_cacheable = edge.manager.supportsAcquireTFast(access_address, a_size)
+    val user_bit_cacheable = s2_pma.cacheable
 
     x.privileged  := s2_req.dprv === PRV.M || user_bit_cacheable
     // if the address is cacheable, enable outer caches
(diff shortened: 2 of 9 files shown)


src/main/scala/amba/apb/ToTL.scala
```diff
@@ -92,8 +92,9 @@ class APBToTL()(implicit p: Parameters) extends LazyModule
         prot.secure     := !in.pprot(1)
         prot.fetch      :=  in.pprot(2)
         prot.bufferable :=  true.B
-        prot.cacheable  :=  true.B
         prot.modifiable :=  true.B
+        prot.readalloc  :=  true.B
+        prot.writealloc :=  true.B
```
terpstra, line 97: This is opposite to the default values used in AMBA. Perhaps these should all be false?

src/main/scala/rocket/ICache.scala
```diff
@@ -448,13 +448,16 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
     val user_bit_cacheable = edge_out.manager.supportsAcquireTFast(refill_paddr, lgCacheBlockBytes.U)
 
     x.privileged  := io.privileged || user_bit_cacheable    // privileged if machine mode or memory port
-    x.cacheable   := user_bit_cacheable
+
+    // enable outer caches for cacheable address PMAs
+    x.bufferable  := user_bit_cacheable
+    x.modifiable  := user_bit_cacheable
+    x.readalloc   := user_bit_cacheable
+    x.writealloc  := user_bit_cacheable
```
terpstra, line 456: I did NOT tie these all to true. I think that is a debate for another PR.

src/main/scala/rocket/DCache.scala
```diff
@@ -557,13 +557,15 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
     val user_bit_cacheable = edge.manager.supportsAcquireTFast(access_address, a_size)
```
terpstra, line 557: The calculation of 'user_bit_cacheable' is potentially expensive. Can we reuse this calculation from the TLB permissions check? @aswaterman

jsmithsf: For rocket/bullet/mallard, yeah we may be able to. Caboose would have to do something else.

src/main/scala/rocket/ICache.scala
terpstra: Since travis failed anyway, I made this change, since it sounds like we now agree this is the right way to go.

Issue #2389, opened 2020-04-02
AXI4.B channel could be abnormally blocked by TLToAXI4


scott-kim-7:
**Type of issue**: other enhancement

**Impact**: functional change

**Development Phase**: proposal

**Other information**
 

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
- Environment: 5 AXI4 Masters are connected to 1 AXI4 slave through TLXbar 
- 4 of AXI4 masters issues 100Kbytes read-burst transactions(ARLEN=7, max-outstanding=16)
- 1 of AXI4 masters issues 100Kbytes write-burst transactions(ARLEN=7, max-outstanding=16)
==>

**What is the current behavior?**
The AXI4 master(writing) could not get AXI4.B response until the other 4 masters have done.
TLToAXI4 gives priority to read-response(AXI4.R), it is fixed priority.
So, a bunch of read-transactions could block all of AXI4.B response channel transactions abnormally.

**What is the expected behavior?**
I propose that AXI4.B channel get a chance to win whenever a read response has done.

**Please tell us about your environment:**

- version: 7c072e9999a1be4028aed383dcf97b03e2773988
- OS: CentOS Linux 7 (Core)
**What is the use case for changing the behavior?**

PR #2391, opened 2020-04-02
ToAXI4.scala fix abnormal AXI4.B channel blocking problem

scott-kim-7:
**Related issue**: AXI4.B channel could be abnormally blocked by TLToAXI4 #2389

**Impact**: functional change

**Release Notes**


scott-kim-7:
We usually use this for the memory controller. So at least, the bandwidth should not be biased.
In my opinion, if an r-b channel arbitration policy has a chance to make biasing, we should not take the policy.

PR #2392, opened 2020-04-03
coreMonitorBundle print trace log wrdst only if instruction writes a register

johningalls-sifive:
**Related issue**: follow-on to #2372
**Impact**: API modification

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -875,8 +875,8 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
   coreMonitorBundle.timer := csr.io.time(31,0)
   coreMonitorBundle.valid := csr.io.trace(0).valid && !csr.io.trace(0).exception
   coreMonitorBundle.pc := csr.io.trace(0).iaddr(vaddrBitsExtended-1, 0).sextTo(xLen)
-  coreMonitorBundle.wrdst := rf_waddr
-  coreMonitorBundle.wren := rf_wen && !(wb_set_sboard && wb_wen)
+  coreMonitorBundle.wrdst := Mux(wb_ctrl.wxd || wb_ctrl.wfd, wb_waddr, 0.U)
+  coreMonitorBundle.wren := wb_wen && !(wb_set_sboard && wb_wen)
   coreMonitorBundle.wrdata := Mux(coreMonitorBundle.wren, rf_wdata, 0.U)
   coreMonitorBundle.rd0src := wb_reg_inst(19,15)
   coreMonitorBundle.rd0val := Reg(next=Reg(next=ex_rs(0)))


src/main/scala/rocket/RocketCore.scala
```diff
@@ -875,8 +875,8 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
   coreMonitorBundle.timer := csr.io.time(31,0)
   coreMonitorBundle.valid := csr.io.trace(0).valid && !csr.io.trace(0).exception
   coreMonitorBundle.pc := csr.io.trace(0).iaddr(vaddrBitsExtended-1, 0).sextTo(xLen)
-  coreMonitorBundle.wrdst := rf_waddr
-  coreMonitorBundle.wren := rf_wen && !(wb_set_sboard && wb_wen)
+  coreMonitorBundle.wrdst := Mux(wb_ctrl.wxd || wb_ctrl.wfd, wb_waddr, 0.U)
```
aswaterman, line 878: I don't think this mux does anything useful, since `wrdst` is a don't-care when `wren` is false, and with your change, `wren` is only true when `wb_ctrl.wxd` is true.

src/main/scala/rocket/RocketCore.scala
```diff
@@ -875,8 +875,8 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
...
+  coreMonitorBundle.wren := wb_wen && !(wb_set_sboard && wb_wen)
```
aswaterman, line 879: This change removes the ability to see long-latency writebacks in the log.  That's OK, but might be useful to add a second printout for those in the future.

src/main/scala/rocket/RocketCore.scala
johningalls-sifive: it's actually the inverse that we want to use to identify long-latency writebacks: when `wb_ctrl.wxd` is true and `wren` is false

src/main/scala/rocket/RocketCore.scala
johningalls-sifive: yeah that is still a hole: logging long-latency writebacks after commit, potentially out-of-order

src/main/scala/rocket/RocketCore.scala
aswaterman: I don't see how that works, though: `wb_ctrl.wxd` and `wb_ctrl.wfd` are themselves don't-cares some of the time, namely, on cycles when there's nothing in the writeback stage and no long-latency writeback happening.

Furthermore, 0 doesn't work as a sentinel value when FP registers are involved, because f0 isn't hardwired.

It seems like you're trying to multiplex an additional control signal on top of this field... why not just add the control signal?

johningalls-sifive: I agree with your principles, and I'll have to think on it / discuss further for specifics.

src/main/scala/rocket/RocketCore.scala
johningalls-sifive: we're thinking of how to do this outside of rocket-chip, and we'll likely pursue the idea of a second printout for long-latency writebacks there

PR #2399, opened 2020-04-06
Bump to Chisel v3.3.0 and FIRRTL v1.3.0

jackkoenig:
* Update some uses of @chiselName to disable new prefixing behavior
* Update Stage/Phase dependencies to new Dependency instead of Class
* Technically bumped to merge-base of each tag with their respective
  release branches: 3.3.x for chisel3 and 1.3.x for firrtl
  This better supports Wit dependencies
* Include all chisel3 class files on FIRRTL run classpath

We include chisel3 on the classpath for running FIRRTL because of use of utils like `chisel3.util.HasBlackBoxResource`. https://github.com/freechipsproject/chisel3/pull/1384 broke up the chisel3 build into multiple directories so I had to update the Makefiles to include the extra directories.

TODO Release Notes Highlights

**Related issue**: 

**Impact**: API modification

**Release Notes**
Bump to Chisel v3.3.0-RC3 and FIRRTL v1.3.0-RC3. Chisel and FIRRTL use a versioning scheme `X.Y.Z` where `Y` signifies major version and thus may contain backwards incompatible API changes. See their respective release notes (currently in draft):
https://github.com/freechipsproject/chisel3/releases
https://github.com/freechipsproject/firrtl/releases

### DIFF
--- a/Makefrag
+++ b/Makefrag
@@ -31,7 +31,13 @@ SHELL := /bin/bash
 FIRRTL_TRANSFORMS := \
 	firrtl.passes.InlineInstances \
 
-ROCKET_CLASSES ?= "$(base_dir)/target/scala-2.12/classes:$(base_dir)/chisel3/target/scala-2.12/*"
+ROCKET_CLASS_DIRS ?= \
+	$(base_dir)/target/scala-2.12/classes \
+	$(base_dir)/chisel3/target/scala-2.12/classes \
+	$(base_dir)/chisel3/core/target/scala-2.12/classes \
+	$(base_dir)/chisel3/macros/target/scala-2.12/classes
+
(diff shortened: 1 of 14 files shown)


jackkoenig:
This has been updated to RC3, it currently will fail the Wake compilation test because I need to update api-chisel3-sifive and api-firrtl-sifive, but the functionality should be correct.

mwachs5:
the two failign tests looked like transient maven issues so i kicked them off again

mwachs5:
Does this include the workaround https://github.com/freechipsproject/firrtl/pull/1556?

jackkoenig:
> Does this include the workaround freechipsproject/firrtl#1556?

Not yet, I was thinking about copy-pasting the code here and then removing it on the next bump but I could also just try to get the fix in and backported and then we can bump along the stable branch, a bit past `1.3.0`.

PR #2403, opened 2020-04-07
Bump hardfloat (revert DivSqrtRecF64 to use Module instead RawModule)

pentin-as:
https://github.com/ucb-bar/berkeley-hardfloat/issues/44

### DIFF
--- a/wit-manifest.json
+++ b/wit-manifest.json
@@ -1,6 +1,6 @@
 [
     {
-        "commit": "3f88b663c7324cf9a27f9e20b82382908bb4cc11",
+        "commit": "22ebeb54301ca85fa087180596d21fa56335fdbc",
         "name": "hardfloat",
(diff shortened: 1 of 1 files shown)


mwachs5:
Note this will not work on any config that tries to instantiate these in a domain with AsyncReset. You'll get an error trying to connect a `AsyncReset` to a `Bool`. The right thing to do before performing this bump is to convert those files to chisel3 (or at least use something like this Compatibility mode workaround) so that they can be used in systems with Async Reset:

https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/devices/tilelink/Plic.scala#L6

mwachs5:
I am going to close and re-open this PR to try to get travis to do its thing

mwachs5:
@pentin-as you will need to update the `wit-manifest.json` to match this -- it is failing Travis regressions

PR #2407, opened 2020-04-08
Connect source to sink in RationalCrossingFull

albert-magyar:
**Impact**: no API changes

I don't think the full "crossing in its own module" was being used.

### DIFF
--- a/src/main/scala/util/RationalCrossing.scala
+++ b/src/main/scala/util/RationalCrossing.scala
@@ -149,6 +149,7 @@ class RationalCrossingFull[T <: Data](gen: T, direction: RationalDirection = Sym
 
   source.io.enq <> io.enq
   io.deq <> sink.io.deq
+  sink.io.enq <> source.io.deq
 }
 
 object ToRational


terpstra:
An alternative would be to remove the unused module.

sequencer:
> An alternative would be to remove the unused module.

I just write a transform to do this:
https://github.com/sequencer/diplomatictester/blob/dependencyapi/src/firrtl/RemoveUnreachableModules.scala

Maybe it will be useful too?

albert-magyar:
It looks like RationalCrossingFull is still around, so perhaps this fix is still worthwhile?

sequencer:
Since this is a bug fix.

sequencer:
Sorry I miss deleted this comment.
> It looks like RationalCrossingFull is still around, so perhaps this fix is still worthwhile?

I think we can safely merge this PR.

tymcauley:
Just found a local workaround for this issue at my company, wanted to bump this :)

albert-magyar:
Bump

jerryz123 (approved):

PR #2343, opened 2020-03-18
Create OMErrorDevice.scala [WIP]

mwachs5:
**Related issue**: 

**Type of change**: feature request

**Impact**: API addition (no impact on existing code) 

**Development Phase**:   implementation

**Release Notes**

Add Object Model for TLError Device.

PR #2410, opened 2020-04-11
Attach TLError LogicalTreeNodes to subsystem LogicalTreeNode

albertchen-sifive:
This is my attempt at adding `OMErrorDevice` to the object model. It seems to work based on visual inspection of the resulting objectModel.json. Looking for feedback on whether this is the best way of plumbing and attaching the logicalTreeNodes.

**Related issue**: 
includes https://github.com/chipsalliance/rocket-chip/pull/2343

**Impact**: API modification
- factor out `CanHaveBuiltInDevices.attachBuiltInDevices` into `attach` helper method in `BuiltInDevices` companion object
- `CanHaveBuiltInDevices` has abstract field `def builtInDevices: BuiltInDevices`
- `TLError` extends `HasLogicalTreeNode`
- `TLZero` extends `HasLogicalTreeNode`
- argument `device` of  `DevNullDevice` is a `protected val`

**Release Notes**

Add Object Model for TLError Device.

### DIFF
--- a/src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
+++ b/src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
@@ -16,14 +16,14 @@ sealed trait BuiltInDevices {
   def zeroOpt: Option[TLZero]
 }
 
-/* Optionally add some built-in devices to a bus wrapper */
-trait CanHaveBuiltInDevices { this: TLBusWrapper =>
-
-  def attachBuiltInDevices(params: HasBuiltInDeviceParams): BuiltInDevices = new BuiltInDevices {
+object BuiltInDevices {
+  def attach(
+    params: HasBuiltInDeviceParams with HasTLBusParams,
+    outwardNode: TLOutwardNode)(implicit p: Parameters) = new BuiltInDevices {
     val errorOpt = params.errorDevice.map { dnp => LazyScope("wrapped_error_device") {
       val error = LazyModule(new TLError(
         params = dnp,
-        beatBytes = beatBytes))
+        beatBytes = params.beatBytes))
 
       error.node := TLBuffer() := outwardNode
       error
@@ -32,10 +32,15 @@ trait CanHaveBuiltInDevices { this: TLBusWrapper =>
     val zeroOpt = params.zeroDevice.map { addr => LazyScope("wrapped_zero_device") {
       val zero = LazyModule(new TLZero(
         address = addr,
-        beatBytes = beatBytes))
-      zero.node := TLFragmenter(beatBytes, blockBytes) := TLBuffer() := outwardNode
+        beatBytes = params.beatBytes))
+      zero.node := TLFragmenter(params.beatBytes, params.blockBytes) := TLBuffer() := outwardNode
       zero
     }}
   }
+}
+
+/* Optionally add some built-in devices to a bus wrapper */
+trait CanHaveBuiltInDevices {
+  def builtInDevices: BuiltInDevices
 }
 
--- a/src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
+++ b/src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
@@ -11,22 +11,30 @@ trait HasBuiltInDeviceParams {
   val errorDevice: Option[DevNullParams]
 }
 
+sealed trait BuiltInDevices {
+  def errorOpt: Option[TLError]
+  def zeroOpt: Option[TLZero]
+}
+
 /* Optionally add some built-in devices to a bus wrapper */
 trait CanHaveBuiltInDevices { this: TLBusWrapper =>
 
-  def attachBuiltInDevices(params: HasBuiltInDeviceParams) {
-    params.errorDevice.foreach { dnp => LazyScope("wrapped_error_device") {
+  def attachBuiltInDevices(params: HasBuiltInDeviceParams): BuiltInDevices = new BuiltInDevices {
+    val errorOpt = params.errorDevice.map { dnp => LazyScope("wrapped_error_device") {
       val error = LazyModule(new TLError(
         params = dnp,
         beatBytes = beatBytes))
+
       error.node := TLBuffer() := outwardNode
+      error
     }}
 
-    params.zeroDevice.foreach { addr => LazyScope("wrapped_zero_device") {
+    val zeroOpt = params.zeroDevice.map { addr => LazyScope("wrapped_zero_device") {
       val zero = LazyModule(new TLZero(
         address = addr,
         beatBytes = beatBytes))
       zero.node := TLFragmenter(beatBytes, blockBytes) := TLBuffer() := outwardNode
+      zero
     }}
   }
 }
--- a/src/main/scala/diplomaticobjectmodel/model/OMErrorDevice.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMErrorDevice.scala
@@ -1,0 +1,11 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.diplomaticobjectmodel.model
+
+
+case class OMErrorDevice(
+  memoryRegions: Seq[OMMemoryRegion],
+  interrupts: Seq[OMInterrupt],
+  specifications: Seq[OMSpecification],
+  _types: Seq[String] = Seq("OMErrorDevice", "OMDevice", "OMComponent", "OMCompoundType")
+) extends OMDevice
--- a/src/main/scala/subsystem/BaseSubsystem.scala
+++ b/src/main/scala/subsystem/BaseSubsystem.scala
@@ -143,8 +143,13 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem
     mbus.builtInDevices,
     cbus.builtInDevices
   )
-  for (builtIn <- builtInDevices; error <- builtIn.errorOpt) {
-    LogicalModuleTree.add(logicalTreeNode, error.logicalTreeNode)
+  builtInDevices.foreach { builtIn => 
+    builtIn.errorOpt.foreach { error => 
+      LogicalModuleTree.add(logicalTreeNode, error.logicalTreeNode)
+     }
+    builtIn.zeroOpt.foreach { zero => 
+      LogicalModuleTree.add(logicalTreeNode, zero.logicalTreeNode)
+    }
   }
 }
 
--- a/src/main/scala/subsystem/BaseSubsystem.scala
+++ b/src/main/scala/subsystem/BaseSubsystem.scala
@@ -134,20 +134,17 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem
     }
   }
 
-  lazy val logicalTreeNode = {
-    val subsystemLogicalTreeNode = new SubsystemLogicalTreeNode()
-    val builtInDevices = Seq(
-      pbus.builtInDevices,
-      fbus.builtInDevices,
-      mbus.builtInDevices,
-      cbus.builtInDevices
-    )
-    for (builtIn <- builtInDevices) {
-      builtIn.errorOpt.foreach { error =>
-        LogicalModuleTree.add(subsystemLogicalTreeNode, error.logicalTreeNode)
-      }
-    }
-    subsystemLogicalTreeNode
+  lazy val logicalTreeNode = new SubsystemLogicalTreeNode()
+
+  private val builtInDevices = Seq(
+    sbus.builtInDevices,
+    pbus.builtInDevices,
+    fbus.builtInDevices,
+    mbus.builtInDevices,
+    cbus.builtInDevices
+  )
+  for (builtIn <- builtInDevices; error <- builtIn.errorOpt) {
+    LogicalModuleTree.add(logicalTreeNode, error.logicalTreeNode)
   }
 }
 
--- a/src/main/scala/subsystem/BaseSubsystem.scala
+++ b/src/main/scala/subsystem/BaseSubsystem.scala
@@ -81,7 +81,7 @@ case object ResetAsynchronousFull extends SubsystemResetScheme
 case object SubsystemResetSchemeKey extends Field[SubsystemResetScheme](ResetSynchronous)
 
 /** Base Subsystem class with no peripheral devices or ports added */
-abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem 
+abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem
     with Attachable {
 
   override val module: BaseSubsystemModuleImp[BaseSubsystem]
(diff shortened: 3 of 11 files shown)


src/main/scala/diplomaticobjectmodel/model/OMErrorDevice.scala
```diff
@@ -0,0 +1,11 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.diplomaticobjectmodel.model
+
+
+case class OMErrorDevice(
+  memoryRegions: Seq[OMMemoryRegion],
+  interrupts: Seq[OMInterrupt],
+  specifications: Seq[OMSpecification],
```
richardxia, line 9: What is this `specifications` field for? You seem to only set it to `Nil`. We should just remove it if there's not defined meaning to it.

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -134,7 +134,21 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem
...
+      pbus.builtInDevices,
+      fbus.builtInDevices,
+      mbus.builtInDevices,
+      cbus.builtInDevices
+    )
+    for (builtIn <- builtInDevices) {
+      builtIn.errorOpt.foreach { error =>
+        LogicalModuleTree.add(subsystemLogicalTreeNode, error.logicalTreeNode)
+      }
+    }
```
richardxia, line 149: Not sure if this is more readable, but how about:

```suggestion
    for (builtIn <- builtInDevices; error <- builtIn.errorOpt) {
      LogicalModuleTree.add(subsystemLogicalTreeNode, error.logicalTreeNode)
    }
```
?

src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
```diff
@@ -11,22 +11,30 @@ trait HasBuiltInDeviceParams {
   val errorDevice: Option[DevNullParams]
 }
 
+sealed trait BuiltInDevices {
```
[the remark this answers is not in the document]
albertchen-sifive: I the main difference I care about vs a final case class is that the case class generates a bunch of methods and extends stuff that I don't think are appropriate/useful. like case class extends `serializable` which `TLError` and `TLZero` definitely are not and it generates stuff like unapply that I don't want exposed because that API will probably be broken.

sealed is for similar reasons. to prevent anyone else from creating `BuiltInDevices` outside of `attachBuiltInDevices` and just narrow down the API surface area.

just my personal preference, maybe I'm being a too cautious or this isn't the right approach? open to changing this.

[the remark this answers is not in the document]
richardxia: > just my personal preference, maybe I'm being a too cautious or this isn't the right approach? open to changing this.

No, I think it's a good idea to limit API surface area. I was mostly wondering `sealed trait` vs `final case class` and not `sealed trait` vs `(not sealed) trait`. I hadn't thought of the extra methods and traits that case classes provide as being extra stuff that consumers could possibly depend on that would become later liabilities, but now that you point it out, I agree, especially because this data structure is, at the moment, an arbitrary bundle of "stuff", and we may want to reserve the right to change this in the future.

src/main/scala/subsystem/BaseSubsystem.scala
albertchen-sifive: looking at this again, I'm not sure why I didn't just do this to be consistent
```suggestion
    builtInDevices.foreach { builtIn =>
      builtIn.errorOpt.foreach { error =>
        LogicalModuleTree.add(subsystemLogicalTreeNode, error.logicalTreeNode)
      }
    }
```
but I dunno. I guess the your version looks the best to me.

mwachs5: I think @albertchen-sifive 's is more what I'm used to in the code base, but either change looks good and symmetrical :)

src/main/scala/diplomaticobjectmodel/model/OMErrorDevice.scala
mwachs5: This is probably due to my PR, which was a blind copy-paste of some other `OM*` from some other device.

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -134,7 +134,21 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem
...
+  lazy val logicalTreeNode = {
+    val subsystemLogicalTreeNode = new SubsystemLogicalTreeNode()
+    val builtInDevices = Seq(
+      pbus.builtInDevices,
+      fbus.builtInDevices,
+      mbus.builtInDevices,
+      cbus.builtInDevices
+    )
+    for (builtIn <- builtInDevices) {
+      builtIn.errorOpt.foreach { error =>
```
mwachs5, line 146: Since this has taken the trouble to do "the hard part" here I am wondering if we shouldn't extend this PR to do the same for builtIn.zeroOpt. I can push to this branch to add that...

mwachs5:
Curious if this is going to massively conflict with @hcook's https://github.com/chipsalliance/rocket-chip/pull/2327... if so would be nice to get this in first...

src/main/scala/subsystem/BaseSubsystem.scala
albertchen-sifive: more importantly. I just realized that we are overriding this field. so I guess this stuff should go in an `attachBuiltInDevicesLogicalTreeNodes` method that can be called in the constructor of the LogicalTreeNode?

richardxia: Ah, I had forgotten about the `lazy val` being overridden by subclasses. Now that you've brought it up, I don't actually think we want any of this stuff happening in the constructor of the LogicalTreeNode. Is there a problem with just moving all of the `LogicalModuleTree.add()` statements below/outside of the `lazy val logicalTreeNode = { }` statement?

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -134,7 +134,21 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem
...
   }
 
-  lazy val logicalTreeNode = new SubsystemLogicalTreeNode()
+  lazy val logicalTreeNode = {
+    val subsystemLogicalTreeNode = new SubsystemLogicalTreeNode()
+    val builtInDevices = Seq(
+      pbus.builtInDevices,
+      fbus.builtInDevices,
+      mbus.builtInDevices,
+      cbus.builtInDevices
```
mwachs5, line 143: is there a reason you left off `sbus.builtInDevices?`

src/main/scala/subsystem/BaseSubsystem.scala
mwachs5: I will push a change that adds the OMZero stuff but won't touch this part of the code since it's still under discussion

src/main/scala/subsystem/BaseSubsystem.scala
albertchen-sifive: no. good catch

src/main/scala/subsystem/BaseSubsystem.scala
albertchen-sifive: that would force evaluation of the `lazy val logicalTreeNode` early? doesn't seem like it matters when the logicalTreeNode gets evaluated though, because it still works if I move it outside of the block.

richardxia: Yeah, that should be fine, since we have a pattern of doing this in a number of places. The logicalTreeNodes are only really meant to create the tree of nodes without fulling evaluated the nodes, somewhat analogous to Diplomatic nodes.

albertchen-sifive: I don't get how `lazy val`s work

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -135,6 +135,17 @@ abstract class BaseSubsystem(implicit p: Parameters) extends BareSubsystem
...
+  private val builtInDevices = Seq(
+    sbus.builtInDevices,
+    pbus.builtInDevices,
+    fbus.builtInDevices,
+    mbus.builtInDevices,
+    cbus.builtInDevices
+  )
+  for (builtIn <- builtInDevices; error <- builtIn.errorOpt) {
+    LogicalModuleTree.add(logicalTreeNode, error.logicalTreeNode)
+  }
```
mwachs5, line 148:
```suggestion
  builtInDevices.foreach { builtIn => 
    builtIn.errorOpt.foreach { error => 
      LogicalModuleTree.add(logicalTreeNode, error.logicalTreeNode)
     }
    builtIn.zeroOpt.foreach { zero => 
      LogicalModuleTree.add(logicalTreeNode, zero.logicalTreeNode)
    }
```

src/main/scala/subsystem/FrontBus.scala
```diff
@@ -20,5 +20,5 @@ class FrontBus(params: FrontBusParams)(implicit p: Parameters)
     with CanHaveBuiltInDevices
     with CanAttachTLMasters
     with HasTLXbarPhy {
-  attachBuiltInDevices(params)
+  val builtInDevices = attachBuiltInDevices(params)
```
mwachs5, line 23: is there a reason this `val` can't go into the trait `CanHaveBuiltInDevices`? Maybe as a `def` or `lazy val`?

albertchen-sifive: this just seemed like the least intrusive way of adding it. the val can go in `CanHaveBuiltInDevices`, but then it would need to have access to a `params: HasBuiltInDeviceParams`. and I wasn't sure if `attachBuiltInDevices` was actually being called everywhere that `CanHaveBuiltInDevices` was being mixed in.

but it does look like the attach method is unconditionally called everywhere that the trait is mixed in. and the devices are already optionally attached based on the params. so yeah, it makes sense to me to have the `val` in `CanHaveBuiltInDevices` and just inline the `attachBuiltInDevices` method.

albertchen-sifive: but that seems like a cake-pattern and we're trying to avoid that?

richardxia: > but that seems like a cake-pattern and we're trying to avoid that?

We should all be clear on _why_ we are (or are not?) avoiding the cake pattern in future code, since it's rarely the case that we would conclude "X is always bad and we should never do X again". Not having a shared understanding of what the root problems actually are could lead to bandwagoning and scapegoating where we dump a solution without knowing why and end up building a new solution that runs into similar problems. My impression of why the way we have used the cake pattern has been problematic is because:

- We end up creating these [god classes](https://en.wikipedia.org/wiki/God_object) which have to know about every possible trait that we may want to use, causing every addition of a new trait or a change to a trait to have to propagate up to the god class. The reason we have the "CanHaveX" pattern is because we've turned the "might have or might not have feature X" into an optional thing everywhere.
- Parameterizing the concrete instances of the cake interfaces becomes challenging due to statement ordering issues and restrictions on constructors. Coupled with the fact that the Chisel DSL essentially requires people to write HDL code directly in the "constructor" of a Scala class, this makes it difficult to provide a consistent interface for concrete implementations of the cake interface.

See https://kubuszok.com/2018/cake-antipattern/ for a detailed blog post on the Scala cake pattern and why the author at least considers it to be an antipattern. See specifically "[Adding dependencies](https://kubuszok.com/2018/cake-antipattern/)", which provides a concrete example of a god class, and "[Initialization order](https://kubuszok.com/2018/cake-antipattern/)", which describes the ordering and initialization issues.

With these points in mind, my personal response would be that to best avoid the two problems above without also making a huge architectural change to the existing code would be to decouple the _interface_ from the _implementation_. We should not let the amount of typing required dictate what the interface should be. If we think that if a module says that it "can have built-in devices" should imply that we can retrieve said built-in devices under a uniform interface, then I think it is a good idea to define a `def`. However, if we think that different modules may want to expose them in different ways, then we should hold off from defining an interface.

mwachs5: @richardxia I'm not sure what you are concretely suggesting, or what you need to know to make a concrete suggestion

richardxia: I need to know what a `CanHaveBuiltInDevices` conceptually represents, and I need to know how we intend for concrete implementations to use it as well as how consumers of the concrete implementations are meant to use it. Otherwise, I don't think it's possible to have a conversation about whether we should pull the `val builtInDevices = attachBuiltInDevices(params)` into the trait definition.

Let me flip this onto you: Is there a reason this val _should_ go into the trait CanHaveBuiltInDevices?

albertchen-sifive: I don't actually know what a `BuiltInDevice` is or the naming-conventions of these traits, but here's what I think:

I would think that `CanHaveBuiltInDevices` would have some field `builtInDevices: BuiltInDevices` so that consumers of the interface can know about and get some reference to the attached `Option[TLError/TLZero]`. right now it just defines a method `attachBuiltInDevices(params)` that can optionally be called to optionally attach devices based on `params`. If `CanHaveBuiltInDevices` doesn't define some interface for getting at the attached built-in devices, I don't quite understand the purpose of that trait. There's no reason `attachBuiltInDevices` needs to be implemented as a method inside a trait with self-type vs a regular function with a few extra arguments. and I don't think the `attachBuiltInDevices` method is helpful to consumers of the interface because if you call it from outside you risk instantiating and attaching devices multiple times, plus you'll also need a reference to the original `LazyScope` if you want the attach to actually happen in the same module that mixed in `CanHaveBuiltInDevices`.

I propose having an abstract `def builtInDevices: BuiltInDevices` in `CanHaveBuiltInDevices`, getting rid of the `TLBusWrapper` self-type, and adding a  `attachBuiltInDevices(params: HasBuiltInParams with HasTLBusParams, outwardNode: TLOutwardNode)` to the `BuiltInDevices` companion object. This way we can avoid adding another `lazy val` to the cake and still provide a useful interface for consumers.

richardxia: That sounds like a reasonable proposal to me, and I agree with the `trait` otherwise being useless since the `attach` method could just be in a standalone function and also that it is dangerous to call it twice.

albertchen-sifive:
I just realized that https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/subsystem/MemoryBus.scala#L60 is a `def` rather than a `val` and now it only get's called once since `outwardNode` is an argument whereas before `outwardNode` was a method that was called twice within `attachBuiltInDevices(params)`. so I think I'm going to change that argument to `=> TLOutwardNode`. or should I just change it to `busWrapper: TLBusWrapper` and call `busWrapper.outwardNode`?

hcook:
@albertchen-sifive I'm not really sure why they are `def`s anymore; I think it is a historical artifact and not a requirement. In my bus refactoring PR I actually changed the code like so, with no apparent ill-effect. Perhaps you could adopt this code in order to reduce my eventual merge conflict:
```
  private def replicate(node: TLInwardNode): TLInwardNode =
    if (params.replicatorMask == 0) node else { node :*=* RegionReplicator(params.replicatorMask) }
  val inwardNode: TLInwardNode = replicate(xbar.node)
  val outwardNode: TLOutwardNode = ProbePicker() :*= xbar.node
```

albertchen-sifive:
> Perhaps you could adopt this code in order to reduce my eventual merge conflict:

I think it might be better if I leave it as is? I haven't changed those lines so they shouldn't conflict with your change, and everything should still work with or without that change. and there's a good chance that I'll copy past that snippet incorrectly and cause a conflict.

PR #2350, opened 2020-03-20
wake: Add DTSJSON to RocketChipGeneratorOutputs.

richardxia:
**Related issue**: https://github.com/sifive/api-generator-sifive/pull/49

**Type of change**: feature request

**Impact**: API addition (no impact on existing code)

**Development Phase**: implementation

**Release Notes**
In the Wake build rules, the `RocketChipGeneratorOutputs` now has a new property which holds the path to the generated DTS JSON file, which a JSON encoding of the same information found in the DTS file.

PR #2413, opened 2020-04-13
use stages api to build rocket-chip in wake flow

kritikbhimani:
**Related issue**: 

**Release Notes**

### DIFF
--- a/build.wake
+++ b/build.wake
@@ -106,7 +106,6 @@ global def runRocketChipGenerator options =
   def targetDir = options.getRocketChipGeneratorOptionsTargetDir
 
   def cmdline =
-    def rootPackage = "_root_"
     def main = "freechips.rocketchip.system.Generator"
     def configs = catWith "," options.getRocketChipGeneratorOptionsConfigNames
     def topModule = options.getRocketChipGeneratorOptionsTopModuleName
@@ -117,10 +116,9 @@ global def runRocketChipGenerator options =
       Some name = name, Nil
       None = Nil
     which "java", "-cp", classpath, main,
-    relTargetDir,
-    rootPackage, topModule,
-    rootPackage, configs,
-    baseFileName
+    "-td", relTargetDir,
+    "-T", topModule,
+    "-C", configs, Nil
 
   def inputs =
     def bootrom = source 'rocket-chip/bootrom/bootrom.img'
@@ -139,7 +137,10 @@ global def runRocketChipGenerator options =
     head, tail = Some head
 
   def allOutputs = generatorJob.getJobOutputs
-  def annoFile    = getFile `.*\.anno\.json`
+  def topModPackage =
+    tokenize `\.` (options.getRocketChipGeneratorOptionsTopModuleName) | reverse | tail | reverse
+    | catWith "." | quote
+  def annoFile    = getFile (regExpCat (`.*`, topModPackage, `.*\.anno\.json`, Nil))
   def firrtlFile  = getFile `.*\.fir`
   def romConfFile = getFile `.*\.rom\.conf`
   def dtsFile     = getFile `.*\.dts`
--- a/build.wake
+++ b/build.wake
@@ -139,12 +139,12 @@ global def runRocketChipGenerator options =
 
   def allOutputs = generatorJob.getJobOutputs
   def baseFileName = match options.getRocketChipGeneratorOptionsBaseFileName
-    Some fileName = fileName
+    Some fileName = fileName.quote, `\.anno\.json`, Nil
     None =
       def topModuleName = options.getRocketChipGeneratorOptionsTopModuleName
       def packageName = topModuleName | tokenize `\.` | reverse | tail | reverse | catWith "."
-      packageName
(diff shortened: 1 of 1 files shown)


build.wake
```diff
@@ -117,10 +116,9 @@ global def runRocketChipGenerator options =
       Some name = name, Nil
       None = Nil
     which "java", "-cp", classpath, main,
-    relTargetDir,
-    rootPackage, topModule,
-    rootPackage, configs,
-    baseFileName
+    "-td", relTargetDir,
+    "-T", topModule,
+    "-C", configs, Nil
```
debs-sifive, line 121: Can you make sure these are the full paths to the top module/configs?

richardxia, line 121: Unless if I'm missing something, I don't think this is a correct translation between the old CLI and the new CLI. This looks like it is throwing out the `rootPackage` argument, which I assume was the package containing the `topModule`, but we're not passing it to the command line interface anymore. Same goes for the configs.

Also, it looks like the `baseFileName` has been dropped altogether, which makes determining the actual base file name more difficult, as you can see in the `def topModPackage` statement below. See https://github.com/chipsalliance/rocket-chip/pull/2350 for how I had attempted to do this prior to the stages API being merged in, but I think it results in an easier way to handle the base name, since we set it on the Wake side rather than trying to derive it from what we think the rocket-chip Generator is trying to do.

I'd suggest that we make the following changes:

1. Construct the full package paths for the top module and config before passing them in to the rocket-chip generator CLI.
2. Continue passing in the `baseFileName` argument
3. Pull in pieces of my changes in https://github.com/chipsalliance/rocket-chip/pull/2350 to compute the derived `outputBaseName`. Feel free to ignore the DTSJSON thing for now.
4. (Nit:) Using the fully-spelled out forms of the command line flags, since they make it easier for someone who is not familiar with the CLI to read. `--top-module` is much more understandable than `-T`, IMO.

kritikbhimani: I think the current layout kind of makes it self-explanatory i.e. td -> target dir, t -> topmodule, c -> configs and looking at code --name for basefilename

kritikbhimani: they are full paths

kritikbhimani: I tested these changes with a bumped api-scala-sifive(https://github.com/sifive/api-scala-sifive/tree/v0.1.1) where it seemed like it didn't require _root_

build.wake
```diff
@@ -139,7 +137,14 @@ global def runRocketChipGenerator options =
...
   def allOutputs = generatorJob.getJobOutputs
-  def annoFile    = getFile `.*\.anno\.json`
+  def topModPackage =
+    tokenize `\.` (options.getRocketChipGeneratorOptionsTopModuleName) | reverse | tail | reverse
+    | catWith "."
+  def annoRegexStr =
+    if (isSome (options.getRocketChipGeneratorOptionsBaseFileName)) then
+      options | getRocketChipGeneratorOptionsBaseFileName | getOrElse "Invalid"
+    else topModPackage
+  def annoFile    = getFile (regExpCat (`.*`, (annoRegexStr | quote), `.*\.anno\.json`, Nil))
```
richardxia, line 147: I think this would be clearer written as:

```suggestion
  def baseFileName = match options.getRocketChipGeneratorOptionsBaseFileName
    Some fileName = fileName
    None =
      def topModuleName = options.getRocketChipGeneratorOptionsTopModuleName
      def packageName = topModuleName | tokenize `\.` | reverse | tail | reverse | catWith "."
      packageName
  def annoFile    = getFile (regExpCat (`.*`, baseFileName.quote, `.*\.anno\.json`, Nil))
```

kritikbhimani:
i don't have merge powers

RajeshVaradharajan:
@mwachs5. Can you merge this PR.

richardxia:
I'm not sure if I agree that the change in https://github.com/chipsalliance/rocket-chip/pull/2413/commits/0b67d3265d2f43eb2946b1b0d64e322f7d2c0ac4 is correct. Shouldn't we remove the `.*` in both cases, not just in the case where the base file name is set? We should have the Wake code mirror the Scala code.

This made me take a closer look at the Scala code, and it looks like when the base file name is not specified, it defaults to a value that combines the package name of the top module with the class names of the config classes passed in, joined by an underscore: https://github.com/chipsalliance/rocket-chip/blame/d76703884d9dbd2965536a28a8499928076b89d5/src/main/scala/stage/RocketChipOptions.scala#L38

For example, if you run:
```
java -cp rocket-chip.jar freechips.rocketchip.system.Generator \
  --top-module freechips.rocketchip.system.TestHarness \
  --configs richard.xia.RichardConfig,freechips.rocketchip.system.WithJtagDTMSystem
```

Then it creates a file named `freechips.rocketchip.system.RichardConfig_WithJtagDTMSystem.anno.json`.

Let me submit a PR against this branch with the changes that I think more closely matches the behavior of the actual rocket-chip CLI.

richardxia:
Thanks for submitting this change, and sorry that it took a bit of time to iterate on it!

PR #2414, opened 2020-04-14
coreMonitorBundle print trace log regs only if read or written

johningalls-sifive:
**Related issue**: fuller implementation of https://github.com/chipsalliance/rocket-chip/pull/2392
**Impact**: API modification
**Release Notes**: “don't print bogus register reads or writes that confuse users and muddy diff-ing logs.”

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -877,7 +877,7 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
   coreMonitorBundle.pc := csr.io.trace(0).iaddr(vaddrBitsExtended-1, 0).sextTo(xLen)
   coreMonitorBundle.wren := wb_wen && !wb_set_sboard
   coreMonitorBundle.wrdst := wb_waddr
-  coreMonitorBundle.wrdata := Mux(coreMonitorBundle.wren, rf_wdata, 0.U)
+  coreMonitorBundle.wrdata := rf_wdata
   coreMonitorBundle.rd0src := wb_reg_inst(19,15)
   coreMonitorBundle.rd0val := Reg(next=Reg(next=ex_rs(0)))
   coreMonitorBundle.rd1src := wb_reg_inst(24,20)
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/RocketCore.scala
```diff
@@ -876,12 +876,12 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
   coreMonitorBundle.valid := csr.io.trace(0).valid && !csr.io.trace(0).exception
   coreMonitorBundle.pc := csr.io.trace(0).iaddr(vaddrBitsExtended-1, 0).sextTo(xLen)
   coreMonitorBundle.wren := wb_wen && !wb_set_sboard
-  coreMonitorBundle.wrdst := wb_waddr
+  coreMonitorBundle.wrdst := Mux(wb_ctrl.wxd || wb_ctrl.wfd, wb_waddr, 0.U)
```
aswaterman, line 879: I still don't agree with this change.  0 is ineffective as a sentinel value because it's a valid FP register.  Regardless.  we have the `wren` signal above.  This signal is a don't-care when `wren` is false.

src/main/scala/rocket/RocketCore.scala
```diff
@@ -876,12 +876,12 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
...
   coreMonitorBundle.wrdata := Mux(coreMonitorBundle.wren, rf_wdata, 0.U)
-  coreMonitorBundle.rd0src := wb_reg_inst(19,15)
-  coreMonitorBundle.rd0val := Reg(next=Reg(next=ex_rs(0)))
-  coreMonitorBundle.rd1src := wb_reg_inst(24,20)
-  coreMonitorBundle.rd1val := Reg(next=Reg(next=ex_rs(1)))
+  coreMonitorBundle.rd0src := Mux(wb_ctrl.rxs1 || wb_ctrl.rfs1, wb_reg_inst(19,15), 0.U)
+  coreMonitorBundle.rd0val := Mux(wb_ctrl.rxs1 || wb_ctrl.rfs1, Reg(next=Reg(next=ex_rs(0))), 0.U)
+  coreMonitorBundle.rd1src := Mux(wb_ctrl.rxs2 || wb_ctrl.rfs2, wb_reg_inst(24,20), 0.U)
+  coreMonitorBundle.rd1val := Mux(wb_ctrl.rxs2 || wb_ctrl.rfs2, Reg(next=Reg(next=ex_rs(1))), 0.U)
```
aswaterman, line 884: Same comment applies here.  Reading FP register is meaningful, so forcing values to 0 when there's no read gives a false sense that 0 means no read.

src/main/scala/rocket/RocketCore.scala
johningalls-sifive: Agree that this change can't and doesn't change function.
Would it be better if I changed the log printout values, instead of here in the coreMonitorBundle?
If we change the printout, then we can use other values than 0, like spaces or hyphens.
Or should we just leave it be and not complicate too much?
Just trying to make the simulation log prettier to read, not changing function.

aswaterman: I think making the printout easier to read is a worthy goal, but I do think those improvements are best isolated to the printing code, so that it’s clear we aren’t trying to add extra semantics or constraints to the monitoring logic.

You might want to shop around the various options for improving the printout. Moving these muxes over there would help right now and isn’t controversial, but other changes that improve the visuals could upset code that parses the printout.

PR #2416, opened 2020-04-14
Add Finder to diplomacy

sequencer:

**Impact**: API addition (no impact on existing code) | API modification

**Release Notes**
Since some `LazyModule` are instantiated by `Parameter`, which are not accessible by outer `LazyModule`, we cannot annotate these modules.
This PR implemented a `LazyModule` and `BaseNode` finder to find a specific element in the diplomatic modules.
When found required module, we can use `asInstanceOf` to cast it to the type it should be and annotate the contents.

### DIFF
--- a/src/main/scala/diplomacy/LazyModule.scala
+++ b/src/main/scala/diplomacy/LazyModule.scala
@@ -118,9 +118,25 @@ abstract class LazyModule()(implicit val p: Parameters)
     children.foreach( _.childrenIterator(iterfunc) )
   }
 
+  def childrenFinder(filter: LazyModule => Boolean): LazyModule = {
+    val found = scala.collection.mutable.ListBuffer[LazyModule]()
+    childrenIterator(lm => if (filter(lm)) found += lm)
+    require(found.nonEmpty, s"no LazyModule found")
+    require(found.size == 1, s"constraint of filter is too loose, found ${found.size} LazyModules")
+    found.head
+  }
+
   def nodeIterator(iterfunc: (BaseNode) => Unit): Unit = {
     nodes.foreach(iterfunc)
     childrenIterator(_.nodes.foreach(iterfunc))
+  }
+
+  def nodeFinder(filter: BaseNode => Boolean): BaseNode = {
+    val found = scala.collection.mutable.ListBuffer[BaseNode]()
+    nodeIterator(node => if (filter(node)) found += node)
+    require(found.nonEmpty, s"no BaseNode found")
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -118,11 +118,27 @@ abstract class LazyModule()(implicit val p: Parameters)
     children.foreach( _.childrenIterator(iterfunc) )
   }
 
+  def childrenFinder(filter: LazyModule => Boolean): LazyModule = {
```
mwachs5, line 121: should this be called "findChild"?

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -118,11 +118,27 @@ abstract class LazyModule()(implicit val p: Parameters)
...
+    require(found.size == 1, s"constraint of filter is too loose, found ${found.size} LazyModules")
+    found.head
+  }
+
   def nodeIterator(iterfunc: (BaseNode) => Unit): Unit = {
     nodes.foreach(iterfunc)
     childrenIterator(_.nodes.foreach(iterfunc))
   }
 
+  def nodeFinder(filter: BaseNode => Boolean): BaseNode = {
```
mwachs5, line 134: should this be called "findNode"?

hcook:
Maybe if you also build the `asInstanceOf` cast into the functions by making it `def childrenFinder[T <: LazyModule](filter: LazyModule => Boolean): T` you are getting closer to having a specific piece of functionality that is worth having its own method name, but there are also built in collection functions `find` and `collectFirst`... why not just use one of those?
```
val childOption: Option[Subclass] = lazyModuleInstance.childIterator.collectFirst { case c if filter(c) => c.asInstanceOf[Subclass] }
```

src/main/scala/diplomacy/LazyModule.scala
hcook: i think it would have to be `findExactlyOneNode`

src/main/scala/diplomacy/LazyModule.scala
hcook: i think it would have to be findExactlyOneChild

sequencer:
Yes I moved these code into my own repository. thanks for reviewing.

sequencer:
> And if you really want to make sure there aren't multiple matches then i suggest using collect, checking the resulting size is one, and then using head. Still should only be two lines of user code.

Yes, I should do so! Thank you for you advice :)

PR #2424, opened 2020-04-19
A Pair of Phase-Ordering Fixes

davidbiancolin:
While bumping Rocket Chip in Chipyard i ran into a couple of order issues with the initial Stage / Phase implementation. There many ways they can be fixed, but this is what i did: 

1) The dependency manager can schedule `TransformAnnotations` behind chisel's `AddImplicitOutputFile` phase, which leads the down stream phases to use the instance of the annotation provided by `AddImplicitOutputFile` and not `TransformAnnotations`. I added a dependency between the two phases to resolve this. 

2) The dependency manager can schedule `AddDefaultTestSuites` ahead of `GenerateFirrtlAnnos`, which leads to `RocketTestSuiteAnnotation`s being serialized when they cannot be. While there are myriad ways to fix this i've simply marked them as unserializable. 

**Release Notes**
- [stage] Fix a bug where unserializable RocketTestSuiteAnnotations were being serialized
- [stage] Fix a bug where the desired output file name was being superseded  by another phase

### DIFF
--- a/src/main/scala/stage/phases/AddDefaultTests.scala
+++ b/src/main/scala/stage/phases/AddDefaultTests.scala
@@ -7,7 +7,7 @@ import chipsalliance.rocketchip.config.Parameters
 import chisel3.stage.phases.Elaborate
 import firrtl.AnnotationSeq
 import firrtl.annotations.NoTargetAnnotation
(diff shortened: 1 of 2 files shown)


davidbiancolin:
No problem. :)

PR #2427, opened 2020-04-21
Object model updates for AHB and AXI

mhtwn:
**Related issue**: 

**Release Notes**

An example representation of a port with AXI protocol:
```
      "signalNamePrefix" : "axi4_mem_port",
      "width" : 32,
      "protocol" : {
        "specification" : {
          "name" : "AXI Protocol",
          "version" : "4",
          "_types" : [ "OMSpecification" ]
        },
        "_types" : [ "AXI", "AMBA", "OMProtocol" ]
      },
```

An example representation of a port with AHB-Lite protocol:

```
      "signalNamePrefix" : "ahb_sys_port",
      "width" : 32,
      "protocol" : {
        "specification" : {
          "name" : "AHB Lite Protocol",
          "version" : "3",
          "_types" : [ "OMSpecification" ]
        },
        "_types" : [ "AHB", "AMBA", "OMProtocol" ]
      },
```

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
@@ -51,7 +51,7 @@ case class AXI4_Lite(
 
 case class AHB_Lite(
   specification: Option[OMSpecification],
-  val _types: Seq[String] = Seq("AHB", "AMBA",  "OMProtocol")
+  val _types: Seq[String] = Seq("AHB_Lite", "AMBA",  "OMProtocol")
 ) extends AMBA
 
 case class AHB(
--- a/src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
@@ -41,12 +41,12 @@ trait AMBA extends OMProtocol
 
 case class AXI4(
   specification: Option[OMSpecification],
-  _types: Seq[String] = Seq("AXI", "AMBA",  "OMProtocol")
+  _types: Seq[String] = Seq("AXI4", "AMBA",  "OMProtocol")
 ) extends AMBA
 
 case class AXI4_Lite(
   specification: Option[OMSpecification],
-  val _types: Seq[String] = Seq("AXI", "AMBA",  "OMProtocol")
+  val _types: Seq[String] = Seq("AXI4_Lite", "AMBA",  "OMProtocol")
 ) extends AMBA
 
 case class AHB_Lite(
--- a/src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
@@ -41,17 +41,17 @@ trait AMBA extends OMProtocol
 
 case class AXI4(
   specification: Option[OMSpecification],
-  _types: Seq[String] = Seq("AXI4", "AMBA",  "OMProtocol")
+  _types: Seq[String] = Seq("AXI", "AMBA",  "OMProtocol")
 ) extends AMBA
 
 case class AXI4_Lite(
   specification: Option[OMSpecification],
-  val _types: Seq[String] = Seq("AXI4_Lite", "AMBA",  "OMProtocol")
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
```diff
@@ -41,17 +41,17 @@ trait AMBA extends OMProtocol
 
 case class AXI4(
   specification: Option[OMSpecification],
-  _types: Seq[String] = Seq("AXI4", "AMBA",  "OMProtocol")
+  _types: Seq[String] = Seq("AXI", "AMBA",  "OMProtocol")
```
richardxia, line 44: Are you sure you want to make this change? I'm certain this is going to break a lot of things that are expecting to match based on that `4` being present in the name.

mwachs5: agree with @richardxia

mhtwn: fair enough - reverting back

mwachs5:
I'd like @aisha-w-sifive and/or @jmilstead-sifive  feedback on this...

src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
```diff
@@ -125,15 +125,15 @@ case class SystemPort(
...
+    APBProtocol -> "APB Protocol",
     TLProtocol -> "TileLink specification"
   )
 
   val protocolSpecificationVersions = Map[ProtocolType, String](
-    AHBProtocol -> "1.0",
-    AXI4Protocol -> "1.0",
+    AHBProtocol -> "3",
+    AXI4Protocol -> "4",
     APBProtocol -> "1.0",
```
mwachs5, line 137: why is APB still 1.0 when before it was "AMBA 3" ? Is this right @terpstra ? Do the AMBA spec version stay alighed with the specific bus interfaces...?

aisha-w-sifive:
So it is "OMAXI4", "OMAXI4_Lite", "OMAHB", "OMAHB_Lite"? If so, LGTM, should run fine in docs.

jmilstead-sifive:
Maybe a naïve question, but why include both `AXI4` and `AMBA` if `AXI4` extends `AMBA`? These are all `AMBA`.

src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
jmilstead-sifive: According to [this wiki page](https://en.wikipedia.org/wiki/Advanced_Microcontroller_Bus_Architecture), they track more or less, but not perfectly.

src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
```diff
@@ -124,21 +124,26 @@ case class SystemPort(
...
-    APBProtocol -> "AMBA 3 APB Protocol",
-    TLProtocol -> "TileLink specification"
-  )
+  val protocolSpecifications: (ProtocolType, SubProtocolType) => String = {
+   case (AHBProtocol, AHBLiteSubProtocol) => "AHB Lite Protocol"
+   case (AHBProtocol, AHBFullSubProtocol) => "AHB Full Protocol"
+   case (AXI4Protocol, AXI4SubProtocol) => "AXI Protocol"
+   case (AXI4Protocol, AXI4LiteSubProtocol) => "AXI Lite Protocol"
+   case (APBProtocol, APBSubProtocol) => "APB Protocol"
+   case (TLProtocol, TL_UHSubProtocol) => "TileLink specification"
```
mwachs5, line 133: why is this called "specification" when the others are called "Protocol"?

mhtwn: Good point - I haven't the description for TL, but I can update it - do you think mentioning UL, UH and CS will also be a good idea?

mwachs5: for that I'd defer to @hcook or @terpstra

PR #2432, opened 2020-04-24
RocketTile: do not distort placement when hartid is not constant

terpstra:
When hartid is not a constant, driving it to the I$ D$ and CSR file directly from a Tile IO distorts placement of the core. These registers alleviate that.

However, when the hartid IS a constant, these registers are unneeded. I believe chisel will eliminate the registers via constant propagation.

**Release Notes**
Pipeline tile hartid input.


johningalls-sifive:
mark as multi-cycle path instead of adding registers?

aswaterman:
Need some Chisel support for that, and a bunch of other infrastructural plumbing for different tools.  Until that exists, the tiny number of flops is fine.

PR #2434, opened 2020-04-24
Object Model: add Sv48 to RocketISA addressTranslationModes

johningalls-sifive:
**Related issue**: 

**Release Notes**
enable Sv48 setting PgLevels=4

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/model/OMISA.scala
+++ b/src/main/scala/diplomaticobjectmodel/model/OMISA.scala
@@ -56,7 +56,7 @@ case class OMVectorExtension(
 )
 
 object OMISA {
-  def rocketISA(tile: RocketTile, xLen: Int): OMISA = {
+  def rocketISA(tile: RocketTile, xLen: Int, pgLevels: Int): OMISA = {
     val coreParams = tile.rocketParams.core
 
     val baseInstructionSet = xLen match {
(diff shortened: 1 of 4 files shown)


src/main/scala/diplomaticobjectmodel/model/OMISA.scala
```diff
@@ -82,10 +82,17 @@ object OMISA {
     }
 
     val addressTranslationModes = xLen match {
```
aswaterman, line 84: This is just stylistic, but you could do `(xLen, pgLevels) match {` to flatten this structure and only have one error message ("invalid XLEN/PgLevels combination")

src/main/scala/rocket/PTW.scala
```diff
@@ -153,7 +153,7 @@ class PTW(n: Int)(implicit edge: TLEdgeOut, p: Parameters) extends CoreModule()(
     Cat(r_pte.ppn, vpn_idx) << log2Ceil(xLen/8)
   }
   val fragmented_superpage_ppn = {
-    val choices = (pgLevels-1 until 0 by -1).map(i => Cat(r_pte.ppn >> (pgLevelBits*i), r_req.addr(pgLevelBits*i-1, 0)))
+    val choices = (pgLevels-1 until 0 by -1).map(i => Cat(r_pte.ppn >> (pgLevelBits*i), r_req.addr(((pgLevelBits*i) min vpnBits)-1, 0).padTo(pgLevelBits*i)))
```
aswaterman, line 156: How can (pgLevels - 1) * pgLevelBits ever exceed vpnBits?

johningalls-sifive: in a heterogeneous system where some cores have vmEnabled=true and some cores have vmEnabled=false, the vmEnabled=false cores still compile PTW.scala with a small number of vpnBits, leading to a Chisel compile error here before this all gets optimized away in Firrtl

aswaterman: Right.  I guess I'm wondering why this wasn't a problem before... maybe all the systems had just enough PA bits so that by coincidence it wasn't a problem.

PR #2439, opened 2020-04-27
Add definitions for ratified version of RISC-V Processor Trace spec

ernie-sifive:
**Related issue**: 

**Release Notes**

[The RISC-V Processor Trace Specification](https://github.com/riscv/riscv-trace-spec/blob/master/riscv-trace-spec.pdf) has recently been ratified.  The definition of the core-to-trace encoder interface has been changed from the legacy definition currently implemented in rocket-chip.  This PR adds a second BundleBridge trace interface that conforms to the new version of the spec that will allow processors that conform to that spec to connect their trace ports to encoders.

The TraceAux bundle is for sideband signals that are useful but not in the spec.  Previously, there has been a `stall` request signal and this PR adds a new `enable` signal to TraceAux so a trace port client can enable the trace logic in a core only when it is being used (as a power saving).  In case there is more than one trace client, the `stall` and `enable` signals are a logical OR of these signals from all the clients.

Finally, the definition of the bpwatch bundle has been parameterized to allow an implementation to replace the default definition, if necessary, by overriding the `getBpwatchParams` function in BaseTile.

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -172,16 +172,26 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
   protected val tlSlaveXbar = LazyModule(new TLXbar)
   protected val intXbar = LazyModule(new IntXbar)
 
+  // Node for legacy instruction trace from core
   val traceSourceNode = BundleBridgeSource(() => Vec(tileParams.core.retireWidth, new TracedInstruction()))
   val traceNode = BundleBroadcast[Vec[TracedInstruction]](Some("trace"))
   traceNode := traceSourceNode
 
+  // Trace sideband signals into core
   val traceAuxNode = BundleBridgeNexus[TraceAux]()
   val traceAuxSinkNode = BundleBridgeSink[TraceAux]()
   val traceAuxDefaultNode = BundleBridgeSource(() => new TraceAux)
   traceAuxSinkNode := traceAuxNode := traceAuxDefaultNode
 
-  val bpwatchSourceNode = BundleBridgeSource(() => Vec(tileParams.core.nBreakpoints, new BPWatch(tileParams.core.retireWidth)))
+  // Node for instruction trace conforming to RISC-V Processor Trace spec V1.0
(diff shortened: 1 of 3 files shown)


src/main/scala/tile/BaseTile.scala
```diff
@@ -172,16 +172,26 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
...
+  // Node for legacy instruction trace from core
   val traceSourceNode = BundleBridgeSource(() => Vec(tileParams.core.retireWidth, new TracedInstruction()))
   val traceNode = BundleBroadcast[Vec[TracedInstruction]](Some("trace"))
   traceNode := traceSourceNode
 
+  // Trace sideband signals into core
   val traceAuxNode = BundleBridgeNexus[TraceAux]()
   val traceAuxSinkNode = BundleBridgeSink[TraceAux]()
   val traceAuxDefaultNode = BundleBridgeSource(() => new TraceAux)
   traceAuxSinkNode := traceAuxNode := traceAuxDefaultNode
```
rmac-sifive, line 184: I know this is older code, but why do we make a `BundleBridgeSink` and `BundleBridgeSource` in the same level of hierarchy? Why not just have a single `BundleBroadcast` and get rid of the `BBSink`/``BBSource` nodes?

ernie-sifive: Let's start by saying I don't have a thorough understanding of BundleBridge.  The thinking at the time this was written was that both the source and sink of TraceAux were optional.  There can be zero or one sinks (cores) and zero or more sources (trace clients).  BundleBroadcast requires exactly one source so does not work in this situation.

The RocketTile level connects I/O signals from the core directly to the traceSourceNode and may or may not connect I/O signals to traceAuxSinkNode.
- If there is a TraceAux sink but no source, such as when there is no trace client in the system, the default source ends up driving the sink.
- If there is a trace client source but the core doesn't connect to the sink, that's also ok.  
- If there is a sink and one or more sources, then the connection is made using the TraceAux logic in BaseTileModuleImp (the logical OR described in the submission).

PR #2446, opened 2020-05-02
Make the TLBroadcast and TLSourceShinker adapters pass through user bits

hcook:
This change makes the `TLBroadcast` and `TLSourceShrinker` transparently pass through any user bits that are received from clients attached to them.

### DIFF
--- a/src/main/scala/tilelink/Broadcast.scala
+++ b/src/main/scala/tilelink/Broadcast.scala
@@ -16,9 +16,13 @@ class TLBroadcast(lineBytes: Int, numTrackers: Int = 4, bufferless: Boolean = fa
 
   val node = TLAdapterNode(
     clientFn  = { cp =>
-      cp.v1copy(clients = Seq(TLMasterParameters.v1(
-        name     = "TLBroadcast",
-        sourceId = IdRange(0, 1 << log2Ceil(cp.endSourceId*4)))))
+      cp.v1copy(
+        clients = Seq(TLMasterParameters.v1(
+          name     = "TLBroadcast",
+          sourceId = IdRange(0, 1 << log2Ceil(cp.endSourceId*4)))),
+        echoFields    = cp.echoFields,
+        requestFields = cp.requestFields,
+        responseKeys  = cp.responseKeys)
     },
     managerFn = { mp =>
       mp.v1copy(
@@ -252,6 +256,8 @@ class TLBroadcastTracker(id: Int, lineBytes: Int, probeCountBits: Int, bufferles
   val param   = Reg(io.in_a.bits.param)
   val size    = Reg(io.in_a.bits.size)
   val source  = Reg(io.in_a.bits.source)
+  val user    = Reg(io.in_a.bits.user)
+  val echo    = Reg(io.in_a.bits.echo)
   val address = RegInit(UInt(id << lineShift, width = io.in_a.bits.address.getWidth))
   val count   = Reg(UInt(width = probeCountBits))
   val idle    = got_e && sent_d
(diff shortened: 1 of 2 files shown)


src/main/scala/tilelink/Broadcast.scala
```diff
@@ -314,6 +317,15 @@ class TLBroadcastTracker(id: Int, lineBytes: Int, probeCountBits: Int, bufferles
   io.out_a.bits.mask    := o_data.bits.mask
   io.out_a.bits.data    := o_data.bits.data
   io.out_a.bits.corrupt := Bool(false)
+  io.out_a.bits.user.lift(AMBAProt).foreach { x =>
```
johningalls-sifive, line 320: could these be passed through from `io.in_a`?

hcook: Maybe they could could, but that is not what was done in the closed source example I was working of off, for reasons I am personally unaware of. Perhaps @terpstra @jsmithsf or @daveparry can comment about the particular selection of bits and the decision to ignore any settings supplied by inward request initiators?

terpstra:
You've made this too complicated. You should not modify the parameters at all. All you needed to add was a 'user :<= in.user' and an 'out.user :<= in.user' ... and the same for echo.

terpstra:
(I am assuming we want to forward the request permissions from the client onwards)

terpstra:
In a real cache, it does not make sense to pass through the request since requests are merged. But here they are not.

terpstra:
OTOH, maybe we want this? 🤷

hcook:
@terpstra I think it makes sense to make both of these as transparent as possible for now. Another adapter to override or provide additional information could be inserted by user code.

This produced the verilog signals I was expecting in my example, but can you re-review?

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -16,9 +16,13 @@ class TLBroadcast(lineBytes: Int, numTrackers: Int = 4, bufferless: Boolean = fa
...
-      cp.v1copy(clients = Seq(TLMasterParameters.v1(
-        name     = "TLBroadcast",
-        sourceId = IdRange(0, 1 << log2Ceil(cp.endSourceId*4)))))
+      cp.v1copy(
+        clients = Seq(TLMasterParameters.v1(
+          name     = "TLBroadcast",
+          sourceId = IdRange(0, 1 << log2Ceil(cp.endSourceId*4)))),
+        echoFields    = cp.echoFields,
+        requestFields = cp.requestFields,
+        responseKeys  = cp.responseKeys)
```
terpstra, line 25: I mean, it does not hurt to list these three lines, but ... they are copied by v1copy already anyway. So they are just dead code.

src/main/scala/tilelink/SourceShrinker.scala
```diff
@@ -18,7 +18,12 @@ class TLSourceShrinker(maxInFlight: Int)(implicit p: Parameters) extends LazyMod
     sourceId = IdRange(0, maxInFlight))
   val node = TLAdapterNode(
     // We erase all client information since we crush the source Ids
-    clientFn  = { cp => TLMasterPortParameters.v1(clients = Seq(client.v1copy(requestFifo = cp.clients.exists(_.requestFifo)))) },
+    clientFn  = { cp => TLMasterPortParameters.v1(
+      clients = Seq(client.v1copy(requestFifo = cp.clients.exists(_.requestFifo))),
+      echoFields = cp.echoFields,
+      requestFields = cp.requestFields,
+      responseKeys = cp.responseKeys
+    )},
```
terpstra, line 26: In contrast, you need it here b/c not copy.

src/main/scala/tilelink/Broadcast.scala
terpstra: You could probably have left the entire node unmodified.

terpstra:
👍 Minimal diff approved.

jerryhethatday:
In general, What does `user bits` mean? Does it mean bits that is custom added by the protocal implementer?

PR #2447, opened 2020-05-02
subsystem: add an optional node to use for assigning hartids

terpstra:
This is useful when the hartid of a core cannot be known statically.

### DIFF
--- a/src/main/scala/subsystem/RocketSubsystem.scala
+++ b/src/main/scala/subsystem/RocketSubsystem.scala
@@ -12,6 +12,8 @@ import freechips.rocketchip.diplomaticobjectmodel.logicaltree._
 import freechips.rocketchip.diplomaticobjectmodel.model._
 import freechips.rocketchip.tile._
 
+case object HartPrefixKey extends Field[Boolean](false)
+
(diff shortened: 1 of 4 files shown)


src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -80,14 +82,29 @@ class RocketSubsystem(implicit p: Parameters) extends BaseSubsystem
   // add Mask ROM devices
   val maskROMs = p(PeripheryMaskROMKey).map { MaskROM.attach(_, cbus) }
 
+  val hartPrefixNode = if (p(HartPrefixKey)) {
```
hcook, line 85: If possible, I would put this in the trait `HasRocketTiles`, which actually currently gets reused across multiple `BaseSubsystem` subclasses, unlike `RocketSubsystem` which is only used here and in `soc-testsocket-sifive`

terpstra: I would do that, but then this PR would break those other subclasses. Since the hartReg() part of this is needed to unblock the pipeline, I think I will do that refactor as a follow-up PR.

PR #2449, opened 2020-05-04
Provide reset values for registers driving cease and wfi

SandeepRajendran:
Registers driving cease and wfi did not have reset values and could possibly be high for one cycle under reset.

**Related issue**: 

**Release Notes**


mwachs5:
yes, but this is shared code for cores that may satisfy that property

PR #2450, opened 2020-05-04
additional TileLink assertions

debs-sifive:
**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/tilelink/Monitor.scala
+++ b/src/main/scala/tilelink/Monitor.scala
@@ -167,11 +167,11 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
     val mask = edge.full_mask(bundle)
     val legal_source = Mux1H(edge.client.find(bundle.source), edge.client.clients.map(c => c.sourceId.start.U)) === bundle.source
 
-    if (!edge.manager.anySupportArithmetic) monAssert (bundle.opcode =/= TLMessages.ArithmeticData, s"''B' channel should not see ArithmeticData")
-    if (!edge.manager.anySupportLogical) monAssert (bundle.opcode =/= TLMessages.LogicalData, s"''B' channel should not see LogicalData")
-    if (!edge.manager.anySupportGet) monAssert (bundle.opcode =/= TLMessages.Get, s"''B' channel should not see Get")
-    if (!edge.manager.anySupportPutFull) monAssert (bundle.opcode =/= TLMessages.PutFullData, s"''B' channel should not see PutFullData")
-    if (!edge.manager.anySupportPutPartial) monAssert (bundle.opcode =/= TLMessages.PutPartialData, s"''B' channel should not see PutPartialData")
+    if (!edge.client.anySupportArithmetic) monAssert (bundle.opcode =/= TLMessages.ArithmeticData, s"''B' channel should not see ArithmeticData")
+    if (!edge.client.anySupportLogical) monAssert (bundle.opcode =/= TLMessages.LogicalData, s"''B' channel should not see LogicalData")
+    if (!edge.client.anySupportGet) monAssert (bundle.opcode =/= TLMessages.Get, s"''B' channel should not see Get")
+    if (!edge.client.anySupportPutFull) monAssert (bundle.opcode =/= TLMessages.PutFullData, s"''B' channel should not see PutFullData")
+    if (!edge.client.anySupportPutPartial) monAssert (bundle.opcode =/= TLMessages.PutPartialData, s"''B' channel should not see PutPartialData")
 
     when (bundle.opcode === TLMessages.Probe) {
       monAssert (edge.client.supportsProbe(bundle.source, bundle.size), "'B' channel carries Probe type unsupported by client" + extra)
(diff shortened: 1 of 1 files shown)


src/main/scala/tilelink/Monitor.scala
```diff
@@ -66,6 +66,18 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
       c.visibility.map(_.contains(address)).reduce(_ || _)
     }.reduce(_ && _)
 
+  def checkFormatOpcode(opcode: UInt, edge: TLEdge, format: String): Unit = {
+    // read params and check the opcode
+    if (!edge.manager.anySupportAcquireT) monAssert (opcode =/= TLMessages.AcquirePerm, s"'${format} channel should not see AcquirePerm")
+    if (!edge.manager.anySupportAcquireB) monAssert (opcode =/= TLMessages.AcquireBlock, s"'${format} channel should not see AcquireBlock")
```
hcook, line 72: unfortunately, Trunk/Branch is orthogonal to Perm/Block. The former are specified in the `param` field, the latter are specified in the `opcode` field. I don't think we have any diplomatic `edge.manager.anySupport` helper methods for distinguishing between AcquirePerm vs AcquireBlock opcode types, so you'll have to assume that if one is allowed the other is too.

@terpstra @brrmorre should be making the `support` methods more granular in the `v2` versions?

debs-sifive: Ok, I kind of expected this to fail b/c I have no idea what `anySupportAcquireT` and `anySupportAcquireB` are supposed to do. I was loosely pattern-matching based off of name and clearly guessed wrong.

My goal with this PR is to look at all the params of a node, and write relevant assertions that don't exist yet.

So I think what I need to figure out is exactly what `anySupportAcquireT` and `anySupportAcquireB` mean (please help me if you know/know where I should look for this info).

jchang0:
Let's remove the two lines that are not quite right and push the rest in.  I think I'd like to have small PRs, especially as we want to push the flow on downstream and get feedback from them in terms of how this improves coverage analysis.  Especially just the few we have already probably goes a long way.

src/main/scala/tilelink/Monitor.scala
```diff
@@ -161,6 +173,8 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
     val mask = edge.full_mask(bundle)
     val legal_source = Mux1H(edge.client.find(bundle.source), edge.client.clients.map(c => c.sourceId.start.U)) === bundle.source
 
+    checkFormatOpcode(bundle.opcode, edge, "B")
```
jchang0, line 176: I think the manager parameters only applied to traffic on channel A (and C, but I don't think we send any of those transactions on channel C)

debs-sifive:
I removed the `acquire` ones on a prior commit. Removing `putpartial` b/c I am seeing that fail travis.

src/main/scala/tilelink/Monitor.scala
debs-sifive: what are "manager parameters"? Looking at the spec, channels A, B, and C seem to have the same signals?

jchang0:
I meant to write a new comment.  Looked at the spec just remove it from C and D channel, I mean.  C and D channel doesn't have those operations.

src/main/scala/tilelink/Monitor.scala
```diff
@@ -173,7 +167,11 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
     val mask = edge.full_mask(bundle)
     val legal_source = Mux1H(edge.client.find(bundle.source), edge.client.clients.map(c => c.sourceId.start.U)) === bundle.source
 
-    checkFormatOpcode(bundle.opcode, edge, "B")
+    if (!edge.manager.anySupportArithmetic) monAssert (bundle.opcode =/= TLMessages.ArithmeticData, s"''B' channel should not see ArithmeticData")
```
jchang0, line 170: For B channel, you'll need edge.client.anySupportArithmetic (or equivalent of)

src/main/scala/tilelink/Monitor.scala
```diff
@@ -320,7 +318,7 @@ class TLMonitor(args: TLMonitorArgs, monitorDir: MonitorDirection = MonitorDirec
     val deny_put_ok = edge.manager.mayDenyPut.B
     val deny_get_ok = edge.manager.mayDenyGet.B
 
-    checkFormatOpcode(bundle.opcode, edge, "D")
+    if (!edge.manager.anySupportHint) monAssert (bundle.opcode =/= TLMessages.Hint, s"''D' channel should not see Hint")
```
jchang0, line 321: similarly here (edge.client.anySupporHint)

debs-sifive:
@terpstra I was told you have a stake in how we do our TL assertions, so I am open to any feedback! My goal with this PR is to write some simple assertions based on what the A/B channels support.

terpstra:
These assertions are redundant as the `when() {` clause for each message type already checks if the operation is supported and also confirms the size of the operation is supported.

PR #2451, opened 2020-05-05
.travis.yml: get rid of request for sudo

mwachs5:
Back at the end of 2017 apparently `sudo: required` gave you a bigger disk image, which we needed. But we're seeing a lot of Travis instability now so reverting this change to see if a) we still need that and b) whether it resolves some of the stability issues

**Related issue**: 

**Release Notes**

For CI, don't request sudo (which gives a larger disk image) from Travis

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 git:
   submodules: false
 language: scala
-sudo: required
+sudo: false
 cache:
   apt: true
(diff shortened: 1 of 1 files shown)


mwachs5:
So, this *did* work and it passed travis on the first try. Will merge EOD today if no one screams

PR #2459, opened 2020-05-10
Use GitHub Actions for CI

richardxia:
**Related issue**: N/A

**Impact**: API modification

**Release Notes**

---

I'm going to make up my own PR template to organize my thoughts:

## Problem Statement

I've occasionally run into friction while using Travis CI, and I think it's worth trying out other CI services to see whether they have nicer experiences. Some of my main issues with Travis CI are:

1. Documentation just doesn't feel that clear or polished. I often find myself not understanding exactly how something works and having to just try it out in order to test it.
2. Parts of their product that feel a bit "Frankenstein" like, where they ended up in a dead-end API-wise, but then had to support both an older API and a newer API simultaneously. For example, jobs configured as Build Stages vs. Jobs configured using a normal build have noticeably different behavior and syntax.
3. Travis CI kills any jobs that don't print to stdout at least once every 10 minutes, treating them as hangs. `travis_wait` is an annoying kludge, since it has the side effect of hiding all of the normal console output of your command until either it completes or until the timeout kicks.
4. You are limited to five parallel jobs at a time.
5. Anecdotally, I feel that I very often see transient failures due to networking issues, and I'm actually inclined to believe that the way Travis CI have set up their server networking, it's often an issue on their end rather than on the other side of the internet.
6. The way caching is handle is confusing

## Why GitHub Actions may be a better solution

Addressing each of my points above, one by one:

1. Although I feel that GitHub Actions' documentation could be better, it already feels a lot more detailed and precise than Travis CI's.
2. GitHub Actions' APIs and schemas feel much more conceptually unified than Travis CI's. I find it less difficult to reason about how to make a configuration change in GitHub Actions.
3. GitHub Actions does not have the restriction on needing to print to stdout at least once every 10 minutes. GitHub Actions also seems to have much longer max timeouts set (I think 5 hours per job).
4. GitHub Actions appears to have a max of 20 concurrent jobs.
5. Anecdotally, I feel that I see errors due to networking less frequently on GitHub Actions.
6. GitHub Actions' caching APIs make a lot more sense to me. You can specify multiple caches, each with a different key, and they also have utility functions for computing hashes for keys based on the contents of files (e.g. build.sbt), which allows us to control cache hits and misses.

## Drawbacks to using GitHub Actions

Although I have had a fairly positive experience with GitHub Actions, I have noticed some things that feel like a step back from Travis CI:

1. You cannot delete caches. You must either way for the cache to automatically be cleaned up (one week with no hits) or you must change the cache key, which requires making a Git commit to change the configuration file.
2. Parallel jobs flood the status indicator at the bottom of a PR. Travis normally has a single bubble for the entire workflow, but GitHub Actions has a bubble for each job in a workflow.
3. GitHub Actions has a much, much smaller amount of disk available (14 GB, including whatever comes on the preinstalled VM image they use, vs. Travis CI's ~50 GB).
4. You cannot restart an individual parallel job; you must restart the whole workflow.
5. Anecdotally, I feel that GitHub Actions has run into a lot of infrastructural issues over the past several months (coincident with COVID-19?). https://www.githubstatus.com/ shows quite a lot of outages. Travis CI's for comparison: https://www.traviscistatus.com/

## What this PR changes

I've done a heavy amount of rebasing and squashing so that my commits are ordered in a specific way.

### Refactoring work that is probably useful to merge in even if we don't want to use GitHub Actions

https://github.com/chipsalliance/rocket-chip/commit/2963b428086bc527c550783b03c5ea2871d42377 - I pulled out the actual `make` commands from the `.travis.yml` file into a bash script so that I could more easily run them in either Travis CI or GitHub Actions. This would affect anyone that wants to update the CI tests in both Travis CI and GitHub Actions, since they must now modify that bash script. I would want this to be easily modifiable by others, so please let me know if you find this more confusing or worse than what we had previously with Travis CI.

https://github.com/chipsalliance/rocket-chip/commit/74e29617717043397d5d0fc72aad51078e8ef26a - I added a `verilator.hash`, modeled after `riscv-tools.hash`, so that I could share that version number between both Travis CI and GitHub Actions and so that I could use that hash to compute the cache key in GitHub Actions.

https://github.com/chipsalliance/rocket-chip/commit/ff21f883239708cb8723d67cea3e45a7eabafbf8 - I modified the `regression/Makefile` targets to invert the dependency between `riscv-tests.stamp` and `rocket-tools_checkout.stamp`. Previously, `riscv-tests.stamp` depended on `rocket-tools_checkout.stamp`, and `riscv-tests.stamp` was essentially a no-op, since it relied on `rocket-tools_checkout.stamp` to actually check out riscv-tests. I modified this because I noticed it was taking 30 minutes to just clone all of riscv-tools' submodules, particularly riscv-gnu-toolchain and fsf-binutils-gdb, even if you have a cache hit on a precompiled riscv-tools, since the GDB tests require the actual riscv-tests source code to be checked out.

I changed this so that now `rocket-tools_checkout.stamp` depends on `riscv-tests.stamp`, and `riscv-tests.stamp` is now responsible for doing the clone of rocket-tools.git and riscv-tests.git. `rocket-tools_checkout.stamp` is still responsible for cloning all the other submodules.  If @aswaterman or someone could let me know if I've done this correctly, that would be great, since I don't think I understand how these targets work or how they are meant to be used in local development.

### Adding the actual GitHub Actions workflow definitions

I broke up the commits such that each commit introduces a different job in the workflow file (e.g. wit submodule check, prepare riscv-tools cache, etc.). I think the main interesting one is the one that actually runs the main tests, which I set up as a matrix job: https://github.com/chipsalliance/rocket-chip/commit/245898ee70e8497e35c04b7fa768fe65bb385358

The other main commit of note is https://github.com/chipsalliance/rocket-chip/commit/093659dfa4c32eb14b6c3e273c31bb64d335b4a7, where I added a `README_GITHUB_ACTIONS.md`.

## Proposed rollout

I'm not sure if we want to make a big change to flip from Travis CI to GitHub Actions, so I tried to develop this PR to support running both CI systems simultaneously. I think it may even be advantages to perpetually run both CI systems, since an outage in one service will (hopefully) not affect the other service, and we can more comfortably waive transient failures from one service with the passing results from the other.

I think we actually do get quite a lot of resilience from using both providers, since they even use different cloud vendors underneath: Travis CI uses Google Cloud Platform, while GitHub Actions uses Microsoft Azure. This means that we're more robust to datacenter-wide failures as well as entire GCP- or Azure-wide failures.

That said, it is more overhead to maintain two CI systems, and I think there is a social risk of letting one system rot if it starts deterministically failing, since we always have the other to rely on.

## Special thanks

@aswaterman, for helping me with working out some issues with the compilation of the fesvr and Verilator when using mismatched versions of g++! That really helped unblock me and get me to the finish line!


aswaterman:
You’re welcome! I’m still curious how the compiler version change resulted in a silent ABI incompatibility...

richardxia:
☹️ My job failed Travis CI with the following error:

```
./regression/run-test-bucket: line 42: travis_wait: command not found
```

According to https://github.com/travis-ci/travis-ci/issues/9336, `travis_wait` is a shell function, not an executable, and therefore it is not something that can be invoked from within a sub shell script. It looks like the recommended approach is to only call `travis_wait` from `.travis.yml`. This unfortunately means that I have to make the `travis_wait` timeouts less fine grained now, since my `run-test-bucket` script is only called once per job, and that means the timeout can only be applied to the whole bucket, not the individual `make` commands within it.

Alternatively, I could have a separate "bucket" per `make` command and call `run-test-bucket` multiple times within a single job.

Lastly, one other approach would be to use an alternative tool to prevent Travis CI from killing jobs. It looks there are a couple others:

- https://github.com/chanzuckerberg/go-travis-wait
- https://github.com/cisagov/travis-wait-improved

Neither has that many stars/commits/forks, but I suppose this would be a pretty simple program and probably wouldn't cause too many issues. If we installed one of these, then we could call that program from within the `run-test-bucket` script.

aswaterman:
If travis_wait is as simple as I assume it is, can't we just include our own script that we can directly call?  Some googling led me to this issue on travis-ci where it seems travis_wait was originally proposed; perhaps we could just crib and lightly modify that: https://github.com/travis-ci/travis-ci/issues/1962

richardxia:
Ah, I had followed a different link, where I saw that [`travis_wait`](https://github.com/travis-ci/travis-build/blob/3af56667b3221448763e79a73007b94a38bb9612/lib/travis/build/templates/header.sh#L213-L271) was defined in a large, templated bash script containing many other function and which is run through a Ruby template engine (ERB). However, upon closer inspection, it looks like they ditched that approach sometime last year, and now there is a [completely standalone bash script](https://github.com/travis-ci/travis-build/blob/73a5393263e0b135f49aceeb40ef6f0d827b9b11/lib/travis/build/bash/travis_wait.bash) that only defines `travis_wait` and which is not set through a Ruby template engine. It does look like that standalone script still [depends](https://github.com/travis-ci/travis-build/blob/73a5393263e0b135f49aceeb40ef6f0d827b9b11/lib/travis/build/bash/travis_wait.bash#L16) on [`travis_jigger.bash`](https://github.com/travis-ci/travis-build/blob/73a5393263e0b135f49aceeb40ef6f0d827b9b11/lib/travis/build/bash/travis_jigger.bash), but I think `travis_jigger.bash` does not depend on anything else.

This looks more promising, so let me see if I can just include that into `run-test-bucket`.

mwachs5:
Cool! Personally, I don't hate the all-the-parallel-jobs-showing-up-in-the-github-status thing. It's one less thing to have to click to see how your jobs are doing.

README_GITHUB_ACTIONS.md
```diff
@@ -0,0 +1,38 @@
...
+rocket-chip makes use of [GitHub Actions](https://github.com/features/actions)
+for running continuous integration (CI) tests. For more information on GitHub
+Actions, please see their [documentation](https://help.github.com/en/actions).
+
+
+## How It Is Configured
+
+The main configuration file describing the continuous integration workflow is
+located at
+[.github/workflows/continuous-integration.yml](.github/workflows/continuous-integration.yml).
```
mwachs5, line 12: If this YML file is only for GitHub Actions, is it possible to add as a comment to the top of that YML file itself? To avoid confusion for someone trying to tweak the travis job?

README_GITHUB_ACTIONS.md
```diff
@@ -0,0 +1,38 @@
...
+## Caching
+
+Caching is handled by the official GitHub Action [actions/cache](https://github.com/marketplace/actions/cache).
+Caches are kept around for one week, and caches are searched in order branch,
+base branch (e.g. PR target), and default branch (e.g. master).
+
+[There is no way to delete a cache](https://github.com/actions/cache/issues/294),
+so the recommended approach for invalidating a cache is to modify the cache key.
+In the rocket-chip GitHub Action workflow configuration file, the cache keys
+each end in a number. These numbers can be incremented to "invalidate" the
```
mwachs5, line 33: I am trying to understand exactly what you have to change. Are you saying you'd change `-v1` to `-v2` e.g. here: https://github.com/chipsalliance/rocket-chip/pull/2459/files#diff-17b09a7990ba40f6a89628ab9402f1f5R150. If so, perhaps you could change this to say "each end in a version, e.g.` -v1`. These version numbers can be incremented..." If you mean something else it might be nice to clarify since this seems like something users may need to do even if they aren't that familiar with the Github Actions API

regression/Makefile
```diff
@@ -150,11 +150,15 @@ stamps/other-submodules.stamp:
...
-	mkdir -p $(dir $@)
+stamps/riscv-tests.stamp:
 	git -C $(abspath $(TOP)) clone -n https://github.com/freechipsproject/rocket-tools.git
 	git -C $(RISCV_TOOLS) checkout $(TOOLS_HASH)
-	git -C $(RISCV_TOOLS) submodule update --init --recursive riscv-isa-sim  riscv-tests riscv-pk riscv-openocd fsf-binutils-gdb
+	git -C $(RISCV_TOOLS) submodule update --init --recursive riscv-tests
+	mkdir -p $(dir $@)
+	date > $@
+
+stamps/rocket-tools_checkout.stamp: | stamps/riscv-tests.stamp
```
mwachs5, line 160: this looks like the right division to me

README_GITHUB_ACTIONS.md
richardxia: Yes, this is only for GitHub Actions (which requires the file to be located at `.github/workflows`). Added a comment in 3becccb04f30d679d347f3c27305510660bb7e5d.

README_GITHUB_ACTIONS.md
richardxia: Thanks for the feedback. I had a feeling that what I wrote wouldn't be clear. I updated the descriptions to give an example of the version number and to include step-by-step instructions for invaliding the cache: 1e2fe619ea35f8bc6a417372049878bd4b1b5679

Can you take a look and let me know if that's any better?

mwachs5: yep, that's very clear. Thanks!

mwachs5: one final thing... what's an example "However, there are situations where
you may need to manually invalidate the caches."

Like, is it a situation where you need to invalidate your cache because you were doing something, then rebased on master which had updated the riscv-tools.hash and therefore you want to grab the master cache instead of your PR cache? But that seems like not a scenario since when you rebsae on master your riscv-tools.hash will get updated...

I guess maybe the answer is "there may be uncommon scenarios"

richardxia: No, the scenario I was thinking of was the scenario I hit over and over again while developing this PR, which was that any changes to the actual configuration of the riscv-tools compile steps (Makefile rules) or dependent packages will _not_ get picked up.

I suppose that I could include the Makefile rules in the hash computation, but I was afraid that would lead to too many false invalidations. I am, however, now thinking that putting the `apt-get install` stuff into a standalone script is not a bad idea, since that would allow us to more properly invalidate the cache if someone changes the dependencies.

I'll probably work on that, but I might now have much time to work on this PR for a couple days.

terpstra:
I don't have anything constructive to add other than: 👍 👍 👍 

Thank god.

richardxia:
I think the caching is getting messed up for some weird reason, and now I'm failing GitHub Actions CI. This ticket seems to be related: https://github.com/actions/cache/issues/208

Sigh. Maybe this is not really ready for primetime yet.

richardxia:
I'm going to attempt to close this PR and create a new one to see if creating a new one improves anything.

PR #2461, opened 2020-05-12
Dynamic local and remote order

terpstra:

**Release Notes**
Improved QoR in designs with large physical address maps.

### DIFF
--- a/src/main/scala/tilelink/BusWrapper.scala
+++ b/src/main/scala/tilelink/BusWrapper.scala
@@ -356,8 +356,7 @@ case class AddressAdjusterWrapperParams(
   blockBytes: Int,
   beatBytes: Int,
   replication: Option[ReplicatedRegion],
-  forceLocal: Seq[AddressSet] = Nil,
-  policy: TLFIFOFixer.Policy = TLFIFOFixer.allVolatile
+  forceLocal: Seq[AddressSet] = Nil
 )
   extends HasTLBusParams
   with TLBusWrapperInstantiationLike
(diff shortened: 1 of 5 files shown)


src/main/scala/tilelink/BusWrapper.scala
```diff
@@ -374,7 +373,7 @@ case class AddressAdjusterWrapperParams(
 class AddressAdjusterWrapper(params: AddressAdjusterWrapperParams, name: String)(implicit p: Parameters) extends TLBusWrapper(params, name) {
   private val address_adjuster = params.replication.map { r => LazyModule(new AddressAdjuster(r, params.forceLocal)) }
   private val viewNode = TLIdentityNode()
-  val inwardNode: TLInwardNode = address_adjuster.map(_.node :*=* TLFIFOFixer(params.policy) :*=* viewNode).getOrElse(viewNode)
```
hcook, line 377: Is the fact that you are now ignoring `policy` going to have implications for throughput to (non-relocatable) outer caches or cacheable memory ports when an AA is put between the tiles and the SBUS?
Or are any such limitations ameliorated by the fact that all such regions are expected to be down a single fork, either local or remote?

terpstra: The policy has only ever affected masters which request FIFO order (ie: not caches).

However, if you have an E-series core that alternates between using local and remote memory, yes it will end up stalling whenever it switches regions.

The FIFOFixer had the option to leave some regions 'unfixed'. I don't think we can do that here, when you have a local 1GiB in a 32PiB region, there are too many cases to describe well statically. This was one of the exponential blow-ups this PR is fixing.

What we could do is perhaps have a mode where the entire thing gets turned into fifoId=None. We remove the requirement for FIFO from the legs and wipe out fifoId in the adapter. This might make sense for main memory.

What do you think?

hcook: The mode option seems like a good dimension of flexibility but your first point ameliorates my concern.

PR #2470, opened 2020-05-15
AddressAdjuster patches

hcook:
Porting some old enhancements to the AddressAdjuster and one bugfix:
- Allow DTS to report a default relocation amount that isn't the 0th region
- Allow the two legs to have divergent user bits with :<=

### DIFF
--- a/src/main/scala/tilelink/AddressAdjuster.scala
+++ b/src/main/scala/tilelink/AddressAdjuster.scala
@@ -7,8 +7,11 @@ import chisel3.util._
 import freechips.rocketchip.config._
 import freechips.rocketchip.diplomacy._
 
-// forceLocal -> used to ensure special devices (like debug) remain reacheable at chip_id=0 even if in params.region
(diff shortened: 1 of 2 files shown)


src/main/scala/tilelink/AddressAdjuster.scala
```diff
@@ -7,18 +7,27 @@ import chisel3.util._
...
   // Which bits are in the mask?
   val bits = AddressSet.enumerateBits(mask)
   // Find the portion of the addresses which correspond to prefix0
   private def prefix0(region: Seq[AddressSet]): Seq[AddressSet] = {
     region.flatMap { _.intersect(params.local) }
   }
-  private def prefixNot0(region: Seq[AddressSet]): Seq[AddressSet] = {
-    region.flatMap { _.subtract(params.local) }
+  // Default region is used to display address map metadata which corresponds to the expected default choice of prefix value
+  private val defaultRegion: AddressSet = params.local.copy(base = params.local.base + localBaseAddressDefault.getOrElse(0))
```
terpstra, line 23: You might consider adding in a require() that the 'localBaseAddressDefault' only uses bits in the replicationMask.
ie: localBaseAddressDefault.foreach { x => require (x & ~mask == 0) }

PR #2477, opened 2020-05-18
RVV 0.9 changes

aswaterman:
- Fractional LMUL
- Tail-agnostic/mask-agnostic bits
- EEW loads/stores
- Some encoding changes
- Add utilities for bitwise shifts by signed shift amounts


src/main/scala/rocket/Instructions.scala
```diff
@@ -319,52 +319,40 @@ object Instructions {
...
-  def VLEFF_V            = BitPat("b???000?10000?????111?????0000111")
-  def VLBUFF_V           = BitPat("b???000?10000?????000?????0000111")
-  def VLHUFF_V           = BitPat("b???000?10000?????101?????0000111")
-  def VLWUFF_V           = BitPat("b???000?10000?????110?????0000111")
-  def VL1R_V             = BitPat("b000000101000?????111?????0000111")
-  def VS1R_V             = BitPat("b000000101000?????111?????0100111")
+  def VLE8_V             = BitPat("b???000?00000?????000?????0000111")
+  def VLE16_V            = BitPat("b???000?00000?????101?????0000111")
+  def VLE32_V            = BitPat("b???000?00000?????110?????0000111")
+  def VLE64_V            = BitPat("b???000?00000?????111?????0000111")
```
solomatnikov, line 325: Why did you define `VLE...` opcodes for every width instead of single `VLE` opcode and then extracting the width from instruction bits directly?

aswaterman: Note all three funct3 bits get used (bit14 is 0 for VLE8, but it’s 1 for the others). This is because FLH/FLW/FLD/FLQ are in the same major opcode. If I replaced funct3 with ???, then VLE would conflict with the FP loads.

Of course this is just for illegal instruction detection. Once you get past that point, I agree you should just extract bits 13:12 to determine the width, rather than treating these as four separate instructions.

PR #2478, opened 2020-05-18
Update GeneratorUtils.scala "did you misspell it" error message

mwachs5:
**Related issue**: 

**Release Notes**

Update GeneratorUtils error message to suggest checking the package path as well as spelling when it can't find a config


jackkoenig:
If someone wanted to have fun with this, they could copy the code from https://github.com/scala/scala/blob/3b28fb7021441f4ab305129923401866a86ab041/src/compiler/scala/tools/nsc/util/EditDistance.scala (or write their own implementation of edit distance) and reflectively find all Configs on the classpath and make suggestions on correct spellings :-P

PR #2479, opened 2020-05-18
AXI4 Deinterleaver: update to chisel3 to get inferred reset

mwachs5:
**Related issue**: None

**Release Notes**

Support Asynchronous Reset for AXI4 Deinterleaver

### DIFF
--- a/src/main/scala/amba/axi4/Deinterleaver.scala
+++ b/src/main/scala/amba/axi4/Deinterleaver.scala
@@ -41,14 +41,22 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
         val qs = Seq.tabulate(endId) { i =>
           val depth = edgeOut.master.masters.find(_.id.contains(i)).flatMap(_.maxFlight).getOrElse(0)
           if (depth > 0) {
-            Module(new Queue(out.r.bits.cloneType, beats)).io
+            val q = Module(new Queue(out.r.bits.cloneType, beats))
+            q.suggestName(s"queue_${i}")
+            q.io
           } else {
-            Wire(new QueueIO(out.r.bits.cloneType, beats))
+            // These are unused IDs and should be never used.
+            // But, to satisfy type checks we must produce a Wire of the
+            // correct type.
+            val q = Wire(new QueueIO(out.r.bits.cloneType, beats))
+            q.suggestName(s"queue_wire_${i}")
+            assert(!q.enq.valid, s"ID ${i} should not be used")
+            // These could just be marked DontCare
+            q.deq :<> q.enq
+            q.count := 0.U
+            q
           }
         }
-
-        // This seems dubious. Are all these queues not actually used?
-        qs.foreach { _ := DontCare }
 
         // Which ID is being enqueued and dequeued?
         val locked = RegInit(false.B)
@@ -85,16 +93,19 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
 
         // Transmit the selected burst to inner
         in.r.valid := locked
-        val deq_bits = WireInit(VecInit(qs.map(_.deq.bits)))
+        val deq_bits = VecInit(qs.map(_.deq.bits))
         in.r.bits  := deq_bits(deq_id)
-        (deq_OH.asBools zip qs) foreach { case (s, q) =>
+        val deq_OH_bools = deq_OH.asBools
+        require(deq_OH_bools.size == qs.size, s"deq_OH.size != qs.size (${deq_OH_bools.size} vs ${qs.size})")
+        (deq_OH_bools zip qs) foreach { case (s, q) =>
           q.deq.ready := s && in.r.fire()
         }
 
+        require(enq_OH_bools.size == qs.size, s"enq_OH.size != qs.size (${enq_OH_bools.size} vs ${qs.size})")
         // Feed response into matching Q
-        val enq_readys = WireInit(VecInit(qs.map(_.enq.ready)))
+        val enq_readys = VecInit(qs.map(_.enq.ready))
         out.r.ready := enq_readys(enq_id)
-        (enq_OH.asBools zip qs) foreach { case (s, q) =>
+        (enq_OH_bools zip qs) foreach { case (s, q) =>
           q.enq.valid := s && out.r.valid
           q.enq.bits := out.r.bits
         }
--- a/src/main/scala/amba/axi4/Deinterleaver.scala
+++ b/src/main/scala/amba/axi4/Deinterleaver.scala
@@ -51,9 +51,7 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
             val q = Wire(new QueueIO(out.r.bits.cloneType, beats))
             q.suggestName(s"queue_wire_${i}")
             assert(!q.enq.valid, s"ID ${i} should not be used")
-            // These could just be marked DontCare
-            q.deq :<> q.enq
-            q.count := 0.U
+            q := DontCare
             q
           }
         }
--- a/src/main/scala/amba/axi4/Deinterleaver.scala
+++ b/src/main/scala/amba/axi4/Deinterleaver.scala
@@ -2,9 +2,9 @@
 
 package freechips.rocketchip.amba.axi4
 
-import Chisel.{defaultCompileOptions => _, _}
-import freechips.rocketchip.util.CompileOptions.NotStrictInferReset
-import chisel3.util.IrrevocableIO
+import chisel3._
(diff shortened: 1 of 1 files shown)


mwachs5:
yeah but I can't actually do that in the GitHub gui :)

mwachs5:
the need for `DontCare`s is killing me. Will push what I have in the "full chisel3 conversion" but getting a bunch of errors about the queues not being fully initialized

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -46,9 +47,12 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
           }
         }
 
+        // This seems dubious. Are all these queues not actually used?
+        qs.foreach { _ := DontCare }
```
mwachs5, line 51: @terpstra why is this OK (if I don't put this i get lots of firrtl errors about things being `VOID`)? At the very least it seems we should be defaulting the readys and valids to 0, not DontCare... down below they only get conditionally set which makes `chisel3` unhappy.

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -81,13 +85,15 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
...
+        val deq_bits = WireInit(VecInit(qs.map(_.deq.bits)))
+        in.r.bits  := deq_bits(deq_id)
         (deq_OH.asBools zip qs) foreach { case (s, q) =>
           q.deq.ready := s && in.r.fire()
         }
 
         // Feed response into matching Q
-        out.r.ready := Vec(qs.map(_.enq.ready))(enq_id)
+        val enq_readys = WireInit(VecInit(qs.map(_.enq.ready)))
+        out.r.ready := enq_readys(enq_id)
```
terpstra, line 96: Why did you need the WireInit?

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -81,13 +85,15 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
 
         // Transmit the selected burst to inner
         in.r.valid := locked
-        in.r.bits  := Vec(qs.map(_.deq.bits))(deq_id)
+        val deq_bits = WireInit(VecInit(qs.map(_.deq.bits)))
+        in.r.bits  := deq_bits(deq_id)
```
terpstra, line 89: Was changing Vec to VecInit not enough?

src/main/scala/amba/axi4/Deinterleaver.scala
mwachs5: you can's just do Vec(Seq) anymore. I think this is what the cihsel3 cookbook says to do now.

src/main/scala/amba/axi4/Deinterleaver.scala
mwachs5: TBH i didn't try it. I will try it out. I was going by memory of fixing these things before, but maybe that's only needed for RegInit

src/main/scala/amba/axi4/Deinterleaver.scala
terpstra: Hmm. I don't understand this, either. The ready and valid seem to be driven unconditionally in the zip+foreach. Is it possible that the zip length is shorter than the 'qs' length?

terpstra: Hmm. I don't understand this, either. The ready and valid seem to be driven unconditionally in the zip+foreach. Is it possible that the zip length is shorter than the 'qs' length?

src/main/scala/amba/axi4/Deinterleaver.scala
jackkoenig: `VecInit` should be sufficient

src/main/scala/amba/axi4/Deinterleaver.scala
mwachs5: ok you're right `VecInit` alone is enough

src/main/scala/amba/axi4/Deinterleaver.scala
mwachs5: confirmed

src/main/scala/amba/axi4/Deinterleaver.scala
mwachs5: i put a require that they be the same length, which did not fail, yet i still get the errors if I dont put in the `DontCare`.

mwachs5: @terpstra did something change about `QueueIO`? in the case where you DON't put down a queue for `qs` and just put down a QueueIO, what makes the deq.bits := enq.bits ?

I think that's what its complaining about. I don't know why this would have worked before

mwachs5: (talking about this line: https://github.com/chipsalliance/rocket-chip/blob/a344f719bd14db9a9e77b868eb645afba616f232/src/main/scala/amba/axi4/Deinterleaver.scala#L46)

mwachs5: ah, the commit that added that line says "axi4: Deinterleaver need not make a Q for an unused AXI id". So presumably these wires are really `DontCare`-able, and we could put an assertion that their enq.fire == false?

Or I will plan to just wire them up so that deq := enq

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -27,28 +28,39 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
...
           } else {
-            Wire(new QueueIO(out.r.bits.cloneType, beats))
+            // These are unused IDs and should be never used.
+            // But, to satisfy type checks we must produce a Wire of the
+            // correct type.
+            val q = Wire(new QueueIO(out.r.bits.cloneType, beats))
+            q.suggestName(s"queue_wire_${i}")
+            assert(!q.enq.valid, s"ID ${i} should not be used")
+            // These could just be marked DontCare
+            q.deq :<> q.enq
```
mwachs5, line 55: @terpstra can you re-review this bit? This is all that is needed to be DontCare, not every queue, but I wired them up to eachother and added an assertion instead

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -81,14 +93,19 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
...
-        (deq_OH.asBools zip qs) foreach { case (s, q) =>
+        val deq_bits = VecInit(qs.map(_.deq.bits))
+        in.r.bits  := deq_bits(deq_id)
+        val deq_OH_bools = deq_OH.asBools
+        require(deq_OH_bools.size == qs.size, s"deq_OH.size != qs.size (${deq_OH_bools.size} vs ${qs.size})")
+        (deq_OH_bools zip qs) foreach { case (s, q) =>
           q.deq.ready := s && in.r.fire()
         }
 
+        require(enq_OH_bools.size == qs.size, s"enq_OH.size != qs.size (${enq_OH_bools.size} vs ${qs.size})")
```
mwachs5, line 104:
```suggestion
        val enq_OH_bools = enq_OH.asBools
        require(enq_OH_bools.size == qs.size, s"enq_OH.size != qs.size (${enq_OH_bools.size} vs ${qs.size})")
```

src/main/scala/amba/axi4/Deinterleaver.scala
terpstra: Ahh. This makes sense. Thanks for figuring it out! I do not think cross connecting them is wise, however. Use a DontCare instead.

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -27,28 +28,39 @@ class AXI4Deinterleaver(maxReadBytes: Int)(implicit p: Parameters) extends LazyM
...
+            // These are unused IDs and should be never used.
+            // But, to satisfy type checks we must produce a Wire of the
+            // correct type.
+            val q = Wire(new QueueIO(out.r.bits.cloneType, beats))
+            q.suggestName(s"queue_wire_${i}")
+            assert(!q.enq.valid, s"ID ${i} should not be used")
+            // These could just be marked DontCare
+            q.deq :<> q.enq
+            q.count := 0.U
+            q
```
mwachs5, line 57:
```suggestion
            q := DontCare
```

src/main/scala/amba/axi4/Deinterleaver.scala
mwachs5: done

PR #2480, opened 2020-05-20
TileLink Xbar Arbiter DecoupledIO Late Cancel

johningalls-sifive:
**Related issue**: 
**Type of change**: timing enhancement
**Impact**: API addition

**Release Notes**
- _Util_: create a new bundle `ReadyValidCancel` similar to `DecoupledIO`, but split `valid` into separate `earlyValid` and `lateCancel` wires.
- _TileLink_: create a new bundle `TLBundle_ACancel` using `ReadyValidCancel` instead of `DecoupledIO` wrapping `TLBundleA`.

**Proposal**
I would like to create a new interface similar to DecoupledIO that replaces `valid` with separate `earlyValid` and `lateCancel` signals, and use it to fix a timing path where we have too much logic on the A-channel request `valid` signal through the TileLink Xbar and Arbiter.
We use `earlyValid` as an early-setup signal for steering arbitration mux selects, and add a late-setup `lateCancel` signal in the same cycle to squash the arbitration output valid into the next pipe stage, but which does not affect the arbitration in that same cycle.  The interface signals look like this:
```
  val earlyValid = Output(Bool())
  val lateCancel = Output(Bool())
  val ready = Input(Bool())
  def valid:     Bool = earlyValid && !lateCancel
  def mightFire: Bool = earlyValid && ready
  def fire:      Bool = valid      && ready
```
I like explicitly including the "early" and "late" in the signal names, so the setup intent is clear in the RTL.
I deliberately chose `valid = earlyValid && !lateCancel` instead of `validQual = valid && !lateCancel` to _not_ be directly compatible with ReadyValidIO's `valid`, to avoid inadvertently forgetting `lateCancel`.
We have chosen this polarity `lateCancel` and not `lateConfirm` because this signal is often driven by `flush` or `replay`.

For reference comparison: Chisel3 Decoupled https://github.com/freechipsproject/chisel3/blob/master/src/main/scala/chisel3/util/Decoupled.scala

For completeness, some other solutions that we considered are:
- From @terpstra: Use the existing `valid` for your arbitration request, and modify `opcode` to a no-op if you decide to cancel your request.  Pro: should work similarly to `lateCancel`.  Con: I would feel best if this was an explicit `nop` message in TileLink, so that queues and various components could know to ignore/drop it.
- Change the TileLink Xbar and Arbiter IO to type `Decoupled(Decoupled(bits))`.  Pro: does not add a new flavor of interface bundle to Rocket-chip.  Con: unclear distinction between two valid signals: `valid` or `valid.bits.valid`.
- Pre-calculate the `lateCancel` term to keep the existing Decoupled valid interface, or if that is prohibitively expensive, add a pipeline stage.  Cons: adds hardware gates for parallelized computation, or latency for serialized computation.
- Refactor the Diplomatic hierarchy to move the `tlMasterXbar` further inside the design, so that I can hand-write this mux myself exactly how I want.  Cons: refactoring module hierarchy is a lot of effort to just work around a timing path like this.


src/main/scala/tilelink/Buffer.scala
```diff
@@ -82,3 +82,67 @@ object TLBuffer
...
+{
+  def this(ace: BufferParams, bd: BufferParams)(implicit p: Parameters) = this(ace, bd, ace, bd, ace)
+  def this(abcde: BufferParams)(implicit p: Parameters) = this(abcde, abcde)
+  def this()(implicit p: Parameters) = this(BufferParams.default)
+
+  val node = new TLBufferNodeAndNotCancel(a, b, c, d, e)
+
+  lazy val module = new LazyModuleImp(this) {
+    (node.in zip node.out) foreach { case ((in, edgeIn), (out, edgeOut)) =>
+      out.a <> a(in.a.asDecoupled)
```
terpstra, line 114: So a TLBufferAndNotCancel is the adapter that converts us back to standard Decoupled? That makes sense.
Where is the adapter from TLNode to TL_ACancel?

src/main/scala/tilelink/Bundle_ACancel.scala
```diff
@@ -0,0 +1,66 @@
...
+import freechips.rocketchip.util._
+import scala.collection.immutable.ListMap
+
+class TLBundle_ACancel(val params: TLBundleParameters) extends Record
+{
+  val a =  ReadyValidCancel(new TLBundleA(params))
+  val b = Flipped(Decoupled(new TLBundleB(params)))
+  val c =         Decoupled(new TLBundleC(params))
+  val d = Flipped(Decoupled(new TLBundleD(params)))
+  val e =         Decoupled(new TLBundleE(params))
```
terpstra, line 18: It seems to me we should really be doing this all or nothing. Why is it still Decoupled on BCDE?

src/main/scala/tilelink/Bundles.scala
```diff
@@ -37,15 +37,15 @@ object TLMessages
   def GrantData      = UInt(5) //                    .         => GrantAck
   def ReleaseAck     = UInt(6) //                    .
   def GrantAck       = UInt(0) //                         .
- 
+
```
terpstra, line 40: Please do not include unnecessary changes into a PR. It lead to spurious conflicts.

src/main/scala/util/ReadyValidCancel.scala
```diff
@@ -0,0 +1,151 @@
...
+import chisel3.experimental.chiselName
+import freechips.rocketchip.util._
+
+/** A [[Bundle]] that adds `earlyValid` and `lateCancel` bits to some data.
+  * This indicates that the user expects a "ValidCancel" interface between a producer and a consumer.
+  * Here, the producer asserts the `earlyValid` bit when data on the `bits` line might contain valid data.
+  * The producer also asserts the `lateCancel` bit low to confirm the valid, or high to squash the valid.
+  * This differs from [[ReadyValidCancel]] as there is no `ready` line that the consumer can use
+  * to put back pressure on the producer.
+  * @param gen the type of data to be wrapped in Valid/Cancel
```
terpstra, line 16: You should either explicitly allow or forbid earlyValid=false and lateCancel=true.

src/main/scala/util/ReadyValidCancel.scala
```diff
@@ -0,0 +1,151 @@
...
+  prevSelectEnc_en := inputEarlyValidVec.orR || prevSelectEnc_q.orR
+  prevSelectEnc_in := Mux(io.out.ready || !io.out.earlyValid,
+                          OHToUInt(selectDcd),
+                          prevSelectEnc_q)
+
+  private val grantVecVec = Wire(Vec(n, Vec(n, Bool())))
+  for (i <- 0 until n) {
+    grantVecVec(i)(i) := true.B
+    for (j <- 1 until n) {
+      val k = (i + j) % n
```
terpstra, line 123: I assume you copy-pasted this code. If so, my complaint is therefore directed at the original code. Doing all this calculations with Bools is unnecessarily slow for simulation. The work could've been done using bulk bitwise operators on UInts. No matter.

src/main/scala/tilelink/Xbar.scala
```diff
@@ -116,93 +219,117 @@ object TLXbar
...
       if (connectBIO(i).exists(x=>x)) {
         io_in(i).b :<> in(i).b
         io_in(i).b.bits.source := trim(in(i).b.bits.source, r.size)
       } else {
-        in(i).b.ready := Bool(true)
-        io_in(i).b.valid := Bool(false)
+        in(i).b.ready := true.B
+        in(i).b.bits  := DontCare
+        io_in(i).b.valid := false.B
+        io_in(i).b.bits  := DontCare
```
terpstra, line 248: Upgrading the TLXbar to chisel3 and DontCares is god's work. Thank you.

src/main/scala/tilelink/Buffer.scala
johningalls-sifive: I hadn't needed an adapter from TLNode to TL_ACancel yet.  I can add one, but it would be un-tested.

src/main/scala/tilelink/Arbiter.scala
```diff
@@ -32,17 +33,40 @@ object TLArbiter
     apply(lowestIndexFirst)(sink, sources.map(s => (edge.numBeats1(s.bits), s)):_*)
   }
 
+  def lowestFromSeq[T <: TLChannel](edge: TLEdge, sink: ReadyValidCancel[T], sources: Seq[ReadyValidCancel[T]]) {
+    applyCancel(lowestIndexFirst)(sink, sources.map(s => (edge.numBeats1(s.bits), s)):_*)
+  }
+
```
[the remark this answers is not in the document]
johningalls-sifive: it already exists for Decoupled immediately above this one

src/main/scala/tilelink/Bundle_ACancel.scala
johningalls-sifive: I'm fine doing it all-or-nothing, but I just figured that I should avoid changing everything at once.  So far the A-channel is the only critical path where I've needed this, but the other channels could plausibly want this in the future too.
Should I make this change now?  It will ripple through a handful of places.

src/main/scala/util/ReadyValidCancel.scala
johningalls-sifive: will convert to UInts

johningalls-sifive:
> Once we have a NoOp message, this whole alternative code path becomes unnecessary.

My understanding is that adding a NoOp message to TileLink v1 would require:
- adding the new opcode to components and verification collateral throughout the protocol and ecosystem
- add a bit to `a_opcode`, presumably resulting in an additional register bit

src/main/scala/diplomacy/Parameters.scala
```diff
@@ -275,6 +275,14 @@ case class BufferParams(depth: Int, flow: Boolean, pipe: Boolean)
...
     }
 
+  def andNotCancel[T <: Data](x: ReadyValidCancel[T]) = {
+    val y = Wire(new DecoupledIO(x.bits))
+    y.valid := x.validQual
+    y.bits  := x.bits
+    x.ready := y.ready
+    apply(y)
+  }
+
```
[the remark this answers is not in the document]
johningalls-sifive: this method is actually unused!  good eye, I'll remove the harmful dead code.

src/main/scala/util/ReadyValidCancel.scala
johningalls-sifive: unable: Firrtl throws combination logic loop warnings with UInt but not Vec(Bool)

PR #2481, opened 2020-05-20
RocketChipStage shouldn't extend ChiselStage

### LINKED ISSUE #2467: RocketChipStage.emitVerilog Doesn't Run FIRRTL
**Type of issue**: bug report

**Impact**: API Modification

**Development Phase**: request

**Other information**
 

Reported via Gitter ([link](https://view.matrix.org/room/!jSbgpvvpsPbLpsLZOC:matrix.org/?anchor=$tBV3UHZTrE9-MEprtYBUYTIqTRxBWmBrBAR2HRQdruo&highlight)) in reference to a usage of `emitVerilog` ([link](https://github.com/Neurodyne/fusionchip/blob/ebd5ca0340b243585de4c30350c3af8257e93e22/src/main/scala/Emit.scala#L38)).

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

The following should just generate Chirrtl when it is expected that it would generate Verilog:

```scala
class Foo extends RawModule

(new RocketChipStage).emitVerilog(new Foo)
```

**What is the current behavior?**

Using certain methods of `RocketChipStage` inherited from `ChiselStage` won't currently work as intended. 

Consider `emitFirrtl` (which emits High FIRRTL IR) and `emitVerilog` (which emits Verilog for a given design). Because these both rely on running `MaybeFirrtlStage`, but that is excluded via overriding `targets` in `RocketChipStage`, these methods don't work if you use them (nor do they really make sense to use here).

Fundamentally, `ChiselStage.targets` (or the actual `ChiselStage` class) should probably be `final` to avoid this problem. `RocketChipStage` may consequently not want to extend `ChiselStage`.

**What is the expected behavior?**

These methods should either not exist in `RocketChipStage` or they should work as expected.

**Please tell us about your environment:**

`master` branch at time of filing this issue (17c09cc948773b00a054190e428ab5800457893f).

**What is the use case for changing the behavior?**

Prevent user confusion.

seldridge:
Add a separate `RocketChiselStage` that adds appropriate modifications to `ChiselStage` that rocket chip needs (only modification is the addition of . Remove `ChiselStage` from `RocketChipStage`'s type hierarchy.

I only spot checked that this passes `DefaultConfig` and that the stack trace trimming didn't appear to break.

I think that this is a better approach than to rely on access to `ChiselStage.run` which is a method that should probably be made `protected`.

**Related issue**: Fixes #2467

**Impact**: API modification

**Release Notes**

- Remove emitVerilog, emitFirrtl, and emitChirrtl methods from RocketChipStage

### DIFF
--- a/src/main/scala/system/RocketChipStageGenerator.scala
+++ b/src/main/scala/system/RocketChipStageGenerator.scala
@@ -43,7 +43,6 @@ class RocketChipStage extends Stage with PreservesAll[Phase] {
 
   override def run(annotations: AnnotationSeq): AnnotationSeq = pm.transform(annotations)
 
-  // TODO: need a RunPhaseAnnotation to inject phases into ChiselStage
 }
 
 object Generator extends StageMain(new RocketChipStage)
--- a/src/main/scala/system/RocketChipStageGenerator.scala
+++ b/src/main/scala/system/RocketChipStageGenerator.scala
@@ -3,18 +3,16 @@
 package freechips.rocketchip.system
 
 import chisel3.stage.{ChiselCli, ChiselStage}
+import firrtl.AnnotationSeq
 import firrtl.options.PhaseManager.PhaseDependency
-import firrtl.options.{Dependency, Phase, PreservesAll, Shell, StageMain}
+import firrtl.options.{Dependency, Phase, PhaseManager, PreservesAll, Shell, Stage, StageMain}
 import firrtl.stage.FirrtlCli
 import freechips.rocketchip.stage.RocketChipCli
 
-class RocketChipStage extends ChiselStage with PreservesAll[Phase] {
+/** Modified ChiselStage that includes the GenerateROMs phase */
+private[freechips] final class RocketChiselStage extends ChiselStage {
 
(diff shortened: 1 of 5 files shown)


src/main/scala/system/RocketChipStageGenerator.scala
```diff
@@ -3,32 +3,46 @@
...
     Dependency[freechips.rocketchip.stage.phases.GenerateTestSuiteMakefrags],
-    Dependency[freechips.rocketchip.stage.phases.GenerateArtefacts],
+    Dependency[freechips.rocketchip.stage.phases.GenerateArtefacts]
   )
 
+  private val pm = new PhaseManager(targets)
+
+  override def run(annotations: AnnotationSeq): AnnotationSeq = pm.transform(annotations)
+
   // TODO: need a RunPhaseAnnotation to inject phases into ChiselStage
```
debs-sifive, line 46: Should this comment be taken out, or do you think this is still something to consider in the future?

seldridge: The PR can remove it, sure. TODOs are probably better tracked in issues.

debs-sifive: thanks!

PR #2482, opened 2020-05-21
BEU report Corrupt+Denied on I-Fetch

johningalls-sifive:
**Related issue**: 

**Release Notes**
add Bus Error Unit cause #1: I-Cache TileLink Bus error (was Reserved)

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -49,6 +49,7 @@ class ICacheErrors(implicit p: Parameters) extends CoreBundle()(p)
     with CanHaveErrors {
   val correctable = (cacheParams.tagCode.canDetect || cacheParams.dataCode.canDetect).option(Valid(UInt(width = paddrBits)))
   val uncorrectable = (cacheParams.itimAddr.nonEmpty && cacheParams.dataCode.canDetect).option(Valid(UInt(width = paddrBits)))
+  val bus = Valid(UInt(width = paddrBits))
 }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/rocket/ICache.scala
```diff
@@ -220,6 +221,8 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
 
     ccover(tl_out.d.bits.corrupt, "D_CORRUPT", "I$ D-channel corrupt")
   }
+  io.errors.bus.valid := tl_out.d.fire() && (tl_out.d.bits.denied || tl_out.d.bits.corrupt)
+  io.errors.bus.bits  := refill_paddr
```
aswaterman, line 225: I think we should do `refill_paddr >> blockOffBits << blockOffBits here` - in addition to shaving off six flops, it keeps things consistent with the D$, which doesn't report the block offset.

johningalls-sifive: okay will do

jerryhethatday: @aswaterman  I know it's long time ago, but I felt confused why you suggested refill_paddr >> blockOffBits << blockOffBits instead of refill_paddr,  I thought these two are exactly the same thing. What do you mean by _in addition to shaving off six flops_ Thanks

aswaterman: We only report the address of the cache line that had an error, rather than trying to be more precise.

jerryhethatday: > We only report the address of the cache line that had an error, rather than trying to be more precise.

Got it, Thanks!

PR #2483, opened 2020-05-21
Standardize IdMap and IdMapEntry

hcook:
Requests for information about the meaning and properties of certain ids used in diplomatic protocols to be stored in the Object Model necessitates having a common representation of id spaces across protocols. I refactored some common code related to displaying this information during elaboration such that an underlying abstract class and trait are now shared across the representations.

**Type of change**: refactoring

### DIFF
--- a/src/main/scala/diplomacy/Parameters.scala
+++ b/src/main/scala/diplomacy/Parameters.scala
@@ -322,3 +322,23 @@ trait DirectedBuffers[T] {
   def copyOut(x: BufferParams): T
   def copyInOut(x: BufferParams): T
 }
+
+trait IdMapEntry {
+  val name: String
+  val from: IdRange
+  val to: IdRange
+  val isCache: Boolean
+  val requestFifo: Boolean
+  def pretty(fmt: String) =
+    if (from ne to) {
+      fmt.format(to.start, to.end, from.start, from.end, s""""$name"""", if (isCache) " [CACHE]" else "", if (requestFifo) " [FIFO]" else "")
+    } else {
+      fmt.format(from.start, from.end, s""""$name"""", if (isCache) " [CACHE]" else "", if (requestFifo) " [FIFO]" else "")
+    }
+}
+
+abstract class IdMap[T <: IdMapEntry] {
+  protected val fmt: String
+  val mapping: Seq[T]
+  def pretty: String = mapping.map(_.pretty(fmt)).mkString(",\n")
+}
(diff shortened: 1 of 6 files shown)


mwachs5:
can you show an example output?

hcook:
During elaboration:
```
axi4_sys_port 0 Id mapping:
	[ 0,  1) debug,
	[ 1,  2) Core 0 ICache,
	[ 2,  3) Core 0 ICache,
	[ 3,  4) axi4_front_port ID#0,
	[ 4,  5) axi4_front_port ID#1,
	[ 5,  6) axi4_front_port ID#2,
	[ 6,  7) axi4_front_port ID#3,
	[ 7,  8) Core 0 DCache MMIO,
	[ 8,  9) <unused>,
	[ 9, 10) <unused>,
	[10, 11) <unused>,
	[11, 12) <unused>,
	[12, 13) <unused>,
	[13, 14) <unused>,
	[14, 15) <unused>,
	[15, 16) <unused>

axi4_periph_port 0 Id mapping:
	[0, 1) TLFragmenter
```

src/main/scala/diplomacy/Parameters.scala
```diff
@@ -322,3 +322,23 @@ trait DirectedBuffers[T] {
   def copyOut(x: BufferParams): T
   def copyInOut(x: BufferParams): T
 }
+
+trait IdMapEntry {
+  val name: String
+  val from: IdRange
+  val to: IdRange
+  val isCache: Boolean
+  val requestFifo: Boolean
```
jackkoenig, line 331: It doesn't matter that much, but why make these `val`s over `def`s?

src/main/scala/diplomacy/Parameters.scala
```diff
@@ -322,3 +322,23 @@ trait DirectedBuffers[T] {
...
+  def pretty(fmt: String) =
+    if (from ne to) {
```
jackkoenig, line 333: Every implementer of `IdMapEntry` that I can see is a `case class` thus using structural equality. I see this is a "pretty" method so perhaps there's good reason to use referential equality but I would suggest documenting why that is because it is not immediately apparent to me.

hcook: Because if a mapping coincidentally maps from the same values to the same values, it is still doing an identity transformation, whereas if they are literally an identical reference then no mapping was done. i will add a comment

PR #2484, opened 2020-05-22
RVV 0.9: Tighten fractional LMUL-SEW constraint

aswaterman:
The spec requires that implementations support fractional LMUL settings
for LMUL ≥ SEW/ELEN.  The old code enforced the looser constraint that
LMUL ≥ SEW/VLEN, so e.g. if VLEN were twice ELEN, we would support one
notch smaller LMUL for any given SEW.

Tightening the constraint will catch software bugs that rely on vtype
settings that aren't guaranteed to be supported across implemenetations.
The space of configurations that need to be verified is also reduced.

**Impact**: API modification


aswaterman:
Ah, thanks for the pointer.  I support tightening the spec (or clarifying, depending on your POV).

PR #2485, opened 2020-05-24
Adding src id to object model

vmurali:
**Related issue**: 

**Impact**: API modification

**Release Notes**

Added IdMap as a field in the OMPorts. IdMap contains the source ID, in addition to other fields.

### DIFF
--- a/src/main/scala/diplomacy/Parameters.scala
+++ b/src/main/scala/diplomacy/Parameters.scala
@@ -343,23 +343,3 @@ abstract class IdMap[T <: IdMapEntry] {
   def pretty: String = mapping.map(_.pretty(fmt)).mkString(",\n")
 }
 
-class IdRangeSerial (val start: Int, val end: Int)
-object IdRangeSerial {
-  def apply(i: IdRange): IdRangeSerial = {
-    new IdRangeSerial(i.start, i.end)
-  }
-}
-
-class IdMapEntrySerial (val namel: String, val froml: IdRangeSerial, val tol: IdRangeSerial, val isCachel: Boolean, val requestFifol: Boolean)
-object IdMapEntrySerial {
-  def apply[T <: IdMapEntry](i: T): IdMapEntrySerial = {
-    new IdMapEntrySerial(i.name, IdRangeSerial(i.from), IdRangeSerial(i.to), i.isCache, i.requestFifo)
-  }
-}
-
-class IdMapSerial (val mapping: Seq[IdMapEntrySerial])
-object IdMapSerial {
-  def apply[T <: IdMapEntry](i: IdMap[T]): IdMapSerial = {
-    new IdMapSerial((i.mapping).map(IdMapEntrySerial(_)))
-  }
-}
--- a/src/main/scala/diplomacy/Parameters.scala
+++ b/src/main/scala/diplomacy/Parameters.scala
@@ -342,3 +342,24 @@ abstract class IdMap[T <: IdMapEntry] {
   val mapping: Seq[T]
   def pretty: String = mapping.map(_.pretty(fmt)).mkString(",\n")
 }
+
+class IdRangeSerial private (val start: Int, val end: Int)
+object IdRangeSerial {
+  def apply(i: IdRange): IdRangeSerial = {
+    new IdRangeSerial(i.start, i.end)
+  }
+}
+
+class IdMapEntrySerial private(val namel: String, val froml: IdRangeSerial, val tol: IdRangeSerial, val isCachel: Boolean, val requestFifol: Boolean)
+object IdMapEntrySerial {
+  def apply[T <: IdMapEntry](i: T): IdMapEntrySerial = {
+    new IdMapEntrySerial(i.name, IdRangeSerial(i.from), IdRangeSerial(i.to), i.isCache, i.requestFifo)
+  }
+}
+
+class IdMapSerial private(val mapping: Seq[IdMapEntrySerial])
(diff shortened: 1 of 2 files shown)


src/main/scala/diplomacy/Parameters.scala
```diff
@@ -322,3 +322,44 @@ trait DirectedBuffers[T] {
...
+  val mapping: Seq[T]
+  def pretty: String = mapping.map(_.pretty(fmt)).mkString(",\n")
+}
+
+class IdRangeSerial (val start: Int, val end: Int)
+object IdRangeSerial {
+  def apply(i: IdRange): IdRangeSerial = {
+    new IdRangeSerial(i.start, i.end)
+  }
+}
```
mwachs5, line 351:
```suggestion
case class IdRangeSerial (val start: Int, val end: Int)

object IdRangeSerial {
  def apply(i: IdRange): IdRangeSerial = IdRangeSerial(i.start, i.end)
}
```

This makes sense as a case class to me

mwachs5: though now that I wrote that, i wonder if it will conflict with the IdRangeSerial.apply() method. 
@jackkoenig thoughts?

I guess worth asking is there a reason you did not want to make this a case class?

vmurali:
Why though?

In my head, case classes are a way to create "constructors" of another
class, by extending that other class in the definition of a case class.
Here, these three things are not entities of another type, but rather types
in their own right, which have nothing to do with each other.

Perhaps, I don't have a good sense of any other use for case classes in
Scala. For me, they are a way to encode Abstract Data Types containing
different constructors.

vmurali:
*Algebraic data types. Typo.

https://en.wikipedia.org/wiki/Algebraic_data_type

src/main/scala/diplomacy/Parameters.scala
```diff
@@ -322,3 +322,44 @@ trait DirectedBuffers[T] {
...
+    }
+}
+
+abstract class IdMap[T <: IdMapEntry] {
+  protected val fmt: String
+  val mapping: Seq[T]
+  def pretty: String = mapping.map(_.pretty(fmt)).mkString(",\n")
+}
+
+class IdRangeSerial (val start: Int, val end: Int)
```
mwachs5, line 346: Is this only to be used for the object model? If so, the pattern is to create something like "OMIDRange", similar to `OMAddressSet`: https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/diplomaticobjectmodel/model/OMAddressing.scala#L18

I'd suggest renaming these to OMIDMap and OMIDRange, respectively, and moving them to
https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/diplomaticobjectmodel/model/OMPorts.scala

Without the `_types` the OM deserialization code doesn't know how to deserialize them

src/main/scala/diplomacy/Parameters.scala
mwachs5: im going to resolve this comment since I think it's irrelevant given my more recent comment

src/main/scala/diplomacy/Parameters.scala
```diff
@@ -342,3 +342,4 @@ abstract class IdMap[T <: IdMapEntry] {
   val mapping: Seq[T]
   def pretty: String = mapping.map(_.pretty(fmt)).mkString(",\n")
 }
+
```
mwachs5, line 345:
```suggestion
```

get rid of whitespace change

src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
```diff
@@ -79,12 +79,43 @@ case class TL_C(
   val _types: Seq[String] = Seq("TL_C", "TL",  "OMProtocol")
 ) extends TL
 
+
+class OMIDRange (val start: Int,
+                 val end: Int,
+                 val _types: Seq[String] = Seq("OMIDRange", "OMCompundType"))
+object OMIDRange {
```
mwachs5, line 86:
```suggestion

object OMIDRange {
```

src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
```diff
@@ -79,12 +79,43 @@ case class TL_C(
...
+  }
+}
+
+class OMIDMapEntry(val name: String,
+                   val from: OMIDRange,
+                   val to: OMIDRange,
+                   val isCache: Boolean,
+                   val requestFifo: Boolean,
+                   val _types: Seq[String] = Seq("OMIDMapEntry", "OMCompoundType"))
+object OMIDMapEntry {
```
mwachs5, line 98:
```suggestion

object OMIDMapEntry {
```

src/main/scala/diplomaticobjectmodel/model/OMPorts.scala
```diff
@@ -79,12 +79,43 @@ case class TL_C(
...
+  def apply[T <: IdMapEntry](i: T): OMIDMapEntry = {
+    new OMIDMapEntry(i.name, OMIDRange(i.from), OMIDRange(i.to), i.isCache, i.requestFifo)
+  }
+}
+
+class OMIDMap (val mapping: Seq[OMIDMapEntry],
+               val _types: Seq[String] = Seq("OMIDMap", "OMCompoundType"))
+object OMIDMap {
```
mwachs5, line 106:
```suggestion

object OMIDMap {
```

PR #2486, opened 2020-05-25
Create RecordMap for better diplomatic I/O naming

mwachs5:
~Optionally allow HeterogeneousBag signals to be named something other than "_0", "_1"...
Also allow HeterogenousBag signals to be indexed by such names.~
Create a RecordMap as an alternative to HeterogenousBag with explicit names for the elements.

This PR does not yet use this API anywhere, but I plan to make it use it for clock bundles as a demonstration, likely in a follow-on PR (see #2487 )

**Related issue**: None

**Release Notes**

Add a RecordMap that allows names to be specified along with the Data elements for better. signal naming. This would result in verilog with different signal names if applied.

### DIFF
--- a/src/main/scala/util/HeterogeneousBag.scala
+++ b/src/main/scala/util/HeterogeneousBag.scala
@@ -6,12 +6,34 @@ import Chisel._
 import chisel3.Record
 import scala.collection.immutable.ListMap
 
-final case class HeterogeneousBag[T <: Data](elts: Seq[T]) extends Record with collection.IndexedSeq[T] {
+final case class HeterogeneousBag[T <: Data](elts: Seq[T], eltNames: Option[Seq[String]] = None)
+    extends Record with collection.IndexedSeq[T] {
+
+  eltNames.foreach{ n =>
+    require(elts.size == n.size,
+    s"If eltNames are specified, they must be specified for all elts " +
+      s" (eltNames.size = ${n.size} but elts.size = ${elts.size})")
+    require(n.distinct.size == n.size, "If eltNames are specified they must all be specified & unique")
+  }
+
+  private def nameFromIndex(i: Int): String = eltNames.map(_(i)).getOrElse(i.toString)
+
+  private def nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
+    val name = nameFromIndex(i)
+    name -> elt
+  }:_*)
+
   def apply(x: Int) = elts(x)
+
+  def apply(s: String) = nameToElts(s)
+
   def length = elts.length
 
-  val elements = ListMap(elts.zipWithIndex.map { case (n,i) => (i.toString, n) }:_*)
-  override def cloneType: this.type = (new HeterogeneousBag(elts.map(_.chiselCloneType))).asInstanceOf[this.type]
+  val elements = ListMap(elts.zipWithIndex.map { case (elt, i) =>
+    (nameFromIndex(i), elt)
+  }:_*)
+
+  override def cloneType: this.type = (new HeterogeneousBag(elts.map(_.chiselCloneType), eltNames)).asInstanceOf[this.type]
 
   // IndexedSeq has its own hashCode/equals that we must not use
   override def hashCode: Int = super[Record].hashCode
--- a/src/main/scala/util/HeterogeneousBag.scala
+++ b/src/main/scala/util/HeterogeneousBag.scala
@@ -18,7 +18,7 @@ final case class HeterogeneousBag[T <: Data](elts: Seq[T], eltNames: Option[Seq[
 
   private def nameFromIndex(i: Int): String = eltNames.map(_(i)).getOrElse(i.toString)
 
-  private def nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
+  private val nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
     val name = nameFromIndex(i)
     name -> elt
   }:_*)
--- a/src/main/scala/util/HeterogeneousBag.scala
+++ b/src/main/scala/util/HeterogeneousBag.scala
@@ -18,7 +18,7 @@ final case class HeterogeneousBag[T <: Data](elts: Seq[T], eltNames: Option[Seq[
 
   private def nameFromIndex(i: Int): String = eltNames.map(_(i)).getOrElse(i.toString)
 
-  private val nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
+  private def nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
     val name = nameFromIndex(i)
     name -> elt
   }:_*)
--- a/src/main/scala/util/HeterogeneousBag.scala
+++ b/src/main/scala/util/HeterogeneousBag.scala
@@ -6,34 +6,12 @@ import Chisel._
 import chisel3.Record
 import scala.collection.immutable.ListMap
 
-final case class HeterogeneousBag[T <: Data](elts: Seq[T], eltNames: Option[Seq[String]] = None)
-    extends Record with collection.IndexedSeq[T] {
-
-  eltNames.foreach{ n =>
-    require(elts.size == n.size,
-    s"If eltNames are specified, they must be specified for all elts " +
-      s" (eltNames.size = ${n.size} but elts.size = ${elts.size})")
-    require(n.distinct.size == n.size, "If eltNames are specified they must all be specified & unique")
-  }
-
-  private def nameFromIndex(i: Int): String = eltNames.map(_(i)).getOrElse(i.toString)
-
-  private def nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
-    val name = nameFromIndex(i)
-    name -> elt
-  }:_*)
-
+final case class HeterogeneousBag[T <: Data](elts: Seq[T]) extends Record with collection.IndexedSeq[T] {
   def apply(x: Int) = elts(x)
-
-  def apply(s: String) = nameToElts(s)
-
   def length = elts.length
 
-  val elements = ListMap(elts.zipWithIndex.map { case (elt, i) =>
-    (nameFromIndex(i), elt)
-  }:_*)
-
-  override def cloneType: this.type = (new HeterogeneousBag(elts.map(_.chiselCloneType), eltNames)).asInstanceOf[this.type]
+  val elements = ListMap(elts.zipWithIndex.map { case (n,i) => (i.toString, n) }:_*)
+  override def cloneType: this.type = (new HeterogeneousBag(elts.map(_.chiselCloneType))).asInstanceOf[this.type]
 
   // IndexedSeq has its own hashCode/equals that we must not use
   override def hashCode: Int = super[Record].hashCode
--- a/src/main/scala/util/RecordListMap.scala
+++ b/src/main/scala/util/RecordListMap.scala
@@ -18,7 +18,7 @@ final case class RecordListMap[T <: Data](eltMap: ListMap[String, T])
   // Required by Record
   val elements = eltMap
 
-  override def cloneType: this.type = (new RecordMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
+  override def cloneType: this.type = (new RecordListMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
 
   // Map has its own hashCode/equals that we must not use
   override def hashCode: Int = super[Record].hashCode
--- a/src/main/scala/util/RecordListMap.scala
+++ b/src/main/scala/util/RecordListMap.scala
@@ -1,0 +1,26 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import Chisel._
+import chisel3.Record
+import scala.collection.immutable.ListMap
+
+final case class RecordListMap[T <: Data](eltMap: ListMap[String, T])
+    extends Record with collection.Map[String, T] {
+
+  // Required by collection.Map[String, T]
+  def get(key: String) = eltMap.get(key)
+  def -(key: String) = eltMap - key
+  def +[V1 >: T](kv: (String, V1)) = eltMap + kv
+  def iterator = eltMap.iterator
+
+  // Required by Record
+  val elements = eltMap
+
+  override def cloneType: this.type = (new RecordMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
+
+  // Map has its own hashCode/equals that we must not use
+  override def hashCode: Int = super[Record].hashCode
+  override def equals(that: Any): Boolean = super[Record].equals(that)
+}
--- a/src/main/scala/util/RecordMap.scala
+++ b/src/main/scala/util/RecordMap.scala
@@ -7,19 +7,18 @@ import scala.collection.immutable.ListMap
 import chisel3.internal.requireIsChiselType
 import chisel3.experimental.DataMirror.internal.chiselTypeClone
 
-final class RecordMap[T <: Data](val eltMap: ListMap[String, T])
+final class RecordMap[T <: Data](eltMap: ListMap[String, T])
     extends Record {
   
   eltMap.foreach { case (name, elt) => requireIsChiselType(elt, name) }
 
-
   def apply(x: Int) = eltMap.values.toSeq(x)
   def apply(x: String) = eltMap.get(x)
   def size = eltMap.size
-  def data = eltMap.values
 
   // This is needed for Record, and doesn't give the actual elements
-  val elements = ListMap[String, T]() ++ eltMap.mapValues(chiselTypeClone(_))  // mapValues return value is lazy
+  val elements = ListMap[String, T]() ++ eltMap.mapValues(chiselTypeClone(_).asInstanceOf[T])  // mapValues return value is lazy
+  def data = elements.values
 
   override def cloneType: this.type = (new RecordMap(eltMap)).asInstanceOf[this.type]
 
--- a/src/main/scala/util/RecordMap.scala
+++ b/src/main/scala/util/RecordMap.scala
@@ -18,7 +18,7 @@ final case class RecordMap[T <: Data](eltMap: ListMap[String, T])
   // Required by Record
   val elements = eltMap
 
-  override def cloneType: this.type = (new RecordListMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
+  override def cloneType: this.type = (new RecordMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
 
   // Map has its own hashCode/equals that we must not use
   override def hashCode: Int = super[Record].hashCode
--- a/src/main/scala/util/RecordMap.scala
+++ b/src/main/scala/util/RecordMap.scala
@@ -12,12 +12,12 @@ final class RecordMap[T <: Data] private (eltMap: ListMap[String, T])
   
   eltMap.foreach { case (name, elt) => requireIsChiselType(elt, name) }
 
+  // This is needed for Record
+  val elements = ListMap[String, T]() ++ eltMap.mapValues(chiselTypeClone(_).asInstanceOf[T])  // mapValues return value is lazy
+
   def apply(x: Int) = elements.values.toSeq(x)
   def apply(x: String) = elements.get(x)
   def size = elements.size
-
-  // This is needed for Record, and doesn't give the actual elements
-  val elements = ListMap[String, T]() ++ eltMap.mapValues(chiselTypeClone(_).asInstanceOf[T])  // mapValues return value is lazy
   def data = elements.values
 
   override def cloneType: this.type = (new RecordMap(eltMap)).asInstanceOf[this.type]
--- a/src/main/scala/util/RecordMap.scala
+++ b/src/main/scala/util/RecordMap.scala
@@ -7,7 +7,7 @@ import scala.collection.immutable.ListMap
 import chisel3.internal.requireIsChiselType
 import chisel3.experimental.DataMirror.internal.chiselTypeClone
 
-final class RecordMap[T <: Data](eltMap: ListMap[String, T])
+final class RecordMap[T <: Data] private (eltMap: ListMap[String, T])
     extends Record {
   
   eltMap.foreach { case (name, elt) => requireIsChiselType(elt, name) }
--- a/src/main/scala/util/RecordMap.scala
+++ b/src/main/scala/util/RecordMap.scala
@@ -12,9 +12,9 @@ final class RecordMap[T <: Data](eltMap: ListMap[String, T])
   
   eltMap.foreach { case (name, elt) => requireIsChiselType(elt, name) }
 
-  def apply(x: Int) = eltMap.values.toSeq(x)
-  def apply(x: String) = eltMap.get(x)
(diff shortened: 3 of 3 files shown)


src/main/scala/util/HeterogeneousBag.scala
```diff
@@ -6,12 +6,34 @@ import Chisel._
...
+  eltNames.foreach{ n =>
+    require(elts.size == n.size,
+    s"If eltNames are specified, they must be specified for all elts " +
+      s" (eltNames.size = ${n.size} but elts.size = ${elts.size})")
+    require(n.distinct.size == n.size, "If eltNames are specified they must all be specified & unique")
+  }
+
+  private def nameFromIndex(i: Int): String = eltNames.map(_(i)).getOrElse(i.toString)
+
+  private def nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
```
hcook, line 21: I would make this a `val` so that you aren't building a new ListMap object every time you `apply(s: String)`

mwachs5, line 21:
```suggestion
  private val nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
```

mwachs5: done, thanks

src/main/scala/util/HeterogeneousBag.scala
```diff
@@ -6,12 +6,34 @@ import Chisel._
 import chisel3.Record
 import scala.collection.immutable.ListMap
 
-final case class HeterogeneousBag[T <: Data](elts: Seq[T]) extends Record with collection.IndexedSeq[T] {
+final case class HeterogeneousBag[T <: Data](elts: Seq[T], eltNames: Option[Seq[String]] = None)
```
hcook, line 9: I would probably make `eltNames: Seq[String]] = Nil` and then replace the `eltNames.foreach`/`eltNames.map` below with `if (eltNames.nonEmpty)`. I say this because AFAICT `Some(Nil)` is not a meaningful case that you want to allow, so it clarifies things to just check emptiness.

src/main/scala/util/HeterogeneousBag.scala
```diff
@@ -6,12 +6,34 @@ import Chisel._
...
+    s"If eltNames are specified, they must be specified for all elts " +
+      s" (eltNames.size = ${n.size} but elts.size = ${elts.size})")
+    require(n.distinct.size == n.size, "If eltNames are specified they must all be specified & unique")
+  }
+
+  private def nameFromIndex(i: Int): String = eltNames.map(_(i)).getOrElse(i.toString)
+
+  private val nameToElts: Map[String, T] = ListMap(elts.zipWithIndex.map { case (elt, i) =>
+    val name = nameFromIndex(i)
+    name -> elt
```
hcook, line 23: nit: no need for a `val` you only use once. i would just `(nameFromIndex(i), elt)` like you did below... and actually i would just use `nameToElts` there directly

src/main/scala/util/HeterogeneousBag.scala
```diff
@@ -6,12 +6,34 @@ import Chisel._
...
+
   def apply(x: Int) = elts(x)
+
+  def apply(s: String) = nameToElts(s)
+
   def length = elts.length
 
-  val elements = ListMap(elts.zipWithIndex.map { case (n,i) => (i.toString, n) }:_*)
-  override def cloneType: this.type = (new HeterogeneousBag(elts.map(_.chiselCloneType))).asInstanceOf[this.type]
+  val elements = ListMap(elts.zipWithIndex.map { case (elt, i) =>
```
hcook, line 32: are `elements` and `nameToElts` the same thing?

src/main/scala/util/HeterogeneousBag.scala
mwachs5: yeah this was probably some intermediate debugging step, will clean it up

src/main/scala/util/HeterogeneousBag.scala
mwachs5: hmm, maybe! I was just replacing the string code but seems like you're right.

src/main/scala/util/HeterogeneousBag.scala
```diff
@@ -6,12 +6,34 @@ import Chisel._
 import chisel3.Record
 import scala.collection.immutable.ListMap
 
-final case class HeterogeneousBag[T <: Data](elts: Seq[T]) extends Record with collection.IndexedSeq[T] {
+final case class HeterogeneousBag[T <: Data](elts: Seq[T], eltNames: Option[Seq[String]] = None)
+    extends Record with collection.IndexedSeq[T] {
```
jackkoenig, line 10: I know it was already like this, but mixing in `IndexedSeq` here is not true because "Indexed sequences support constant-time or near constant-time element access and length computation." (https://www.scala-lang.org/api/2.12.10/scala/collection/IndexedSeq.html). It's not a huge deal, just misleading about performance, I think it would be better to change this to extending `Seq`.

jackkoenig: The observation that this is not upholding the API is that `apply: Int => T` and `length` below both delegate to `elts` which is itself just a `Seq`.

jackkoenig:
Is it better to add this optional API to HeterogeneousBag vs. creating a new type? I'm guessing the issue is that we hardcode HetBag throughout rocket-chip, but I believe this is more of a historic need than something we should have now. 

IIRC, HetBag originally existed to make it possible to construct IOs from Chisel objects that _already existed and were already connected to_. This is because it predates `MultiIOModule` and `RawModule` so it made it possible to use the cake pattern, connecting to IOs, and yet later group them up into a single HetBag `val io`. This pattern is now much better served by using `MultiIOModule` and even `chisel3.experimental.IO` for arbitrary programmatic `IO` materialization. I suspect HetBag is no longer serving it's original purpose, but it remains pervasive in the API.

My suspicion, is that we simply should stop mandating `HetBag` throughout, instead using merely `T <: Data`, then provide a different subclass of `Record` that has user-provided names, perhaps `RecordMap` or `CustomBundle`. This type could conceivably even be integrated into `chisel3.util`, there is a more modern of `HeterogeneousBag` there called [`MixedVec`](https://github.com/freechipsproject/chisel3/blob/387274784115bee2bf7167547a7ee459219e1413/src/main/scala/chisel3/util/MixedVec.scala).

jackkoenig:
**TLDR** of my above, this PR is kind of like adding a `names` field to a type like `List` because you want the `List` to act like a `Map`, when I think the better approach would be to use a different data structure, like a `Map`.

mwachs5:
I don't think there is any mandation (?) of using `HeterogenousBag`. Now that I implemented an example, I realize the different diplomatic classes choose to use it for creating their `Bundle`s, but since those diplomatic classes have to be modified to pass names in anyway, you're right that it could just do something other than putting down a HeterogenousBag. As long as the API for whatever it is you do with the HeterogenousBag (e.g. connecting to some other Seq with `zip`s and assuming stuff is in the expected order) still does the right thing. 

I thought about creating a new class but seemed like that would be less DRY than adding to this, but it does seem awkward to pass in two lists of things instead of a Map, like you say.

hcook:
Thinking back, my original concept was indeed to make a new class, but that got lost somewhere in the jira creation process. I agree with @jackkoenig 's feedback about making this a potentially upstream-able contribution, and @mwachs5 is right that we can roll it out as we use it.

@jackkoenig A related refactoring that would "clean up" a lot of usages is to switch things to [makeIOs](https://github.com/chipsalliance/rocket-chip/blob/8acb22459a2aa9ea8dc70d9c4472e7e5d5177ff8/src/main/scala/diplomacy/Nodes.scala#L643) while also changing the implementation of that. That change plus relaxing some arguments to be `Record` would be a good followup.

mwachs5:
Ok, by "upstreamable" you mean "part of chisel itself"?
I *think* the concrete suggestion is to make a new subclass of `Record`, `RecordMap[T]`.

Isn't it important to the API that the items are still ordered properly so they can be zipped up together? So should this be explicitly `RecordListMap[T]`?

src/main/scala/util/HeterogeneousBag.scala
mwachs5: ok, this may become irrelevant if i end up making a new class, do you still think i should fix this here?

src/main/scala/util/HeterogeneousBag.scala
mwachs5: this was all reverted

src/main/scala/util/HeterogeneousBag.scala
mwachs5: this was all reverted

src/main/scala/util/HeterogeneousBag.scala
mwachs5: this wasall reverted

src/main/scala/util/RecordListMap.scala
```diff
@@ -0,0 +1,26 @@
...
+  // Required by collection.Map[String, T]
+  def get(key: String) = eltMap.get(key)
+  def -(key: String) = eltMap - key
+  def +[V1 >: T](kv: (String, V1)) = eltMap + kv
+  def iterator = eltMap.iterator
+
+  // Required by Record
+  val elements = eltMap
+
+  override def cloneType: this.type = (new RecordMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
```
mwachs5, line 21:
```suggestion
  override def cloneType: this.type = (new RecordListMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
```

hcook:
Yes, I meant "upstream into chisel itself"

I think further enforcement of ordering is never a bad choice, so ListMap seems appropriate.

It it interesting to think about how one might use the `+` and `-` operators to generate new Records as a result of a circuit that does some filtering or adds extra information.

Also, as future work, I think we could consider using [selectDynamic](https://stackoverflow.com/questions/15799811/how-does-type-dynamic-work-and-how-to-use-it) to provide `.` operator accessors to the members

src/main/scala/util/RecordListMap.scala
```diff
@@ -0,0 +1,26 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import Chisel._
+import chisel3.Record
+import scala.collection.immutable.ListMap
+
+final case class RecordListMap[T <: Data](eltMap: ListMap[String, T])
```
jackkoenig, line 9: A bit of a nit, but in anticipation of [`SeqMap` in Scala 2.13](https://www.scala-lang.org/api/2.13.2/scala/collection/SeqMap.html) with the faster `VectorMap` concrete implementation (vs. slower `ListMap`), I think it might better to call this `RecordSeqMap`. The type of `elements` will remain `ListMap`, but we could change it to `SeqMap` once we upgrade to 2.13

jackkoenig: Alternatively this can be `ListMap` and we can introduce a new one in rocket-chip Scala 2.13

src/main/scala/util/RecordListMap.scala
```diff
@@ -0,0 +1,26 @@
...
+    extends Record with collection.Map[String, T] {
+
+  // Required by collection.Map[String, T]
+  def get(key: String) = eltMap.get(key)
+  def -(key: String) = eltMap - key
+  def +[V1 >: T](kv: (String, V1)) = eltMap + kv
+  def iterator = eltMap.iterator
```
jackkoenig, line 16: Also what do these mean in the context of `RecordListMap` extending `Data`? The existence of these methods imply some utility but they just return to you a `ListMap[String, T]` that means what exactly?

I think we should either return a new `RecordListMap` when calling these operations, or we should not extend `collection.Map`.

src/main/scala/util/RecordListMap.scala
mwachs5: should I just call it "RecordMap"?

src/main/scala/util/RecordListMap.scala
mwachs5: to be clear, you're fine with `get` just returning `T`, but the `-` and `+` should return RecordListMap[T], right? what should the iterator type be?

src/main/scala/util/RecordListMap.scala
aswaterman: RecordMap would be my preference - we can just make it part of the API description that it's ordered

mwachs5: I renamed this to RecordMap

src/main/scala/util/RecordListMap.scala
mwachs5: I updated the + and -, but not sure about the iterator

mwachs5: nope, that was wrong. Struggling with the contravariance (?) required of the `+` operator. 

Should I just NOT extend the Map API?

jackkoenig: > Should I just NOT extend the Map API?

I think that would be the easiest thing to do, just implement whatever methods you think are useful.

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,27 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import Chisel._
+import chisel3.Record
+import scala.collection.immutable.ListMap
+
+final case class RecordMap[T <: Data](eltMap: ListMap[String, T])
+    extends Record with collection.Seq[T] {
```
jackkoenig, line 10: I don't think we should extend `Seq`, I think the only thing we should extend is `Record`. `Seq` is less problematic than `Map`, but there is the issue that it may be surprising.

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,27 @@
...
+  def length = eltMap.size
+  def iterator = eltMap.values.iterator
+
+  def apply(x: String) = eltMap.get(x)
+
+  // Required by Record
+  val elements = eltMap
+
+  override def cloneType: this.type = (new RecordMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
+
```
jackkoenig, line 23:
```suggestion
  import chisel3.internal.requireIsChiselType
  import chisel3.experimental.internal.chiselTypeClone
  
  eltMap.foreach { case (name, elt) => requireIsChiselType(elt, name) }

  val elements = ListMap() ++ eltMap.mapValues(chiselTypeClone) // mapValues return value is lazy

  override def cloneType: this.type = (new RecordMap(eltMap)).asInstanceOf[this.type]

```

I know this is following the style of `HeterogeneousBag`, but I think this is wrong. In general, custom `Record` implementations should provide fresh types to `elements` rather than the ones passed by the user. This is because the `elements` must be _bound in place_ by Chisel, ie. they will be mutated. `HeterogeneousBag` did the "copy on clone" technique because it was intentionally allowing things to be used as hardware before the HeterogeneousBag was created. This was because all ports had to be packed into `val io`. Since that is no longer true, this pattern should no longer be used. Furthermore, the pattern won't work in `import chisel3._` anyway.

This suggestion also implies that we could make the `eltMap` argument an `Iterable[String, T]` and then provide a companion object with a varargs version as well.

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,27 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import Chisel._
+import chisel3.Record
+import scala.collection.immutable.ListMap
+
+final case class RecordMap[T <: Data](eltMap: ListMap[String, T])
```
jackkoenig, line 9: This should not be a `case class` because Chisel types do not conform to structural equality.

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,27 @@
...
+  def apply(x: String) = eltMap.get(x)
+
+  // Required by Record
+  val elements = eltMap
+
+  override def cloneType: this.type = (new RecordMap(eltMap.map{case (k, v) => k -> v.chiselCloneType})).asInstanceOf[this.type]
+
+  // Map has its own hashCode/equals that we must not use
+  override def hashCode: Int = super[Record].hashCode
+  override def equals(that: Any): Boolean = super[Record].equals(that)
```
jackkoenig, line 26: I see that you're fixing the `case class` issue down here, but I think better is to make this not be a case class and then delete both of these.

src/main/scala/util/RecordMap.scala
mwachs5: so the common pattern is to want to zip these things together. Can I still do that if it doesnt' extend Seq?

mwachs5: though there is an argument to be made that zipping them up sequentially (vs pairing them by name) is exactly the buggy case we want to avoid... thoughts @hcook ?

jackkoenig: `.elements` is of type `ListMap` which implements `Iterable` so you can always `zip` those, eg.
```scala
whatever.zip(myRecordMap.elements)
```

src/main/scala/util/RecordMap.scala
mwachs5: I tried the Iterable[String, T] but then that was annoying to write an accessor by the string name, which I assume we would want at some point. I still wrote the varargs version.

I did not use your suggestion directly, but I believe I captured the request... please take a look.

This doesnt work in my use case however, I get an exception on the chiselTypeClone call. WIll point it out on #2487

src/main/scala/util/RecordMap.scala
mwachs5: removed the case class

src/main/scala/util/RecordMap.scala
mwachs5: got rid of the case class

src/main/scala/util/RecordMap.scala
mwachs5: To be clear... `elements` is not the REAL elements, right? Given your suggestion below, elements returns some sort of cloned type version thing, wheras eltMap.values gives... the real data? 

I wrote a `val data = eltMap.values` function separately from `elements` because it seemed more correct/different for the use case when you actually want to connect to the bundle, vs clone its type. But I'm having difficulty getting this right.

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  override def cloneType: this.type = (new RecordMap(eltMap)).asInstanceOf[this.type]
+
+}
+
+object RecordMap {
+
+  def apply[T <: Data](eltMap: ListMap[String, T]) = new RecordMap(eltMap)
+
+  def apply[T <: Data](elements: (String, T)*) {
```
hcook, line 31: `def apply[T <: Data](elements: (String, T)*): RecordMap[T] = {`?

mwachs5: sorry are you just suggesting that I add the return type to the function declaration? If so I agree

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,34 @@
...
+  val elements = ListMap[String, T]() ++ eltMap.mapValues(chiselTypeClone(_).asInstanceOf[T])  // mapValues return value is lazy
+  def data = elements.values
+
+  override def cloneType: this.type = (new RecordMap(eltMap)).asInstanceOf[this.type]
+
+}
+
+object RecordMap {
+
+  def apply[T <: Data](eltMap: ListMap[String, T]) = new RecordMap(eltMap)
```
mwachs5, line 29:
```suggestion
  def apply[T <: Data](eltMap: ListMap[String, T]): RecordMap[T] = new RecordMap(eltMap)
```

src/main/scala/util/RecordMap.scala
mwachs5, line 31:
```suggestion
  def apply[T <: Data](elements: (String, T)*): RecordMap[T] =  {
```

oh are you pointing out that I am just missing an `=` and this is a pointless, untested bit of code... yes.

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,34 @@
...
+import chisel3.experimental.DataMirror.internal.chiselTypeClone
+
+final class RecordMap[T <: Data](eltMap: ListMap[String, T])
+    extends Record {
+  
+  eltMap.foreach { case (name, elt) => requireIsChiselType(elt, name) }
+
+  def apply(x: Int) = eltMap.values.toSeq(x)
+  def apply(x: String) = eltMap.get(x)
+  def size = eltMap.size
```
jackkoenig, line 17:
```suggestion
  def apply(x: Int) = elements.values.toSeq(x)
  def apply(x: String) = elements.get(x)
  def size = elements.size
```

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,34 @@
...
+
+package freechips.rocketchip.util
+
+import chisel3._
+import scala.collection.immutable.ListMap
+import chisel3.internal.requireIsChiselType
+import chisel3.experimental.DataMirror.internal.chiselTypeClone
+
+final class RecordMap[T <: Data](eltMap: ListMap[String, T])
+    extends Record {
```
jackkoenig, line 11:
```suggestion
final class RecordMap[T <: Data] private (eltMap: ListMap[String, T])
    extends Record {
```
It's not strictly necessary to make the constructor private, but if we ever wanted to refactor `RecordMap` into an interface with different implementations (eg. in Scala 2.13 add one that uses `VectorMap` instead of `ListMap`), this would make that refactoring easier.

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,34 @@
...
+  override def cloneType: this.type = (new RecordMap(eltMap)).asInstanceOf[this.type]
+
+}
+
+object RecordMap {
+
+  def apply[T <: Data](eltMap: ListMap[String, T]) = new RecordMap(eltMap)
+
+  def apply[T <: Data](elements: (String, T)*) {
+    new RecordMap[T](ListMap[String, T](elements:_*))
```
jackkoenig, line 32: Are the type parameters needed here? Does the following now just work?
```scala
new RecordMap(ListMap(elements:_*))
```

src/main/scala/util/RecordMap.scala
mwachs5: so then you coul donly make this with the companion object, is the idea?

src/main/scala/util/RecordMap.scala
mwachs5: it probably just works. will try

src/main/scala/util/RecordMap.scala
jackkoenig: Exactly. Doing it via the companion object (aka a factory method) has a huge advantage that you can change the concrete subtype without breaking source and binary compatibility. So if we later wanted to make `RecordMap` a `trait`, that would break every use of `new RecordMap`, but the factory method would be just fine.

jackkoenig: Also if we wanted to do anything before or after object construction, you can do that in a factory method, can't do that in a constructor. Basically the default should always be private constructors.

src/main/scala/util/RecordMap.scala
```diff
@@ -0,0 +1,34 @@
...
+final class RecordMap[T <: Data] private (eltMap: ListMap[String, T])
+    extends Record {
+  
+  eltMap.foreach { case (name, elt) => requireIsChiselType(elt, name) }
+
+  def apply(x: Int) = elements.values.toSeq(x)
+  def apply(x: String) = elements.get(x)
+  def size = elements.size
+
+  // This is needed for Record, and doesn't give the actual elements
```
jackkoenig, line 19: Does this comment still apply? `val elements` are indeed the actual elements.

PR #2487, opened 2020-05-26
PRCI: use new RecordMap API to name signals

mwachs5:
**Related issue**: This builds on top of #2486 and will be rebased on top of it as that API changes.

This demonstrates the use of the new RecordListMap. API to get better names for clock group signals

In the comments I'll include a diff before and after this change.

**Type of change**:feature request

**Impact**: API modification

(Generated RTL will have different output)

**Release Notes**

I/O and internal signals in the clock related modules (ClockSource, ClockGroup, etc) will have more meaninguful I/O names vs simply indexed names.

### DIFF
--- a/src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
+++ b/src/main/scala/diplomaticobjectmodel/logicaltree/LogicalTreeNode.scala
@@ -36,7 +36,7 @@ object LogicalModuleTree {
 
   def rootLogicalTreeNode: LogicalTreeNode = {
     val roots = tree.collect { case (k, _) if !tree.exists(_._2.contains(k)) => k }
-    assert(roots.size == 1, "Logical Tree contains more than one root.")
+    require(roots.size == 1, s"Logical Tree contains more than one root:\n$roots")
     roots.head
   }
 
--- a/src/main/scala/groundtest/Configs.scala
+++ b/src/main/scala/groundtest/Configs.scala
@@ -5,36 +5,67 @@ package freechips.rocketchip.groundtest
 
 import Chisel._
 import freechips.rocketchip.config.Config
+import freechips.rocketchip.devices.tilelink.{CLINTKey, PLICKey}
+import freechips.rocketchip.devices.debug.{DebugModuleKey}
 import freechips.rocketchip.subsystem._
 import freechips.rocketchip.system.BaseConfig
 import freechips.rocketchip.rocket.{DCacheParams}
-import freechips.rocketchip.tile.{MaxHartIdBits, XLen}
+import freechips.rocketchip.tile.{XLen}
 
 /** Actual testing target Configs */
 
-class TraceGenConfig extends Config(new WithTraceGen(List.fill(2){ DCacheParams(nSets = 16, nWays = 1) }) ++ new BaseConfig)
+class TraceGenConfig extends Config(
+  new WithTraceGen(2)() ++
+  new GroundTestBaseConfig
+)
 
-class TraceGenBufferlessConfig extends Config(new WithBufferlessBroadcastHub ++ new TraceGenConfig)
+class TraceGenBufferlessConfig extends Config(
+  new WithBufferlessBroadcastHub ++
+  new TraceGenConfig
+)
 
 /* Composable Configs to set individual parameters */
 
-class WithTraceGen(params: Seq[DCacheParams], nReqs: Int = 8192) extends Config((site, here, up) => {
-  case GroundTestTilesKey => params.map { dcp => TraceGenParams(
-    dcache = Some(dcp),
-    wordBits = site(XLen),
-    addrBits = 32,
-    addrBag = {
-      val nSets = dcp.nSets
-      val nWays = dcp.nWays
-      val blockOffset = site(SystemBusKey).blockOffset
-      val nBeats = site(SystemBusKey).blockBeats
-      List.tabulate(nWays) { i =>
-        Seq.tabulate(nBeats) { j => BigInt((j * 8) + ((i * nSets) << blockOffset)) }
-      }.flatten
-    },
-    maxRequests = nReqs,
-    memStart = site(ExtMem).get.master.base,
-    numGens = params.size)
+class GroundTestBaseConfig extends Config(
+  new BaseConfig().alter((site,here,up) => {
+    case DebugModuleKey => None
+    case CLINTKey => None
+    case PLICKey => None
+  })
+)
+
+class WithTraceGen(
+  n: Int = 2,
+  overrideIdOffset: Option[Int] = None,
+  overrideMemOffset: Option[BigInt] = None)(
+  params: Seq[DCacheParams] = List.fill(n){ DCacheParams(nSets = 16, nWays = 1) },
+  nReqs: Int = 8192
+) extends Config((site, here, up) => {
+  case TilesLocated(InSubsystem) => {
+    val prev = up(TilesLocated(InSubsystem), site)
+    val idOffset = overrideIdOffset.getOrElse(prev.size)
+    val memOffset: BigInt = overrideMemOffset.orElse(site(ExtMem).map(_.master.base)).getOrElse(0x0L)
+    params.zipWithIndex.map { case (dcp, i) =>
+      TraceGenTileAttachParams(
+        tileParams = TraceGenParams(
+          hartId = i + idOffset,
+          dcache = Some(dcp),
+          wordBits = site(XLen),
+          addrBits = 32,
+          addrBag = {
+            val nSets = dcp.nSets
+            val nWays = dcp.nWays
+            val blockOffset = site(SystemBusKey).blockOffset
+            val nBeats = site(SystemBusKey).blockBeats
+            List.tabulate(nWays) { i =>
+              Seq.tabulate(nBeats) { j => BigInt((j * 8) + ((i * nSets) << blockOffset)) }
+            }.flatten
+          },
+          maxRequests = nReqs,
+          memStart = memOffset,
+          numGens = params.size),
+        crossingParams = RocketCrossingParams()
+      )
+    } ++ prev
   }
-  case MaxHartIdBits => log2Up(params.size)
 })
--- a/src/main/scala/groundtest/GroundTestSubsystem.scala
+++ b/src/main/scala/groundtest/GroundTestSubsystem.scala
@@ -3,29 +3,23 @@
 package freechips.rocketchip.groundtest
 
 import Chisel._
-import freechips.rocketchip.config.{Field, Parameters}
-import freechips.rocketchip.diplomacy._
-import freechips.rocketchip.diplomaticobjectmodel.model.OMInterrupt
-import freechips.rocketchip.interrupts._
-import freechips.rocketchip.subsystem._
-import freechips.rocketchip.tilelink._
-import freechips.rocketchip.tile._
+import chisel3.dontTouch
+import freechips.rocketchip.config.{Parameters}
+import freechips.rocketchip.diplomacy.{AddressSet, LazyModule}
+import freechips.rocketchip.interrupts.{IntSinkNode, IntSinkPortSimple}
+import freechips.rocketchip.subsystem.{BaseSubsystem, BaseSubsystemModuleImp, HasTiles, CanHaveMasterAXI4MemPort}
+import freechips.rocketchip.tilelink.{TLRAM, TLFragmenter}
 
-import scala.math.max
-
-case object TileId extends Field[Int]
-
-class GroundTestSubsystem(implicit p: Parameters) extends BaseSubsystem
-    with CanHaveMasterAXI4MemPort {
-  val tileParams = p(GroundTestTilesKey)
-  val tiles = tileParams.zipWithIndex.map { case(c, i) => LazyModule(c.build(i, p)) }
-
-  tiles.map(_.masterNode).foreach { m =>
-    sbus.fromTile(None, buffer = BufferParams.default){ m }
-  }
-
+class GroundTestSubsystem(implicit p: Parameters)
+  extends BaseSubsystem
+  with HasTiles
+  with CanHaveMasterAXI4MemPort
+{
   val testram = LazyModule(new TLRAM(AddressSet(0x52000000, 0xfff), beatBytes=pbus.beatBytes))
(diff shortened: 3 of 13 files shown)


src/main/scala/prci/ClockNodes.scala
```diff
@@ -73,7 +73,7 @@ object ClockSourceNode
 object ClockGroupImp extends SimpleNodeImp[ClockGroupSourceParameters, ClockGroupSinkParameters, ClockGroupEdgeParameters, ClockGroupBundle]
 {
   def edge(pd: ClockGroupSourceParameters, pu: ClockGroupSinkParameters, p: Parameters, sourceInfo: SourceInfo) = ClockGroupEdgeParameters(pd, pu, p, sourceInfo)
-  def bundle(e: ClockGroupEdgeParameters) = new ClockGroupBundle(e.bundle)
+  def bundle(e: ClockGroupEdgeParameters) = new ClockGroupBundle(e.bundle, Some(e.bundle.members.zipWithIndex.map{ case (_, i) => s"${e.sink.name}_${i}"}))
```
mwachs5, line 76: @hcook I am right that only the `sink` has a name, right? there is no `source` so I can't do something like "source_to_sink" right?

mwachs5:
attaching the diff for DefaultConfig:
[diff.txt](https://github.com/chipsalliance/rocket-chip/files/4684320/diff.txt)

For example, old:

```
module ClockGroupAggregator_1( // @[:freechips.rocketchip.system.DefaultConfig.fir@28996.2]
  input   auto_in_member_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@28997.4]
  input   auto_in_member_0_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@28997.4]
  output  auto_out_member_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@28997.4]
  output  auto_out_member_0_reset // @[:freechips.rocketchip.system.DefaultConfig.fir@28997.4]
);
```

new:
```
module ClockGroupAggregator( // @[:freechips.rocketchip.system.DefaultConfig.fir@44.2]
  input   auto_in_member_subsystem_sbus_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_subsystem_sbus_0_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_0_member_subsystem_sbus_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_0_member_subsystem_sbus_0_reset // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
);
```

mwachs5:
old:
```
module ClockGroupAggregator( // @[:freechips.rocketchip.system.DefaultConfig.fir@44.2]
  input   auto_in_member_5_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_5_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_4_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_4_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_3_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_3_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_2_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_2_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_1_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_1_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  input   auto_in_member_0_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_3_member_1_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_3_member_1_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_3_member_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_3_member_0_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_2_member_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_2_member_0_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_1_member_1_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_1_member_1_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_1_member_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_1_member_0_reset, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_0_member_0_clock, // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
  output  auto_out_0_member_0_reset // @[:freechips.rocketchip.system.DefaultConfig.fir@45.4]
);
```

new:

...? Actually not sure what happened to this one in the new world.

mwachs5:
so maybe I screwed something up here...

src/main/scala/prci/ClockNodes.scala
hcook: Correct, though such functionality would be an argument for adding such a parameter. The source-side parameters are very underdeveloped. We could also think about adding the NodePath functionality that some of the other protocols have, then  lieu of a parameterized name you might be able to get the ValName of the source node or the module name of it's parent.

hcook: But even just knowing the sink is useful IMO

src/main/scala/prci/ClockBundles.scala
```diff
@@ -10,7 +10,7 @@ class ClockBundle(val params: ClockBundleParameters) extends Bundle
   val reset = Reset()
 }
 
-class ClockGroupBundle(val params: ClockGroupBundleParameters) extends Bundle
+class ClockGroupBundle(val params: ClockGroupBundleParameters, val names: Option[Seq[String]] = None) extends Bundle
```
hcook, line 13: Would it be simpler to have the names be calculated by `ClockGroupBundleParameters` itself?

mwachs5: I persued this route, but take a look at the code (will comment on it)

src/main/scala/prci/ClockGroup.scala
```diff
@@ -22,7 +22,9 @@ class ClockGroup(groupName: String)(implicit p: Parameters) extends LazyModule
     require (node.in.size == 1)
     require (in.member.size == out.size)
 
-    (in.member zip out) foreach { case (i, o) => o := i }
+    (in.member zip out) foreach { case ((iName, i) , o) =>
+      o := i
+    }
```
mwachs5, line 27:
```suggestion
    (in.member zip out) foreach { case ((iName, i) , o) => o := i }
```

src/main/scala/prci/ClockNodes.scala
```diff
@@ -8,7 +8,9 @@ import freechips.rocketchip.diplomacy._
 
 object ClockImp extends SimpleNodeImp[ClockSourceParameters, ClockSinkParameters, ClockEdgeParameters, ClockBundle]
 {
-  def edge(pd: ClockSourceParameters, pu: ClockSinkParameters, p: Parameters, sourceInfo: SourceInfo) = ClockEdgeParameters(pd, pu, p, sourceInfo)
+  def edge(pd: ClockSourceParameters, pu: ClockSinkParameters, p: Parameters, sourceInfo: SourceInfo) = {
+    ClockEdgeParameters("foo", pd, pu, p, sourceInfo)
```
mwachs5, line 12: @hcook i don't know what to put for the name here. Is this never used? I did not end up with any "foo" signals in the design so hard to understand the code flow exactly here.

src/main/scala/prci/ClockParameters.scala
```diff
@@ -83,25 +86,32 @@ case class ClockGroupDriverParameters(
...
   }
 
-  def driveFromIOs: DriveFn = { (groups, num, p, vn) =>
+  def driveFromIOs()(implicit valName: ValName): DriveFn = { (groups, num, p, vn) =>
     implicit val pp = p
     val ioClockGroupSourceNode = ClockGroupSourceNode(List.fill(num) { ClockGroupSourceParameters() })
     groups :*= ioClockGroupSourceNode
-    InModuleBody { ioClockGroupSourceNode.makeIOs()(vn) }
+    InModuleBody {
+      val bundlesAndEdges = ioClockGroupSourceNode.out
```
mwachs5, line 109: this is basically a rewrite of `makeIOs` from diplomacy.Nodes, which assumes Heterogeneous Bag

src/main/scala/prci/ClockNodes.scala
hcook: You would have to add `name` as a parameter to the individual `ClockSinkParameters`; right now only groups get named, not individual sinks.

Alternatively, you/we could add the `nodePath` parameters that many of the other diplomatic protocol types have [e.g. interrupts](https://github.com/chipsalliance/rocket-chip/blob/bf0e85422db1ea1f1ef4160963c6d9858e254eb1/src/main/scala/interrupts/Parameters.scala#L28) and then use the node path to derive the name (from the val name of the node or the node's `parent` module). That seems like a useful extension but probably as a separate PR

hcook: As far as usage goes, this code gets used to make edges that are presumed never to be a group, i.e. only ever can accept a single clock. You would see this for example in a sifive-blocks device that has an asynchronous clock crossing inserted between it and the bus. However, even in that case, a `ClockBundle` is produced, and it is only `ClockGroupBundles` that currently use the name information in creating `members`.

hcook: It's not clear to me that you actually _had_ to add `name` to `ClockEdgeParameters` and `ClockBundleParameters`. You could have made `ClockGroupEdgeParameters.members` an e.g. `Map[String, ClockEdgeParameters]" and therbey kept the naming concept only at the "Group" level

src/main/scala/prci/ClockParameters.scala
hcook: Yeah in a follow up PR we should just entirely replace that use of `HeterogeneousBag` with `RecordListMap`, I think. Will require sed/updating a moderate amount of boilerplate closed source code to change arguments like "`io: HeterogeneousBag`" to "`io: Record`" though

src/main/scala/prci/ClockBundles.scala
```diff
@@ -12,5 +14,8 @@ class ClockBundle(val params: ClockBundleParameters) extends Bundle
 
 class ClockGroupBundle(val params: ClockGroupBundleParameters) extends Bundle
 {
-  val member = HeterogeneousBag(params.members.map(p => new ClockBundle(p)))
+  val member: RecordListMap[ClockBundle] = {
+    val nameToBundleMap = ListMap(params.members.map{p => p.name -> new ClockBundle(p)}: _*)
```
hcook, line 18: As i mentioned below, I think you could make `ClockGroupBundleParameters.members` itself be a `Map[String, ClockBundleParams] and then just iterate through it, turning each item in `values` into a `ClockBundle` here

mwachs5: thanks, have done that

src/main/scala/prci/ClockNodes.scala
mwachs5: yep, have done that

mwachs5:
This is currently giving me a compile error (just running `make verilog`) that I don't understand:

```
info] [0.001] Elaborating design...
[error] java.lang.ClassCastException: freechips.rocketchip.prci.ClockBundle cannot be cast to scala.runtime.Nothing$
[error] 	...
[error] ocketchiat freechips.rocketchip.util.RecordMap.$anonfun$elements$1(RecordMap.scala:22)
[error] 	at scala.collection.MapLike$MappedValues.$anonfun$foreach$3(MapLike.scala:256)
[error] 	at scala.collection.TraversableLike$WithFilter.$anonfun$foreach$1(TraversableLike.scala:877)
```

mwachs5:
Here's the current example of what this creates:

Screen Shot 2020-06-15 at 3 37 47 PM

I am still a little confused why we end up with only one output instead of 6 in the SimpleClockGroupSource, I think I am checking for duplicates in the wrong place in RecordMap (does a ListMap already not allow duplicates?)

Screen Shot 2020-06-15 at 3 42 13 PM

mwachs5:
Just as an update I'm still very weirded out by the before-and-after diff. If I look at the `fir` that comes out, the diff looks as I expect: all the same wires and I/Os are there, they just have better names ("member.sbus_0.clock" vs "member.0.clock" and so on). But in the verilog most of the signals are removed. I am going to do a run with the split-trace on to see when they diverge.

mwachs5:
I think I figured out what was wrong /different with this. 

In the original, this line does something:
https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/diplomacy/LazyModule.scala#L292

In this new version, that is tied off to 1'b0:
Screen Shot 2020-06-21 at 2 46 13 PM

Which may be why things get constant-propped away. Still looking at the FIRRTL intermediate to see where things diverge

mwachs5:
im not sure why this PR is closed but also says "merged and closed". Something is weird in github-land. Anyway, I have a suspicion what is wrong here... somewhere I am combining HeterogeneousBags of ClockGroupBundles whose names don't match up. I think the auto-bundle-zipping of diplomacy  (`:=`) is then not matching them up. In other words, somewhere I am trying to attach a `HeterogenousBag[ClockGroupBundle] := HeterogeneousBag[ClockGroupBundle]` and since the two ClockGroupBundles have different RecordMap parameters, the signals don't just magically get zipped together. Need to figure out where this is happening and fix it.

PR #2490, opened 2020-05-27
Fix Travis CI Shell Options

seldridge:
This fixes two issues when using the new regression scripts on Travis infrastructure:

1. `travis_setup_env.bash` may have an undefined variable `TRAVIS_ENABLE_INFRA_DETECTION`. This script needs to continue past this and have unset variable checking disabled.
2. `travis_wait.bash` is written such that it needs to continue after the first failure to do useful things like print out the log of execution.

**Related issue**: None.

**Type of change**: bug fix

**Release Notes**

None.

### DIFF
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -277,7 +277,7 @@ ifdef SEED
 SEED_ARG = --seed $(SEED)
 endif
 
-JTAG_DTM_SIM_ARGS = +verbose +jtag_rbb_enable=1 $(SEED_ARG)
+JTAG_DTM_SIM_ARGS = +jtag_rbb_enable=1 $(SEED_ARG)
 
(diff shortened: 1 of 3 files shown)


seldridge:
@richardxia: I have no idea what that failure was before I pushed. I assume that's spurious.

I did notice that the log for the JTAG tests is large (>500MiB). I added a commit that turns off `+verbose`, but was unsure if that was correct or would help with stability.

richardxia:
Yeah, I agree that it's probably spurious, since I can't see why any of the changes here would cause that test to specifically fail.

seldridge:
Ping @richardxia. Okay to merge this? (I technically have the power to do it, but that seems inappropriate.)

richardxia:
> Ping @richardxia. Okay to merge this? (I technically have the power to do it, but that seems inappropriate.)

Yep, I think the build was still running when I left my review, but I can hit the merge button now. Thanks again!

jackkoenig:
Need that Mergify `[Please Merge]` label

PR #2493, opened 2020-05-28
PseudoLRU non-power-of-2

johningalls-sifive:
**Related issue**: 
**Type of change**: performance enhancement

**Release Notes**
- Formerly, the PseudoLRU algorithm would not replace all entries in a non-power-of-2 ways cache (or TLB).  For example, PLRU with 6 ways results in way 5 (the sixth way) never being replaced unless it is otherwise invalidated.  However, the way index is always "in bounds", so this is purely a performance bug, not a functional bug.
- The new PseudoLRU algorithm is more verbose Chisel code, with extra if-statements to recursively descend an un-balanced binary tree.
- I wrote tests to exercise the old and the new PLRU behavior.  @hcook do I need to do anything else to include these in CI regressions?
- While I was at it, I upgraded Replacement.scala to Chisel3.
- I also added TrueLRU as an option for a replacement policy, based on code we had outside Rocket-chip.
- Lastly, I exposed some more internals of the ReplacementPolicy classes as public function defs to use outside Rocket-chip.

### DIFF
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -61,7 +61,7 @@ endif
 
 ifeq ($(SUITE),UnittestSuite)
 PROJECT=freechips.rocketchip.unittest
-CONFIGS=$(PROJECT).AMBAUnitTestConfig $(PROJECT).TLSimpleUnitTestConfig $(PROJECT).TLWidthUnitTestConfig $(PROJECT).PowerQueueTestConfig
+CONFIGS=$(PROJECT).AMBAUnitTestConfig $(PROJECT).TLSimpleUnitTestConfig $(PROJECT).TLWidthUnitTestConfig
 endif
 
 ifeq ($(SUITE), JtagDtmSuite)
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -61,7 +61,7 @@ endif
 
 ifeq ($(SUITE),UnittestSuite)
 PROJECT=freechips.rocketchip.unittest
-CONFIGS=$(PROJECT).AMBAUnitTestConfig $(PROJECT).TLSimpleUnitTestConfig $(PROJECT).TLWidthUnitTestConfig $(PROJECT).PLRUUnitTestConfig $(PROJECT).PowerQueueTestConfig
+CONFIGS=$(PROJECT).AMBAUnitTestConfig $(PROJECT).TLSimpleUnitTestConfig $(PROJECT).TLWidthUnitTestConfig $(PROJECT).PowerQueueTestConfig
 endif
 
 ifeq ($(SUITE), JtagDtmSuite)
--- a/regression/Makefile
+++ b/regression/Makefile
@@ -61,7 +61,7 @@ endif
 
 ifeq ($(SUITE),UnittestSuite)
 PROJECT=freechips.rocketchip.unittest
-CONFIGS=$(PROJECT).AMBAUnitTestConfig $(PROJECT).TLSimpleUnitTestConfig $(PROJECT).TLWidthUnitTestConfig $(PROJECT).ScatterGatherTestConfig $(PROJECT).PLRUUnitTestConfig $(PROJECT).PowerQueueTestConfig
+CONFIGS=$(PROJECT).AMBAUnitTestConfig $(PROJECT).TLSimpleUnitTestConfig $(PROJECT).TLWidthUnitTestConfig $(PROJECT).PLRUUnitTestConfig $(PROJECT).PowerQueueTestConfig
 endif
 
 ifeq ($(SUITE), JtagDtmSuite)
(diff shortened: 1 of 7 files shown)


regression/Makefile
```diff
@@ -61,7 +61,7 @@ endif
 
 ifeq ($(SUITE),UnittestSuite)
 PROJECT=freechips.rocketchip.unittest
-CONFIGS=$(PROJECT).AMBAUnitTestConfig $(PROJECT).TLSimpleUnitTestConfig $(PROJECT).TLWidthUnitTestConfig
+CONFIGS=$(PROJECT).AMBAUnitTestConfig $(PROJECT).TLSimpleUnitTestConfig $(PROJECT).TLWidthUnitTestConfig $(PROJECT).ScatterGatherTestConfig $(PROJECT).PLRUUnitTestConfig $(PROJECT).PowerQueueTestConfig
```
BrianGraysonSiV, line 64: This is probably too long of a line for lint et al, right?

src/main/scala/util/Replacement.scala
```diff
@@ -37,43 +63,147 @@ class SeqRandom(n_ways: Int) extends SeqReplacementPolicy {
...
+    if (this_ways > 2) {
+      val half_ways: Int = 1 << (log2Ceil(this_ways) - 1)
+      if (this_ways > 3) {
+        Cat(!way(log2Ceil(this_ways)-1),
+            Mux(way(log2Ceil(this_ways)-1),
+                get_next_state(state(this_ways-3,half_ways-1), way(log2Ceil(this_ways-half_ways)-1,0), this_ways-half_ways),
+                state(this_ways-3,half_ways-1)),
+            Mux(way(log2Ceil(this_ways)-1),
+                state(half_ways-2,0),
+                get_next_state(state(half_ways-2,0), way(log2Ceil(half_ways)-1,0), half_ways)))
```
BrianGraysonSiV, line 165: @johningalls-sifive can you add some comments here or in another file that explains _**in detail**_ what is going on? The software-engineer in me sees all these magic -3, -2, -1 values, and the lack of comments, and although I trust you that it does the right thing, I don't know if anyone besides you can maintain this function as it currently is! :)

src/main/scala/util/Replacement.scala
```diff
@@ -37,43 +63,147 @@ class SeqRandom(n_ways: Int) extends SeqReplacementPolicy {
...
+  def access(ways: Seq[Valid[UInt]]) {
+    when (ways.map(_.valid).orR) {
+      state_reg := get_next_state(state_reg, ways)
+    }
+    for (i <- 1 until ways.size) {
+      cover(PopCount(ways.map(_.valid)) === i.U, s"PLRU_UpdateCount$i", s"PLRU Update $i simultaneous")
+    }
+  }
+
+  def get_next_state(state: UInt, way: UInt, this_ways: Int): UInt = {
```
BrianGraysonSiV, line 153: I'm a fan of descriptive parameter and variable names. Would either n_ways or this_nways be better than just "this_ways", which (to me at least) does not clearly convey that it is the total number of ways?

Similarly, I'd change "way" to "touched_way", and if typing that is too annoying, say val w = touched_way, and use w in the code below for conciseness. That way, people fully understand the parameters to get_next_state() even without any comments.

regression/Makefile
johningalls-sifive: not sure about lint in makefiles, but the line keeps getting shorter as I remove tests that I tried to add but fail in CI regressions!  🙃

johningalls-sifive:
> I think if you actually want them to run on every PR CI you have to add them to `UnittestSuite` list in `regression/Makefile`

this (and other) unit test that I tried to add to the PR CI regression Makefile fails with
```
/home/runner/work/rocket-chip/rocket-chip/src/main/resources/csrc/verilator.h:29:1: fatal error: /home/runner/work/rocket-chip/rocket-chip/emulator/generated-src/freechips.rocketchip.unittest.<name>Config.plusArgs: No such file or directory
 #endif
 ^
compilation terminated.
```

src/main/scala/util/Replacement.scala
johningalls-sifive: I changed it to `tree_nways` and `touch_way`

src/main/scala/util/Replacement.scala
johningalls-sifive: I agree completely and went through the PseudoLRU code again.  Let me know if it's improved now!

johningalls-sifive:
the PLRU unit test is passing running this command locally:
```
make -C vsim run PROJECT=freechips.rocketchip.unittest CONFIG=freechips.rocketchip.unittest.PLRUUnitTestConfig
```

PR #2497, opened 2020-06-01
BundleBridge enhancements

hcook:
This PR adds two new capabilities to the diplomatic BundleBridge API.

`BundleBroadcast` is generalized into a `BundleBridgeNexus` that now can accept multiple inputs and a user-defined function for reducing the inputs into a single wire value that is to be broadcast, as well as a user-defined function for broadcasting that value.
- the confusingly named `BundleBridgeNexus` case class is renamed to `BundleBridgeNexusNode`
- `BundleBridgeNexus` is now the base class of the implementation of the reduction and broadcast wiring. It defaults to or-reducing its inputs and broadcasting identical copies of the reduction.
- `BundleBridgeNexus` can also be supplied a default wire value in cases where there are no input edges. This value is treated as an input by `inputFn`.
- aiding backwards compatibility,`BundleBridgeNexus.apply` produces a `BundleBridgeNexusNode`
- `BundleBroadcast.appy` is now just a factory of `BundleBridgeNexus` that supplies an `inputFn` that `require`s that only a single input edge be present
- `object BundleBridgeNexus` supplies some example input and output functions, including ones that insert timing closure registers on either input or output sides.

`BundleBridgeSink` is enhanced to also be able to optionally supply a `gen` function for producing bundles in the graph. If both source and sink supply gen functions, they must pass
`chisel3.DataMirror.checkTypeEquivalence`. If only sources or sinks supply a `gen` function, it is the one used. Neither supplying the function throws an exception. `BundleBridgeNexus` checks the type equivalence of all input and output edges.

**Impact**: API modification

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -178,14 +178,7 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
   traceNode := traceSourceNode
 
   // Trace sideband signals into core
-  val traceAuxNode = BundleBridgeNexus[TraceAux](
-    inputFn = (seq: Seq[TraceAux]) => {
-      val aux = Wire(new TraceAux)
-      aux.stall := seq.map(_.stall).foldLeft(false.B)(_||_)
-      aux.enable := seq.map(_.enable).foldLeft(false.B)(_||_)
-      aux
-    }
-  )
+  val traceAuxNode = BundleBridgeNexus[TraceAux]()
   val traceAuxSinkNode = BundleBridgeSink[TraceAux]()
   val traceAuxDefaultNode = BundleBridgeSource(() => new TraceAux)
   traceAuxSinkNode := traceAuxNode := traceAuxDefaultNode
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -178,7 +178,14 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
   traceNode := traceSourceNode
 
   // Trace sideband signals into core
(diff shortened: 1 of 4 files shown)


src/main/scala/tile/BaseTile.scala
```diff
@@ -264,10 +264,6 @@ abstract class BaseTileModuleImp[+L <: BaseTile](val outer: L) extends LazyModul
 
   outer.traceAuxDefaultNode.bundle.stall := false.B
   outer.traceAuxDefaultNode.bundle.enable := false.B
```
ernie-sifive, line 266: With the new default wire value for BundleBridgeNexus, it seems like the traceAuxDefaultNode is no longer needed.  Its purpose was to provide a default value in case there were no other input edges (from trace encoders).

hcook: Correct! But you do have to supply the default value to the nexus. For e.g. `UInt` nexuses doing so is quite pithy, but for the aux node nexus I needed
```
val traceAuxNode = BundleBridgeNexus[TraceAux](
  default = Some({
    val aux = Wire(new TraceAux)
    aux. stall := false.B
    aux. enable := false.B
  })
)
```
and I wasn't sure that that was better than leaving the existing code as is since it is actually equal or more lines of code (modulo some futzing with whitespace). Is having the extra node onerous or does it help explain where the value is coming from when e.g. looking at the graphml. I could be convinced either way

ernie-sifive: Understood.  I don't really have an opinion either way, so leaving it as is seems fine to me.

hcook: i later decided to go with using the default because those wires became the only ones left in the `BaseTileModuleImp` class

PR #2503, opened 2020-06-05
TLB: Check for PutPartial support separately from PutFull

aswaterman:
Previously, the TLB would report that M_PWR (PutPartial) was permissible
anywhere that M_XWR (PutFull) was.  This is fine for cache transactions,
since the caches can handle PutPartial internally.  It's not fine for
uncached transactions, since PutFull support does not imply PutPartial
support.

So, check for PutPartial support separately, similar to AMOs.  If the
access is cached, assume the access can proceed, since it'll be handled
in cache.  If the access is uncached, check the PMA.

Rocket never issues M_PWR, so there's no functional change on that front.

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -897,7 +897,6 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
       // We could consider turning some of these into dynamic PMA checks.
       require(!m.supportsAcquireB || m.supportsGet, "With a vector unit, cacheable memory must support Get")
       require(!m.supportsAcquireT || m.supportsPutPartial, "With a vector unit, cacheable memory must support PutPartial")
-      require(!m.supportsPutFull || m.supportsPutPartial, "With a vector unit, writable memory must support PutPartial")
     }
   }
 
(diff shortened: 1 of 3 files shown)


johningalls-sifive:
`TLSimpleUnitTestConfig` has now failed twice in `Continuous Integration / Test (1) (pull_request)`

aswaterman:
Similar to https://github.com/chipsalliance/rocket-chip/pull/2502 -- "Killed", so probably OOM, not a definitive test failure.

src/main/scala/rocket/DCache.scala
```diff
@@ -877,15 +877,17 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
...
-    amoalu.io.rhs := pstore1_data
-    pstore1_storegen_data := (if (!usingDataScratchpad) amoalu.io.out else {
+    val amoalus = (0 until coreDataBits / xLen).map { i =>
+      val amoalu = Module(new AMOALU(xLen))
+      amoalu.io.mask := pstore1_mask >> (i * xBytes)
+      amoalu.io.cmd := (if (usingAtomicsInCache) pstore1_cmd else M_XWR)
+      amoalu.io.lhs := s2_data_word >> (i * xLen)
+      amoalu.io.rhs := pstore1_data >> (i * xLen)
+      amoalu
+    }
```
johningalls-sifive, line 887: yep, I had to do this same thing elsewhere too

PR #2504, opened 2020-06-05
Standardize Tile Instantiation and Attachment

hcook:
`HasTiles` is refactored to eagerly instantiate tiles of all types by pulling instances of `CanAttachTile` out of a new `Field` named `TilesLocated`.

This change has the following benefits:
- Any class `<: BaseTile` can be `Config`-urably instantiated inside a design without needing to mix in additional traits to the subsystem class hierarchy.
- Various subsystem user code that would like to operate over the `tiles` list (e.g. `nTiles`) does not have to jump through hoops to ensure that it doesn't get called before all tile-adding traits are mixed in, because those traits no longer exist.

In order to make a tile instantiate-able via the new field, you must supply a class like this one:
```
case class RocketTileAttachParams(
  tileParams: RocketTileParams,
  crossingParams: RocketCrossingParams,
  lookup: LookupByHartIdImpl
) extends CanAttachTile { type TileType = RocketTile }
```
where
- `tileParams` satisfies the new base class `InstantiableTileParams[TileType]` (you provide a function that instantiates an instance of your `TileType`)
- `crossingParams` satisfies `TileCrossingParamsLike`, for which you could just use the example `RocketCrossingParams`.

Backwards compatibility:
- Tiles can still be instantiated and attached manually via traits:
```
trait HasMyTiles extends HasTiles {
  val myTiles: Seq[MyTile] = p(MyTileField).map { ... }
  override val tiles = myTiles ++ super[HasTiles].tiles
```
- However, if you do so, be aware that you will have to also manually connect your tiles.
- The `Fields` `RocketTilesKey` and `RocketCrossingKey` are maintained for backwards compatibility and `LegacyTileFieldHelper` is applied to `TilesLocated` in `BaseConfig` to ensure that `Config` alterations that manipulate these fields continue to function
- `BaseTile` now `extends HasLogicalTreeNode`, but a `GenericLogicalTreeNode()` is supplied, which can be overridden.

**Impact**:  API modification

**Release Notes**
`val tiles` in `trait HasTiles` is now populated eagerly via the `TilesLocated` `Field`.

### DIFF
--- a/src/main/scala/subsystem/Configs.scala
+++ b/src/main/scala/subsystem/Configs.scala
@@ -41,6 +41,8 @@ class BaseSubsystemConfig extends Config ((site, here, up) => {
   case DebugModuleKey => Some(DefaultDebugModuleParams(site(XLen)))
   case CLINTKey => Some(CLINTParams())
   case PLICKey => Some(PLICParams())
+  case TilesLocated(InSubsystem) => 
+    LegacyTileFieldHelper(site(RocketTilesKey), site(RocketCrossingKey), RocketTileAttachParams.apply _)
 })
 
 /* Composable partial function Configs to set individual parameters */
@@ -410,3 +412,17 @@ class WithScratchpadsOnly extends Config((site, here, up) => {
         scratch = Some(0x80000000L))))
   }
 })
+
+/** Boilerplate code for translating between the old XTilesParamsKey/XTilesCrossingKey pattern and new TilesLocated pattern */
+object LegacyTileFieldHelper {
+  def apply[TPT <: InstantiatableTileParams[_], TCT <: TileCrossingParamsLike, TAP <: CanAttachTile](
+    tileParams: Seq[TPT],
+    tcp: Seq[TCT],
+    apply: (TPT, TCT, LookupByHartIdImpl) => TAP): Seq[TAP] =
+  {
+    val crossingParams = heterogeneousOrGlobalSetting(tcp, tileParams.size)
+    tileParams.zip(crossingParams).map { case (t, c) =>
+      apply(t, c, PriorityMuxHartIdFromSeq(tileParams))
+    }
+  }
+}
--- a/src/main/scala/subsystem/Configs.scala
+++ b/src/main/scala/subsystem/Configs.scala
@@ -77,8 +77,10 @@ class WithCoherentBusTopology extends Config((site, here, up) => {
       l2 = site(BankedL2Key)))
 })
 
-class WithNBigCores(n: Int, idOffset: Int = 0) extends Config((site, here, up) => {
+class WithNBigCores(n: Int, overrideIdOffset: Option[Int] = None) extends Config((site, here, up) => {
   case RocketTilesKey => {
+    val prev = up(RocketTilesKey, site)
+    val idOffset = overrideIdOffset.getOrElse(prev.size)
     val big = RocketTileParams(
       core   = RocketCoreParams(mulDiv = Some(MulDivParams(
         mulUnroll = 8,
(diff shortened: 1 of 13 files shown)


src/main/scala/subsystem/Configs.scala
```diff
@@ -41,6 +41,8 @@ class BaseSubsystemConfig extends Config ((site, here, up) => {
   case DebugModuleKey => Some(DefaultDebugModuleParams(site(XLen)))
   case CLINTKey => Some(CLINTParams())
   case PLICKey => Some(PLICParams())
+  case TilesLocated(InSubsystem) =>
```
jerryz123, line 44: Will there be an example for what direct manipulation of the TilesLocated pattern looks like? (Not using the LegacyTileFieldHelper)

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -42,41 +101,94 @@ trait HasTiles extends HasCoreMonitorBundles { this: BaseSubsystem =>
...
+}
+
+/** HasTiles adds a Config-urable sequence of tiles of any type
+  *   to the subsystem class into which it is mixed.
+  */
+trait HasTiles extends HasCoreMonitorBundles with DefaultTileContextType
+{ this: BaseSubsystem => // TODO: ideally this bound would be softened to Attachable
+  implicit val p: Parameters
+  val tileAttachParams: Seq[CanAttachTile] = p(TilesLocated(location))
+  val tiles: Seq[BaseTile] = tileAttachParams.map(_.instantiate(p)) // actual tile creation
```
jerryz123, line 113: Is the expectation that `p(TilesLocated(location))` lists tiles by order of hartid? If so, it would be nice to check hartid sanity here. We've run into problems in the past where instantiating and connecting tiles out of hartid order results in a broken DTS.

hcook: Currently that is the expectation, but now that you mention it, now that all subtypes are ostensibly in one list I would be comfortable using `sortBy` to ensure that property. Would you prefer that to expecting/`requiring`?

src/main/scala/subsystem/Configs.scala
hcook: Good suggestion, will do

src/main/scala/subsystem/HasTiles.scala
jerryz123: Yes, that would be fine. A sanity check for hartid uniqueness would still be useful.

src/main/scala/subsystem/Configs.scala
hcook: I ended up doing this example in `WithTraceGen` for the `TraceGenTile` subtype. I'm not very keen to mess too much with the existing `RocketTilesKey` `Config` alterations since it it seems hard to change them while remaining backwards compatibly with external alterations of `RocketTilesKey`. Check out `WithTraceGen` and `HeterogeneousTileExampleConfig` and let me know if that is sufficiently helpful.

jerryz123: Thanks, this is pretty clear.

I noticed that manipulation of core parameters is more verbose without direct access through `CoreTilesKey`, since the type of the core must be matched, and since `AttachParams` adds another layer of case-class nesting. Given that manipulating tile-level parameters is the common use-case of the core-specific `TilesKey` s, it may be worth defining a generic macro to modify the `TileParams` matching some `desired` Tile type.

src/main/scala/tile/BaseTile.scala
```diff
@@ -29,6 +32,11 @@ trait TileParams {
   val name: Option[String]
 }
 
+abstract class InstantiatableTileParams[TileType <: BaseTile] extends TileParams {
```
jerryz123, line 35: Can this be shortened to `Instantiable`?

src/main/scala/system/Configs.scala
```diff
@@ -39,6 +39,15 @@ class DualChannelDualBankConfig extends Config(
 
 class RoccExampleConfig extends Config(new WithRoccExample ++ new DefaultConfig)
 
+class HeterogeneousTileExampleConfig extends Config(
+  new WithTraceGen (n = 2, idOffset = 3)() ++
```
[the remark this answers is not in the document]
jerryz123: Hmm, this config is a useful example, but wouldn't the TraceGen tiles prevent this design from running any software?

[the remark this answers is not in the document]
hcook: > Couldn't idOffset be dynamically generated with up(TilesLocated(InSubsystem), site)? Assuming that the configs place down the tiles in order of hartid.
That is an assumption, but definitely a common case. I will make it work both ways
> Hmm, this config is a useful example, but wouldn't the TraceGen tiles prevent this design from running any software?
I'll offset the memory address that the trace generators are pounding on such that "small enough" programs can be loaded into the same main memory at the base, but you're correct that it is a somewhat suspect design overall 😎

src/main/scala/subsystem/Configs.scala
hcook: if you can figure out a good pattern for applying a ["lens"](https://medium.com/zyseme-technology/functional-references-lens-and-other-optics-in-scala-e5f7e2fdafe) to nested case classes in Config alterations I will buy you a beer

src/main/scala/tile/BaseTile.scala
hcook: I am attempting do this but apparently i cannot spell

src/main/scala/subsystem/Configs.scala
hcook: But yes something specific to this particular use case is probably also possible

PR #2505, opened 2020-06-05
Supporthelperchange

brrmorre:
**Related issue**: 

**Release Notes**


src/main/scala/tilelink/Parameters.scala
```diff
@@ -876,62 +912,92 @@ class TLMasterPortParameters private(
...
+  // Available during RTL runtime, checks to see if (id, size) is supported by the master's (master's)  diplomatic parameters
+  private def safety_helper(member: TLMasterParameters => TransferSizes)(id: UInt, lgSize: UInt) = {
+    val allSame = masters.map(member(_) == member(masters(0))).reduce(_ && _)
+    if (allSame) member(masters(0)).containsLg(lgSize) else {
+      // Find the master associated with ID and returns whether that particular master is able to receive transaction of lgSize
+      Mux1H(find(id), masters.map(member(_).containsLg(lgSize)))
     }
   }
 
+  private def emitHelper(
```
jchang0, line 973: Don't insert this code between the definition of safety_helper and usage of safety_helper, because it seems like the safety_helper is only used here.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -876,62 +912,92 @@ class TLMasterPortParameters private(
...
+    safe:    Boolean,
+    member: TLMasterParameters => TransferSizes,
+    sourceId: UInt,
+    lgSize:  UInt,
+    range:   Option[TransferSizes]): Bool = {
+    def trim(x: TransferSizes) = range.map(_.intersect(x)).getOrElse(x)
```
jchang0, line 979: why do we need trim?

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1085,27 +1151,58 @@ case class TLEdgeParameters(
...
+  def expectsGetMasterToSlaveFast       (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsGet(sourceId, lgSize, range)         && slave.supportsGetFast(address, lgSize, range)
+  def expectsPutFullMasterToSlaveFast   (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsPutFull(sourceId, lgSize, range)     && slave.supportsPutFullFast(address, lgSize, range)
+  def expectsPutPartialMasterToSlaveFast(sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsPutPartial(sourceId, lgSize, range)  && slave.supportsPutPartialFast(address, lgSize, range)
+  def expectsHintMasterToSlaveFast      (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsHint(sourceId, lgSize, range)        && slave.supportsHintFast(address, lgSize, range)
+
+  //We don't use Safe vs Fast because we don't have the equivalent of address decoder for sourceIds
+// TODO (can use sourceId here to see if the specific master can support this transaction)
+  def expectsProbeSlaveToMaster      = master.supportsProbe
```
hcook, line 1183: Duality: these B channel messages from `SlaveToMaster` are:
- being routed to a client based on a `sourceId`, so they need to be passed to a helper that checks whether the _master_ that owns certain _sourceIds_ claimed to _support_ transactions of this size/type
- being sent from the manager of a particular address, so they need to be passed to a helper that checks with the _slave_ that owns certain _addresses_  claimed to _emit_ transactions of this size/type

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1085,27 +1151,58 @@ case class TLEdgeParameters(
...
   // Sanity check the link...
-  require (maxTransfer >= manager.beatBytes, s"Link's max transfer (${maxTransfer}) < ${manager.managers.map(_.name)}'s beatBytes (${manager.beatBytes})")
-
-  val bundle = TLBundleParameters(client, manager)
-  def formatEdge = client.infoString + "\n" + manager.infoString
+  require (maxTransfer >= slave.beatBytes, s"Link's max transfer (${maxTransfer}) < ${slave.slaves.map(_.name)}'s beatBytes (${slave.beatBytes})")
+
+  // For emits, check that the source is allowed to send this transactions
+// TODO emitAcquire, source => sourceId (or don't need sourceId at all)
+//      Probe, source => address
```
hcook, line 1162: These A channel messages from MasterToSlave are:

being routed to a slave based on a address, so they need to be passed to a helper that checks whether the `slave` that owns certain `addresses` claimed to `support` transactions of this size/type
being sent from the client using a sourceId, so they need to be passed to a helper that checks with the `master` that owns certain `sourceIds` claimed to `emit` transactions of this size/type

src/main/scala/tilelink/Parameters.scala
hcook: It seems asymmetric to "ignore" the `range` argument that is passed into the `expectsXYX` functions below. The helper that helps with sourceId-based calculations and the helper that helps with address-based calculations should probably treat range in the same way?

src/main/scala/tilelink/Parameters.scala
```diff
@@ -876,62 +912,92 @@ class TLMasterPortParameters private(
...
 
-  def requestFifo(id: UInt) = Mux1H(find(id), clients.map(c => Bool(c.requestFifo)))
+  def requestFifo(id: UInt) = Mux1H(find(id), masters.map(c => Bool(c.requestFifo)))
 
-  private def safety_helper(member: TLClientParameters => TransferSizes)(id: UInt, lgSize: UInt) = {
-    val allSame = clients.map(member(_) == member(clients(0))).reduce(_ && _)
-    if (allSame) member(clients(0)).containsLg(lgSize) else {
-      Mux1H(find(id), clients.map(member(_).containsLg(lgSize)))
+  // Available during RTL runtime, checks to see if (id, size) is supported by the master's (master's)  diplomatic parameters
+  private def safety_helper(member: TLMasterParameters => TransferSizes)(id: UInt, lgSize: UInt) = {
```
hcook, line 965: We need to reason about which features of the proposed`emitHelper` need to be merged with this function and whether they actually produce different logic in valuable ways.

I recommend changing the names to be `addressHelper` and `sourceIdHelper`, since I believe the supports/emits aspect is orthogonal (i.e. it is whichever `TLMasterParameters => TransferSizes` you happen to pass in, which could be either about supporting or about emitting).

src/main/scala/tilelink/Parameters.scala
```diff
@@ -876,62 +912,92 @@ class TLMasterPortParameters private(
...
+    val allSame = masters.map(member(_) == member(masters(0))).reduce(_ && _)
+    if (allSame) member(masters(0)).containsLg(lgSize) else {
```
hcook, line 967: I think this `if` statement is a coarse generalization of the `groupBy` in the `emitsHelper` version; the case where there is only one group.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,20 +537,53 @@ class TLSlavePortParameters private(
...
+    // Grouping the slaves based on the actual transfer size range they support
+    // intersecting the range and checking their membership
+    // FOR SUPPORTCASES instead of returning the list of slaves, you are returning a map from transfer size to the set of address sets that are supported for that transfer size
+
+    // find all the slaves that support a certain type of operation and then group their addresses by the supported size
+    // for every size there could be multiple address ranges
+    // safety is a trade off between checking between all possible addresses vs only the addresses that are known to have supported sizes
+    // the trade off is 'checking all addresses is a more expensive circuit but will always give you the right answer even if you give it an illegal address'
+    // the not safe version is a cheaper circuit but if you give it an illegal address then it might produce the wrong answer
+    // fast presumes address legality
```
hcook, line 568: Please remember to clean up this commentary.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,20 +537,53 @@ class TLSlavePortParameters private(
   // Does this Port manage this ID/address?
   def containsSafe(address: UInt) = findSafe(address).reduce(_ || _)
 
+  // all we need to change about this is adding 2 members, one for the tlmaster parametrs and one for the tlslave parameters
+  // All wee need to do is change the membership function to instead of only slave pamaters it looks at the intersection of both master and slave
+  // do we need to pass in two different arguments
```
hcook, line 542: I think we concluded below that two different helper functions are necessary and that we will `&&` the results of the duals together in the edge function. So this one should remain focused on `TLSlaveParameters => TransferSizes` cases that look at `address`.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -876,75 +912,111 @@ class TLMasterPortParameters private(
...
   def contains(id: UInt) = find(id).reduce(_ || _)
 
-  def requestFifo(id: UInt) = Mux1H(find(id), clients.map(c => Bool(c.requestFifo)))
+  def requestFifo(id: UInt) = Mux1H(find(id), masters.map(c => Bool(c.requestFifo)))
 
-  private def safety_helper(member: TLClientParameters => TransferSizes)(id: UInt, lgSize: UInt) = {
-    val allSame = clients.map(member(_) == member(clients(0))).reduce(_ && _)
-    if (allSame) member(clients(0)).containsLg(lgSize) else {
-      Mux1H(find(id), clients.map(member(_).containsLg(lgSize)))
+  // Available during RTL runtime, checks to see if (id, size) is supported by the master's (master's)  diplomatic parameters
```
jchang0, line 964: sic

terpstra:
As a follow-up PR we might want to split fast/safe into:
- a VIP checker API that consults both the full address (not assumed to be legal) and the full source (not assumed to be legal)
- a what possible sizes of Get might flow over this edge API (that does not look at address/source; just an intersection of the unions of master with the unions of slaves)
- a check for whether or not a message is legal for a given address (assumed to be legal) -- used by slaves
- a check for whether or not a message is legal for a given source (assumed to be legal) -- used by slaves

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1085,27 +1157,73 @@ case class TLEdgeParameters(
...
+  def expectsPutFullMasterToSlaveCheckerSafe   (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsPutPartialChecker(sourceId, lgSize, range)  && slave.supportsPutPartialCheckerSafe(address, lgSize, range)
+  def expectsPutPartialMasterToSlaveCheckerSafe(sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsPutPartialChecker(sourceId, lgSize, range)  && slave.supportsPutPartialCheckerSafe(address, lgSize, range)
+  def expectsHintMasterToSlaveCheckerSafe      (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsHintChecker(sourceId, lgSize, range)        && slave.supportsHintCheckerSafe(address, lgSize, range)
+
+  def expectsAcquireTMasterToSlaveCheckerFast  (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsAcquireTChecker(sourceId, lgSize, range)    && slave.supportsAcquireTCheckerFast(address, lgSize, range)
```
brrmorre, line 1186: Delete this

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1085,27 +1157,73 @@ case class TLEdgeParameters(
...
+  //      Probe, source => address
+  //These A channel messages from MasterToSlave are:
+  //being routed to a slave based on address bits,
+  //    so they need to be passed to a helper that checks whether the slave that owns certain addresses
+  //    claimed to support transactions of this size/type
+  //being sent from a master using sourceId bits,
+  //    so they need to be passed to a helper that checks whether the master that owns certain sourceIds
+  //    claimed to emit transactions of this size/type
+
+  def expectsAcquireTMasterToSlaveCheckerSafe  (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsAcquireTChecker(sourceId, lgSize, range)    && slave.supportsAcquireTCheckerSafe(address, lgSize, range)
```
brrmorre, line 1177: Get rid of all mentions of safe

brrmorre: Change name of checker to something else

brrmorre: expectsVipChecker

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,48 +537,81 @@ class TLSlavePortParameters private(
...
+    // fast presumes address legality
+    val supportCases = groupByIntoSeq(slaves)(m => trim(member(m))).map { case (k: TransferSizes, vs: Seq[TLSlaveParameters]) =>
       k -> vs.flatMap(_.address)
     }
+    // the rest of this function is just processing that more
+    // safe produces a circuit that compares against all possible addresses
+    // whereas fast presumes that the address is legal
+    // It creates an address decoder that only works if the address is supported
     val mask = if (safe) ~BigInt(0) else AddressDecoder(supportCases.map(_._2))
+    // simplified...?
```
brrmorre, line 577: remove this

brrmorre: Make these comments complete sentences

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1085,27 +1157,73 @@ case class TLEdgeParameters(
...
+  //We don't use Safe vs Fast because we don't have the equivalent of address decoder for sourceIds
+  //TODO (can use sourceId here to see if the specific master can support this transaction)
+  //Duality: these B channel messages from SlaveToMaster are:
+  //being routed to a master based on sourceId bits,
+  //    so they need to be passed to a helper that checks whether the master that owns certain sourceIds
+  //    claimed to support transactions of this size/type
+  //being sent from a slave owning certain addresses,
+  //    so they need to be passed to a helper that checks with the slave that owns certain addresses claimed
+  //    to emit transactions of this size/type
+  def expectsProbeSlaveToMasterChecker      = master.supportsProbeChecker
```
brrmorre, line 1204: Add the dual in

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1085,27 +1157,73 @@ case class TLEdgeParameters(
...
+  def expectsGetMasterToSlaveCheckerFast       (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsGetChecker(sourceId, lgSize, range)         && slave.supportsGetCheckerFast(address, lgSize, range)
+  def expectsPutFullMasterToSlaveCheckerFast   (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsPutFullChecker(sourceId, lgSize, range)     && slave.supportsPutFullCheckerFast(address, lgSize, range)
+  def expectsPutPartialMasterToSlaveCheckerFast(sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsPutPartialChecker(sourceId, lgSize, range)  && slave.supportsPutPartialCheckerFast(address, lgSize, range)
+  def expectsHintMasterToSlaveCheckerFast      (sourceId: UInt, address: UInt, lgSize: UInt, range: Option[TransferSizes] = None) = master.emitsHintChecker(sourceId, lgSize, range)        && slave.supportsHintCheckerFast(address, lgSize, range)
+
+  //We don't use Safe vs Fast because we don't have the equivalent of address decoder for sourceIds
```
hcook, line 1195: This is now irrelevant because we always use safe for the "vip" versions

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1085,27 +1157,73 @@ case class TLEdgeParameters(
...
+  //TODO (can use sourceId here to see if the specific master can support this transaction)
```
hcook, line 1196: Please do this TODO, i.e. add the arguments to make the below SlaveToMaster functions match the above MasterToSlave functions, and cal the `slave.emitsXChecker` function using some of those arguments.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -1085,27 +1157,73 @@ case class TLEdgeParameters(
...
   // Sanity check the link...
-  require (maxTransfer >= manager.beatBytes, s"Link's max transfer (${maxTransfer}) < ${manager.managers.map(_.name)}'s beatBytes (${manager.beatBytes})")
-
-  val bundle = TLBundleParameters(client, manager)
-  def formatEdge = client.infoString + "\n" + manager.infoString
+  require (maxTransfer >= slave.beatBytes, s"Link's max transfer (${maxTransfer}) < ${slave.slaves.map(_.name)}'s beatBytes (${slave.beatBytes})")
+
+  // For emits, check that the source is allowed to send this transactions
+  // TODO emitAcquire, source => sourceId (or don't need sourceId at all)
+  //      Probe, source => address
```
hcook, line 1168: I think the above TODO can be removed

src/main/scala/tilelink/Parameters.scala
```diff
@@ -876,75 +912,111 @@ class TLMasterPortParameters private(
...
-  val supportsHint       = safety_helper(_.supportsHint)       _
+  val supportsProbeChecker      = sourceIdHelper(_.supportsProbe)      _
+  val supportsArithmeticChecker = sourceIdHelper(_.supportsArithmetic) _
+  val supportsLogicalChecker    = sourceIdHelper(_.supportsLogical)    _
+  val supportsGetChecker        = sourceIdHelper(_.supportsGet)        _
+  val supportsPutFullChecker    = sourceIdHelper(_.supportsPutFull)    _
+  val supportsPutPartialChecker = sourceIdHelper(_.supportsPutPartial) _
+  val supportsHintChecker       = sourceIdHelper(_.supportsHint)       _
+
+  // TODO: Merge sourceIdHelper2 with sourceIdHelper
```
hcook, line 988: Please resolve this

src/main/scala/tilelink/Parameters.scala
hcook: We agreed that we do not need to be concerned with range trimming for the version of these functions used for VIP circuitry.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,48 +537,81 @@ class TLSlavePortParameters private(
...
 
-  private def supportHelper(
+  private def addressHelper(
       safe:    Boolean,
-      member:  TLManagerParameters => TransferSizes,
+      member:  TLSlaveParameters => TransferSizes,
       address: UInt,
       lgSize:  UInt,
-      range:   Option[TransferSizes]): Bool = {
+      range:   Option[TransferSizes]): Bool = { // What does the range argument do?
```
hcook, line 545: `// range provides a limit on the sizes that are expected to be sent, which might reduce circuit complexity`

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,48 +537,81 @@ class TLSlavePortParameters private(
   // Does this Port manage this ID/address?
   def containsSafe(address: UInt) = findSafe(address).reduce(_ || _)
 
-  private def supportHelper(
+  private def addressHelper(
       safe:    Boolean,
```
hcook, line 541: `// setting safe to false indicates that all addresses are expected to be legal, which might reduce circuit complexity`

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,48 +537,81 @@ class TLSlavePortParameters private(
...
-      member:  TLManagerParameters => TransferSizes,
+      member:  TLSlaveParameters => TransferSizes,
```
hcook, line 542: `// member filters out the sizes being checked based on the opcode being emitted or supported`

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,48 +537,81 @@ class TLSlavePortParameters private(
...
       address: UInt,
       lgSize:  UInt,
-      range:   Option[TransferSizes]): Bool = {
+      range:   Option[TransferSizes]): Bool = { // What does the range argument do?
+    // trim is just processing the range argument
```
hcook, line 546: `// trim reduces circuit complexity by intersecting checked sizes with the range argument

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,48 +537,81 @@ class TLSlavePortParameters private(
...
+    // the right answer even if you give it an illegal address'
+    // the not safe version is a cheaper circuit but if you give it an illegal address then it might produce the wrong answer
+    // fast presumes address legality
+    val supportCases = groupByIntoSeq(slaves)(m => trim(member(m))).map { case (k: TransferSizes, vs: Seq[TLSlaveParameters]) =>
       k -> vs.flatMap(_.address)
     }
+    // the rest of this function is just processing that more
+    // safe produces a circuit that compares against all possible addresses
+    // whereas fast presumes that the address is legal
+    // It creates an address decoder that only works if the address is supported
```
hcook, line 575:
```
// safe produces a circuit that compares against all possible addresses,
// whereas fast presumes that the address is legal but uses an efficient address decoder
```

src/main/scala/tilelink/Parameters.scala
hcook:
```
// This groupByIntoSeq deterministically groups all address sets for which a given `member` transfer size applies.
// In the resulting Map of cases, the keys are transfer sizes and the values are all address sets which emit or support that size.
```

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,48 +537,81 @@ class TLSlavePortParameters private(
...
     val mask = if (safe) ~BigInt(0) else AddressDecoder(supportCases.map(_._2))
+    // simplified...?
+    // simplified is doing some unification with the addresses to combine them with the mask
```
hcook, line 578: // Simplified crates the most concise possible representation of each cases' address sets based on the mask.

src/main/scala/tilelink/Parameters.scala
```diff
@@ -534,48 +537,81 @@ class TLSlavePortParameters private(
...
     val simplified = supportCases.map { case (k, seq) => k -> AddressSet.unify(seq.map(_.widen(~mask)).distinct) }
-    simplified.map { case (s, a) =>
+    simplified.map { case (s, a) => // once you've done that, you are returning
+    // s is a size, you are checking for this size either the size of the operation is in s
+    // you are checking if the wire address named `address` is contained within the calculated address set
+    // you are also checking if the size is contained within the size in the address set pair
+    // you are looking at that for every pair and then reducing them
+    // as long as you have an address that supports that size
+    // thats the circuit you are returning
```
hcook, line 586: `// We return an or-reduction of all the cases, checking whether any contains both the dynamic size and dynamic address on the wire.`

hcook:
@jchang0 @brrmorre Because the ` diplomacy information for the edge` contains the static tile id, this means that the monitors on the tile master crossbar and master ports breaks deduplication of multicore tiles...

jackkoenig:
Chisel 3.4 should support instance-specific annotations no longer breaking deduplication. TBD how to handle annotated signals that get constant propagated, but this should make it possible make things like monitors link against the DUT via annotations and thus not break dedup when they have custom code per instance.

Just mentioning some nearish possibilities, but for now I think we need to remove that part of the info String.

src/main/scala/tilelink/Parameters.scala
ingallsj: done in https://github.com/chipsalliance/rocket-chip/pull/2754

PR #2508, opened 2020-06-08
Disallow writes to MSTATUS.XS

jerryz123:
The ISA specifies the XS field to be read-only. This PR modifies Rocket's CSR file to hardware XS to 3 (dirty) when RoCC is present, and 0 otherwise.

**Impact**: API modification

**Development Phase**: mplementation

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -297,6 +297,7 @@ class CSRFile(
   val reset_mstatus = Wire(init=new MStatus().fromBits(0))
   reset_mstatus.mpp := PRV.M
   reset_mstatus.prv := PRV.M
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/CSR.scala
```diff
@@ -297,6 +297,7 @@ class CSRFile(
   val reset_mstatus = Wire(init=new MStatus().fromBits(0))
   reset_mstatus.mpp := PRV.M
   reset_mstatus.prv := PRV.M
+  reset_mstatus.xs := if (usingRoCC) UInt(3) else UInt(0)
```
aswaterman, line 300: I think this won't compile unless you add parens around the if/else.

jerryz123: :man_facepalming:

PR #2509, opened 2020-06-08
Broadcast filters

terpstra:
This adds the ability to add Probe filters to the Broadcast coherence manager.

### DIFF
--- a/src/main/scala/tilelink/Broadcast.scala
+++ b/src/main/scala/tilelink/Broadcast.scala
@@ -71,9 +71,9 @@ class TLBroadcast(lineBytes: Int, numTrackers: Int = 4, bufferless: Boolean = fa
 
       // Create the probe filter
       val filter = Module(filterFactory(ProbeFilterParams(
-        mshrs       = numTrackers,
-        caches      = caches.size,
-        addressBits = log2Ceil(edgeIn.manager.maxAddress) - lineShift)))
+        mshrs  = numTrackers,
+        caches = caches.size,
+        blockAddressBits = log2Ceil(edgeIn.manager.maxAddress) - lineShift)))
 
       // Create the request tracker queues
       val trackers = Seq.tabulate(numTrackers) { id =>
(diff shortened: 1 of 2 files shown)


src/main/scala/tilelink/Broadcast.scala
```diff
@@ -159,119 +184,214 @@ class TLBroadcast(lineBytes: Int, numTrackers: Int = 4, bufferless: Boolean = fa
...
+        TLMessages.PutFullData    -> true.B,
+        TLMessages.PutPartialData -> true.B,
+        TLMessages.ArithmeticData -> true.B,
+        TLMessages.LogicalData    -> true.B,
+        TLMessages.Get            -> false.B,
+        TLMessages.Hint           -> MuxLookup(in.a.bits.param, Wire(Bool()), Array(
+          TLHints.PREFETCH_READ   -> false.B,
+          TLHints.PREFETCH_WRITE  -> true.B)),
+        TLMessages.AcquireBlock   -> acq_needT,
+        TLMessages.AcquirePerm    -> acq_needT))
```
hcook, line 236: These seem like helper functions that could be defined on the `edge`. They're globally true for all TL1.9 caching adapters that have to deal with these opcode/param types, right?

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -159,119 +184,214 @@ class TLBroadcast(lineBytes: Int, numTrackers: Int = 4, bufferless: Boolean = fa
...
-      out.c.valid := Bool(false)
-      out.e.valid := Bool(false)
+      out.b.ready := true.B
+      out.c.valid := false.B
+      out.e.valid := false.B
     }
   }
 }
 
+// addressBits is after removing the block offset (typically lowest 6 bits for 64 bytes)
```
hcook, line 274: isn't this a "tag"?

terpstra: Well, you would have to strip sector bits and row bits as well before it becomes a tag. I agree addressBits is a bad name. I could not think of a better one, though. If you have a suggestion, I'll change it.

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -159,119 +184,214 @@ class TLBroadcast(lineBytes: Int, numTrackers: Int = 4, bufferless: Boolean = fa
...
+        probe_todo  := todo
+        probe_line  := filter.io.response.bits.address
+        probe_perms := Mux(filter.io.response.bits.needT, TLPermissions.toN, TLPermissions.toB)
+      }
+
+      // Inform the tracker of the number of probe targets
+      val responseCache = filter.io.response.bits.cacheOH | filter.io.response.bits.allocOH
+      val responseCount = PopCount(todo)
+      val responseMSHR = UIntToOH(filter.io.response.bits.mshr, numTrackers).asBools
+      val sack = filter.io.response.fire() && leaveB && others =/= 0.U
```
hcook, line 258: what is `sack`

src/main/scala/tilelink/Broadcast.scala
terpstra: Sure. Fixed.

src/main/scala/tilelink/Broadcast.scala
terpstra: shared acknowledge.

src/main/scala/tilelink/Broadcast.scala
hcook: `blockAddressBits` or `lineAddressBits` maybe?

terpstra: Changed to blockAddressBits.

PR #2511, opened 2020-06-09
Add `sbt test` to regressions

seldridge:
This adds the few Scalatest tests to the regression suite of both Travis and GitHub Actions.

This is largely cargo cult-ed from existing tests, but I expect @richardxia can straighten me out.

**Related issue**: None

**Release Notes**

None.

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -123,7 +123,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        bucket: [1, 2, 3, 4, 5, 6, 7]
+        bucket: [1, 2, 3, 4, 5, 6, 7, 8]
     steps:
       - uses: actions/checkout@v2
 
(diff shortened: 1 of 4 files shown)


richardxia:
The GitHub Actions changes look about right to me, but it looks like they failed because we haven't done the recursive submodule init. I think the Makefiles currently handle this, which is why the other tests didn't need to explicitly do the submodule init step.

One thing you could try is adding a Makefile target that runs the `sbt test` command but that depends on the git submodule checkout step. We have the git submodule checkouts split into a few different targets so that we don't have to clone all of GCC all the time, so I think the one that you want to target to grab the other dependencies like Chisel is this one: https://github.com/chipsalliance/rocket-chip/blob/711851d24ed556b2e6838ee5fd316a3e075b39f7/regression/Makefile#L148

seldridge:
Ping @richardxia. Would you be able to take a look? Your suggestion seems to work great.

PR #2520, opened 2020-06-12
BundleBridge: preserve width; don't use RegNext

terpstra:
RegNext erases the width of a UInt.
This leads to:

java.lang.IllegalArgumentException: requirement failed: hartPrefixNode.node (A nexus node with parent 'hartPrefixNode/topMod/topDesign/gen' at  (BundleBridge.scala:144:31)) requires all outputs have equivalent Chisel Data types, but got
UInt<10>(Wire in BundleBridgeNexus)
vs
UInt(Reg in BundleBridgeNexus)

Using safeRegNext, we can keep the useful requirement check and still have
BundleBridgeNexus work when registered=true with a UInt.

@jackkoenig FYI

### DIFF
--- a/src/main/scala/diplomacy/BundleBridge.scala
+++ b/src/main/scala/diplomacy/BundleBridge.scala
@@ -122,18 +122,24 @@ class BundleBridgeNexus[T <: Data](
 }
 
 object BundleBridgeNexus {
+  def safeRegNext[T <: Data](x: T): T = {
+    val reg = Reg(chiselTypeOf(x))
+    reg := x
+    reg
+  }
+
   def requireOne[T <: Data](registered: Boolean)(seq: Seq[T]): T = {
     require(seq.size == 1, "BundleBroadcast default requires one input")
-    if (registered) RegNext(seq.head) else seq.head
+    if (registered) safeRegNext(seq.head) else seq.head
   }
 
   def orReduction[T <: Data](registered: Boolean)(seq: Seq[T]): T = {
     val x = seq.reduce((a,b) => (a.asUInt | b.asUInt).asTypeOf(seq.head))
-    if (registered) RegNext(x) else x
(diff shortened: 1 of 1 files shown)


mwachs5:
Is there a chisel issue for "RegNext erases width of UInt"?

aswaterman:
@mwachs5 IIRC, it has been discussed, and both the current and alternate behaviors are controversial, but since a change would break existing code that relies on this behavior, it remains as-is.  cc @jackkoenig

src/main/scala/diplomacy/BundleBridge.scala
```diff
@@ -122,18 +122,24 @@ class BundleBridgeNexus[T <: Data](
 }
 
 object BundleBridgeNexus {
+  def safeRegNext[T <: Data](x: T): T = {
```
mwachs5, line 125: should this be moved to rocket-chip Utils?

terpstra: The fact that I needed this is due to a detail of the BundleBroadcast. I'd rather not promote an alternative RegNext, especially, since it might get fixed in the future.

mwachs5: ok

seldridge:
Prior discussion about this: https://github.com/freechipsproject/rfcs/issues/10

This is complicated and the current solution was to just document the behavior in https://github.com/freechipsproject/chisel3/pull/1318.

src/main/scala/diplomacy/BundleBridge.scala
seldridge: Yeah, I agree with @terpstra here. Thanks for your trepidation about promoting an alternative `RegNext`. Trying to keep Rocket Chip-dialect Chisel close to upstream Chisel is a noble goal. There does need to be an upstream solution to this.

PR #2521, opened 2020-06-13
Use BundleBridge to propagate tile input "constants"

hcook:
The "TileInputConstants" are increasingly less constant in certain configurations. Meanwhile, use of `hartId: Int` to generate circuits within the tile continues to occasionally foil deduplication across tiles. This PR attempts to address both issues by:
- deprecating `HasTileParmas.hartId` and providing the more clearly named `staticIdForMetadataUseOnly` to replace it
- using the new BundleBridgeNexus features in the subsystem to collect reset vectors from different sources (e.g. bootROMs), or per-tile subsystem inputs, and then broadcasting them to the tiles
- using the new BundleBridgeNexus features in the subsystem to provide default hart id constant values into the tile, optionally merge them with prefixes driven by other components, or replace them with per-tile subsystem inputs, and then broadcasting them to the tiles
- using BundleBroadcast within the tile to further propagate hart id and reset vector wires to tile submodules that have already been made diplomatic (i.e. everything other than the Rocket core pipeline itself).
- misc cleanups of the parameterization of related tile width parameters
- updating how ROMs are attached to the subsystem

**Impact**:  API modification

**Release Notes**
- `HasPeripheryBootROM` and `HasPeripheryBootROMModuleImp` are removed  and replaced by a call to `BootROM.attach`
- `BootROMParams` `Field` is removed and replaced with `BootROMLocated` `Field`
- `MaskROMLocated` `Field` is added
- `SubsystemExternalResetVectorKey`, `SubsystemExternalHartIdWidthKey` and `InsertTimingClosureRegistersOnHartIds ` `Fields` are added
- Unused `ResetVectorBits` `Field` is removed
- `HasExternallyDrivenTileConstants` bundle mixin is removed
- `HasResetVectorWire` subsystem trait is removed
- `HasTileInputConstants` and `InstantiatesTiles` subsystem traits are added
- `BaseTile` exposes `val hartIdNode: BundleBridgeNode[UInt]` and `resetVectorNode: BundleBridgeNode[UInt]` and these are automatically connected to in `HasTiles`.
- `rocket.Frontend, rocket.ICache, rocket.DCache, rocket.NDCache` now have `BundleBridgeSink[UInt]` for their reset vector or hartid wire inputs. If you instantiate them manually, i.e. not using the traits e.g. `rocket.HasHellaCache`, you will have to manually connect up those nodes to the aforementioned `BaseTile` nodes.


src/main/scala/system/ExampleRocketSystem.scala
```diff
@@ -16,12 +16,15 @@ class ExampleRocketSystem(implicit p: Parameters) extends RocketSubsystem
     with CanHaveMasterAXI4MemPort
     with CanHaveMasterAXI4MMIOPort
     with CanHaveSlaveAXI4Port
-    with HasPeripheryBootROM {
+{
+  // optionally add ROM devices
+  val bootROM  = p(BootROMLocated(location)).map { BootROM.attach(_, this, CBUS) }
+  val maskROMs = p(MaskROMLocated(location)).map { MaskROM.attach(_, this, CBUS) }
```
hcook, line 22: @rmac-sifive These are candidate targets for a more generic `HasDevices` trait in rocket-chip, but note that the `BootROM` one needs more than just `Attachable`

src/main/scala/diplomacy/BundleBridge.scala
```diff
@@ -42,16 +43,26 @@ case class BundleBridgeSink[T <: Data](genOpt: Option[() => T] = None)
...
+  }
+  def makeIO(name: String): T = makeIO()(ValName(name))
 }
 
 case class BundleBridgeSource[T <: Data](genOpt: Option[() => T] = None)(implicit valName: ValName) extends SourceNode(new BundleBridgeImp[T])(Seq(BundleBridgeParams(genOpt)))
 {
   def bundle: T = out(0)._1
 
-  def makeIO()(implicit valName: ValName): T = makeIOs()(valName).head
-  def makeIO(name: String): T = makeIOs()(ValName(name)).head
```
hcook, line 54: @jackkoenig before this PR, using the HeterogenousBag IO maker built in to the generic diplomacy `makeIOs` i ended up with `reset_vector_N_0` 😂 

@mwachs5 This is a case where we could potentially use RecordMap to have the reset vectors be named for the id of the tile they are attached to rather than just the index of the IO (and hoping those coincidentally match up if all the static ids are contiguous)

src/main/scala/devices/tilelink/BootROM.scala
```diff
@@ -58,25 +57,31 @@ class TLROM(val base: BigInt, val size: Int, contentsDelayed: => Seq[Byte], exec
...
-    val romdata = Files.readAllBytes(Paths.get(params.contentFileName))
-    val rom = ByteBuffer.wrap(romdata)
-    rom.array() ++ dtb.contents
-  }
-  def resetVector: BigInt = params.hang
+case class BootROMLocated(loc: HierarchicalLocation) extends Field[Option[BootROMParams]](None)
 
-  val bootrom = LazyModule(new TLROM(params.address, params.size, contents, true, cbus.beatBytes))
+object BootROM {
+  def attach(params: BootROMParams, subsystem: BaseSubsystem with HasTiles, where: TLBusWrapperLocation)
```
mwachs5, line 63: maybe it's worth adding a scaladoc comment here to the effect of if you that if you call this `attach` methodit takes over your reset vector of `subsystem` parameter. It's always been a bit tricky for me to figure out where that happens

```
subsystem.tileResetVectorNode
```

src/main/scala/devices/tilelink/BootROM.scala
```diff
@@ -58,25 +57,31 @@ class TLROM(val base: BigInt, val size: Int, contentsDelayed: => Seq[Byte], exec
...
+      LazyModule(new TLROM(params.address, params.size, contents, true, cbus.beatBytes))
+    }
 
-/** Subsystem will power-on running at 0x10040 (BootROM) */
-trait HasPeripheryBootROMModuleImp extends LazyModuleImp
-    with HasResetVectorWire {
-  val outer: HasPeripheryBootROM
-  global_reset_vector := outer.resetVector.U
+    bootrom.node := cbus.coupleTo("bootrom"){ TLFragmenter(cbus) := _ }
+    subsystem.tileResetVectorNode := bootROMResetVectorSourceNode
```
mwachs5, line 78:
```suggestion
    // Drive the `subsystem` reset vector to the `hang` address of this Boot ROM.
    subsystem.tileResetVectorNode := bootROMResetVectorSourceNode
```

src/main/scala/rocket/Frontend.scala
```diff
@@ -59,15 +59,15 @@ class FrontendIO(implicit p: Parameters) extends CoreBundle()(p) {
   val perf = new FrontendPerfEvents().asInput
 }
 
-class Frontend(val icacheParams: ICacheParams, hartid: Int)(implicit p: Parameters) extends LazyModule {
+class Frontend(val icacheParams: ICacheParams, staticIdForMetadata: Int)(implicit p: Parameters) extends LazyModule {
```
mwachs5, line 62: I am missing something fundamental in trying to rename this parameter to better capture its intended use. What is "Metadata" meant to encompass? In a sentence is it meant to say "this is the id for metadata but dont use it to generate hardware"?

src/main/scala/rocket/Frontend.scala
```diff
@@ -59,15 +59,15 @@ class FrontendIO(implicit p: Parameters) extends CoreBundle()(p) {
...
   lazy val module = new FrontendModule(this)
-  val icache = LazyModule(new ICache(icacheParams, hartid))
+  val icache = LazyModule(new ICache(icacheParams, staticIdForMetadata))
   val masterNode = icache.masterNode
   val slaveNode = icache.slaveNode
+  val resetVectorSinkNode = BundleBridgeSink[UInt](Some(() => Output(UInt(masterNode.edges.out.head.bundle.addressBits.W))))
```
mwachs5, line 67: i guess i need to read up on what BundleBridge parameters take these days... why does the Sink take  a function that generates an Output?

src/main/scala/rocket/Frontend.scala
```diff
@@ -349,9 +349,11 @@ class FrontendModule(outer: Frontend) extends LazyModuleImp(outer)
 /** Mix-ins for constructing tiles that have an ICache-based pipeline frontend */
 trait HasICacheFrontend extends CanHavePTW { this: BaseTile =>
   val module: HasICacheFrontendModule
-  val frontend = LazyModule(new Frontend(tileParams.icache.get, hartId))
+  val frontend = LazyModule(new Frontend(tileParams.icache.get, staticIdForMetadata))
```
mwachs5, line 352: what is this used for?

src/main/scala/rocket/ICache.scala
```diff
@@ -52,13 +52,14 @@ class ICacheErrors(implicit p: Parameters) extends CoreBundle()(p)
   val bus = Valid(UInt(width = paddrBits))
 }
 
-class ICache(val icacheParams: ICacheParams, val hartId: Int)(implicit p: Parameters) extends LazyModule {
+class ICache(val icacheParams: ICacheParams, val staticIdForMetadata: Int)(implicit p: Parameters) extends LazyModule {
```
mwachs5, line 55: maybe we should call this "staticIdForMetadataOnly"

src/main/scala/subsystem/Configs.scala
```diff
@@ -37,7 +37,8 @@ class BaseSubsystemConfig extends Config ((site, here, up) => {
     beatBytes = site(XLen)/8,
     blockBytes = site(CacheBlockBytes))
   // Additional device Parameters
-  case BootROMParams => BootROMParams(contentFileName = "./bootrom/bootrom.img")
+  case BootROMLocated(InSubsystem) => Some(BootROMParams(contentFileName = "./bootrom/bootrom.img"))
+  case SubsystemExternalResetVectorKey => false
```
mwachs5, line 41: should BootROMLocated basically `require` that SubsystemExternalResetVectorKey` is false?

src/main/scala/subsystem/Configs.scala
```diff
@@ -307,7 +308,7 @@ class WithFPUWithoutDivSqrt extends Config((site, here, up) => {
 })
 
 class WithBootROMFile(bootROMFile: String) extends Config((site, here, up) => {
-  case BootROMParams => up(BootROMParams, site).copy(contentFileName = bootROMFile)
+  case BootROMLocated(x) => up(BootROMLocated(x), site).map(_.copy(contentFileName = bootROMFile))
```
mwachs5, line 311: should this automatically set BootROMLocated(...) to false?

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -10,13 +10,18 @@ import freechips.rocketchip.devices.tilelink.{BasicBusBlocker, BasicBusBlockerPa
...
+import freechips.rocketchip.tile.{BaseTile, LookupByHartIdImpl, TileParams, InstantiableTileParams, MaxHartIdBits}
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.util._
 
 /** Entry point for Config-uring the presence of Tiles */
 case class TilesLocated(loc: HierarchicalLocation) extends Field[Seq[CanAttachTile]](Nil)
 
+/** Whether a global hart id prefix is going to be dynamically applied to the static tile hart ids */
+case object HartPrefixKey extends Field[Boolean](false)
+case object SubsystemExternalHartIdWidthKey extends Field[Option[Int]](None)
```
mwachs5, line 22: I dont get this set of knobs. HartPrefixKey is binary yes or no, so if I set it to yes, how does SubsystemExternalHartIdWidthKey get used? How do I know how wide the prefix is vs the static part?

Could we add a comment with a simple ASCII art something like this:

```
----------------------------------------------------------------------
| <---dynamic_hart_prefix_bits ---> | <---- static_tile_id_bits ---> |
---------------------------------------------------------------------
dynamic_hart_prefix_bits + static_tile_id_bits-1                     0
                     HartId
```
----------------------------------------------------------------
| <---dynamic hart prefix bits ---> | <---- statid tile id bits ---> |
----------

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -10,13 +10,18 @@ import freechips.rocketchip.devices.tilelink.{BasicBusBlocker, BasicBusBlockerPa
...
+case object SubsystemExternalResetVectorKey extends Field[Boolean](true)
```
mwachs5, line 23:
```suggestion
/** Whether the Subsystem module has I/Os that control the tile's power-on reset vectors
case object SubsystemExternalResetVectorKey extends Field[Boolean](true)
```

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,6 +91,49 @@ trait HasTileInterruptSources
   }
 }
 
+/** These are sources of "constants" that are driven into the tile.
+  * While they are not expected to change dyanmically while the tile is executing code,
+  * they may be either tied to a contant value or programmed during boot or reset.
+  * They need to be instantiated before tiles are attached to the subsystem containing them.
```
mwachs5, line 97:
```suggestion
  * They need to be instantiated before tiles are attached within the subsystem containing them.
```

src/main/scala/subsystem/Configs.scala
hcook: Because I used a `BundleBroadcast` (i.e. a `BundleBridgeNexus` that `requires` that there only be a single bridge source input) the build will already fail if you don't set them compatibly.

src/main/scala/subsystem/Configs.scala
hcook: Not sure what you mean by this

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,6 +91,49 @@ trait HasTileInterruptSources
...
+  val tileHartIdNode = BundleBridgeEphemeralNode[UInt]()
+  val tileHartIdNexusNode = BundleBridgeNexus[UInt](
+    inputFn = BundleBridgeNexus.orReduction[UInt](p(HartPrefixKey)) _,
+    outputFn = (prefix: UInt, n: Int) =>  Seq.tabulate(n) { i =>
+      val y = dontTouch(prefix | hartIdList(i).U(p(MaxHartIdBits).W))
+      if (p(HartPrefixKey)) BundleBridgeNexus.safeRegNext(y) else y
+    },
+    default = Some(() => 0.U(p(MaxHartIdBits).W))
+  )
+  // TODO: this also needs to be wired into the DebugModule's hart selection circuit?
```
mwachs5, line 110: The debug module as currently written requires in its parameters a function UInt <-> UInt to map hartID (as seen by the core in its `mhartid` CSR) to hart index. 

However in the special case that you only have one core, the debug module is able to ignore dynamic hart ID's. Otherwise it doesn't work with dynamic hart IDs. Not sure if you are able to solve this in this PR? If so add @ernie-sifive as a reviewer...

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,6 +91,49 @@ trait HasTileInterruptSources
...
+
+  // Collect a reset vector address and distribute to the tiles
+  val tileResetVectorNode = BundleBridgeEphemeralNode[UInt]()
```
mwachs5, line 113: does this mean that all tiles have the same reset vector? has this always been the case?

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,6 +91,49 @@ trait HasTileInterruptSources
...
+  val tileResetVectorNexusNode = BundleBroadcast[UInt]()
+
+  // Create subsystem IOs that are based on the number of tiles
+  //   or, if such IOs don't exist, rely on connections of drivers to internal nexus nodes
+  val tileHartIdIONodes: Seq[BundleBridgeSource[UInt]] = p(SubsystemExternalHartIdWidthKey) match {
+    case Some(w) => Seq.fill(tiles.size) {
+      val hartIdSource = BundleBridgeSource(() => UInt(w.W))
+      tileHartIdNode := hartIdSource
+      hartIdSource
```
mwachs5, line 122: should there be an InModuleBody check here to make sure tileHartIdNexusNode doesn't have any connections to it?

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,6 +91,49 @@ trait HasTileInterruptSources
...
+    }
+    case None => { tileHartIdNode :*= tileHartIdNexusNode; Nil }
+  }
+
+  val tileResetVectorIONodes: Seq[BundleBridgeSource[UInt]] = p(SubsystemExternalResetVectorKey) match {
+    case true => Seq.fill(tiles.size) {
+      val resetVectorSource = BundleBridgeSource[UInt]()
+      tileResetVectorNode := resetVectorSource
+      resetVectorSource
```
mwachs5, line 131: same Q, should there be an InModuleBody check to make sure no one is trying to connect to tileResetVectorNexusNode?

src/main/scala/rocket/Frontend.scala
hcook: Ah, this is a bit of a kludge; BundleBridge is written assuming it is bridging an IO defined from the source/"master"'s perspective. In this case, the sink is defining the bundle to be bridged (because only sinks know how wide to make the bundle in this case), and the bridge makes an external IO at the source side by `Flipping` the bundle. But if you just give it an undirected `UInt` instead of some "IO" Bundle class with directionality, Chisel complains when doing the flip.

It's plausible I could build something smarter into the `BundleBridgeSource.makeIO` function, where it assumes it should make an input IO if the direction it was given is undefined? @jackkoenig any thoughts

src/main/scala/subsystem/HasTiles.scala
hcook: Adding some more complicated comments here in a local commit

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -234,20 +258,49 @@ trait CanAttachTile {
...
 }
 
-/** Provides some Chisel connectivity to certain tile IOs */
-trait HasTilesModuleImp extends LazyModuleImp with HasPeripheryDebugModuleImp {
-  val outer: HasTiles with HasTileInterruptSources
+/** InstantiatesTiles adds a Config-urable sequence of tiles of any type
+  *   to the subsystem class into which it is mixed.
+  */
+trait InstantiatesTiles { this: BaseSubsystem =>
+  // Actually instantiate all tiles, in order based on statically-assigned hartids
```
mwachs5, line 273:
```suggestion
  // Actually instantiate all tiles, in order based on statically-assigned hartids
  // Note that these hartIds may not be those actually reflected at runtime in e.g. the $mhartid CSR
```

src/main/scala/subsystem/HasTiles.scala
hcook: The only way to get them to have separate reset vectors is to flip the `SubsystemExternalResetVectorKey` flag, which has historically been the only way to accomplish that outcome. In the closed source, flipping the flag would be overridden by Chisel last connect semantics by mixing in different traits adding roms or etc that override the inputs to all be the same. Now you have to choose whether to connect the tiles to the NexusNode (all same) or to the IONodes (all different, unless the parent drives them to be the same).

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -234,20 +258,49 @@ trait CanAttachTile {
...
+    params.connect(t.asInstanceOf[params.TileType], this.asInstanceOf[params.TileContextType])
   }
+}
+
+/** Provides some Chisel connectivity to certain tile IOs */
+trait HasTilesModuleImp extends LazyModuleImp with HasPeripheryDebugModuleImp {
+  val outer: HasTiles with HasTileInterruptSources with HasTileInputConstants
 
-  val tile_inputs = outer.tiles.map(_.module.constants)
+  val reset_vector = outer.tileResetVectorIONodes.zipWithIndex.map { case (n, i) => n.makeIO(s"reset_vector_$i") }
```
mwachs5, line 302: i was confused by the code that this connects to. This makes it look like each tile can have a different reset vector, but I didn't understand how that would work above

is the idea that `tileResetVectorIONodes` won't be present if p(SubsystemHasExternalResetVectorKey) = false?

src/main/scala/system/ExampleRocketSystem.scala
```diff
@@ -16,12 +16,15 @@ class ExampleRocketSystem(implicit p: Parameters) extends RocketSubsystem
     with CanHaveMasterAXI4MemPort
     with CanHaveMasterAXI4MMIOPort
     with CanHaveSlaveAXI4Port
-    with HasPeripheryBootROM {
+{
+  // optionally add ROM devices
```
mwachs5, line 20:
```suggestion
  // optionally add ROM devices
  // Note that setting BootROMLocated will override the reset_vector for all tiles
```

src/main/scala/tile/BaseTile.scala
```diff
@@ -66,10 +65,22 @@ trait HasNonDiplomaticTileParameters {
...
-  def hartIdLen: Int = p(MaxHartIdBits)
+  // Use staticIdForMetdata only to emit build information or identify a component to diplomacy.
+  //   Including it in a constructed Chisel circuit will prevent us
+  //   from being able to deduplicate otherwise-homogeneous tiles.
+  def staticIdForMetadata: Int = tileParams.hartId
+  @deprecated("use hartIdSinkNode.bundle or staticIdForMetadata", "rocket-chip 1.3")
+  def hartId: Int = staticIdForMetadata
+  lazy val hartIdLen: Int = {
+    val len = p(MaxHartIdBits)
+    require (log2Up(staticIdForMetadata + 1) <= len, s"p(MaxHartIdBits) of $len is not enough for static hart id $staticIdForMetadata")
```
mwachs5, line 81: im a little confused by this requirement, given that staticIdForMetadata is only supposed to be used for metadata. But it still seems like it's getting used in the hardware somewhere. See my previous request for ASCII art for this :)

src/main/scala/tile/BaseTile.scala
```diff
@@ -132,7 +143,11 @@ trait HasNonDiplomaticTileParameters {
   */
 trait HasTileParameters extends HasNonDiplomaticTileParameters {
   protected def tlBundleParams = p(TileVisibilityNodeKey).edges.out.head.bundle
-  def paddrBits: Int = tlBundleParams.addressBits
+  lazy val paddrBits: Int = {
+    val bits = tlBundleParams.addressBits
+    require(bits <= maxPAddrBits, s"asked $bits paddr bits, but since xLen is $xLen only $maxPAddrBits can fit")
```
mwachs5, line 148:
```suggestion
    require(bits <= maxPAddrBits, s"Requested $bits paddr bits, but since xLen is $xLen only $maxPAddrBits will fit")
```

src/main/scala/tile/BaseTile.scala
```diff
@@ -148,7 +163,12 @@ trait HasTileParameters extends HasNonDiplomaticTileParameters {
   def vpnBitsExtended: Int = vpnBits + (vaddrBits < xLen).toInt
   def vaddrBitsExtended: Int = vpnBitsExtended + pgIdxBits
 
-  def resetVectorLen: Int = paddrBits
+  lazy val resetVectorLen: Int = {
+    val externalLen = paddrBits
+    require(externalLen <= xLen, s"External reset vector length ($externalLen) must be <= XLEN ($xLen)")
+    require(externalLen <= vaddrBitsExtended, s"External reset vector length ($externalLen must be <= virtual address bit width ($vaddrBitsExtended)")
```
mwachs5, line 169:
```suggestion
    require(externalLen <= vaddrBitsExtended, s"External reset vector length ($externalLen) must be <= virtual address bit width ($vaddrBitsExtended)")
```

src/main/scala/tile/BaseTile.scala
```diff
@@ -265,24 +300,4 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
...
-}
-
-/** Some other non-tilelink but still standard inputs */
-trait HasExternallyDrivenTileConstants extends Bundle with HasTileParameters {
-  val hartid = UInt(INPUT, hartIdLen)
-  val reset_vector = UInt(INPUT, resetVectorLen)
-}
-
-class TileInputConstants(implicit val p: Parameters) extends Bundle with HasExternallyDrivenTileConstants
+abstract class BaseTileModuleImp[+L <: BaseTile](val outer: L) extends LazyModuleImp(outer) with HasTileParameters
```
mwachs5, line 303: 🎉

src/main/scala/subsystem/HasTiles.scala
mwachs5: by "flip" you mean "set SubsystemExternalResetVectorKey -> true"? If you set that, then all tiles get separate reset vectors?

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,6 +91,49 @@ trait HasTileInterruptSources
...
+  )
+  // TODO: this also needs to be wired into the DebugModule's hart selection circuit?
+
+  // Collect a reset vector address and distribute to the tiles
+  val tileResetVectorNode = BundleBridgeEphemeralNode[UInt]()
+  val tileResetVectorNexusNode = BundleBroadcast[UInt]()
+
+  // Create subsystem IOs that are based on the number of tiles
+  //   or, if such IOs don't exist, rely on connections of drivers to internal nexus nodes
+  val tileHartIdIONodes: Seq[BundleBridgeSource[UInt]] = p(SubsystemExternalHartIdWidthKey) match {
```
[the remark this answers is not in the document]
mwachs5: but in the case where p(SubsystemExternalHartIdWidthKey) is None aren't they NOT IOs? Oooh, in that case the list is just empty. I see. So anything that does show up in the list *is* an IO.

Ok, agree with your naming then.

src/main/scala/subsystem/HasTiles.scala
mwachs5: i undestand the code now, the answer to my question is yes, it would be an empty list in that case.

src/main/scala/subsystem/HasTiles.scala
hcook: Yes

src/main/scala/subsystem/Configs.scala
hcook: Ah, well, `WithBootROMFile` as written doesn't add a boot rom; if there is a boot rom in `up` it changes what file it uses to fill in the contents; but if there isn't one it does nothing. Your point is well taken though, since there are now some conjoined parameter settings I can make a `WithBootRom` that sets both

src/main/scala/subsystem/HasTiles.scala
hcook: Yeah I think dealing with this kind of boot-time dynamism is something that is going to have to be solved as part of a certain internal project, but not this PR.

src/main/scala/subsystem/Configs.scala
mwachs5: ah, yeah i see. Ok, well you can ignore my comment here then

src/main/scala/subsystem/Configs.scala
mwachs5: ok, that sounds good.

src/main/scala/diplomacy/BundleBridge.scala
mwachs5: ok hoping to clean up the PRCI on Record Map PR so that RecordMap could go in first...

src/main/scala/rocket/Frontend.scala
mwachs5: yeah, once I read more down it made sense. I can resolve this comment.

src/main/scala/tile/BaseTile.scala
mwachs5: doesn't p(MaxHartIdBits) also get used in $mhartid CSR? Well I guess i can RTFC to figure out how the width of that CSR is set, I just feel like i stumbled over this in the past

mwachs5: ok, yeah, so currently CSR.scala uses `hartIdLen` to figure out how wide to make the mhartid CSR. Where is that coming from now...?

mwachs5: https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/rocket/CSR.scala#L196

mwachs5: basically I am saying the important thing is whatever CSR.scala is using to set that width of the `mhartid` CSR needs to be checking the actual width of whatever I wanted shown in my ASCII art diagram. That's what we want this require to actually be checking, I believe.

src/main/scala/subsystem/HasTiles.scala
hcook: chose to handle it at the diplomacy layer using a nexus node feature, will push shortly

src/main/scala/diplomacy/BundleBridge.scala
hcook: Oh yeah, i just mention it as more of a "possible future work that could maybe also use the same feature." Since it changes a subsystem IO name I think we would have to figure out how to "productize" the change...

src/main/scala/tile/BaseTile.scala
hcook: The CSR's parent Core also has an IO sized by `hartIdLen`, i added a requirement checking the widths using Chisels `getWidth`

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -234,20 +258,49 @@ trait CanAttachTile {
...
   }
+}
+
+/** Provides some Chisel connectivity to certain tile IOs */
+trait HasTilesModuleImp extends LazyModuleImp with HasPeripheryDebugModuleImp {
+  val outer: HasTiles with HasTileInterruptSources with HasTileInputConstants
 
-  val tile_inputs = outer.tiles.map(_.module.constants)
+  val reset_vector = outer.tileResetVectorIONodes.zipWithIndex.map { case (n, i) => n.makeIO(s"reset_vector_$i") }
+  val tile_hartids = outer.tileHartIdIONodes.zipWithIndex.map { case (n, i) => n.makeIO(s"tile_hartids_$i") }
```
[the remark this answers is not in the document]
jackkoenig: Right so if you're going to suggest names of IOs, Chisel requires you to provide unique names (it will not do `_#` suffixing for you). Now since these values are top-level vals in the `LazyModuleImp`, you could let Chisel reflectively name them rather them rather than using `ValName`. The reason why you're having to provide the name yourself is because `ValName` just picks the name of the val, it assumes Chisel will take care of the suffixiing.

src/main/scala/rocket/Frontend.scala
jackkoenig: I don't entirely understand, clarifying questions:
1. Is `resetVectorSinkNode` still the sink or are you inverting the direction of the BundleBridge?
2. Is is just that the sink is defining the type and for that to work it has to give the `UInt` a direction?

hcook: 1. It is still the sink
2. Yes, the sink is defining the type, and the way the BundBridgeSource.makeIO is currently written, it `Flipped`s the bundle (because it was written assuming the source is defining a directionalized IO).

So I think more concretely my question is, does it seem sensible to treat an undirected bundle as something that needs to be flipped? I guess it is actually based on whether the source or the defined the type? :rubber_duck: I think I can probably do something sensible with sufficient application of `DataMirror.directionOf`

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,7 +108,67 @@ trait HasTileInterruptSources
...
+  val tileHartIdNexusNode = BundleBridgeNexus[UInt](
+    inputFn = BundleBridgeNexus.orReduction[UInt](registered = p(HartPrefixKey)) _,
+    outputFn = (prefix: UInt, n: Int) =>  Seq.tabulate(n) { i =>
+      val y = dontTouch(prefix | hartIdList(i).U(p(MaxHartIdBits).W))
+      if (p(HartPrefixKey)) BundleBridgeNexus.safeRegNext(y) else y
+    },
+    default = Some(() => 0.U(p(MaxHartIdBits).W)),
+    inputRequiresOutput = true // guard against this being driven but ignored in tileHartIdIONodes below
+  )
+  // TODO: do the set of dynamic hart ids also need to be wired into the DebugModule's hart selection circuit?
```
mwachs5, line 133:
```suggestion
  // TODO: Replace the DebugModuleHartSelFuncs config key with logic to consume the dynamic hart IDs
```

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,7 +108,67 @@ trait HasTileInterruptSources
   }
 }
 
+/** These are sources of "constants" that are driven into the tile.
+  * While they are not expected to change dyanmically while the tile is executing code,
```
mwachs5, line 112:
```suggestion
  * 
  * While they are not expected to change dyanmically while the tile is executing code,
```

src/main/scala/rocket/Frontend.scala
hcook: @jackkoenig if you get a chance can you take a look at https://github.com/chipsalliance/rocket-chip/pull/2521/commits/9c10a2c8e8590b2d8ddd31a0e92f8cea07690a3d to see if it looks kosher

jackkoenig: Yeah I think that looks good. Yeah I'm not sure if there's something Chisel should be doing here to help you or if this is just the right way to do it but 9c10a2c LGTM!

src/main/scala/subsystem/HasTiles.scala
mwachs5: I now understand that a "prefix" is really any number of "HartId" inputs of the same width `SubsystemExternalHartIdWidth`, and you can connect any number of them via the tileHartIdNexusNode

I'll resolve this comment to rescind my request for ASCII art and see if there are places I should suggest comments that would have made this more obvious to me.

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,7 +108,68 @@ trait HasTileInterruptSources
...
+  * They need to be instantiated before tiles are attached within the subsystem containing them.
+  */
+trait HasTileInputConstants extends InstantiatesTiles { this: BaseSubsystem =>
+  /** tileHartIdNode is used to collect publishers and subscribers of hartids. */
+  val tileHartIdNode = BundleBridgeEphemeralNode[UInt]()
+
+  /** tileHartIdNexusNode is a BundleBridgeNexus that collects dynamic hart prefixes,
+    *   orReduces them, and combines the reduction with the static ids assigned to each tile,
+    *   producing a unique, dynamic hartid for each tile.
+    */
```
mwachs5, line 124:
```suggestion
  /** tileHartIdNexusNode is a BundleBridgeNexus that collects dynamic hart "prefixes".
    * 
    *  Each "prefix" input is actually the same full width as the hartId, the expected usage
    *  is that each prefix source would set only some non-overlapping portion of the bits to something
    *  other than zero. 
    *  Thus, the tileHartIdNexusNode orReduces all the incoming "prefix" wires along with the static ids
    *  assigned to each tile via hartIdList, producing a unique, dynamic hartid for each tile.
    */
```

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -86,7 +108,68 @@ trait HasTileInterruptSources
...
+  /** tileResetVectorNexusNode is a BundleBridgeNexus that accepts a single reset vector source, and broadcasts it to all tiles. */
+  val tileResetVectorNexusNode = BundleBroadcast[UInt](
+    inputRequiresOutput = true // guard against this being driven but ignored in tileResetVectorIONodes below
+  )
+
+  /** tileHartIdIONodes may generate subsystem IOs, one per tile, allowing the parent to assign unique hart ids.
+    *
+    *   Or, if such IOs are not configured to exist, tileHartIdNexusNode is used to supply an id to each tile.
+    */
+  val tileHartIdIONodes: Seq[BundleBridgeSource[UInt]] = p(SubsystemExternalHartIdWidthKey) match {
```
mwachs5, line 148: I am still struggling with the relationships between p(MaxHartIdWidth) and p(SubsystemExternalHartIdWidthKey). 

I believe that the former is the width that will be used in $mhartid CSR. So should p(SubsystemExternalHartIdWidthKey) (if defined) be <= p(MaxHartIdBits)? Or is it allowed to be smaller?

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -10,25 +10,47 @@ import freechips.rocketchip.devices.tilelink.{BasicBusBlocker, BasicBusBlockerPa
...
 import freechips.rocketchip.interrupts._
-import freechips.rocketchip.tile.{BaseTile, LookupByHartIdImpl, TileParams, HasExternallyDrivenTileConstants, InstantiableTileParams, PriorityMuxHartIdFromSeq}
+import freechips.rocketchip.tile.{BaseTile, LookupByHartIdImpl, TileParams, InstantiableTileParams, MaxHartIdBits}
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.util._
 
 /** Entry point for Config-uring the presence of Tiles */
 case class TilesLocated(loc: HierarchicalLocation) extends Field[Seq[CanAttachTile]](Nil)
 
+/** Whether a global hart id prefix is going to be dynamically applied to the static tile hart ids.
```
mwachs5, line 20:
```suggestion
/** Whether a global hart id prefix is going to be dynamically applied to the static tile hart ids via external inputs.
```

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -10,25 +10,47 @@ import freechips.rocketchip.devices.tilelink.{BasicBusBlocker, BasicBusBlockerPa
...
+  *
+  * This information is used to conditionally add timing-closure registers along the path of
+  * the dynamically-configurable hart id as it makes its way into the tile.
```
mwachs5, line 23:
```suggestion
  * the dynamically-configurable hart id prefix as it makes its way into the tile.
```

src/main/scala/subsystem/HasTiles.scala
mwachs5: I'm not sure if these width statements are correct.

src/main/scala/subsystem/HasTiles.scala
mwachs5: Looking at how this is used, shouldn't this be called something like "RegisterHartPrefixKey" ? It has nothing to do with whether to actually apply the prefix or not, that's given by whoever connects to tileHartIdeNexusNode

src/main/scala/tile/RocketTile.scala
```diff
@@ -154,16 +154,13 @@ class RocketTileModuleImp(outer: RocketTile) extends BaseTileModuleImp(outer)
...
   outer.bpwatchSourceNode.bundle <> core.io.bpwatch
-  outer.frontend.module.io.reset_vector := constants.reset_vector
-
-  def regHart(x: UInt): UInt = if (p(HartPrefixKey)) RegNext(x) else x
-  core.io.hartid := regHart(constants.hartid)
-  outer.dcache.module.io.hartid := regHart(constants.hartid)
-  outer.frontend.module.io.hartid := regHart(constants.hartid)
+  core.io.hartid := outer.hartIdSinkNode.bundle
+  require(core.io.hartid.getWidth >= outer.hartIdSinkNode.bundle.getWidth,
+    s"core hartid wire (${core.io.hartid.getWidth}b) truncates external hartid wire (${outer.hartIdSinkNode.bundle.getWidth}b)")
```
mwachs5, line 163: ok I see that this is where the width check is happening to make sure it doesnt get truncated, but it seems like people hooking up to tileHartIdNexusNode can do as they please and it won't be checked until later, wheras we COULD have checked the widths of the wires in the tileHartIdNexusNode to ensure none of them were > p(MaxHartIdBits)?

hcook: The nexus node internally contains usages of `chisel3.DataMirror` that ensures that all the bundles on input and output edges are identical widths to one another, but you are right that it doesn't check against `p(MaxHartIdBits)`. But since now only the core is using ``p(MaxHartIdBits)`, it seems right to me that only the core checks against ``p(MaxHartIdBits)`? Like if we changed it to be part of `CoreParams` or `TileParams`, why should we have to change code in `subsystem` package, ideally?

src/main/scala/subsystem/HasTiles.scala
hcook: Previously it did both, now it only does the what comment says. I just kept the name the same since closed source is using it, but you're right that it would be better to update it and it's usages to be more clearly named. I'll change the name but I don't agree with your comment; most of the registers that are added contain the entire bit-wise-or'd hart id, not just the prefix.

src/main/scala/subsystem/HasTiles.scala
hcook: But in the closed source, these prefixes are driven by devices that may or may not be located in the subsystem and bundle-bridged to this node. So I would not necessarily say they are "external".

src/main/scala/subsystem/HasTiles.scala
hcook: IMO its allowed to be smaller.

src/main/scala/subsystem/HasTiles.scala
mwachs5: cool

src/main/scala/subsystem/HasTiles.scala
hcook: You are right about the widths.

But currently (and this is just me preserving the extant behavior) it isn't just the inputs that are registered, but rather the inputs, the or'd value fanning out to the tile, the tile inputs, and the value fanning out within the tile (when `p(HartPrefixKey) == true`).

src/main/scala/tile/RocketTile.scala
mwachs5: yeah, I agree with this. This is the place that cares so it seems right that this is the place that checks.

src/main/scala/subsystem/HasTiles.scala
mwachs5: Ok, and if it were larger, the truncation check would catch it, so I think this is OK

src/main/scala/subsystem/HasTiles.scala
mwachs5: ok, do you want me to edit my suggestion to try to capture that?

hcook: No thank you, I will do it now as I change the name of the `HartPrefixKey`

src/main/scala/tile/RocketTile.scala
hcook: Actually looking at the code again, I'm currently supply `0.U(p(MaxHartIdBits)` as the default value, which is treated as an input for the purposes of the `DataMirror` equivalency check. So it's already handled. I'll leave the more specific check in the core code though.

src/main/scala/subsystem/HasTiles.scala
hcook: After further digging around in the closed source side, i found this:
`case MaxHartIdBits => 10 // maximum possible before debug unit requirements fail`
which implies to me that not only can it be smaller it in fact MUST be smaller.

This does make me wonder whether I should switch the core IOs over to just using `XLen` though, rather than have this somewhat extraneous parameter floating around in core-parameter-land just to make a layer or two of still-narrow IOs that then get put into an `XLen`-sized CSR?

mwachs5: that might be a question for @aswaterman

hcook: Oh actually it looks like the physical CSR is actually sized by the width of this IO wire as well, so i will leave the `hartIdLen` parameter present to control that.

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -10,25 +10,48 @@ import freechips.rocketchip.devices.tilelink.{BasicBusBlocker, BasicBusBlockerPa
...
 /** Entry point for Config-uring the presence of Tiles */
 case class TilesLocated(loc: HierarchicalLocation) extends Field[Seq[CanAttachTile]](Nil)
 
+/** Whether to add timing-closure registers along the path of the hart id
+  * as it propagates through the subsystem and into the tile.
+  *
+  * These are typically only desirable when a dynamically programmable prefix is being combined
+  * with the static hart id via [[freechips.rocketchip.subsystem.HasTiles.tileHartIdNexusNode]].
+  */
+case object InsertTimingClosureRegistersOnHartIds extends Field[Boolean](false)
```
mwachs5, line 26: 👍

mwachs5:
can you update the release notes sectoin of the PR comments?

hcook:
This PR's CI seems to be permanently stuck. Github action/cache always hits, but returns cache sizes of 0MB. Opening a new PR #2531 to replace this one seems to have magically fixed it.

src/main/scala/diplomacy/BundleBridge.scala
```diff
@@ -56,8 +61,12 @@ case class BundleBridgeSource[T <: Data](genOpt: Option[() => T] = None)(implici
...
 
+  private def inferInput = bundle.getElements.forall { elt =>
+    DataMirror.directionOf(elt) == ActualDirection.Unspecified
+  }
+
   def makeIO()(implicit valName: ValName): T = {
-    val io: T = IO(Flipped(chiselTypeClone(bundle)))
+    val io: T = IO(if (inferInput) Input(chiselTypeOf(bundle)) else Flipped(chiselTypeClone(bundle)))
     io.suggestName(valName.name)
     bundle <> io
```
jerryhethatday, line 71: I just felt confused here.  By code reading, I felt like the direction of makeIO of BundleBridgeSourceNode is Input, While the direction of makeIO of BundleBridgeSinkNode is Output. I felt like this is strange, the source node only has output bundles and the sink only gets input right? Why this is opposite here. Also,  if inferInput is false in the makeIO of BundleBridgeSource, why we need to Flipped the orignal direction, while inside the BundleBridgeSink, there is no Flipped?  it is wired. Seems like  the io after calling makeIO in corresponding node is opposite with its original IO, why? 
@hcook @mwachs5 @jackkoenig

jerryhethatday: > I just felt confused here. By code reading, I felt like the direction of makeIO of BundleBridgeSourceNode is Input, While the direction of makeIO of BundleBridgeSinkNode is Output. I felt like this is strange, the source node only has output bundles and the sink only has input right? Why this is opposite here. Also, if inferInput is false in the makeIO of BundleBridgeSource, why we need to Flipped the orignal direction, while inside the BundleBridgeSink, there is no Flipped? it is wired. Seems like the io after calling makeIO in corresponding node is opposite with its original IO, why?
> @hcook @mwachs5 @jackkoenig

I answer (more or less) this question in stackoverflow https://stackoverflow.com/questions/66963837/how-to-understand-the-flip-in-autobundle-and-in-makeios

PR #2530, opened 2020-06-23
github ci: attempt to use cache@v2

hcook:
maybe using the official release v2 will somehow magically improve the hit rate?


aswaterman:
🤷

hcook:
It got 7/8, which is better than my other PR has been doing :/

PR #2533, opened 2020-06-25
Bundle Bridge Tile MMIO Address Prefix

hcook:
In some configurations we want to be able to relocate the apparent base address of MMIO devices inside the tile. This is problematic for the fast path that the RocketCore has to the ITIM and DTIM, which has to know what the overridden base address should be. To address this limitation this PR adds another (optional) tile input "constant" that is or'd with the base address used in the ITIM and DTIM hit calculations.

This PR also cleans up a few loose ends left over from #2521 and #2497 

**Impact**: API modification

**Release Notes**
- `BaseTile.mmioAddressPrefixNode` propagates a prefix or supplies a default `0.U`
- `TileCrossingParamsLike.controlBaseAddressPrefixWhere` to lookup prefix information location for tiles
- `mmioAddressPrefixSinkNodeOpt` and `hartIdSinkNodeOpt` in `HellaCache` and `ICache`, defined iff the scratchpads are
- certain `TLBusWrapper` subclasses will propagate their `prefix` information

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -55,7 +55,7 @@ class ICacheErrors(implicit p: Parameters) extends CoreBundle()(p)
 class ICache(val icacheParams: ICacheParams, val staticIdForMetadataUseOnly: Int)(implicit p: Parameters) extends LazyModule {
   lazy val module = new ICacheModule(this)
   val hartIdSinkNodeOpt = icacheParams.itimAddr.map(_ => BundleBridgeSink[UInt]())
-  val mmioAddressPrefixSinkNodeOpt= icacheParams.itimAddr.map(_ => BundleBridgeSink[UInt]())
+  val mmioAddressPrefixSinkNodeOpt = icacheParams.itimAddr.map(_ => BundleBridgeSink[UInt]())
   val useVM = p(TileKey).core.useVM
   val masterNode = TLClientNode(Seq(TLMasterPortParameters.v1(
     clients = Seq(TLMasterParameters.v1(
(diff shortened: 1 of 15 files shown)


src/main/scala/rocket/ICache.scala
```diff
@@ -54,7 +54,8 @@ class ICacheErrors(implicit p: Parameters) extends CoreBundle()(p)
 
 class ICache(val icacheParams: ICacheParams, val staticIdForMetadataUseOnly: Int)(implicit p: Parameters) extends LazyModule {
   lazy val module = new ICacheModule(this)
-  val hartIdSinkNode = BundleBridgeSink[UInt]()
+  val hartIdSinkNodeOpt = icacheParams.itimAddr.map(_ => BundleBridgeSink[UInt]())
+  val mmioAddressPrefixSinkNodeOpt= icacheParams.itimAddr.map(_ => BundleBridgeSink[UInt]())
```
mwachs5, line 58:
```suggestion
  val mmioAddressPrefixSinkNodeOpt = icacheParams.itimAddr.map(_ => BundleBridgeSink[UInt]())
```

src/main/scala/tile/BaseTile.scala
```diff
@@ -201,6 +201,13 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
   val resetVectorSinkNode = BundleBridgeSink[UInt](Some(() => UInt(visiblePhysAddrBits.W)))
   resetVectorSinkNode := resetVectorNode
 
+  /** Node for prefixing base addresses of MMIO slaves. */
```
mwachs5, line 204: does this prefix work the same as the other (it is OR-ed in) or is it actually a prefix?

hcook: Yeah similarly or'd. Open to a better name for describing this pattern but it seems to be what we are converging on at the wiring level.

PR #2535, opened 2020-06-25
build.wake: Fix string interpolation of bootrom path.

richardxia:
**Related issue**: 

**Release Notes**

I can't actually add @mmjconolly as a reviewer, since he does not have write access to this repo, so @jackkoenig will have to be one to actually approve this PR.

In https://github.com/chipsalliance/rocket-chip/pull/2526, one of the strings that we changed was a single-quoted string, which does not do string interpolation, which resulted in a string that literally had a value of `{rocketChipRoot}/bootrom/bootrom.img`. This fixes it to be a double-quoted string, which _is_ interpolated.

BTW, if either of you knows how to set up a Vim syntax highlighting file to specially highlight interpolated expressions within a string, it'd be nice to get this into the Wake Vim syntax highlighting file so that it's more obvious when a string is not interpolated: https://github.com/sifive/wake/blob/96a0323d49ab12a9313ec0b4c385329e2d0f4232/share/doc/wake/syntax/vim/syntax/wake.vim

### DIFF
--- a/build.wake
+++ b/build.wake
@@ -130,7 +130,7 @@ global def runRocketChipGenerator options =
     baseFileName
 
   def inputs =
-    def bootrom = source '{rocketChipRoot}/bootrom/bootrom.img'
+    def bootrom = source "{rocketChipRoot}/bootrom/bootrom.img"
     def extras = options.getRocketChipGeneratorOptionsExtraSources
     (bootrom, targetDir, extras) ++ jars
 


mmjconolly:
Ah I’ll be more careful with that in future

richardxia:
> Ah I’ll be more careful with that in future

No worries, I've totally made that mistake too, and in other languages as well.

PR #2538, opened 2020-06-25
Add CoreMonitorBundles to for late latency writes

abhinay-kayastha:
**Related issue**: 
The current coreMonitorBundle does not detect the late latency load writes and fp writes. This is adding the CoreMonitorBundle definitions to detect such cases.

**Release Notes**

### DIFF
--- a/src/main/scala/tile/FPU.scala
+++ b/src/main/scala/tile/FPU.scala
@@ -136,6 +136,9 @@ class FPUDecoder(implicit p: Parameters) extends FPUModule()(p) {
 }
 
 class FPUCoreIO(implicit p: Parameters) extends CoreBundle()(p) {
+  val hartid = Input(UInt(hartIdLen.W))
+  val time = Input(UInt(xLen.W))
+
   val inst = Bits(INPUT, 32)
   val fromint_data = Bits(INPUT, xLen)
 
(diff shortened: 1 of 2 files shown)


src/main/scala/tile/FPU.scala
```diff
@@ -610,15 +613,15 @@ class MulAddRecFNPipe(latency: Int, expWidth: Int, sigWidth: Int) extends Module
...
+
     val postmul_regs = if(latency>0) 1 else 0
     mulAddRecFNToRaw_postMul.io.fromPreMul   := Pipe(io.validin, mulAddRecFNToRaw_preMul.io.toPostMul, postmul_regs).bits
     mulAddRecFNToRaw_postMul.io.mulAddResult := Pipe(io.validin, mulAddResult, postmul_regs).bits
     mulAddRecFNToRaw_postMul.io.roundingMode := Pipe(io.validin, io.roundingMode, postmul_regs).bits
     roundingMode_stage0                      := Pipe(io.validin, io.roundingMode, postmul_regs).bits
     detectTininess_stage0                    := Pipe(io.validin, io.detectTininess, postmul_regs).bits
     valid_stage0                             := Pipe(io.validin, false.B, postmul_regs).valid
-    
+
```
aswaterman, line 624: Please remove these unrelated whitespace changes in MulAddRecFNPipe

abhinay-kayastha: Those spaces were actually there before which I removed hence the diff :)

PR #2543, opened 2020-06-29
HartId/ResetVector: Correct the input width calculation

mwachs5:
**Related issue**: Follow up to #2531 which introduced this bug (https://github.com/chipsalliance/rocket-chip/blame/master/src/main/scala/tile/Core.scala#L134)

**Impact**:  API modification

**Release Notes**

Fix a bug in the computation of HartID and ResetVector width going into the Tile due to improper use of Chisel compatibility mode functions.

### DIFF
--- a/src/main/scala/tile/Core.scala
+++ b/src/main/scala/tile/Core.scala
@@ -3,6 +3,7 @@
 package freechips.rocketchip.tile
 
 import Chisel._
+
 import freechips.rocketchip.config._
 import freechips.rocketchip.rocket._
 import freechips.rocketchip.util._
--- a/src/main/scala/tile/Core.scala
+++ b/src/main/scala/tile/Core.scala
@@ -3,7 +3,6 @@
 package freechips.rocketchip.tile
 
 import Chisel._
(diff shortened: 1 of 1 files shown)


src/main/scala/tile/Core.scala
```diff
@@ -3,7 +3,7 @@
 package freechips.rocketchip.tile
 
 import Chisel._
-
+import Chisel.ImplicitConversions._
```
jackkoenig, line 6: I don't actually understand how this provides `.W`

jackkoenig: Via indirection, this is just https://github.com/freechipsproject/chisel3/blob/387274784115bee2bf7167547a7ee459219e1413/src/main/scala/chisel3/util/ImplicitConversions.scala#L12 which provides implicit conversion from `Int` to `UInt` and `Boolean` to `Bool`.

mwachs5: ok, so I should just get rid of this import?

mwachs5, line 6:
```suggestion
```

PR #2547, opened 2020-07-02
SynchronizerPrimitiveShiftReg: fix Dedup Issue

mwachs5:
It was intended that the SynchronizerPrimitiveShiftReg be identical, as they have suggestName applied that was intended to capture all variance in terms of functionality.

However, because they had inferred reset type (the reset is cast appropriately *inside* the Primitive module), the FIRRTL would not be identical at the point of Dedup and you could end up with 2 versions. The final verilog (since verilog dosen't distinguish between reset types) would be identical.

Since the Primitives are intended to be easy to replace with library-specific macro sells for synchronization, this is undesirable.

This change pushes the reset type up into the SynchronizerShiftReg (which still may end up with multiple copies) to make sure you only get one SynchronizerPrimitiveShiftReg of each type.

**Related issue**: None

**Impact**:  API modification

**Release Notes**

Correct the dedup behavior for the *ResetSynchronizerPrimitiveShiftReg so that you only end up with one copy.

### DIFF
--- a/src/main/scala/util/SynchronizerReg.scala
+++ b/src/main/scala/util/SynchronizerReg.scala
@@ -32,6 +32,8 @@ object SynchronizerResetType extends Enumeration {
 }
 
 
+// Note: this should not be used directly.
+// Use the companion object to generate this with the correct reset type mixin.
 private class SynchronizerPrimitiveShiftReg(
   sync: Int,
   init: Boolean,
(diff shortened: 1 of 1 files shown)


src/main/scala/util/SynchronizerReg.scala
```diff
@@ -101,13 +108,17 @@ object AsyncResetSynchronizerShiftReg {
     apply (in, sync, init.litValue.toInt, None)
 }
 
+// Note: This module may end up with a non-Bool type reset.
```
mwachs5, line 111: hmm either this comment is wrong or the `with RequireSyncReset` should be removed

src/main/scala/util/SynchronizerReg.scala
```diff
@@ -101,13 +108,17 @@ object AsyncResetSynchronizerShiftReg {
...
+// But the Primitives within will always have AsyncReset type.
 @deprecated("SyncResetSynchronizerShiftReg is unecessary with Chisel3 inferred resets. Use ResetSynchronizerShiftReg which will use the inferred reset type.", "rocket-chip 1.2")
-class SyncResetSynchronizerShiftReg(w: Int = 1, sync: Int, init: Int) extends AbstractPipelineReg(w) {
+class SyncResetSynchronizerShiftReg(w: Int = 1, sync: Int, init: Int) extends AbstractPipelineReg(w) with RequireSyncReset {
```
mwachs5, line 114:
```suggestion
class SyncResetSynchronizerShiftReg(w: Int = 1, sync: Int, init: Int) extends AbstractPipelineReg(w) {
```

src/main/scala/util/SynchronizerReg.scala
```diff
@@ -101,13 +108,17 @@ object AsyncResetSynchronizerShiftReg {
     apply (in, sync, init.litValue.toInt, None)
 }
 
+// Note: This module may end up with a non-Bool type reset.
+// But the Primitives within will always have AsyncReset type.
```
mwachs5, line 112:
```suggestion
// But the Primitives within will always have Bool reset type.
```

mwachs5:
@mergifyio update

src/main/scala/util/SynchronizerReg.scala
```diff
@@ -100,13 +107,17 @@ object AsyncResetSynchronizerShiftReg {
     apply (in, sync, init.litValue.toInt, None)
 }
 
+// Note: This module may end up with a non-Bool type reset.
+// But the Primitives within will always have Bool reset type.
 @deprecated("SyncResetSynchronizerShiftReg is unecessary with Chisel3 inferred resets. Use ResetSynchronizerShiftReg which will use the inferred reset type.", "rocket-chip 1.2")
```
jackkoenig, line 112: Since you now have to set the reset type, is this message correct?

hcook (approved):

jackkoenig (approved):

src/main/scala/util/SynchronizerReg.scala
mwachs5, line 112:
```suggestion
// Note: This module may end up with a non-Bool type reset.
// But the Primitives within will always have Bool reset type.
@deprecated("SyncResetSynchronizerShiftReg usage is discouraged. Use ResetSynchronizerShiftReg which will use the inferred reset type. Cast  the reset to Bool if you truly want to enforce SyncResetSynchronizerShiftReg functionality.", "rocket-chip 1.2")
```

Hmm, good point. What do you think of this instead (at least I should learn how to spell unnecessary)?

PR #2550, opened 2020-07-11
Tile PRCI Domain Wrappers

hcook:
Tiles are now instantiated inside of a wrapper whose purpose is to contain logic related to their Power, Clock, Reset and/or Interrupt domain. In particular:
- The layer has a diplomatically-derived clock and reset, which are used to drive the tile clock and reset.
- The portion of TileLink clock crossings that are supposed to be in the tile clock domain are instantiated in this layer.
- Interrupt synchronizers are instantiated in this layer.
- Boundary buffers and isolation cells related to hierarchical P&R are to be instantiated in this layer.
- In the future core-local interrupt handlers could be instantiated in this layer.

**Related issue**: 

**Impact**:  API modification

**Release Notes**
- All tiles are now instantiated inside of a wrapper that contains logic related to their power, reset, clock and interrupt domain. This change adds a level to the module hierarchy path to the tile.

### DIFF
--- a/src/main/scala/subsystem/HasTiles.scala
+++ b/src/main/scala/subsystem/HasTiles.scala
@@ -50,6 +50,8 @@ trait TileCrossingParamsLike {
   def mmioBaseAddressPrefixWhere: TLBusWrapperLocation
   /** Inject a clock/reset management subgraph */
   def injectClockNode(context: Attachable)(implicit p: Parameters): ClockNode
+  /** Keep the tile clock separate from the interconnect clock (e.g. even if they are synchronous to one another) */
+  def forceSeparateClockReset: Boolean
 }
 
 /** An interface for describing the parameterization of how a particular tile port is connected to an interconnect */
@@ -333,7 +335,9 @@ trait CanAttachTile {
     domain.clockNode :=
       crossingParams.injectClockNode(context) :=
       (crossingParams.crossingType match {
-        case _: SynchronousCrossing => tlBusToGetClockDriverFrom.fixedClockNode
+        case s: SynchronousCrossing =>
+          if (crossingParams.forceSeparateClockReset) tlBusToGetClockDriverFrom.clockNode
+          else tlBusToGetClockDriverFrom.fixedClockNode
         case _: RationalCrossing => tlBusToGetClockDriverFrom.clockNode
         case _: AsynchronousCrossing => {
           val tileClockGroup = ClockGroup()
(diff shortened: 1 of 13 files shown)


src/main/scala/subsystem/HasTiles.scala
```diff
@@ -223,98 +226,121 @@ trait CanAttachTile {
...
-    }
+    tlBusToGetPrefixFrom.prefixNode.foreach { tile.mmioAddressPrefixNode := _ }
+  }
+
+  /** Connect power/reset/clock resources. */
+  def connectPRC(domain: TilePRCIDomain[TileType], context: TileContextType): Unit = {
+    implicit val p = context.p
+    val tlBusToGetClockDriverFrom = context.locateTLBusWrapper(crossingParams.master.where)
+    domain.clockNode :=
+      crossingParams.injectClockNode(context) :=
```
rmac-sifive, line 334: I think the answer is "yes" based on my read but want to confirm: this supports injecting unique sequences of clock nodes per tile, right?

rmac-sifive: Also, hierarchically where do the injected clock nodes reside?

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -348,8 +375,8 @@ trait HasTiles extends InstantiatesTiles with HasCoreMonitorBundles with Default
   implicit val p: Parameters
 
   // connect all the tiles to interconnect attachment points made available in this subsystem context
-  tileAttachParams.zip(tiles).foreach { case (params, t) =>
-    params.connect(t.asInstanceOf[params.TileType], this.asInstanceOf[params.TileContextType])
+  tileAttachParams.zip(tile_prci_domains).foreach { case (params, td) =>
+    params.connect(td.asInstanceOf[TilePRCIDomain[params.TileType]], this.asInstanceOf[params.TileContextType])
```
rmac-sifive, line 379: it seems desirable that someday we might want to support more than 1 tile per PRCI domain wrapper

src/main/scala/subsystem/HasTiles.scala
hcook: Yes, and new wrapper that is hierarchically now the parent of each tile.

src/main/scala/subsystem/HasTiles.scala
hcook: I certainly wouldn't rule that out, but if those tiles share some resources like e.g. a shared cache probably all three of them live inside a ClusteredTilesWrapper, which might either have its own clock sink node or just be wrapped in one of these clock domain wrappers itself. Alternatively we can always (configurably?) make wrappers around these wrappers... put another way, maybe this wrapper is just a `DevicesSubsystem` with a single tile "device" inside it, if you upstreamed that and we re-expressed it that way :)

PR #2553, opened 2020-07-12
Port LazyRoCC examples to chisel3

kammoh:
Convert LazyRoCC examples to use chisel3 syntax instead of legacy Chisel compatibility.

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/tile/LazyRoCC.scala
+++ b/src/main/scala/tile/LazyRoCC.scala
@@ -3,12 +3,12 @@
 
 package freechips.rocketchip.tile
 
-import Chisel._
+import chisel3._
+import chisel3.util._
 import chisel3.util.HasBlackBoxResource
 import chisel3.experimental.IntParam
-
+import chisel3.internal.naming.chiselName
 import freechips.rocketchip.config._
-import freechips.rocketchip.subsystem._
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.rocket._
 import freechips.rocketchip.tilelink._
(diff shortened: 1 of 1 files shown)


src/main/scala/tile/LazyRoCC.scala
```diff
@@ -182,26 +183,27 @@ class AccumulatorExampleModuleImp(outer: AccumulatorExample)(implicit p: Paramet
...
   override lazy val module = new TranslatorExampleModuleImp(this)
 }
 
+@chiselName
 class TranslatorExampleModuleImp(outer: TranslatorExample)(implicit p: Parameters) extends LazyRoCCModuleImp(outer)
     with HasCoreParameters {
-  val req_addr = Reg(UInt(width = coreMaxAddrBits))
-  val req_rd = Reg(io.resp.bits.rd)
+  val req_addr = Reg(UInt(coreMaxAddrBits.W))
+  val req_rd = Reg(io.resp.bits.rd.cloneType)
```
seldridge, line 200: I think use of `cloneType` is discouraged in favor of `chiselTypeOf`.

```suggestion
  val req_rd = Reg(chiselTypeOf(io.resp.bits.rd))
```

src/main/scala/tile/LazyRoCC.scala
```diff
@@ -248,32 +251,32 @@ class CharacterCountExampleModuleImp(outer: CharacterCountExample)(implicit p: P
...
   private val beatOffset = log2Up(cacheDataBits/8)
 
-  val needle = Reg(UInt(width = 8))
-  val addr = Reg(UInt(width = coreMaxAddrBits))
-  val count = Reg(UInt(width = xLen))
-  val resp_rd = Reg(io.resp.bits.rd)
+  val needle = Reg(UInt(8.W))
+  val addr = Reg(UInt(coreMaxAddrBits.W))
+  val count = Reg(UInt(xLen.W))
+  val resp_rd = Reg(io.resp.bits.rd.cloneType)
```
seldridge, line 257:
```suggestion
  val resp_rd = Reg(chiselTypeOf(io.resp.bits.rd))
```

src/main/scala/tile/LazyRoCC.scala
```diff
@@ -357,7 +360,7 @@ class BlackBoxExampleModuleImp(outer: BlackBoxExample, blackBoxFile: String)(imp
                    ) ) with HasBlackBoxResource {
         val io = IO( new Bundle {
                       val clock = Input(Clock())
-                      val reset = Input(Bool())
+                      val reset = Input(Reset())
```
seldridge, line 363: [The reference RoCC is currently hard-coded to synchronous reset](https://github.com/chipsalliance/rocket-chip/blob/52843da3d8daaba5c32c488ee21d20aa2f640772/src/main/resources/vsrc/RoccBlackBox.v#L170).

I think this means that this should stay as `Bool` here or this needs to mix-in the `RequireSyncReset` trait.

kammoh: I had overlooked that! `RequireSyncReset` mixin sounds neat.

kammoh: > Thanks for doing this @kammoh! Generally, looks fine. Comments inline.
> 
> By chance have you tested this with any RoCCs? (By inspection, I see no issues.)

Thanks for the review! I had only tested `AccumulatorExample` and `CharacterCountExample`.  `AccumulatorExample` seemed to run the `chipyard` example program correctly. `CharacterCountExample` did not work correctly even before the change in this PR. It seems the current code only works when the target strings is aligned to cache line size, i.e. It works when adding `__attribute__ ((aligned (CacheBlockBytes)))` to `string`.

kammoh:
@hcook is there anything you'd like me to change here?

PR #2555, opened 2020-07-14
Add CreditedIO to support register-register interfacing for AXI4 and TL

terpstra:
This PR adds a new kind of flow control, CreditedIO, which fills the same role for credit/debit systems that DecoupledIO fills for ready/valid systems. Included in this PR are conversion methods to and from Decoupled as well as adapters for TileLink and AXI4.

The advantage of a credit/debit system is that you can add an arbitrary number of fanout=1 fanin=1 pipeline stages in both the data delivery and credit return paths. By comparison, adding ready/valid pipeline stages invariably includes at least a fan of degree 2.

See the following design diagram: https://app.lucidchart.com/invitations/accept/8271ee7b-98ba-44a4-ab3e-e8ae5af9fb43

**Type of change**:  feature request

### DIFF
--- a/src/main/scala/interrupts/CrossingHelper.scala
+++ b/src/main/scala/interrupts/CrossingHelper.scala
@@ -4,6 +4,7 @@ package freechips.rocketchip.interrupts
 
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.util.CreditedDelay
 
 case class IntInwardCrossingHelper(name: String, scope: LazyScope, node: IntInwardNode) {
   def apply(xing: ClockCrossingType = NoCrossing, alreadyRegistered: Boolean = false)(implicit p: Parameters): IntInwardNode = {
@@ -14,6 +15,8 @@ case class IntInwardCrossingHelper(name: String, scope: LazyScope, node: IntInwa
         node :*=* scope { IntSyncRationalCrossingSink() :*=* IntSyncNameNode(name) } :*=* IntSyncNameNode(name) :*=* IntSyncCrossingSource(alreadyRegistered)
       case SynchronousCrossing(_) =>
         node :*=* scope { IntSyncSyncCrossingSink() :*=* IntSyncNameNode(name) } :*=* IntSyncNameNode(name) :*=* IntSyncCrossingSource(alreadyRegistered)
+      case CreditedCrossing(CreditedDelay(sourceDebit, _), CreditedDelay(sinkDebit, _)) =>
+        node :*=* scope { IntSyncSyncCrossingSink(/*sinkDebit==0*/) :*=* IntSyncNameNode(name) } :*=* IntSyncNameNode(name) :*=* IntSyncCrossingSource(sourceDebit==0)
     }
   }
 }
@@ -27,6 +30,8 @@ case class IntOutwardCrossingHelper(name: String, scope: LazyScope, node: IntOut
         IntSyncRationalCrossingSink() :*=* IntSyncNameNode(name) :*=* scope { IntSyncNameNode(name) :*=* IntSyncCrossingSource(alreadyRegistered) } :*=* node
       case SynchronousCrossing(buffer) =>
         IntSyncSyncCrossingSink() :*=* IntSyncNameNode(name) :*=* scope { IntSyncNameNode(name) :*=* IntSyncCrossingSource(alreadyRegistered) } :*=* node
+      case CreditedCrossing(CreditedDelay(sourceDebit, _), CreditedDelay(sinkDebit, _)) =>
+        IntSyncSyncCrossingSink(/*sinkDebit==0*/) :*=* IntSyncNameNode(name) :*=* scope { IntSyncNameNode(name) :*=* IntSyncCrossingSource(sourceDebit==0) } :*=* node
     }
   }
 }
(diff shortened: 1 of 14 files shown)


src/main/scala/util/CreditedIO.scala
```diff
@@ -0,0 +1,152 @@
...
+    val res = Wire(CreditedIO(genType))
+    if (debitDelay <= 0) {
+      credit := ShiftRegister(res.credit, creditDelay)
+      res.debit := debit
+      res.bits  := bits
+    } else {
+      // We can't use ShiftRegister, because we want debit-gated enables
+      val out = pipeline(debitDelay-1, creditDelay)
+      out.credit := res.credit
+      res.debit := RegNext(out.debit)
```
aswaterman, line 96: Need to reset this, right?

src/main/scala/util/CreditedIO.scala
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  /** Add register stages to the sender and receiver paths.
+    * Apply this method to the producer/sender-facing bundle.
+    * The round-trip-time (RTT) is increased by sender+receiver.
+    */
+  def pipeline(debitDelay: Int, creditDelay: Int): CreditedIO[T] = {
+    val res = Wire(CreditedIO(genType))
+    if (debitDelay <= 0) {
+      credit := ShiftRegister(res.credit, creditDelay)
```
aswaterman, line 89: If this also needs to be reset, you can do `Pipe(res.credit, false.B, creditDelay).valid`, which generates a shfit register whose contents is reset.

src/main/scala/util/CreditedIO.scala
terpstra: Indeed. Thanks.

src/main/scala/util/CreditedIO.scala
terpstra: I just used ShiftRegister's extra args. Seems more readable than creating a Valid chain.

aswaterman: agreed, forgot there was such a thing

terpstra:
I'll hold off on merging this a bit to see what @hcook and @rmac-sifive think. I've noticed that adding a new ClockCrossingType causes quite a number of unmatched case warnings in sifive-blocks. It seems like that codebase used a lot of copy-paste. If I refactor that, the number of places that actually need to be touched in the larger ecosystem seems to only be 3.

src/main/scala/util/CreditedIO.scala
```diff
@@ -0,0 +1,152 @@
...
+{
+  override def cloneType: this.type = new CreditedIO(genType).asInstanceOf[this.type]
+  def genType: T = gen
+
+  val credit = Input (Bool()) // 1: a credit is given to the sender by the receiver
+  val debit  = Output(Bool()) // 1: a credit is consumed by the sender to transfer 'bits'
+  val bits   = Output(genType)
+
+  /** Provide a DecoupledIO interface for sending CreditedIO[Data].
+    * Convert an IrrevocableIO input to DecoupledIO via Decoupled().
```
hcook, line 48: `IrrevocableIO`?

src/main/scala/interrupts/CrossingHelper.scala
```diff
@@ -27,6 +30,8 @@ case class IntOutwardCrossingHelper(name: String, scope: LazyScope, node: IntOut
         IntSyncRationalCrossingSink() :*=* IntSyncNameNode(name) :*=* scope { IntSyncNameNode(name) :*=* IntSyncCrossingSource(alreadyRegistered) } :*=* node
       case SynchronousCrossing(buffer) =>
         IntSyncSyncCrossingSink() :*=* IntSyncNameNode(name) :*=* scope { IntSyncNameNode(name) :*=* IntSyncCrossingSource(alreadyRegistered) } :*=* node
+      case CreditedCrossing(CreditedDelay(sourceDebit, _), CreditedDelay(sinkDebit, _)) =>
+        IntSyncSyncCrossingSink(/*sinkDebit==0*/) :*=* IntSyncNameNode(name) :*=* scope { IntSyncNameNode(name) :*=* IntSyncCrossingSource(sourceDebit==0) } :*=* node
```
hcook, line 34: why is sinkDebit ignored here?

src/main/scala/util/CreditedIO.scala
terpstra: What are you asking? You want me to remove that line of commentary?
While the sender function returns a DecoupledIO, you may want to connect an IrrevocableIO.

src/main/scala/interrupts/CrossingHelper.scala
terpstra: The IntSyncCrossingSink does not support a register stage. I could change that. Should I?

src/main/scala/util/CreditedIO.scala
hcook: I just didn't see `IrrevocableIO` mentioned anywhere in this particular function. `this` is `CreditedIO` and as you the say the return type is `DecoupledIO`. I guess I would find "To connect an IrrevocableIO sender, first convert it with `Decoupled()`" to be clearer phrasing to me?

src/main/scala/interrupts/CrossingHelper.scala
hcook: No I guess I would just leave it, though I confess I don't really understand the semantics of IntSync are as it relates to what circuits present "safe" crossings. Would a user explicitly controlling `sinkDebit` be surprised at the overall result in either direction?

PR #2558, opened 2020-07-15
plusarg default bit width

ingallsj:

**Release Notes**
fix VCS Lint warning message:
```
Lint-[WMIA-L] Width mismatch in assignment
plusarg_reader.v, 18
  Width mismatch between LHS and RHS is found in assignment:
  The following 32-bit wide expression is assigned to a 6-bit LHS target:
  Source info: myplus = DEFAULT;
  Expression: myplus
```


ingallsj:
as far as I'm aware of, yes, but that is a risk that we'll have to take for us to use DEFAULT values over 32 bits wide

PR #2560, opened 2020-07-17
Mwachs Nodes.scala Scaladoc edits (see #2308)

mwachs5:
**Related issue**: Should really be a PR to #2308 , but that is @sequencer 's fork

**Release Notes**

Adding documentation to diplomacy package especially Nodes


sequencer:
Oh, I'm so sorry, I totally missed this.
I'll close #2308 and work on this.

sequencer:
Just rebased onto master.

sequencer:
I just noticed the ascii was reverted. can I keep `Anatomy of a Node` doc? I think it's helpful to understand val and def intrinsic connection inside nodes.

sequencer:
Just recall of that graph :p
If not appropriate, just rebase and drop it.

sequencer:
rebased some commits for better history.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -5,123 +5,328 @@ package freechips.rocketchip.diplomacy
...
+
+import scala.collection.immutable
 import scala.collection.mutable.ListBuffer
-import scala.util.matching._
 
+/** A field available in [[Parameters]] used to determine whether [[InwardNodeImp.monitor]] will be called.
+ */
 case object MonitorsEnabled extends Field[Boolean](true)
+
+/** When rendering the edge in graphical format (e.g. GraphML), flip the direction. */
```
terpstra, line 18: This comment is not quite correct. The arrow for the edge stays in the same direction.

Normally, yEd tries to put the source node vertically above the sink node. RenderFlipped inverts this. In other words, the sink gets place vertically above the source. This is useful for things like Tiles which include both masters and slaves. You want the tiles at the top of the diagram. The master interfaces for the tile will be correctly placed vertically above the rest of the SoC. However, the slave interfaces would normally get placed near the bottom of the SoC. RenderFlipped on the edges leading to control tile slaves ensures that the tile can be placed at the top of the diagram despite this.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -5,123 +5,328 @@ package freechips.rocketchip.diplomacy
...
+  * @tparam U type of the Upwards flowing parameters of the node
+  * @tparam E Edge Parameters describing the connections on either side of the node
+  * @tparam B Bundle type generated on edges connecting to this node.
+  *
+  *           If your edges have the same type in both directions, using this saves you some typing.
+  */
 abstract class SimpleNodeImp[D, U, E, B <: Data]
   extends NodeImp[D, U, E, E, B]
 {
+  /** Create an edge that that passes through this node.
```
terpstra, line 151: Edges do not "pass through nodes".

edge creates the edge parameters out of the downward flowing and upward flowing parameters for edges that connect to this node.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -5,123 +5,328 @@ package freechips.rocketchip.diplomacy
...
+    */
+  def bundleI(e: E): B = bundle(e)
 }
 
+/** [[BaseNode]] is the abstract base class of diplomacy node system.
+  * 
+  *  While [[*NodeImp]] associates the different types in a diplomatic
+  *  system, the [[BaseNode]] uses the parameter types to build
+  *  actual nodes and connect them with edges in the graph through binding
+  *  operations.
```
terpstra, line 194: I would not say that BaseNode uses anything. Rather, sometimes one wants to view the entire diplomacy graph in a way where you do not care about the specific types of the edges. BaseNodes are type-erased and provide this view.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -5,123 +5,328 @@ package freechips.rocketchip.diplomacy
...
+  *  [[BaseNode]]s are expected to be instantiated only within [[LazyModule]]s.
```
terpstra, line 195: No one instantiates BaseNodes. You always instantiate Nodes of concrete type.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -130,223 +335,765 @@ trait FormatNode[I <: FormatEdge, O <: FormatEdge] extends BaseNode {
...
+/** Enumeration of types of binding operations. */
 sealed trait NodeBinding
+
+/** Only connects a single edge. */
 case object BIND_ONCE  extends NodeBinding
+
+/** Connects N (N >= 0) edges.
+  * 
+  *  The number of edges N is determined by the left side via other connections.
+  */
```
terpstra, line 583: It's not left/right. The star can be on either side.
BIND_QUERY means the OTHER side of the edge determines cardinality.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -130,223 +335,765 @@ trait FormatNode[I <: FormatEdge, O <: FormatEdge] extends BaseNode {
...
 case object BIND_QUERY extends NodeBinding
+
+/** Connect N (N >= 0) edges.
+  *  The number of edges N is determined by the right side via other connections.
+  */
 case object BIND_STAR  extends NodeBinding
```
terpstra, line 589: This one means that our side of the edge will decide the cardinality.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -558,45 +1375,58 @@ class MixedAdapterNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
     val dangles = super.instantiate()
-    (out zip in) map { case ((o, _), (i, _)) => o <> i }
+    (out zip in) foreach { case ((o, _), (i, _)) => o <> i }
     dangles
-  } 
+  }
 }
 
-// EphemeralNodes disappear from the final node graph
+/** EphemeralNodes are used for temporary connections, but disappear from the final node graph. */
```
terpstra, line 1410: The connections are not temporary, the node is. An ephemeral node provides a mechanism to directly connect two nodes to each other where neither node knows about the other, but both know about an ephemeral node they can use to facilitate the connection.

sequencer:
Thanks @terpstra!
I move my updates in #2604, wish can be reviewed there.

hcook:
Closed in favor of #2604

src/main/scala/diplomacy/Nodes.scala
jerryhethatday: > This one means that our side of the edge will decide the cardinality.
I feel confused about this, Am I right about this: :*= is  BINDSTAR for the lhs, and BINDQUERY for the rhs. :=* is the BINDQUERY for the lhs and BINDSTAR for the rhs.

src/main/scala/diplomacy/Nodes.scala
jerryhethatday: > It's not left/right. The star can be on either side.
> BIND_QUERY means the OTHER side of the edge determines cardinality.

This one means that our side of the edge will decide the cardinality.
I feel confused about this, Am I right about this: := is BINDSTAR for the lhs, and BINDQUERY for the rhs. := is the BINDQUERY for the lhs and BINDSTAR for the rhs.

jerryhethatday: :\*= does mean the cadinality N is surely determined by the left side, right?

PR #2566, opened 2020-07-21
Add Reset Stretcher in PRCI Domain for async reset systems

ernie-sifive:
**Related issue**: 

**Release Notes**

The Rocket Tile is not currently written to support an asynchronous reset.  When the rest of a system has an asynchronous reset configured by SubsystemResetSchemeKey, a module is inserted into the reset path in the PRCI domain that synchronizes and stretches the core_reset to be a minimum of 16 cycles as required by the core design.  The Tile and everything it contains is generated with synchronous reset.

### DIFF
--- a/src/main/scala/subsystem/RocketSubsystem.scala
+++ b/src/main/scala/subsystem/RocketSubsystem.scala
@@ -17,11 +17,12 @@ case class RocketCrossingParams(
   crossingType: ClockCrossingType = SynchronousCrossing(),
   master: TileMasterPortParams = TileMasterPortParams(),
   slave: TileSlavePortParams = TileSlavePortParams(),
-  mmioBaseAddressPrefixWhere: TLBusWrapperLocation = CBUS
+  mmioBaseAddressPrefixWhere: TLBusWrapperLocation = CBUS,
+  stretchResetCycles: Option[Int] = None
 ) extends TileCrossingParamsLike {
   def injectClockNode(context: Attachable)(implicit p: Parameters): ClockNode = {
-    if (p(SubsystemResetSchemeKey) != ResetSynchronous) {
-      val rs = LazyModule(new ResetStretcher(16))
+    if (stretchResetCycles.isDefined) {
+      val rs = LazyModule(new ResetStretcher(stretchResetCycles.get))
       rs.node
     } else {
       ClockTempNode()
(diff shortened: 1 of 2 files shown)


src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -20,7 +20,12 @@ case class RocketCrossingParams(
   mmioBaseAddressPrefixWhere: TLBusWrapperLocation = CBUS
 ) extends TileCrossingParamsLike {
   def injectClockNode(context: Attachable)(implicit p: Parameters): ClockNode = {
-    ClockTempNode() // TODO reset stretcher could go here, parameterized by args to this case class
+    if (p(SubsystemResetSchemeKey) != ResetSynchronous) {
+      val rs = LazyModule(new ResetStretcher(16))
```
colinschmidt, line 24: Can this 16 be a parameter?

src/main/scala/subsystem/RocketSubsystem.scala
```diff
@@ -20,7 +20,12 @@ case class RocketCrossingParams(
   mmioBaseAddressPrefixWhere: TLBusWrapperLocation = CBUS
 ) extends TileCrossingParamsLike {
   def injectClockNode(context: Attachable)(implicit p: Parameters): ClockNode = {
-    ClockTempNode() // TODO reset stretcher could go here, parameterized by args to this case class
+    if (p(SubsystemResetSchemeKey) != ResetSynchronous) {
```
hcook, line 23: Rather than implicitly coupling usage of this class to `SubsystemResetSchemeKey` specifically, can we make `stretchReset: Boolean = false` or `stretchResetCycles: Option[Int] = None` be one of the case class parameters above, and then set that accordingly in places where we instantiate a `RocketCrossingParams`, e.g.
```
case XCrossingKey => RocketCrossingParams(..., stretchReset = site(SubsystemResetSchemeKey) != ResetSynchronous))
``` (there seem to be only a couple relevant usages in closed source)

src/main/scala/subsystem/RocketSubsystem.scala
hcook: Haha jinx.

While this class is called `RocketCrossingParams` I think it is currently being used for a lot of different types of large-ish Tiles that might have different reset length requirements...

ernie-sifive: Changed to Option[Int] parameter.

src/main/scala/subsystem/RocketSubsystem.scala
ernie-sifive: Changed to Option[Int] parameter.

PR #2577, opened 2020-07-27
Add some debug info to diplomacy.

sequencer:

**Release Notes**:
`BundleBridgeNexusNode` is using `MixedNexusNode`, debug without nearby nodes is hard.
This PR gives better debug infomation by printing nearby nodes `require` is false.

### DIFF
--- a/src/main/scala/diplomacy/Nodes.scala
+++ b/src/main/scala/diplomacy/Nodes.scala
@@ -190,20 +190,18 @@ trait InwardNodeHandle[DI, UI, EI, BI <: Data] extends NoHandle
   def :*=*[EY](h: OutwardNodeHandle[DI, UI, EY, BI])(implicit p: Parameters, sourceInfo: SourceInfo): NoHandle = { bind(h, BIND_FLEX);  NoHandleObject }
 }
 
-sealed trait NodeBinding {
-  def serialize: String
-}
+sealed trait NodeBinding
 case object BIND_ONCE  extends NodeBinding {
-  def serialize: String = "once"
+  override def toString: String = "once"
 }
 case object BIND_QUERY extends NodeBinding {
-  def serialize: String = "query"
+  override def toString: String = "query"
 }
 case object BIND_STAR  extends NodeBinding {
-  def serialize: String = "star"
+  override def toString: String = "star"
 }
 case object BIND_FLEX  extends NodeBinding {
-  def serialize: String = "flex"
+  override def toString: String = "flex"
 }
 
 trait InwardNode[DI, UI, BI <: Data] extends BaseNode
@@ -273,8 +271,8 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
 
   /** debug info of nodes binding. */
   def bindingInfo: String =
-    s"""${oBindings.size} outward nodes binded: [${oBindings.map(n => s"${n._3.serialize}-${n._2.name}").mkString(",")}]
-       |${iBindings.size} inward nodes binded: [${iBindings.map(n => s"${n._3.serialize}-${n._2.name}").mkString(",")}]
+    s"""${oBindings.size} outward nodes bound: [${oBindings.map(n => s"${n._3}-${n._2.name}").mkString(",")}]
+       |${iBindings.size} inward nodes bound: [${iBindings.map(n => s"${n._3}-${n._2.name}").mkString(",")}]
        |""".stripMargin
 
   /** debug info of ports connecting. */
@@ -395,12 +393,12 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       oParamsCycleGuard = true
       val o = mapParamsD(oPorts.size, diParams)
       require (o.size == oPorts.size,
-        s"""Diplomacy detected a problem with your graph:
+        s"""Diplomacy has detected a problem with your graph:
+           |At the following node, the number of outward ports should equal the number of produced outward parameters.
            |$context
            |$connectedPortsInfo
            |Downstream inward parameters: [${diParams.mkString(",")}]
            |Produced outward parameters: [${o.mkString(",")}]
-           |Outward ports size should equal to produced outward parameters size.
            |""".stripMargin)
       o.map(outer.mixO(_, this))
     } catch {
@@ -416,12 +414,12 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       iParamsCycleGuard = true
       val i = mapParamsU(iPorts.size, uoParams)
       require (i.size == iPorts.size,
-        s"""Diplomacy detected a problem with your graph:
+        s"""Diplomacy has detected a problem with your graph:
+           |At the following node, the number of inward ports size should equal the number of produced inward parameters.
            |$context
            |$connectedPortsInfo
            |Upstream outward parameters: [${uoParams.mkString(",")}]
            |Produced inward parameters: [${i.mkString(",")}]
-           |Inward ports size should equal to produced inward parameters size.
            |""".stripMargin)
       i.map(inner.mixI(_, this))
     } catch {
@@ -564,9 +562,11 @@ class MixedJunctionNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   override def description = "junction"
   protected[diplomacy] def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
     require (iKnown == 0 || oKnown == 0,
-      s"""$context
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node appears left of a :=* or a := AND right of a :*= or :=. Only one side may drive multiplicity.
+         |$context
          |$bindingInfo
-         |appears left of a :=* or a := AND right of a :*= or :=. Only one side may drive multiplicity.""".stripMargin)
+         |""".stripMargin)
     multiplicity = iKnown max oKnown
    (multiplicity, multiplicity)
   }
@@ -600,35 +600,53 @@ class MixedAdapterNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   protected[diplomacy] override def flexibleArityDirection = true
   protected[diplomacy] def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
     require (oStars + iStars <= 1,
-      s"""$context
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node appears left of a :*= $iStars times and right of a :=* $oStars times; at most once is allowed.
+         |$context
          |$bindingInfo
-         |appears left of a :*= $iStars times and right of a :=* $oStars times; at most once is allowed""".stripMargin)
+         |""".stripMargin)
     if (oStars > 0) {
       require (iKnown >= oKnown,
-        s"""$context
+        s"""Diplomacy has detected a problem with your graph:
+           |The following node has $oKnown outputs and $iKnown inputs; cannot assign ${iKnown-oKnown} edges to resolve :=*
+           |$context
            |$bindingInfo
-           |has $oKnown outputs and $iKnown inputs; cannot assign ${iKnown-oKnown} edges to resolve :=*""".stripMargin)
+           |""".stripMargin)
       (0, iKnown - oKnown)
     } else if (iStars > 0) {
       require (oKnown >= iKnown,
-        s"""$context
+        s"""Diplomacy has detected a problem with your graph:
+           |The following node has $oKnown outputs and $iKnown inputs; cannot assign ${oKnown-iKnown} edges to resolve :*=
+           |$context
            |$bindingInfo
-           |has $oKnown outputs and $iKnown inputs; cannot assign ${oKnown-iKnown} edges to resolve :*=""".stripMargin)
+           |""".stripMargin)
       (oKnown - iKnown, 0)
     } else {
       require (oKnown == iKnown,
-        s"""$context
+        s"""Diplomacy has detected a problem with your graph:
+           |The following node has $oKnown outputs and $iKnown inputs; these must match.
+           |$context
            |$bindingInfo
-           |has $oKnown outputs and $iKnown inputs; these do not match""".stripMargin)
+           |""".stripMargin)
       (0, 0)
     }
   }
   protected[diplomacy] def mapParamsD(n: Int, p: Seq[DI]): Seq[DO] = {
-    require(n == p.size, s"$context has ${p.size} inputs and ${n} outputs; they must match")
+    require(n == p.size,
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node has ${p.size} inputs and $n outputs; they must match.
+         |$context
+         |$bindingInfo
+         |""".stripMargin)
     p.map(dFn)
   }
   protected[diplomacy] def mapParamsU(n: Int, p: Seq[UO]): Seq[UI] = {
-    require(n == p.size, s"$context has ${n} inputs and ${p.size} outputs; they must match")
+    require(n == p.size,
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node has $n inputs and ${p.size} outputs; they must match
+         |$context
+         |$bindingInfo
+         |""".stripMargin)
     p.map(uFn)
   }
 }
@@ -681,18 +699,20 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
     def resolveStarInfo: String =
       s"""$context
          |$bindingInfo
-         |inward nodes already known: $iKnown
-         |outward nodes already known: $oKnown
-         |query by nodes other nodes: $iStars
-         |query to other nodes: $oStars
+         |number of known := bindings to inward nodes: $iKnown
+         |number of known := bindings to outward nodes: $oKnown
+         |number of binding queries from inward nodes: $iStars
+         |number of binding queries from outward nodes: $oStars
          |""".stripMargin
     require(!outputRequiresInput || oKnown == 0 || iStars + iKnown != 0,
-      s"""$resolveStarInfo
-         |This node has $oKnown outward connections and no inward connections. At least one inward connection was required."
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node has $oKnown outward connections and no inward connections. At least one inward connection was required.
+         |$resolveStarInfo
          |""".stripMargin)
     require(!inputRequiresOutput || iKnown == 0 || oStars + oKnown != 0,
-      s"""$resolveStarInfo
-         |This node has $iKnown inward connections and no outward connections. At least one outward connection was required."
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node node has $iKnown inward connections and no outward connections. At least one outward connection was required.
+         |$resolveStarInfo
          |""".stripMargin)
     if (iKnown == 0 && oKnown == 0) (0, 0) else (1, 1)
   }
@@ -717,35 +737,41 @@ class SourceNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(po: Seq
     def resolveStarInfo: String =
       s"""$context
          |$bindingInfo
-         |inward nodes already known: $iKnown
-         |outward nodes already known: $oKnown
-         |query by nodes other nodes: $iStars
-         |query to other nodes: $oStars
+         |number of known := bindings to inward nodes: $iKnown
+         |number of known := bindings to outward nodes: $oKnown
+         |number of binding queries from inward nodes: $iStars
+         |number of binding queries from outward nodes: $oStars
          |${po.size} parameterized sinks: [${po.map(_.toString).mkString(",")}]
          |""".stripMargin
     require(oStars <= 1,
-      s"""$resolveStarInfo
-         |This node appears right of a :*= $oStars times; at most once is allowed.
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node appears right of a :*= $oStars times; at most once is allowed.
+         |$resolveStarInfo
          |""".stripMargin)
     require(iStars == 0,
-      s"""$resolveStarInfo
-         |This node cannot appear left of a :*=
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node cannot appear left of a :*=
+         |$resolveStarInfo
          |""".stripMargin)
     require(iKnown == 0,
-      s"""$resolveStarInfo
-         |This node cannot appear left of a :=
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node cannot appear left of a :=
+         |$resolveStarInfo
          |""".stripMargin)
     if (oStars == 0)
       require(po.size == oKnown,
-        s"""$resolveStarInfo
-           |$oKnown outputs were connected but ${po.size} sources were specified to the node constructor.
-           |Outputs nodes size should be equal to sinks, or have this node appears right of a :=*
+        s"""Diplomacy has detected a problem with your graph:
+           |The following node has $oKnown outward bindings connected to it, but ${po.size} sources were specified to the node constructor.
+           |Either the number of outward := bindings should be exactly equal to the number of sources, or connect this node on the right-hand side of a :=*
+           |$resolveStarInfo
            |""".stripMargin)
     else
       require(po.size >= oKnown,
-        s"""$resolveStarInfo
-           |This node has $oKnown outputs, but ${po.size} sources were specified to the node constructor.
-           |cannot assign ${po.size - oKnown} edges to resolve :*=""".stripMargin
+        s"""Diplomacy has detected a problem with your graph:
+           |The following node had $oKnown outward bindings connected to it, but ${po.size} sources were specified to the node constructor.
+           |Cannot assign ${po.size - oKnown} edges to resolve :*=
+           |$resolveStarInfo
+           |""".stripMargin
       )
     (0, po.size - oKnown)
   }
@@ -770,35 +796,41 @@ class SinkNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(pi: Seq[U
     def resolveStarInfo: String =
       s"""$context
          |$bindingInfo
-         |inward nodes already known: $iKnown
-         |outward nodes already known: $oKnown
-         |query by nodes other nodes: $iStars
-         |query to other nodes: $oStars
+         |number of known := bindings to inward nodes: $iKnown
+         |number of known := bindings to outward nodes: $oKnown
+         |number of binding queries from inward nodes: $iStars
+         |number of binding queries from outward nodes: $oStars
          |${pi.size} upward parameters: [${pi.map(_.toString).mkString(",")}]
          |""".stripMargin
     require (iStars <= 1,
-      s"""$resolveStarInfo
-         |This node appears left of a :*= $iStars times; at most once is allowed.
+      s"""Diplomacy has detected a problem with your graph:
+         |The following node appears left of a :*= $oStars times; at most once is allowed.
+         |$resolveStarInfo
          |""".stripMargin)
     require (oStars == 0,
-      s"""$resolveStarInfo
-         |This node cannot appear right of a :*=
+      s"""Diplomacy has detected a problem with your graph:
+         |$resolveStarInfo
+         |The following node cannot appear right of a :*=
          |""".stripMargin)
     require (oKnown == 0,
-      s"""$resolveStarInfo
-         |This node cannot appear right of a :=
+      s"""Diplomacy has detected a problem with your graph:
+         |$resolveStarInfo
+         |The following node cannot appear right of a :=
          |""".stripMargin)
     if (iStars == 0)
       require(pi.size == iKnown,
-        s"""$resolveStarInfo
-           |$oKnown inputs were connected but ${pi.size} sinks were specified to the node constructor.
-           |Input nodes size should be equal to sinks, or have this node appears left of a :=*
+        s"""Diplomacy has detected a problem with your graph:
+           |The following node has $iKnown inward bindings connected to it, but ${pi.size} sinks were specified to the node constructor.
+           |Either the number of inward := bindings should be exactly equal to the number of sink, or connect this node on the right-hand side of a :=*
+           |$resolveStarInfo
            |""".stripMargin)
     else
       require(pi.size >= iKnown,
-        s"""$resolveStarInfo
-           |This node has $iKnown inputs, but ${pi.size} sinks were specified to the node constructor.
-           |cannot assign ${pi.size - iKnown} edges to resolve :*=""".stripMargin
+        s"""Diplomacy has detected a problem with your graph:
+           |The following node had $iKnown inward bindings connected to it, but ${pi.size} sinks were specified to the node constructor.
+           |cannot assign ${pi.size - iKnown} edges to resolve :*=
+           |$resolveStarInfo
+           |""".stripMargin
       )
     (pi.size - iKnown, 0)
   }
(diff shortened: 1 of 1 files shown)


src/main/scala/diplomacy/Nodes.scala
```diff
@@ -411,11 +411,11 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   private var bundlesSafeNow = false
   // Accessors to the result of negotiation to be used in LazyModuleImp:
   def out: Seq[(BO, EO)] = {
-    require(bundlesSafeNow, s"${name}.out should only be called from the context of its module implementation")
+    require(bundlesSafeNow, s"${name}.out should only be called from the context of its LazyModule(${scope.get.name}) implementation, currently (${chisel3.Module.currentModule.get.name})")
```
hcook, line 414: Are you sure `chisel3.Module.currentModule.get` is always defined when this error case can arise?
```
class Parent extends SimpleLazyModule {
   val child = LazyModule(new SimpleLazyModule { val node = ... })
   val error = child.node.out.head
}
```
Referencing `child.node.out` in Parent's LazyModule scope should trip this requirement, but I don't think `chisel3.Module.currentModule` is defined, or if it is, it will be the very top-level testharness Module or something confusing like that?

sequencer: sorry my bad, I'm looking into it.

sequencer: `scope.get.name` should always being defined, and I use `LazyModule.scope` to get current scope of LazyModule, maybe it will fall back to `None`, but generally solve this problem.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -411,11 +419,11 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   private var bundlesSafeNow = false
   // Accessors to the result of negotiation to be used in LazyModuleImp:
   def out: Seq[(BO, EO)] = {
-    require(bundlesSafeNow, s"${name}.out should only be called from the context of its module implementation")
+    require(bundlesSafeNow, s"${name}.out should only be called from the context of its LazyModule(${scope.get.name}) implementation, currently (${LazyModule.scope})")
```
hcook, line 422: Let's change this to `${name}.out should only be called from within the context of ${scope.get.name}'s LazyModuleImp, but the current scope is ${LazyModule.scope}`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -369,7 +373,11 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       if (iParamsCycleGuard) throw UpwardCycleException()
       iParamsCycleGuard = true
       val i = mapParamsU(iPorts.size, uoParams)
-      require (i.size == iPorts.size, s"Diplomacy error: $context has ${i.size} != ${iPorts.size} up/down inner parameters")
+      require (i.size == iPorts.size,
+        s"""Diplomacy error: $context has ${i.size} != ${iPorts.size} down/up outer parameters:
+           |up-in parameters: [${i.mkString(",")}]
+           |iPorts: [${iPorts.map(_._2.name).mkString(",")}]
```
hcook, line 379: How about:
```
s"""Diplomacy Error: $context produced ${i.size} parameters for ${iPorts.size} inner ports. These sizes must match.
           | Inward parameters produced: [${i.mkString(",")}]
           | Inward ports connected: [${iPorts.map(_._2.name).mkString(",")}]
```

I am not totally clear on whether there is a meaningful distinction between "inner"/"inward"/"input" when talking about ports vs parameters here that it would be useful to include in this message.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -355,7 +355,11 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       if (oParamsCycleGuard) throw DownwardCycleException()
       oParamsCycleGuard = true
       val o = mapParamsD(oPorts.size, diParams)
-      require (o.size == oPorts.size, s"Diplomacy error: $context has ${o.size} != ${oPorts.size} down/up outer parameters")
+      require (o.size == oPorts.size,
+        s"""Diplomacy error: $context has ${o.size} != ${oPorts.size} down/up outer parameters:
+           |down-out parameters: [${o.mkString(",")}]
+           |oPorts: [${oPorts.map(_._2.name).mkString(",")}]
```
hcook, line 361: How about:

s"""Diplomacy Error: $context produced ${o.size} parameters for ${oPorts.size} outward ports. These sizes must match.
           | Outward parameters produced: [${o.mkString(",")}]
           | Outward ports connected: [${oPorts.map(_._2.name).mkString(",")}]
I am not totally clear on whether there is a meaningful distinction between "outer"/"outward"/"output" when talking about ports vs parameters here that it would be useful to make clear in this message. @terpstra ?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -611,8 +619,8 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   override def description = "nexus"
   protected[diplomacy] def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
     // a nexus treats :=* as a weak pointer
-    require (!outputRequiresInput || oKnown == 0 || iStars + iKnown != 0, s"$context has $oKnown required outputs and no possible inputs")
-    require (!inputRequiresOutput || iKnown == 0 || oStars + oKnown != 0, s"$context has $iKnown required inputs and no possible outputs")
+    require (!outputRequiresInput || oKnown == 0 || iStars + iKnown != 0, s"$context has $oKnown required outputs(${oBindings.map(_._2.name).mkString(",")}) and no possible inputs")
```
hcook, line 622: How about 
`s"$context has $oKnown required outputs (named ${oBindings.map(_._2.name).mkString(",")}) and no inputs. At least one is required."`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -611,8 +619,8 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
+    require (!inputRequiresOutput || iKnown == 0 || oStars + oKnown != 0, s"$context has $iKnown required inputs(${iBindings.map(_._2.name).mkString(",")}) and no possible outputs")
```
hcook, line 623: How about
s"$context has $iKnown inward connections (named ${iBindings.map(_._2.name).mkString(",")}) and no outward connections. At least one outward connection was required."

hcook:
@terpstra How do you feeling about preferring the terms "inward connections" and "outward connections" to "inputs" and "outputs", at least in terms of reporting these node error messages? I feel like "input" and "output" are kind of overloaded due to being used as wire IO directions...

terpstra:
Yes, inward and outward are better.

sequencer:
@hcook added more infos to `resolveStar`, I think this might be more easy for debugging?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -253,6 +265,27 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   val inward = this
   val outward = this
 
+  /** debug info of nodes binding. */
+  def bindingInfo: String =
+    s"""${oBindings.size} outward nodes binded: [${oBindings.map(n => s"${n._3.serialize}-${n._2.name}").mkString(",")}]
+       |${iBindings.size} inward nodes binded: [${iBindings.map(n => s"${n._3.serialize}-${n._2.name}").mkString(",")}]
```
hcook, line 271: Should be "outward nodes bound" instead of "binded" ... sorry english is awful! 😅

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -182,11 +184,21 @@ trait InwardNodeHandle[DI, UI, EI, BI <: Data] extends NoHandle
   def :*=*[EY](h: OutwardNodeHandle[DI, UI, EY, BI])(implicit p: Parameters, sourceInfo: SourceInfo): NoHandle = { bind(h, BIND_FLEX);  NoHandleObject }
 }
 
-sealed trait NodeBinding
-case object BIND_ONCE  extends NodeBinding
-case object BIND_QUERY extends NodeBinding
-case object BIND_STAR  extends NodeBinding
-case object BIND_FLEX  extends NodeBinding
+sealed trait NodeBinding {
+  def serialize: String
```
hcook, line 188: I think if you called this method `toString` then it would just magically do the right thing when interpolated into a string without you even having to call `serialize` when you do so?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -355,7 +388,14 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
       oParamsCycleGuard = true
       val o = mapParamsD(oPorts.size, diParams)
-      require (o.size == oPorts.size, s"Diplomacy error: $context has ${o.size} != ${oPorts.size} down/up outer parameters")
+      require (o.size == oPorts.size,
+        s"""Diplomacy detected a problem with your graph:
+           |$context
+           |$connectedPortsInfo
+           |Downstream inward parameters: [${diParams.mkString(",")}]
+           |Produced outward parameters: [${o.mkString(",")}]
+           |Outward ports size should equal to produced outward parameters size.
```
hcook, line 397: Let's move the problem up to the top of the message, for example:
```
Diplomacy has detected a problem with your graph:
At the following node, outward ports size should equal to produced outward parameters size.
$context
$connectedPortsInfo
...
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -369,7 +409,14 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
       iParamsCycleGuard = true
       val i = mapParamsU(iPorts.size, uoParams)
-      require (i.size == iPorts.size, s"Diplomacy error: $context has ${i.size} != ${iPorts.size} up/down inner parameters")
+      require (i.size == iPorts.size,
+        s"""Diplomacy detected a problem with your graph:
+           |$context
+           |$connectedPortsInfo
+           |Upstream outward parameters: [${uoParams.mkString(",")}]
+           |Produced inward parameters: [${i.mkString(",")}]
+           |Inward ports size should equal to produced inward parameters size.
```
hcook, line 418: As above, let's move the problem up to the top of the message.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -512,7 +559,10 @@ class MixedJunctionNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
 
   override def description = "junction"
   protected[diplomacy] def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
-    require (iKnown == 0 || oKnown == 0, s"$context appears left of a :=* or a := AND right of a :*= or :=. Only one side may drive multiplicity.")
+    require (iKnown == 0 || oKnown == 0,
+      s"""$context
```
hcook, line 563:
```
Diplomacy has detected a problem with your graph:
The following node appears left of a :=* or a := AND right of a :*= or :=. Only one side may drive multiplicity.
$context
...
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -545,15 +595,27 @@ class MixedAdapterNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   override def description = "adapter"
   protected[diplomacy] override def flexibleArityDirection = true
   protected[diplomacy] def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
-    require (oStars + iStars <= 1, s"$context appears left of a :*= $iStars times and right of a :=* $oStars times; at most once is allowed")
+    require (oStars + iStars <= 1,
```
hcook, line 598: As above, change the order for these and preface with `"Diplomacy has detected a problem with your graph:"`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -545,15 +595,27 @@ class MixedAdapterNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
+      s"""$context
```
hcook, line 599: As above, change the statement order for these and preface with "Diplomacy has detected a problem with your graph:"

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -611,8 +673,22 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
   protected[diplomacy] def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
     // a nexus treats :=* as a weak pointer
-    require (!outputRequiresInput || oKnown == 0 || iStars + iKnown != 0, s"$context has $oKnown required outputs and no possible inputs")
-    require (!inputRequiresOutput || iKnown == 0 || oStars + oKnown != 0, s"$context has $iKnown required inputs and no possible outputs")
+    def resolveStarInfo: String =
+      s"""$context
+         |$bindingInfo
+         |inward nodes already known: $iKnown
+         |outward nodes already known: $oKnown
+         |query by nodes other nodes: $iStars
```
hcook, line 681: maybe `number of binding queries from inwards nodes:`?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -611,8 +673,22 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
+         |query to other nodes: $oStars
```
hcook, line 682: maybe `number of binding queries to outward nodes:`?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -611,8 +673,22 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   override def description = "nexus"
   protected[diplomacy] def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
     // a nexus treats :=* as a weak pointer
-    require (!outputRequiresInput || oKnown == 0 || iStars + iKnown != 0, s"$context has $oKnown required outputs and no possible inputs")
-    require (!inputRequiresOutput || iKnown == 0 || oStars + oKnown != 0, s"$context has $iKnown required inputs and no possible outputs")
+    def resolveStarInfo: String =
+      s"""$context
+         |$bindingInfo
+         |inward nodes already known: $iKnown
```
hcook, line 679: maybe `number of known bindings from inward nodes:`?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -611,8 +673,22 @@ class MixedNexusNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
+         |outward nodes already known: $oKnown
```
hcook, line 680: maybe `number of known bindings to outward nodes:`?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -633,11 +709,39 @@ class SourceNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(po: Seq
...
-    require (iKnown == 0, s"$context cannot appear left of a :=")
-    require (po.size == oKnown || oStars == 1, s"$context has only ${oKnown} outputs connected out of ${po.size}")
-    require (po.size >= oKnown, s"$context has ${oKnown} outputs out of ${po.size}; cannot assign ${po.size - oKnown} edges to resolve :=*")
+    def resolveStarInfo: String =
+      s"""$context
+         |$bindingInfo
+         |inward nodes already known: $iKnown
+         |outward nodes already known: $oKnown
+         |query by nodes other nodes: $iStars
+         |query to other nodes: $oStars
```
hcook, line 718: Same changes as above if you think they are correct

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -633,11 +709,39 @@ class SourceNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(po: Seq
...
+         |${po.size} parameterized sinks: [${po.map(_.toString).mkString(",")}]
+         |""".stripMargin
+    require(oStars <= 1,
+      s"""$resolveStarInfo
+         |This node appears right of a :*= $oStars times; at most once is allowed.
```
hcook, line 723:
```
Diplomacy detected a problem with your graph:
The following node appears right of a :*= $oStars times; at most once is allowed.
```

src/main/scala/diplomacy/Nodes.scala
sequencer: Yep, so sorry for this, trying hard to improve these days.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -633,11 +709,39 @@ class SourceNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(po: Seq
...
+         |""".stripMargin)
+    require(iKnown == 0,
+      s"""$resolveStarInfo
+         |This node cannot appear left of a :=
+         |""".stripMargin)
+    if (oStars == 0)
+      require(po.size == oKnown,
+        s"""$resolveStarInfo
+           |$oKnown outputs were connected but ${po.size} sources were specified to the node constructor.
+           |Outputs nodes size should be equal to sinks, or have this node appears right of a :=*
```
hcook, line 737: `|Either the number of outward := bindings should be exactly equal to the number of sources, or have this node appear right of a :=*`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -633,11 +709,39 @@ class SourceNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(po: Seq
...
+         |This node cannot appear left of a :*=
+         |""".stripMargin)
+    require(iKnown == 0,
+      s"""$resolveStarInfo
+         |This node cannot appear left of a :=
+         |""".stripMargin)
+    if (oStars == 0)
+      require(po.size == oKnown,
+        s"""$resolveStarInfo
+           |$oKnown outputs were connected but ${po.size} sources were specified to the node constructor.
```
hcook, line 736: `|$oKnown outward bindings were connected but ${po.size} sources were specified to the node constructor.`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -633,11 +709,39 @@ class SourceNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(po: Seq
...
+           |Outputs nodes size should be equal to sinks, or have this node appears right of a :=*
+           |""".stripMargin)
+    else
+      require(po.size >= oKnown,
+        s"""$resolveStarInfo
+           |This node has $oKnown outputs, but ${po.size} sources were specified to the node constructor.
```
hcook, line 742: `|The following node had $oKnown outward bindings connected to it, but ${po.size} sources were specified to the node constructor.`

src/main/scala/diplomacy/Nodes.scala
hcook: Note: `:=*`!

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -658,11 +762,39 @@ class SinkNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(pi: Seq[U
...
+         |query to other nodes: $oStars
+         |${pi.size} upward parameters: [${pi.map(_.toString).mkString(",")}]
+         |""".stripMargin
+    require (iStars <= 1,
+      s"""$resolveStarInfo
+         |This node appears left of a :*= $iStars times; at most once is allowed.
+         |""".stripMargin)
+    require (oStars == 0,
+      s"""$resolveStarInfo
+         |This node cannot appear right of a :*=
```
hcook: Isn't this `:=*`? It was in the original message.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -633,11 +709,39 @@ class SourceNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(po: Seq
...
+      require(po.size == oKnown,
+        s"""$resolveStarInfo
+           |$oKnown outputs were connected but ${po.size} sources were specified to the node constructor.
+           |Outputs nodes size should be equal to sinks, or have this node appears right of a :=*
+           |""".stripMargin)
+    else
+      require(po.size >= oKnown,
+        s"""$resolveStarInfo
+           |This node has $oKnown outputs, but ${po.size} sources were specified to the node constructor.
+           |cannot assign ${po.size - oKnown} edges to resolve :*=""".stripMargin
```
hcook: `|${oKnown - po.size} additional sources are required to resolve :*=.`

Was this subtraction backwards in the original code? Wouldn't it always be a negative number? Seems very confusing.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -658,11 +762,39 @@ class SinkNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(pi: Seq[U
...
+         |This node cannot appear right of a :*=
+         |""".stripMargin)
+    require (oKnown == 0,
+      s"""$resolveStarInfo
+         |This node cannot appear right of a :=
+         |""".stripMargin)
+    if (iStars == 0)
+      require(pi.size == iKnown,
+        s"""$resolveStarInfo
+           |$oKnown inputs were connected but ${pi.size} sinks were specified to the node constructor.
```
hcook: `|The following node had $oKnown inward bindings connected to it, but ${pi.size} sinks were specified to the node constructor.`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -658,11 +762,39 @@ class SinkNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(pi: Seq[U
...
+           |Input nodes size should be equal to sinks, or have this node appears left of a :=*
```
hcook: `|Either the number of inward := bindings should be exactly equal to the number of sources, or connect this node on the left-hand side of a :*=`

Note: opposite of source node: `sink :*= inwards` means sink picks the number of connections

hcook: Note: opposite of source node: `sink :*= inwards` means that the sink picks the number of connections, so it can be any number

src/main/scala/diplomacy/Nodes.scala
hcook: Note: iKnown not oKnown

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -658,11 +762,39 @@ class SinkNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(pi: Seq[U
...
+           |""".stripMargin)
+    else
+      require(pi.size >= iKnown,
+        s"""$resolveStarInfo
+           |This node has $iKnown inputs, but ${pi.size} sinks were specified to the node constructor.
```
hcook: `|After being connected outwards of :=*, the following node was seen outwards of := $iKnown times, but only ${pi.size} sinks were specified to the node constructor.`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -658,11 +762,39 @@ class SinkNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(pi: Seq[U
...
+           |cannot assign ${pi.size - iKnown} edges to resolve :*=""".stripMargin
```
hcook: `|${iKnown - pi.size} additional sinks are required to resolve :=* successfully.`

hcook: Again it kind of seems to me like the subtraction in the original string is backwards / makes a negative number / is confusing.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -253,6 +265,27 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
+       |${iPorts.size} inward ports connected: [${iPorts.map(_._2.name).mkString(",")}]
+       |""".stripMargin
+
+  /** debug info of parameters propagations. */
+  def parametersInfo: String =
+    s"""${doParams.size} downstream outward parameters: [${doParams.mkString(",")}]
+       |${uoParams.size} upstream outward parameters: [${uoParams.mkString(",")}]
+       |${diParams.size} downstream inward parameters: [${diParams.mkString(",")}]
+       |${uiParams.size} upstream inward parameters: [${uiParams.mkString(",")}]
+       |""".stripMargin
```
hcook, line 284: I am a little worried that for some protocols the amount of parameters printed out here will be very large. For some tilelink nodes you will get a list of every master and slave in the graph, for example! But maybe in the failure cases you really do want as much info as possible.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -545,15 +595,27 @@ class MixedAdapterNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   override def description = "adapter"
   protected[diplomacy] override def flexibleArityDirection = true
   protected[diplomacy] def resolveStar(iKnown: Int, oKnown: Int, iStars: Int, oStars: Int): (Int, Int) = {
-    require (oStars + iStars <= 1, s"$context appears left of a :*= $iStars times and right of a :=* $oStars times; at most once is allowed")
+    require (oStars + iStars <= 1,
+      s"""$context
+         |$bindingInfo
+         |appears left of a :*= $iStars times and right of a :=* $oStars times; at most once is allowed""".stripMargin)
```
hcook: `| The following node appears inwards of a :*= $iStars times and outwards of a :=* $oStars times; at most one of either is allowed."`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -545,15 +595,27 @@ class MixedAdapterNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
     if (oStars > 0) {
-      require (iKnown >= oKnown, s"$context has $oKnown outputs and $iKnown inputs; cannot assign ${iKnown-oKnown} edges to resolve :=*")
+      require (iKnown >= oKnown,
+        s"""$context
+           |$bindingInfo
+           |has $oKnown outputs and $iKnown inputs; cannot assign ${iKnown-oKnown} edges to resolve :=*""".stripMargin)
```
hcook: `| After being connected inwards of :=*, the following node appears outwards of a := $iKnown times and inwards of a := $oKnown times. ${iKnown-oKnown} additional inward := bindings are required to resolve :*= successfully."`

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -545,15 +595,27 @@ class MixedAdapterNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
       (0, iKnown - oKnown)
     } else if (iStars > 0) {
-      require (oKnown >= iKnown, s"$context has $oKnown outputs and $iKnown inputs; cannot assign ${oKnown-iKnown} edges to resolve :*=")
+      require (oKnown >= iKnown,
+        s"""$context
+           |$bindingInfo
+           |has $oKnown outputs and $iKnown inputs; cannot assign ${oKnown-iKnown} edges to resolve :*=""".stripMargin)
```
hcook:
```
| After being connected outwards of :*=, the following node appears outwards of a := $iKnown times and inwards of a := $oKnown times. 
| ${oKnown-iKnown} additional outward := bindings are required to resolve :*= successfully."""
```

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -545,15 +595,27 @@ class MixedAdapterNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
       (oKnown - iKnown, 0)
     } else {
-      require (oKnown == iKnown, s"$context has $oKnown outputs and $iKnown inputs; these do not match")
+      require (oKnown == iKnown,
+        s"""$context
+           |$bindingInfo
+           |has $oKnown outputs and $iKnown inputs; these do not match""".stripMargin)
```
hcook: `The following node appears outwards of a := $iKnown times and inwards of a := $oKnown times. The number of bindings on both sides of the node must match.`

src/main/scala/diplomacy/Nodes.scala
hcook: `Appears inwards of a :*= $iStars times.`

src/main/scala/diplomacy/Nodes.scala
hcook: `Appears outwards of a :=* $oStars times.`

src/main/scala/diplomacy/Nodes.scala
hcook: `Transitively appears inwards of := $oKnown times.`

src/main/scala/diplomacy/Nodes.scala
hcook: `Transitively appears outwards of := $iKnown times.`

src/main/scala/diplomacy/Nodes.scala
sequencer: Yes I do think so, trying to print as much as possible information for debugging use.

src/main/scala/diplomacy/Nodes.scala
sequencer: I think left/right here should be more intuitional?

src/main/scala/diplomacy/Nodes.scala
sequencer: changed to `To resolve :*=, size of inward parameters can not be less than bindings.`

src/main/scala/diplomacy/Nodes.scala
hcook: Left/right might be more clear in terms of what to look for in the source code. Inwards/Outwards is more clear in terms of what diplomacy calls things, like `DI` `UO` etc. But maybe for an error message left/right is better? 

But what if I have:
```
a :*= b
b  :*=* c
c :*=* d
d :=* e
```
`c` will get the error, but is c "right" of `:*=` or "left" of `:=*`. Maybe "transitively left"?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -361,7 +406,14 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       if (oParamsCycleGuard) throw DownwardCycleException()
       oParamsCycleGuard = true
       val o = mapParamsD(oPorts.size, diParams)
-      require (o.size == oPorts.size, s"Diplomacy error: $context has ${o.size} != ${oPorts.size} down/up outer parameters")
+      require (o.size == oPorts.size,
+        s"""Diplomacy has detected a problem with your graph:
+           |At the following node, the number of outward ports should equal the number of produced outward parameters.
+           |$context
+           |$connectedPortsInfo
+           |Downstream inward parameters: [${diParams.mkString(",")}]
```
hcook, line 414: Let's make this `Downstreamed` to match `Produced` (both verbs)

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -375,7 +427,14 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       if (iParamsCycleGuard) throw UpwardCycleException()
       iParamsCycleGuard = true
       val i = mapParamsU(iPorts.size, uoParams)
-      require (i.size == iPorts.size, s"Diplomacy error: $context has ${i.size} != ${iPorts.size} up/down inner parameters")
+      require (i.size == iPorts.size,
+        s"""Diplomacy has detected a problem with your graph:
+           |At the following node, the number of inward ports should equal the number of produced inward parameters.
+           |$context
+           |$connectedPortsInfo
+           |Upstream outward parameters: [${uoParams.mkString(",")}]
```
hcook, line 435: Let's make this `Upstreamed` to match `Produced` (both verbs)

src/main/scala/diplomacy/Nodes.scala
hcook: I see you have lots of `lefts` and `rights` already, I don't think you have to change them again. Let's try this out as it is now.

PR #2579, opened 2020-07-28
Mark Certain LazyModules For Inlining

hcook:
There are a number of situations where we insert a node in a diplomatic graph, but under certain edge cardinalities or parameterizations the resulting module only connects inputs to outputs, i.e. becomes "passthrough". This PR provides infrastructure to:
- publicly flag when a given subclass of `BaseNode` will have no effect on the generated module circuitry (`circuitIdentity`)
- publicly flag when a given `LazyModule`'s `module` implementation should be inlined by FIRRTL (`shouldBeInlined`)
- actually apply an extant FIRRTL annotation to the `module` to convey this intent to FIRRTL
- report both the `identity` and `shouldBeInlined` flags in the `graphml` output (as empty ellipses and dotted-line rectangles respectively). 

The goal with this top-down approach is to explicitly capture designer intent for when certain LazyModules should be elided (in contrast to past bottom-up attempts to infer it by lifting all input-output connections). For example, this approach has the beneficial side effect of only inlining modules that contain nodes for which we already know we do not insert Monitors. Chisel modules like `OptimizationBarrier` are unaffected.

The policy by which the flags are set remains open to the instantiater. The generic inlining policy for all `LazyModules` is now: 
```
def shouldBeInlined: Boolean = nodes.forall(_.identity) && children.forall(_.shouldBeInlined)
```
This policy allows for recursive inlining of nested wrappers. The question becomes, which nodes are marked as `circuitIdentity`, and/or when do we override `shouldBeInlined` directly? Some node types (`IdentityNode, NameNode, EphemeralNode`) are definitionally `circuitIdentity` based on their kind. But for other types of nodes, I was less comfortable claiming that just because the node might e.g. appear to do nothing to the parameters, the matching circuit implementation is guaranteed to be desirable to inline.

Currently I am trying three different approaches in three different use cases:
- Overriding `circuitIdentity` for certain specific `TLAdapterNodes`, `TLNexusNodes` and `TLJunctionNodes`. For the  adapter cases, there is an node-specific function on the edge parameters (possibly when combined with an independent sizing parameter) that results in the adapter being passthrough. For the `NexusNodes` and `JunctionNodes` `circuitIdentity` often occurs when there is only a single input and output, or a ratio of 1. This cleans up a lot of confusingly named "xbars" that happen to frequently have 1:1 connections.
- Overriding `shouldBeInlined` for certain `TLBusWrappers`. These usually contain a child with a `NexusNode` that might be `circuitIdentity`, but also a `ClockSinkNode` that never is (but serves no function in the nexus `circuitIdentity` case).

While it is tempting to set `circuitIdentity` for certain node types based purely on their inward/outward down/up parameters (i.e. cases where the parameters on both side of an AdapterNode are identical) I wasn't comfortable making that the default behavior. One could always write a circuit generator that modifies the data flowing through a node while not capturing the change in diplomatic parameters; but is doing so an anti-pattern that should require explicit control to prevent inlining? Even if so, it seems safer to having the `circuitIdentity` identification handled on a per-node basis. Some modules like `TLFilter` only change the parameters but not the actual circuit.

Some code generation results with SiFive datapoints: 8-20% reduction in the number of Module classes, 3-6% reduction in lines of verilog emitted. Obviously these modules were "empty" and the actual impact on final area has yet to be evaluated, but it can't hurt to spend less time processing this worthless connectivity in downstream tools.


src/main/scala/diplomacy/Nodes.scala
```diff
@@ -420,7 +421,6 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
   }
 
   // Used by LazyModules.module.instantiate
```
sequencer, line 423: this line should be moved or removed too

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -578,7 +581,7 @@ class IdentityNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])()(imp
   extends AdapterNode(imp)({ s => s }, { s => s })
 {
   override def description = "identity"
-  protected override val identity = true
+  override def identity = true
```
sequencer, line 584: should it be `final` to the `IdentityNode`?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -591,6 +594,7 @@ class EphemeralNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])()(im
   extends AdapterNode(imp)({ s => s }, { s => s })
 {
   override def description = "ephemeral"
+  override def identity = true
```
sequencer, line 597: should it be `final` too?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -578,7 +586,7 @@ class IdentityNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])()(imp
   extends AdapterNode(imp)({ s => s }, { s => s })
 {
   override def description = "identity"
-  protected override val identity = true
+  override final def identity = true
```
terpstra, line 589: `final` is a bit dramatic, is it not?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -121,13 +128,13 @@ trait FormatEdge {
...
   def edges: Edges[I,O]
-  override def formatNode = {
+  override def formatNode = if (!identity) {
     edges.out.map(currEdge =>
       "On Output Edge:\n\n" + currEdge.formatEdge).mkString +
     "\n---------------------------------------------\n\n" +
     edges.in.map(currEdge =>
       "On Input Edge:\n\n" + currEdge.formatEdge).mkString
-  }
+  } else ""
```
terpstra, line 137: Why was this necessary?
If it is necessary, it might be more readable as: 

```
if (identity) "" else {
...
}
```

src/main/scala/diplomacy/Nodes.scala
terpstra: I see now that you were prompted to add these by a prior reviewer.

I think a lot hinges on what this 'identity: Boolean' means. This PR uses it to mean not only that the parameters are identity, but that the wires are an identity. With that definition, it definitely is not correct to say that all IdentityNodes are circuit identity.

Perhaps 'identity' should be named something that makes it clear the circuit is also doing nothing?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -533,6 +538,9 @@ class JunctionNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(
   uFn: Seq[U] => Seq[U])(
   implicit valName: ValName)
     extends MixedJunctionNode[D, U, EI, B, D, U, EO, B](imp, imp)(dFn, uFn)
+{
+  override def identity = (outputs.isEmpty && inputs.isEmpty) || (uRatio == 1 && dRatio == 1)
```
terpstra, line 542: This smells unsafe. This does not assure circuit identity.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -76,10 +76,17 @@ abstract class BaseNode(implicit val valName: ValName)
...
-  def name = scope.map(_.name).getOrElse("TOP") + "." + valName.name
-  def omitGraphML = outputs.isEmpty && inputs.isEmpty
+  def name: String = scope.map(_.name).getOrElse("TOP") + "." + valName.name
+  def omitGraphML: Boolean = outputs.isEmpty && inputs.isEmpty
   lazy val nodedebugstring: String = ""
 
+  /** Mark whether the node represents an identity function over the parameters moving through it.
+    *
+    * This information may be used to elide monitors or inline the parent module,
+    * since it indicates that the circuitry generated by this node should have no consequence.
```
terpstra, line 86: This comment contradicts itself. There are plenty of circuits that have identity for parameters but have a consequential circuit.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -626,6 +635,9 @@ class NexusNode[D, U, EO, EI, B <: Data](imp: NodeImp[D, U, EO, EI, B])(
   outputRequiresInput: Boolean = true)(
   implicit valName: ValName)
     extends MixedNexusNode[D, U, EI, B, D, U, EO, B](imp, imp)(dFn, uFn, inputRequiresOutput, outputRequiresInput)
+{
+  override def identity = (outputs.isEmpty && inputs.isEmpty) || (outputs.size == 1 && inputs.size == 1)
```
terpstra, line 639: Smells unsafe.

src/main/scala/tilelink/Buffer.scala
```diff
@@ -17,7 +17,7 @@ class TLBufferNode (
     managerFn = { p => p.v1copy(minLatency = p.minLatency + a.latency + d.latency) }
 ) {
   override lazy val nodedebugstring = s"a:${a.toString}, b:${b.toString}, c:${c.toString}, d:${d.toString}, e:${e.toString}"
-
+  override def identity = List(a,b,c,d,e).forall(_ == BufferParams.none)
```
terpstra, line 20: This is clearly correct. Circuit identity.

src/main/scala/diplomacy/Nodes.scala
terpstra: Actually, I now recall that IdentityNodes differ from AdapterNodes in that it directly cross connects the bundles for you. Therefore, there can indeed be no hardware in the middle, so this is safe.

PR #2582, opened 2020-07-31
SRAM: repipeline the TLRAM into a 3 cycle RMW state machine

terpstra:
Here is a picture of the change to the pipeline:
  https://app.lucidchart.com/invitations/accept/da44b89c-a93c-45e9-ba5e-cb6f9140d84e

Compared to the old pipeline, occupancy is increased from 2 cycles to 3 for:
 - atomics
 - sub-ECC-granularity writes
 - repaired ECC values

In exchange for this occupancy increase, a new register (REG) was added:
  sram data output => *REG* => ecc-correction => ALU => sram write setup
This path was sufficiently long that it limited fMAX on many designs.
In designs without ECC and without atomics, this pipeline is optimized away.

Compared to the old pipeline, response latency is unchanged (by default) for:
 - reads   (1)
 - atomics (1)
 - writes  (1)
 - ECC-repaired reads   (2)
 - ECC-repaired atomics (2)

Added a knob (sramReg) to set latency for all operations to 2.
With this knob disabled (the default), as in the old pipeline:
  - output data can flow uncorrected from the SRAM
  - output valid depends on correct ECC decode of SRAM output

With the knob enabled:
  - data flows from an ECC correction fed by registers
  - valid is a register

**Release Notes**
Improved cycle time for designs involving TLRAM.

### DIFF
--- a/src/main/scala/tilelink/SRAM.scala
+++ b/src/main/scala/tilelink/SRAM.scala
@@ -2,7 +2,8 @@
 
 package freechips.rocketchip.tilelink
 
-import Chisel._
+import chisel3._
+import chisel3.util._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.diplomaticobjectmodel.logicaltree.{BusMemoryLogicalTreeNode, LogicalModuleTree, LogicalTreeNode}
@@ -23,6 +24,7 @@ class TLRAM(
     atomics: Boolean = false,
     beatBytes: Int = 4,
     ecc: ECCParams = ECCParams(),
+    sramReg: Boolean = false, // drive SRAM data output directly into a register => 1 cycle longer response
     val devName: Option[String] = None,
     val dtsCompat: Option[Seq[String]] = None
   )(implicit p: Parameters) extends DiplomaticSRAM(address, beatBytes, devName, dtsCompat)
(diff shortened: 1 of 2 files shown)


terpstra:
There were some follow-up changes I wanted to make that put down a combination of Xbar + TLRAM, but I think keeping those separate is probably a good idea.

src/main/scala/tilelink/SRAM.scala
```diff
@@ -23,6 +24,7 @@ class TLRAM(
     atomics: Boolean = false,
     beatBytes: Int = 4,
     ecc: ECCParams = ECCParams(),
+    sramReg: Boolean = false, // drive SRAM data output directly into a register => 1 cycle longer response
```
aswaterman, line 27: `pipeline` instead of `sramReg`?

src/main/scala/tilelink/SRAM.scala
```diff
@@ -173,73 +219,107 @@ class TLRAM(
...
+    val dat    = Mux(d_wb, d_data,    Mux(r_replay, r_data,    a_data))
+    val poison = Mux(d_wb, d_poisonv, Mux(r_replay, r_poisonv, a_poisonv))
+    val coded  = VecInit((dat zip poison) map { case (d, p) =>
       if (code.canDetect) code.encode(d, p) else code.encode(d)
     })
 
-    d_raw_data := mem.read(addr, ren)
-    when (wen) { mem.write(addr, coded, sel) }
+    val index = Cat(mask.zip((addr >> log2Ceil(beatBytes)).asBools).filter(_._1).map(_._2).reverse)
+    r_raw_data := mem.read(index, ren) holdUnless RegNext(ren)
```
aswaterman, line 316: Not that it matters much, but a while back I defined `mem.readAndHold(index, ren)` to summarize this pattern

src/main/scala/tilelink/SRAM.scala
```diff
@@ -74,48 +75,100 @@ class TLRAM(
...
+     *
+     * For reads:
+     *   - stage A sets up the read
+     *   - stage R drives the uncorrected data with valid based on ECC validity
+     *   - stage D sets up the correction, if any
+     *
+     * When stage D needs to perform a write (AMO, sub-ECC write, or ECC correction):
+     *   - there is a WaW or WaR hazard vs. the operation in stage R
+     *     - for sub-ECC writes and atomics, we ensure stage R has a bubble
+     *     - for ECC correction, we cause stage R to be replayed (and reject stage A twice)
```
aswaterman, line 105: What's the rationale for not using replay for both cases?  If using a bubble instead of a replay would improve performance, I'd understand.  But since there's a structural hazard from the D stage on the following cycle anyway, is the bubble actually better than the replay?

terpstra: You're right that the performance in terms of latency and throughput would be the same. However, if I replay it requires reading the SRAM twice. I assumed that sub-ecc-granule writes might be common, so paying 2x power for them was unwise.

src/main/scala/tilelink/SRAM.scala
terpstra: I'm not sure that name is much better. What is actually happening is that I remove the 'ecc-ok' fast path.

BTW, it has been suggested to me offline that perhaps in the pipelined case we should increase ECC repair latency to 3. Thus, we could take the output data uncorrected from registers. Of course, the valid signal would still need to include a suppression for the correction-required case. Do you think this change is worth it?

terpstra: I think the trade-off is between latency in valid vs. data. The current design has a flop for valid and correction on the data. The proposal have flops for data and detection on valid. This is the same trade-off we make for the old 'fast path'. If it was better there, why is it not also better here?

PR #2584, opened 2020-08-04
CI: compile using wake 0.18.1

terpstra:

Stop regressing on wake 0.18 support; compile using 0.18.1 instead of 0.17.2.

### DIFF
--- a/ci-tests/wake_scala_compilation
+++ b/ci-tests/wake_scala_compilation
@@ -3,8 +3,8 @@ set -ex
 
 echo "Installing Wake"
 
-wget https://github.com/sifive/wake/releases/download/v0.17.2/ubuntu-16-04-wake_0.17.2-1_amd64.deb
-sudo dpkg -i ubuntu-16-04-wake_0.17.2-1_amd64.deb
+wget https://github.com/sifive/wake/releases/download/v0.18.1/ubuntu-16-04-wake_0.18.1-1_amd64.deb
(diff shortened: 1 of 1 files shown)


terpstra:
This one turned out to be very easy.

ci-tests/wake_scala_compilation
```diff
@@ -34,5 +34,5 @@ echo "Compile Scala"
 
 export WAKE_PATH=$PATH
 wake --init .
-wake --no-tty -p1 -dv 'compileScalaModule rocketchipScalaModule | getPathResult'
+wake -x 'compileScalaModule rocketchipScalaModule | getPathResult'
```
richardxia, line 37: Is there a reason you got rid of the `--no-tty`? I think this affects the carriage return thing for "progress bars", and in CI, when you don't really have a real TTY, it often ends up creating a newline for each carriage return as well.

terpstra: Wake does not need to be told there is no tty. It auto-detects this.
The only reason to use --no-tty is if you actually DO have a tty, but don't want the bards.

richardxia: Ah, I see that you're right. I took a quick look at the actual [CI console log from this PR](https://github.com/chipsalliance/rocket-chip/pull/2584/checks?check_run_id=946222730) and saw some progress bars, but after taking a closer look, I see that they're actually from `curl`.

PR #2586, opened 2020-08-05
Remove hex symlink rule from test makefrag generation

colinschmidt:
riscv-tests hasn't built or installed the hex files this rule is trying to symlink for years.

This dead code is kind of a wart for a downstream rule in chipyard that also wants to generate this same file (obviously it is easily worked around by just naming it something different).

**Type of change**: dead code removal


colinschmidt:
Thanks for the fast approval I'll let someone from your side press the button.

PR #2589, opened 2020-08-06
Change wrdata width to support xLen != fLen case

abhinay-kayastha:
**Related issue**: 
The wdata was collecting incomplete value for fp cases when xlen was different from flen for example xlen = 32 and flen = 64 because the data width was based on xlen.

**Release Notes**

### DIFF
--- a/src/main/scala/util/CoreMonitor.scala
+++ b/src/main/scala/util/CoreMonitor.scala
@@ -15,7 +15,7 @@ class CoreMonitorBundle(val xLen: Int) extends Bundle with Clocked {
   val valid = Bool()
   val pc = UInt(width = xLen.W)
   val wrdst = UInt(width = 5.W)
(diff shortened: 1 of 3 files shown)


src/main/scala/util/CoreMonitor.scala
```diff
@@ -15,7 +15,7 @@ class CoreMonitorBundle(val xLen: Int) extends Bundle with Clocked {
   val valid = Bool()
   val pc = UInt(width = xLen.W)
   val wrdst = UInt(width = 5.W)
-  val wrdata = UInt(width = xLen.W)
+  val wrdata = UInt(width = 64.W)
```
ingallsj, line 18: could this be `(xLen max fLen)`?

abhinay-kayastha: Yes. I'll have to add flen as the parameter. Will update that.

PR #2595, opened 2020-08-07
Add params for nTLBBasePageSectors and nTLBSuperpages for both I and D TLBs

BrianGraysonSiV:
**Release Notes**
The design also supports parametrization of both the sectoring and the number of superpages in the L1 TLBs. The code has always supported the different values, but the parameters to control it were not exposed up the hierarchy.

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -23,6 +23,8 @@ case class ICacheParams(
     nWays: Int = 4,
     rowBits: Int = 128,
     nTLBEntries: Int = 32,
+    nTLB4KSectors: Int = 4,
+    nTLBSuperpages: Int = 4,
     cacheIdBits: Int = 0,
     tagECC: Option[String] = None,
     dataECC: Option[String] = None,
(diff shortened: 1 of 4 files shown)


src/main/scala/rocket/ICache.scala
```diff
@@ -23,6 +23,8 @@ case class ICacheParams(
     nWays: Int = 4,
     rowBits: Int = 128,
     nTLBEntries: Int = 32,
+    nTLB4KSectors: Int = 4,
```
aswaterman, line 26: I'd consider calling this `nTLBBasePageSectors` (more symmetric with "Superpage" and a more generic name overall).  LGTM otherwise

BrianGraysonSiV: Along those lines, the preexisting nTLBEntries should really be nTLBBasePageEntries, but that's too intrusive of a change for now I would hazard....

aswaterman: Yeah, agreed on both counts.

Issue #2599, opened 2020-08-10
DefaultFPGAConfig part not found


Sinbiosis:
**Type of issue**: bug report

**Impact**: FPGA Building

**Other information**
 
lore@Helium:~/src/rocket-chip/vsim$ make verilog CONFIG=DefaultFPGAConfig
mkdir -p /home/lore/src/rocket-chip/vsim/generated-src/
cd /home/lore/src/rocket-chip && java -Xmx2G -Xss8M -XX:MaxPermSize=256M -jar /home/lore/src/rocket-chip/sbt-launch.jar "runMain freechips.rocketchip.system.Generator -td /home/lore/src/rocket-chip/vsim/generated-src -T freechips.rocketchip.system.TestHarness -C DefaultFPGAConfig"
OpenJDK 64-Bit Server VM warning: Ignoring option MaxPermSize; support was removed in 8.0
[info] Loading settings for project rocket-chip-build from plugins.sbt ...
[info] Loading project definition from /home/lore/src/rocket-chip/project
[info] Loading settings for project rocketchip from build.sbt ...
[info] Loading settings for project hardfloat from build.sbt ...
[info] Loading settings for project api-config-chipsalliance from build.sbt ...
[info] Loading settings for project chisel from build.sbt ...
Using addons: 
[info] Set current project to rocketchip (in build file:/home/lore/src/rocket-chip/)
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
[info] running freechips.rocketchip.system.Generator -td /home/lore/src/rocket-chip/vsim/generated-src -T freechips.rocketchip.system.TestHarness -C DefaultFPGAConfig
[error] (run-main-0) java.lang.Exception: Unable to find part "DefaultFPGAConfig" from "ArrayBuffer(DefaultFPGAConfig)", did you misspell it or specify the wrong package path?
[error] java.lang.Exception: Unable to find part "DefaultFPGAConfig" from "ArrayBuffer(DefaultFPGAConfig)", did you misspell it or specify the wrong package path?
[error] 	at Chisel.package$throwException$.apply(compatibility.scala:436)
[error] 	at freechips.rocketchip.util.HasRocketChipStageUtils.$anonfun$getConfig$1(GeneratorUtils.scala:19)
[...]

PR #2603, opened 2020-08-13
configs: update BaseFPGAConfig and DefaultFPGAConfig

hcook:
`DefaultFPGAConfig` was broken by the configurable topology refactor, because it did not have a topology specified in its `Config`.

`DefaultFPGAConfig` is not protected by regression here, nor does it contain sufficient design material to actually make a working FPGA image for any particular FPGA board, as far as I know. Perhaps it should be retired entirely instead? Looks like neither https://github.com/sifive/freedom nor https://github.com/firesim/firesim use it vs `BaseConfig`...

I added `WithoutTLMonitors` alteration to `BaseFPGAConfig` since they are know to break most FPGA flows.

**Related issue**: Closes https://github.com/chipsalliance/rocket-chip/issues/2599

**Type of change**: bug fix

**Impact**:  API addition 

**Release Notes**

### DIFF
--- a/src/main/scala/system/Configs.scala
+++ b/src/main/scala/system/Configs.scala
@@ -83,5 +83,13 @@ class MMIOPortOnlyConfig extends Config(
   new DefaultConfig
 )
 
-class BaseFPGAConfig extends Config(new BaseConfig)
-class DefaultFPGAConfig extends Config(new WithNSmallCores(1) ++ new BaseFPGAConfig)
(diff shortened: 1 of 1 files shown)


terpstra:
I had to manually add those topology settings in soc-freedom-sifive just a few commits ago. Can you try validating this change against that repository?

sequencer:
Discussed in the dev meeting: we think these API can be removed directly?

PR #2614, opened 2020-08-20
Bump Verilator to 4.040, show seed on failure

seldridge:
Bump Verilator from 4.028 to 4.040 to pull in a change that adds
callbacks for when the simulation ends. This helps with debugging if
you have an intermittent failure and you hit a Chisel assertion.
Previously, you would have to be running verbosely to know what the
seed was. This change dumps the seed on exit so you always have a
reproducible seed ready to go.

This is specifically trying to help with the following problem:

1. You make some change to rocket chip
2. Local tests pass
3. CI tests fail intermittently
4. You then have a reproducibility problem where you need to search for a seed that produces the problem.

If you always know the seed at finish, then you aren't stuck in a loop searching for a counterexample.

Alternatively, this could be solved without bumping Verilator and by unconditionally printing the seed at simulation start as opposed to gating this on `--verbose`.

**Related issue**: None

**Release Notes**

None.


seldridge:
This PR should probably wait for a Chisel3 update. Verilator 4.040 doesn't have the `Verilated::flushCall` method which Chisel3's `top.cpp` uses. It would be unideal to have the Rocket Chip version of Verilator be incompatible with what the Chisel3 submodule can use.

sequencer:
We are on verilator 5 now!

PR #2615, opened 2020-08-21
Diplomatic adder example + basic mdoc infrastructure

debs-sifive:
**NOTE:** The `docs` directory right now is set up to just contain the raw documentation `src` files. We may want to include the `generated` documents in the future, but doing this should also include some sort of check to make sure the generated docs are up-to-date with the source.

**Related issue**:  N/A

**Type of change**: other enhancement (documentation)

**Release Notes**

Add documentation directory and functionality to generate documents from markdown using embedded `mdoc`. Embedded `mdoc` code will compile against the codebase as part of CI checks.


docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+wants to send or receive parameter information must have **one or more nodes**.
+
+Between each pair of nodes, there may be **one or more edges**. The edge does
+the negotiation work to ensure parameter agreement between the nodes it connects.
+
+Let's pass around our parameters like so:
+
+![adder diplomatic nodes](diagrams/adder_nodes.pdf)
+
+Because the graph diplomacy creates is **acyclic**, it is directional. The
```
hcook, line 59: I would rephrase this is `Diplomacy expects the user to define a directed graph, and this graph must be **acyclic**.`

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+                                       NexusNode, SinkNode, LazyModule, LazyModuleImp}
+```
+
+### Parameter negotiation and passing
+
+We have a width parameter we would like our modules to share. Any module that
+wants to send or receive parameter information must have **one or more nodes**.
+
+Between each pair of nodes, there may be **one or more edges**. The edge does
+the negotiation work to ensure parameter agreement between the nodes it connects.
```
hcook, line 53: `Edges are directed, and if we draw their direction with an arrow we can then define the tail of the arrow as the **source** of the edge and the head of the arrow as the **sink** of the edge.`

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+
+Because the graph diplomacy creates is **acyclic**, it is directional. The
+direction towards the sinks is **downward**, and the direction towards the
+sources is **upward**.
+
+#### Parameters
+
+In our example, we want all of our circuitry to agree upon a Scala `Int` to use
+as a wire width. For clarity, we will define distinct case classes for our
+parameters, although they all contain the same type.
```
hcook, line 67: i would say `the same type of information`

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+In our "node implementation", or `NodeImp`, we describe how parameters flow
+throughout our graph and how parameters are negotiated between nodes.
+
+We will use `SimpleNodeImp`, which performs the same parameter negotiation and
+passes the same bundles along an edge, regardless of whether the edge points
+into our out of the node.
+
+Our **edge parameter** (`E`) describes the data type that needs to be passed along
+the edges in our graph; in our case we only need to keep track of one `Int` that
+describes our width, which will live inside our `EdgeParam` class.
```
hcook, line 86: "describes our width" => "describes the final, negotiated width used to actually create the hardware wire"

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+    if (pd.width < pu.width) EdgeParam(pd.width) else EdgeParam(pu.width)
+  }
+  def bundle(e: EdgeParam) = UInt(e.width.W)
+  def render(e: EdgeParam) = RenderedEdge("blue", s"width = ${e.width}")
+}
+```
+
+Notice that the `edge` function does the actual negotiation between nodes. In our
+case, it compares the parameters traveling both upwards and downwards, and chooses
+the one with the smaller width to pass along.
```
hcook, line 106: I don't think we should say "pass along" here. Like, I would describe the downwards and upwards ones as "getting passed" since they actually move through the graph from node to node. But the edge ones are only visible to the two nodes connected to this specific edge. Like if you made your edges always pick `width = 3`, all the other parameters would still get passed around, they would just be ignored. Maybe `and chooses
the one with the smaller width as the final result to make visible from each node.`?

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+
+The `render` function is required by `NodeImp`s and holds metadata for graph
+generation.
```
hcook, line 109: "for graph generation" => "for rendering a view of the negotiated information in some graphical format"

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+
+#### Nodes
+
+Nodes can receive or generate parameters. A module parameterized through
+Diplomacy must have one or more nodes in order to receive or send parameters.
+Here, we will create our nodes for the modules we would like to parameterize.
```
hcook, line 115: "create our nodes for the module" => "create some different types of nodes for use in the modules"

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+
+Let's start off with creating a node for our driver. We will make it a
+`SourceNode` with the node implementation shown in the previous section, since
+`SourceNode`s have only outward pointing edges.
```
hcook, line 119: "have only outward pointing edges" -> "only generate downward-flowing parameters along outward edges."

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+
+For our `AdderDriverNode`, `widths` of type `Seq[DownwardParam]` represents
+the outputs of this node. We are using a `Seq` here because each node will send
```
hcook, line 122: `because each node can send multiple outputs, and in our example will send two identical outputs:`

hcook, line 122: "the outputs of this node" => "the output wires of this node's module?"

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+/** node for [[AdderDriver]] (source) */
+class AdderDriverNode(widths: Seq[DownwardParam])(implicit valName: ValName)
+  extends SourceNode(AdderNodeImp)(widths) {
+  require(widths.length == 2)
+  require(widths.head.width == widths.tail.head.width)
+}
+```
+
+Our monitor node follows the same pattern, this time extending `SinkNode`, a type
+of node that only has inward edges. The parameter it takes in is `width` of type
```
hcook, line 136: "that only has inward edges" -> "only generates upward-flowing parameters along inward edges."

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+
+"Lazy" evaluation refers to deferring the evaluation of an expression until
+it is needed.
+
+Parameter negotiation in Diplomacy is done lazily, after the Diplomacy graph
+is created. Because of this, the concrete hardware that we want to parameterize
+must also be generated lazily. Diplomacy provides the `LazyModule` construct for
+writing lazily-evaluated hardware modules.
+
+Let's use `LazyModule` to define our `Adder`. Notice that the creation of the
```
hcook, line 171: "the creation of the components used to define the Diplomacy graph itself"?

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+  val node = new AdderNode (
+    { case dps: Seq[DownwardParam] =>
+      require(dps.forall(dp => dp.width == dps.head.width))
+      dps.head
+    },
+    { case ups: Seq[UpwardParam] =>
+      require(ups.forall(up => up.width == ups.head.width))
+      ups.head
+    }
+  )
```
hcook, line 195: maybe add a `require` on `node.in.size`?

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+sources are on the right.
+
+```scala mdoc
+/** top-level connector */
+class AdderTestHarness()(implicit p: Parameters) extends LazyModule {
+  val adder = LazyModule(new Adder)
+  val driver = LazyModule(new AdderDriver(8))    // passing in downward-traveling width to Diplomacy
+  val checker = LazyModule(new AdderMonitor(4))  // passing in upward-traveling width
+
+  // node connections
```
hcook, line 273: `// Create edges via binding operators between nodes, in order to define a complete graph.`

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+
+Now we are ready to generate the Verilog for our circuit. The `DummyConfig` is
+empty, since we don't need any parameters specified through a `Config`.
+
+```scala mdoc:silent
+/** empty config */
+class DummyConfig extends Config((_,_,_) => { case _ if false => "" })
+
+val verilog = (new ChiselStage).emitVerilog(
+  LazyModule(new AdderTestHarness()(new DummyConfig)).module
```
hcook, line 304: I thiink you could just use `Parameters.empty` here

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+)
+```
+
+Below is the generated Verilog for our modules! Note the parameterized ports
+in `Adder`, `AdderDriver`, and `AdderChecker` all get the lower width (8).
+The LFSR is also parameterized to 8 bits.
```
hcook, line 310: Wasn't 4 the smaller width?

debs-sifive: oops yes, I shortened the widths in the code because the generated LSFRs were taking up too much space :P

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+```scala mdoc
+/** driver (source) */
+class AdderDriver(width: Int)(implicit p: Parameters) extends LazyModule {
+  val nodeA = new AdderDriverNode(Seq(DownwardParam(width), DownwardParam(width)))
+  val nodeB = new AdderDriverNode(Seq(DownwardParam(width), DownwardParam(width)))
+
+  lazy val module = new LazyModuleImp(this) {
+    // generate random addends
+    val randomAddendA = FibonacciLFSR.maxPeriod(nodeA.out.head._2.width)  // notice that we are using negotiated
+    val randomAddendB = FibonacciLFSR.maxPeriod(nodeB.out.head._2.width)  // parameters here
```
hcook, line 217: It's a little off that you take `head` here but then `foreach` below. i'd maybe recommend

```
def generateRandomAddend(width: Int) = FibonacciLFSR.maxPeriod(width)

nodeA.out.foreach { case (outputAddend, edgeParam) => outputAddend := generateRandomAddend(edgeParam.width) }
nodeB.out.foreach { case (outputAddend, edgeParam) => outputAddend := generateRandomAddend(edgeParam.width) }
```

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+
+Our `AdderDriver` randomly generates addends each cycle. Note that each node
+passes on two identical `width` parameters to be consumed by the `Adder` and
+`AdderMonitor`.
+
+```scala mdoc
+/** driver (source) */
+class AdderDriver(width: Int)(implicit p: Parameters) extends LazyModule {
+  val nodeA = new AdderDriverNode(Seq(DownwardParam(width), DownwardParam(width)))
+  val nodeB = new AdderDriverNode(Seq(DownwardParam(width), DownwardParam(width)))
```
hcook, line 212: It feels a bit weird to me to make one `val` per node in the graph inside this driver module, rather than having one node _per_ driver module and then making as many drivers as i want in the DUT, if you see what I mean? But it's not wrong per se, and I am not sure it is worth changing the figures and stuff over. It just means that this driver will only work for one specific graph, rather than being a tool that can be used as a component to make and test even bigger graphs. But maybe this change is something we could explore in future examples?

hcook: I would also consider instead having one node for driving the monitors and one node for driving the adder, where each sequence length is based on the number of "drivers" you want to have in the DUT. That would let you write
```
val driver = AdderDriver(width = 8, num = 2)
adder.node :=* driver.adderNode
monitor.inputsNode :=* driver.monitorNode
monitor.sumNode := adder.node
```
But again this is trending towards a more complicated example

debs-sifive: I don't think it would be _too_ much effort to make these changes, and I do think it's better to have an example that aligns better with a more "canonical" diplomatic graph. I will give it a go!

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+`SourceNode`s have only outward pointing edges.
+
+For our `AdderDriverNode`, `widths` of type `Seq[DownwardParam]` represents
+the outputs of this node. We are using a `Seq` here because each node will send
+two identical outputs: one to the adder, and the other to the monitor. These
+properties are `require`d by the node.
+
+```scala mdoc
+/** node for [[AdderDriver]] (source) */
+class AdderDriverNode(widths: Seq[DownwardParam])(implicit valName: ValName)
```
hcook, line 128: In real code I would make you take a single `width: Int` argument, and then pass `Seq.fill(2){ DownwardParam(width) }` to `SourceNode`. Then there is is no way that the user can ever get it wrong, and so you don't even need the `require` checks! But do you think that would be helping or hurting the example?

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,322 @@
...
+    when (checker.module.io.error) {
+      printf("something went wrong")
+    }
+  }
+
+  override lazy val desiredName = "AdderTestHarness"
+}
+```
+
+We have used the most typical node binding here, but other types of bindings
```
hcook, line 291: `the most typical node binding operator (:=) here`

docs/src/diplomacy/adder_tutorial.md
debs-sifive: I don't think that would hurt the example. But for type-checking reasons in the NodeImp, wouldn't I need to do pass in `width: DownwardParam` instead of `Int`? Like so:

```scala
class AdderDriverNode(width: DownwardParam)(implicit valName: ValName)
  extends SourceNode(AdderNodeImp)(Seq(width))
```

and then I could do `Seq.fill(2){ width }`. Or is there some other way to go about this?

docs/src/diplomacy/adder_tutorial.md
debs-sifive: in your code snippet, wouldn't that generate different addends per node? i.e. `nodeA` would pass two different random numbers into the adder/checker instead of the same random number.

That's why I called the LFSR once with the `head` width, not sure if you know of a better way to do it.

docs/src/diplomacy/adder_tutorial.md
debs-sifive: I feel like that could be misleading. Will try "the desired widths for the output wires of the module instantiating this node".

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,327 @@
...
+  extends SinkNode(AdderNodeImp)(Seq(width))
+```
+
+Our adder node will take inputs from two `AdderDriverNode`s (the two addends)
+and pass one output (the sum) into the monitor. Because the number of inputs and
+outputs differ, we will make this node extend a `NexusNode`.
+
+`dFn` maps **incoming downward** parameters into **outgoing downward** parameters.
+Similarly, `uFn` maps **incoming upward** parameters into **outgoing upward** ones.
+
```
vmurali, line 159: Can you clarify what an "incoming" and an "outgoing" parameter is?

debs-sifive: Added some clarification here: https://github.com/chipsalliance/rocket-chip/pull/2615/commits/4dbe86e7af38246bb928d4e0b218dabc8883b865

Hope that helps!

docs/src/diplomacy/adder_tutorial.md
debs-sifive: Ohhh ignore my previous comment, I see what you are doing now. I think that is a neat way of doing it, but a but hard to parse for diplomacy newbies. I think it could be useful to include both versions, and explain how they are doing the same thing.

debs-sifive: Actually upon more consideration, I think your way of doing it is better to show.

docs/src/diplomacy/adder_tutorial.md
hcook: `uFn` maps **upward** parameters along **inward edges** into **upward** parameters along **outward edges**.`
=>
`uFn` maps **upward** parameters along **outward edges** into **upward** parameters along **inward edges**.`

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,324 @@
...
+  extends SinkNode(AdderNodeImp)(Seq(width))
+```
+
+Our adder node will take inputs from two `AdderDriverNode`s (the two addends)
+and pass one output (the sum) into the monitor. Because the number of inputs and
+outputs differ, we will make this node extend a `NexusNode`.
+
+`dFn` maps **downward** parameters along **inward edges** into **downward**
+parameters along **outward edges**. Similarly, `uFn` maps **upward**
+parameters along **inward edges** into **upward** parameters along **outward
```
hcook, line 161:
```
`uFn` maps **upward** parameters along **inward edges** into **upward** parameters along **outward edges**.``
```
=>
```
`uFn` maps **upward** parameters along **outward edges** into **upward** parameters along **inward edges**.`
```

docs/src/diplomacy/adder_tutorial.md
hcook: Yeah I think we are building up an example that shows lots of different ways of doing things in different cases, which, while it makes the example code less uniform, has the benefit of suggesting those different options as available to readers

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,324 @@
...
+passes it to `numOutputs` sources.
+
+```scala mdoc
+/** driver (source)
+  * drives one random number on multiple outputs */
+class AdderDriver(width: Int, numOutputs: Int = 2)(implicit p: Parameters) extends LazyModule {
+  val nodeSeq = Seq.fill(numOutputs)(new AdderDriverNode(DownwardParam(width)))
+
+  lazy val module = new LazyModuleImp(this) {
+    // generate random addend (notice the use of the negotiated width)
```
hcook, line 226: In theory different pieces of the graph could reach different conclusions, and I think it is better to admit to that case explicitly:
```
val negotiatedWidths = nodeSeq.flatMap(_.edges.out).map(_.width)
require(widths.forall(_ == widths.head.width, "outputs must all have agreed on same width"))
val finalWidth = widths.head.width // or skip the require and use widths.min?
val randomAddend = FibonacciLFSR.maxPeriod(finalWidth)
```

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,324 @@
...
+    val randomAddend = FibonacciLFSR.maxPeriod(nodeSeq.head.out.head._2.width)
+
+    // drive signals
+    nodeSeq.foreach { node => node.out.head._1 := randomAddend }
```
hcook, line 230:
```
 nodeSeq.map(._out).foreach { case (outputBundle, _) => outputBundle := randomAddend }
```
would let this code keep working if you ever changed the number of edges connected to each node

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,324 @@
...
+
+Our `AdderMonitor` prints out the results of the `Adder` each cycle and
+signals an error if the `Adder` returns an incorrect result. It has two
+`AdderMonitorNode`s for taking in addends from the `AdderDriver`, and one
+`AdderMonitorNode` to take in the sum from the `Adder`.
+
+```scala mdoc
+/** monitor (sink) */
+class AdderMonitor(width: Int)(implicit p: Parameters) extends LazyModule {
+  val nodeSeq = Seq.fill(2) { new AdderMonitorNode(UpwardParam(width)) }
```
hcook, line 245: Here "2" should be equivalent to the number of driver modules, so I think you are right to manually create one node for each one of them, expecting one input edge each. Basically, we are saying "every Adder Monitor rightly expects that there are exactly 2 input numbers each coming from a different driver." But contrast this comment to what I am about to say for `AdderDriverNode` above

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,324 @@
...
+```
+
+Our `AdderDriver` randomly generates an addend of `width` bits each cycle, and
+passes it to `numOutputs` sources.
+
+```scala mdoc
+/** driver (source)
+  * drives one random number on multiple outputs */
+class AdderDriver(width: Int, numOutputs: Int = 2)(implicit p: Parameters) extends LazyModule {
+  val nodeSeq = Seq.fill(numOutputs)(new AdderDriverNode(DownwardParam(width)))
```
hcook, line 223: Here `numOutputs` is equivalent to the number of places we want our single number to go. But the driver is really sending out the same number twice, not two different numbers. So I would make this side different from the `AdderMonitorNode` side, which is expecting two different numbers:
```
val node = new AdderDriverNode(Seq.fill(numOutputs)(DownwardParam(width)))
```
Now we have a single node that expects to be on the right hand side of `:=` twice, whereas the `AdderMonitor` has two nodes that each expect to be on the left hand side of `:=` once.

Again, this is a fairly subtle difference and the RTL that is ultimately made will be logically equivalent, but I think it would make the example a bit more rich, and clarify that the 2 here is Adder+Monitor, wheres the 2 in `AdderMonitor` is Driver+Driver

docs/src/diplomacy/adder_tutorial.md
hcook: Just `node.out.foreach` if you make my above suggested change

docs/src/diplomacy/adder_tutorial.md
hcook: Just `node.edges.out.map(_.width)` if you make my above suggested change

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,324 @@
...
+/** top-level connector */
+class AdderTestHarness()(implicit p: Parameters) extends LazyModule {
+  val adder = LazyModule(new Adder)
+  val drivers = Seq.fill(2) { LazyModule(new AdderDriver(width = 8)) } // 8 will be a downward-traveling width
+  val checker = LazyModule(new AdderMonitor(width = 4))                // 4 will be an upward-traveling width
+
+  // create edges via binding operators between nodes in order to define a complete graph
+  drivers.foreach{ driver => adder.node := driver.nodeSeq(0) }
+
+  drivers.zip(checker.nodeSeq).foreach { case (driver, checkerNode) => checkerNode := driver.nodeSeq(1) }
```
hcook, line 284: `driver.nodeSeq(0)` and `driver.nodeSeq(1)` both become `driver.node` if you take my above suggestion

docs/src/diplomacy/adder_tutorial.md
hcook: But maybe there should it should also be an argument like `numOperands`?

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,324 @@
...
+smaller of the two widths to actually be used.
+
+In this module, we also bind our nodes. Sinks are on the left-hand side, while
+sources are on the right.
+
+```scala mdoc
+/** top-level connector */
+class AdderTestHarness()(implicit p: Parameters) extends LazyModule {
+  val adder = LazyModule(new Adder)
+  val drivers = Seq.fill(2) { LazyModule(new AdderDriver(width = 8)) } // 8 will be a downward-traveling width
```
hcook, line 278: this 2 could also be `numOperands`?

docs/src/diplomacy/adder_tutorial.md
```diff
@@ -0,0 +1,324 @@
...
+      dps.head
+    },
+    { case ups: Seq[UpwardParam] =>
+      require(ups.forall(up => up.width == ups.head.width))
+      ups.head
+    }
+  )
+  lazy val module = new LazyModuleImp(this) {
+    require(node.in.size == 2)
+    node.out.head._1 := node.in.head._1 + node.in.tail.head._1
```
hcook, line 209: If we wanted to, we could make this accept any number of operands by making it
```
require(node.in.size >= 2, "adder must have at least two operands")
node.out.head._1 := node.in.unzip._1.reduce(_ + _)
```

docs/src/diplomacy/adder_tutorial.md
hcook: Hmm I guess youd have to make the below logic more complicated too, so nevermind for now

debs-sifive: I think it does look nicer to generalize all the modules to `n` args, so I think I can finagle this one, too!

docs/src/diplomacy/adder_tutorial.md
debs-sifive: 🙏 that looks SO much nicer

PR #2617, opened 2020-08-22
Update to Chisel 3.4.0 and FIRRTL 1.4.0

jackkoenig:
This PR kills the FIRRTL fat jar, instead using https://github.com/freechipsproject/chisel3/pull/1563 to include FIRRTL as a managed dependency in SBT. It also supports toggling between source and published dependencies, if you delete `.sbtopts` it will use the published dependencies.

TODO
- [x] Add test for ability to toggle between source and published dependencies (this currently would not work because RC1 has a bug that breaks rocket-chip)
- [x] FIRRTL test utils

**Related issue**: 

**Impact**: API modification (to build API and Makefiles)

**Release Notes**

Change build flow to build FIRRTL as SBT-managed dependency and use rocket-chip fat jar in Makefiles

### DIFF
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -98,7 +98,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        bucket: [1, 2, 3, 4, 5, 6, 7, 8, 9]
+        bucket: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10]
     steps:
       - uses: actions/checkout@v2
 
--- a/README.md
+++ b/README.md
@@ -14,6 +14,7 @@ the RISC-V Rocket Core. For more information on Rocket Chip, please consult our 
     + [Pushing a Rocket core through the VLSI tools](#vlsi)
 + [How can I parameterize my Rocket chip?](#param)
 + [Debugging with GDB](#debug)
++ [Building Rocket Chip with an IDE](#ide)
 + [Contributors](#contributors)
 
 ## <a name="quick"></a> Quick Instructions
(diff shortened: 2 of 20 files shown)


README.md
```diff
@@ -669,6 +670,32 @@ Now we can proceed as with Spike, debugging works in a similar way:
...
+This file works when invoking SBT in the shell, but because it uses an environment variable `$PWD`, it is not interpreted correctly by IDEs.
+One can fix this issue by using IDE-specific import settings to override the `sbt.workspace` JVM system property to point to the root directory of Rocket Chip with an absolute path, or one can edit the file to expand the variable.
+
+This can be done in `bash` simply as follows:
+
+```bash
+sed -i "s|\$PWD|$PWD|" .sbtopts
+```
+
+So long as this is set properly, the Rocket Chip SBT project should work in any IDE that supports SBT.
```
hcook, line 694: So long as this is set properly, or `.sbtopts` does not exist, ... ?

hcook (approved):
Maybe add something to the readme about when to use/who should use `./scripts/swap-sbt-build`?

README.md
jackkoenig: Yeah that's right, I'll clarify the text

jackkoenig:
> Maybe add something to the readme about when to use/who should use ./scripts/swap-sbt-build?

`./scripts/swap-sbt-build` is only really intended for CI*, how should I clarify that?

*I guess it could be used by me or whoever when bumping Chisel and FIRRTL if it involves switching between building from source and from published artifacts but that's rare and it doesn't save much effort there.

hcook:
Right, I was thinking you were using it to switch between the two. Maybe another way of framing the the question is: you are adding a discussion to the readme of how to modify `.sbtopts` for use with IDEs, if it exists. But does it exist by default, or not? If it does not exist by default, I think that IDE section should have further disclaimers along the lines of "most users dont have to care about the following, but...". If it does exist by default, then I think more explanation about how/why one would get rid of it to depend on the published jars is merited in some other section.

jackkoenig:
Right, it's a little subtle so I understand your confusion and I'm trying to figure out the best way to express it.

Basically, the existence of `.sbtopts` controls if we're building from source or published artifacts. Whether we're building from source or published artifacts is a _property of a particular commit of rocket-chip_, it is **not** something the user should be deleting or adding (that changes the versions of things being used, it's like just checking out a different commit of chisel3 today). So it may exist, or may not, in a given commit of rocket-chip. If it exists, IDE users must do something special to make the IDE work, otherwise they need not do anything except import the project.

hcook:
So then do you view this more as a one way street, where after a certain point in any given branch timeline only the published artifacts are used / the file no longer exists? Or is this something that is going to get toggled back and forth on different branches depending on the workflow of the particular user (i.e. devs working on chisel and firrtl versus everyone else?)

jackkoenig:
> So then do you view this more as a one way street, where after a certain point in any given branch timeline only the published artifacts are used / the file no longer exists? Or is this something that is going to get toggled back and forth on different branches depending on the workflow of the particular user

I'm honestly not sure. My hope is that it's more-or-less a one way street; master (and hopefully someday releases) always work with released versions of chisel and firrtl while weird backport or tapeout branches will likely build from source. That being said, this is the first time moving to published versions so it's hard to say how it will pan out.

.github/workflows/continuous-integration.yml
```diff
@@ -98,7 +98,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        bucket: [1, 2, 3, 4, 5, 6, 7, 8]
+        bucket: [1, 2, 3, 4, 5, 6, 7, 8, 9]
```
richardxia, line 101: Sorry for causing you a merge conflict, but you'll need to renumber your additional test bucket and append a `10` to this list.

kenzhang82:
Hi there,

Thanks for the update! I am just wondering how would the project like https://github.com/sifive/freedom work with the new building system? I guess the other way to ask the question is:

How do I bump up rocket-chip to the newer version in sifive freedom repo with this change?

Thanks,
Ken

PR #2621, opened 2020-08-27
L1TLB: fix support for configurable Sets/Ways

ingallsj:
**Related issue**: follow-on after https://github.com/chipsalliance/rocket-chip/pull/2574

**Type of change**: feature enhancement

**Release Notes**
1. Restore the old field, `nTLBEntries`, by computing what it would be based on the new parameters.  Otherwise, this was a backwards-incompatible change for downstream users that were relying on this.
2. Propagate renaming `nEntries` to `nWays` in `TLBConfig`.
3. Fix a bug where TLB's `sectored_plru` was too wide (`nWays` when it should be `nWays/nSectors`).
4. Add a `Seq` variant for the `access` method in `SetAssocLRU` to support multiple simultaneous updates.


src/main/scala/util/Replacement.scala
```diff
@@ -301,6 +302,17 @@ class SetAssocLRU(n_sets: Int, n_ways: Int, policy: String) extends SetAssocRepl
     state_vec(set) := logic.get_next_state(state_vec(set), touch_way)
   }
 
+  def access(sets: Seq[UInt], touch_ways: Seq[Valid[UInt]]) = {
+    require(sets.size == touch_ways.size)
```
richardxia, line 306: Do you think you could add in the second argument to `require()`, which is the assertion message? It's pretty painful to debug the error messages and stack traces coming from these things as a person who doesn't actually work on the Chisel code.

aswaterman: If it’s an internal consistency check, I don’t think the message is necessary—but if a particular set of configuration parameters could cause the assertion to fire, I totally agree.

ingallsj: it's an internal consistency check

richardxia: I still think it's a good idea to err on the side of adding messages, even for "internal consistency checks". Whether or not something "internal" or "externally configurable" may change over time, and the person who makes that change may not know about all the individual `require()` statements that were internal.

Anyway, thanks @ingallsj for adding a message!

aswaterman: Fair enough

PR #2622, opened 2020-08-29
Reduce common dependencies on the system implicit clock and reset

jerryz123:
This PR changes frequently used peripherals to acquire their clocks diplomatically.

The CLINT, Plic, and BootROM, are now instantiated as children of their attached buses. An alternative implementation could be to wrap these devices in their own `ClockDomains`, instead. 

The interrupt crossings into the tiles are now clocked by the clocks of their source and sink domains, instead of the implicit system clock.

**Related issue**: 

**Impact**: API modification

**Release Notes**

### DIFF
--- a/src/main/scala/devices/tilelink/BootROM.scala
+++ b/src/main/scala/devices/tilelink/BootROM.scala
@@ -74,7 +74,7 @@ object BootROM {
       rom.array() ++ subsystem.dtb.contents
     }
 
-    val bootrom = subsystem {
+    val bootrom = cbus {
       LazyModule(new TLROM(params.address, params.size, contents, true, cbus.beatBytes))
     }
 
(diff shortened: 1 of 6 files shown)


davidbiancolin:
So i foolishly duplicated the work of this PR, though i also had to modify the InterruptBus and PeripheryDebug mixin. @hcook Is there anything we can do to help this along?

hcook:
I think it will have to be changed to the version where they each have a clock domain wrapper and the various extra bits get put in that scope, not in the scope of the bus to which their tilelink port is attached.

davidbiancolin:
> I think it will have to be changed to the version where they each have a clock domain wrapper and the various extra bits get put in that scope, not in the scope of the bus to which their tilelink port is attached.

Coincidentally my implementation does that (mostly) -- i'll open it up, and you can see what you think.

hcook:
Closed in favor of https://github.com/chipsalliance/rocket-chip/pull/2719

PR #2627, opened 2020-09-03
Add maxTransactionsInFlight to IDMapEntry and OMIdMapEntry

hcook:
This PR adds a `maxTransactionsInFlight` field to `IDMapEntry` and `OMIdMapEntry`. This field offers an upper-bound on the number of transactions possibly in flight from a given master associated with a range of ids. The calculation is protocol specific; while it is exactly the number of source ids for TileLink, for AXI4 incomplete information forces us to make a conservative overestimate: assume that every single ID could possibly generate the maximum number of transactions (which may or may not be true), and that each source could generate simultaneous read and write transactions using the same ID (which may or may not be true).

Related issue:

Type of change: feature request

Impact: API addition (no impact on existing code)

Development Phase: implementation

Release Notes

### DIFF
--- a/src/main/scala/amba/axi4/Parameters.scala
+++ b/src/main/scala/amba/axi4/Parameters.scala
@@ -216,7 +216,7 @@ class AXI4IdMap(axi4: AXI4MasterPortParameters) extends IdMap[AXI4IdMapEntry] {
   }
 }
 
-case class AXI4IdMapEntry(axi4Id: IdRange, name: String, maxTransactionsInFlight: Option[Int]) extends IdMapEntry {
+case class AXI4IdMapEntry(axi4Id: IdRange, name: String, maxTransactionsInFlight: Option[Int] = None) extends IdMapEntry {
   val from = axi4Id
   val to = axi4Id
   val isCache = false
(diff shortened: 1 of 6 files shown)


src/main/scala/amba/axi4/Parameters.scala
```diff
@@ -210,11 +210,13 @@ class AXI4IdMap(axi4: AXI4MasterPortParameters) extends IdMap[AXI4IdMapEntry] {
...
   val mapping: Seq[AXI4IdMapEntry] = sorted.map { case c =>
-    AXI4IdMapEntry(c.id, c.name)
+    // to conservatively state max number of transactions, assume every id has up to c.maxFlight and reuses ids between AW and AR channels
+    val maxTransactionsInFlight = c.maxFlight.map(_ * c.id.size * 2)
+    AXI4IdMapEntry(c.id, c.name, maxTransactionsInFlight)
   }
 }
 
-case class AXI4IdMapEntry(axi4Id: IdRange, name: String) extends IdMapEntry {
+case class AXI4IdMapEntry(axi4Id: IdRange, name: String, maxTransactionsInFlight: Option[Int]) extends IdMapEntry {
```
richardxia, line 219: For backwards compatibility, should this default to `None`? Or is case class this only used in this file, and if so, then should this case class be marked as private/protected?

richardxia (approved):
LGTM, with a minor nitpick about backwards compatibility and public/privateness, which are more for my edification about Scala best practices than anything else.

src/main/scala/amba/axi4/Parameters.scala
hcook: It's plausible it should be private, but I decided to just make it take default None for now

PR #2628, opened 2020-09-04
Make SimAXIMem start at expected memory base instead of zero

KireinaHoro:

**Release Notes**

`SimAXIMem` currently assumes physical memory always starts at zero; this is not the case as it is possible to modify `p(ExtMem).get.master.base` for scenarios where the memory does not actually start there (e.g. when instantiating on Zynq FPGA and using the upper half of DRAM).  This change adds support for this situation by reading the base of `ExtMem` and `ExtBus` from the parameters for instantiation.

### DIFF
--- a/src/main/scala/system/SimAXIMem.scala
+++ b/src/main/scala/system/SimAXIMem.scala
@@ -5,14 +5,14 @@ package freechips.rocketchip.system // TODO this should really be in a testharne
 import chisel3._
 import freechips.rocketchip.amba._
 import freechips.rocketchip.amba.axi4._
-import freechips.rocketchip.config.{Parameters}
+import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
-import freechips.rocketchip.subsystem.{CanHaveMasterAXI4MMIOPort, CanHaveMasterAXI4MemPort, ExtMem}
+import freechips.rocketchip.subsystem.{CanHaveMasterAXI4MMIOPort, CanHaveMasterAXI4MemPort, ExtBus, ExtMem}
 
 /** Memory with AXI port for use in elaboratable test harnesses. */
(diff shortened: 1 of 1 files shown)


src/main/scala/system/SimAXIMem.scala
```diff
@@ -5,14 +5,14 @@ package freechips.rocketchip.system // TODO this should really be in a testharne
...
 import freechips.rocketchip.amba.axi4._
-import freechips.rocketchip.config.{Parameters}
+import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
-import freechips.rocketchip.subsystem.{CanHaveMasterAXI4MMIOPort, CanHaveMasterAXI4MemPort, ExtMem}
+import freechips.rocketchip.subsystem.{CanHaveMasterAXI4MMIOPort, CanHaveMasterAXI4MemPort, ExtBus, ExtMem}
 
 /** Memory with AXI port for use in elaboratable test harnesses. */
-class SimAXIMem(edge: AXI4EdgeParameters, size: BigInt)(implicit p: Parameters) extends SimpleLazyModule {
+class SimAXIMem(edge: AXI4EdgeParameters, base: BigInt, size: BigInt)(implicit p: Parameters) extends SimpleLazyModule {
```
hcook, line 13: For backwards compatibility, can you make this `(edge: AXI4EdgeParameters, size: BigInt, base: BigInt = 0)` instead?

KireinaHoro: Done.

KireinaHoro:
Hmm, I have no idea why the test failed; it passed before the default parameter change and the logs aren't so obvious about exactly what failed

sequencer (approved):
Just re-trigger the test, seems passed yesterday. 
Checked this module is non-used in our internal design nor chipyard, so it won't break anything.
@hcook any further suggestions?

hcook (approved):

PR #2631, opened 2020-09-05
Add built Verilator to $PATH in sbt test regression

seldridge:
Fixes a bug where running the `sbt test` regression can fail if a user adds tests that try to call Verilator (e.g., some Chisel testing library). This PR sets the `$PATH` to include the built Verilator before running `sbt test`.

_Note that no existing, upstream test actually uses Verilator._

**Related issue**: None

**Type of change**: bug fix


sequencer:
no more sbt ;p

PR #2634, opened 2020-09-11
Relax master parameter of RocketCrossingParams

jsmithsf:
**Related issue**: 

**Impact**: API modification

**Release Notes**

This change is meant to allow for using more generic or even specialized implementations of the master parameter of RocketCrossingKey.

@hcook I'm a little bit more concerned about my change to subsystem/Configs.scala. Right now just TileMasterPortParams is used. If another subclass gets used then WithIncoherentTiles might not do the right thing.  I wasn't sure if this was the best approach.

### DIFF
--- a/src/main/scala/subsystem/Configs.scala
+++ b/src/main/scala/subsystem/Configs.scala
@@ -234,7 +234,10 @@ class WithBufferlessBroadcastHub extends Config((site, here, up) => {
  */
 class WithIncoherentTiles extends Config((site, here, up) => {
   case RocketCrossingKey => up(RocketCrossingKey, site) map { r =>
-    r.copy(master = r.master.copy(cork = Some(true)))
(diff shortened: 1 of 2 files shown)


hcook:
Yeah, an unfortunate consequence of relaxing the type is the loss of guarantee that there exists a notional cache cork to add or remove. But the `WithIncoherentTiles` Config alteration is also already specific to RocketCrossingKey anyway... I think it has to be up to the SoC integrator to ensure that their equivalent of such an alteration contains the full cross product of tile crossing `Fields` x `TilePortParams` subclasses that they might be exposed to.

A more conservative implementation would be to throw an exception upon encountering a `TilePortParamsLike` subclass that the alteration doesn't understand.

jsmithsf:
@hcook does this look more acceptable?

hcook (approved):

PR #2635, opened 2020-09-11
Remove Verilog delay from DPI outputs and use falling edge instead

jackkoenig:
Verilator ignores these delays (after warning the user), so they can
result in race conditions.

This fixes an issue visible on https://github.com/chipsalliance/rocket-chip/pull/2617 where some assertions fire erroneously due to some optimizations Verilator makes that are incorrect due to the race conditions.

Thank you to @ernie-sifive for doing this work 

**Related issue**: 

**Type of change**: bug fix

**Release Notes**

Remove Verilog delay statements from SimJTAG.v and SimDTM.v

### DIFF
--- a/src/main/resources/vsrc/SimDTM.v
+++ b/src/main/resources/vsrc/SimDTM.v
@@ -35,10 +35,10 @@ module SimDTM(
 
   bit r_reset;
 
-  wire #0.1 __debug_req_ready = debug_req_ready;
-  wire #0.1 __debug_resp_valid = debug_resp_valid;
-  wire [31:0] #0.1 __debug_resp_bits_resp = {30'b0, debug_resp_bits_resp};
-  wire [31:0] #0.1 __debug_resp_bits_data = debug_resp_bits_data;
+  wire __debug_req_ready = debug_req_ready;
+  wire __debug_resp_valid = debug_resp_valid;
+  wire [31:0] __debug_resp_bits_resp = {30'b0, debug_resp_bits_resp};
+  wire [31:0] __debug_resp_bits_data = debug_resp_bits_data;
 
   bit __debug_req_valid;
   int __debug_req_bits_addr;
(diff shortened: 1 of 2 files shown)


src/main/resources/vsrc/SimDTM.v
```diff
@@ -19,26 +19,34 @@ module SimDTM(
...
+  reg [31:0] __debug_resp_bits_resp;
+  reg [31:0]  __debug_resp_bits_data;
+
+//  // Delay sending inputs to DPI to avoid race condition failure in Verilator
+//  always @(negedge clk) begin
+  assign  __debug_req_ready = debug_req_ready;
+  assign  __debug_resp_valid = debug_resp_valid;
+  assign  __debug_resp_bits_resp = {30'b0, debug_resp_bits_resp};
+  assign  __debug_resp_bits_data = debug_resp_bits_data;
+//  end
```
ernie-sifive, line 49: Oops ... looks like some stuff left from the first attempt.
```suggestion
  wire __debug_req_ready = debug_req_ready;
  wire __debug_resp_valid = debug_resp_valid;
  wire [31:0] __debug_resp_bits_resp = {30'b0, debug_resp_bits_resp};
  wire [31:0] __debug_resp_bits_data = debug_resp_bits_data;
```

aswaterman:
In general, using negedge clocking on half of a ready/valid interface is functionally incorrect.  It might be the case that this interface is constrained in such a way that it's OK in this case, but I don't know.

ernie-sifive (approved):
All checks passed on latest commit.  Changed output back to posedge to confine any negedge to within the SimDTM module.

jackkoenig:
Is there a fundamental reason why the logic needs to be different for Verilator? Why does the `negedge` timing work for Verilator but not VCS?

aswaterman:
> Why does the `negedge` timing work for Verilator but not VCS?

This is a red flag for me, too.

aswaterman:
I haven't thought too hard about the details, but superficially, this looks good to me.

jackkoenig:
Let me verify that this works on the 3.4 branch then hopefully we're good to go!

jackkoenig:
It's working on the 3.4 branch!

jackkoenig:
I believe this is right so I'm going to merge

PR #2638, opened 2020-09-15
axi4: remove <>s from metadata

hcook:
Any string that could potentially show up as metadata in the graphml cannot have text bracketed by `<...>` without that text being interpreted as a XML tag. Such a string was getting inserted in some cases for designs with AXIIdIndexers, a situation which was introduced by by me in https://github.com/chipsalliance/rocket-chip/pull/2483/ (though I don't think the data in question was being ingested into the graphml at that point?)

**Related issue**: 


terpstra:
This happens even though it's inside a string?!

terpstra (approved):

rmac-sifive (approved):

mwachs5:
https://xkcd.com/327/

PR #2640, opened 2020-09-16
subsystem: update InterruptBusWrapper crossing code

hcook:
This updates `InterruptBusWrapper` in the following backwards-compatible ways:
- It becomes a `SimpleLazyModule with HasClockCrossings`
- But the wrapper layer is later inlined by FIRRTL
- Parameterized clock crossing helper methods are exposed
- But the legacy connection functions are preserved
- No more use of deprecated `IntXing`

The beneficial effect beyond just modernization of APIs and more parameterizable crossing types is that the synchronization registers made by each crossing type will now be instantiated inside the `InterruptBusWrapper` hierarchically, each taking its clock and reset as their own. While in many cases the wrapper's clock and reset are the same as the subsystem clock and reset in which the `from` methods are presumed to be called (e.g. by the peripherals' `attachTo` methods), in certain cases these methods do get called inside some other scope. Thus this change should ensure that the synchronizing registers are always put in the destination clock domain, i.e. the one containing the InterruptBusWrapper and the PLIC, and not the hierarchy where the peripheral itself is attached.

**Release Notes**
Synchronizers


rmac-sifive (approved):
I was initially thinking it'd be nice to keep the optionality to instantiate synchronizers in the source domain, but on second thought I can't think of scenarios in which a designer would ever want to do that

PR #2618, opened 2020-08-24
Expose control over take field for all ClockSinkParameters in system

jerryz123:
This PR exposes control to set the `take` fields of all ClockSinks in the system. This lets us programmatically generate PLL models by inspecting all desired frequencies at a single `ClockGroupSourceNode`.

A `clockSinkParams` field is added to `TileParams`, and parameterizes the `ClockSinkNode` in the `TilePRCIDomain`.
Additionally, the `CoherenceManagerWrapper` is now set with the same `dtsFrequency` as the `SystemBus`, which drives its clocks

**Related issue**: https://github.com/ucb-bar/chipyard/pull/614

**Type of change**: feature request

**Impact**: API addition (no impact on existing code)

**Development Phase**: implementation

**Release Notes**

PR #2641, opened 2020-09-17
Define ResetCrossingType and use with BlockDuringReset in TilePRCIDomain

hcook:
This PR combines and merges the effects of #2618 and #2623 while also partially reverting #2611. The overall goals are:
- Avoid tunneling the "raw" subsystem reset into `RocketTile`, instead, move all logic requiring raw reset into the `TilePRCIWrapper` 
- Said logic largely takes the form of registers that block tile outputs from appearing valid while the tile is being held in an extended reset, in which case those registers need to be asynchronously reset for certain `ResetSchemes` (when a `ResetStretcher` is injected) and then they need to maintain the blockage for the duration of any stretched reset.
-  `TileResetDomain` is added to manage diplomatic control of the Tile's `reset` input. All other clock crossing logic in `TilePRCIDomain` goes back to being on the raw subsystem reset (but still on the tile clock domain). `TileResetDomain` is inlined by firrtl to prevent further disruption to the module hierarchy.
- `ResetCrossingType` is defined to be analogous to `ClockCrossingType`, the two defined reset crossing types are currently `NoResetCrossing` and `StretchedResetCrossing(cycles: Int)`
- The diplomatic `CrossingHelper` functionality for TL and Interrupts are extended to also deal with `ResetCrossingType`
- Specifically, for `StretchedResetCrossing` they will inject `BlockDuringReset` logic in the scope where the helper is invoked; in this specific case, that scope is `TilePRCIDomain`, ensuring the logic gets the raw reset.
- Various other minor parameterization cleanups and helper nodes.

@aswaterman, @ernie-sifive can you double check that my changes to the actual `BlockDuringReset` logics look kosher?

**Impact**:  API modification

**Release Notes**

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -14,6 +14,7 @@ import freechips.rocketchip.interrupts._
 import freechips.rocketchip.rocket._
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.util._
+import freechips.rocketchip.prci.{ClockSinkParameters}
 
 case object TileVisibilityNodeKey extends Field[TLEphemeralNode]
 case object TileKey extends Field[TileParams]
@@ -28,6 +29,7 @@ trait TileParams {
   val beuAddr: Option[BigInt]
   val blockerCtrlAddr: Option[BigInt]
   val name: Option[String]
+  val clockSinkParams: ClockSinkParameters
 }
 
 abstract class InstantiableTileParams[TileType <: BaseTile] extends TileParams {
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -247,20 +247,23 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
   val mmioAddressPrefixNode: BundleBridgeInwardNode[UInt] =
     mmioAddressPrefixNexusNode :=* BundleBridgeNameNode("mmio_address_prefix")
 
-  protected def traceRetireWidth = tileParams.core.retireWidth
-  protected def traceCoreParams = new TraceCoreParams()
-  protected def traceCoreSignalName = "tracecore"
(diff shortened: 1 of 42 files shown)


src/main/scala/tile/BaseTile.scala
```diff
@@ -248,11 +249,9 @@ abstract class BaseTile private (val crossing: ClockCrossingType, q: Parameters)
 
   /** Node for the core to drive legacy "raw" instruction trace. */
   val traceSourceNode = BundleBridgeSource(() => Vec(traceRetireWidth, new TracedInstruction()))
-  /** Node to broadcast legacy "raw" instruction trace. */
-  val traceNexusNode = BundleBroadcast[Vec[TracedInstruction]]()
   /** Node for external consumers to source a legacy instruction trace from the core. */
   val traceNode: BundleBridgeOutwardNode[Vec[TracedInstruction]] =
-    BundleBridgeNameNode("trace") :*= traceNexusNode := traceSourceNode
+    BundleBridgeNameNode("trace") := traceSourceNode
```
ernie-sifive, line 254: So this moves `traceNexusNode` to the PRCI layer so it can be blocked during reset.  There's some technical debt to apply the same treatment to `traceCoreNexusNode` (below).  This bundle doesn't have a valid signal.  Either we will have to add a `valid` to `TraceCoreInterface` (bending the definition of this interface from the RISC-V Trace Spec) or add another form of `BlockDuringReset` specifically for this (which would zero out `iretire` and `itype`).

ernie-sifive (approved):
I think the reset treatment matches what we discussed.

src/main/scala/tile/BaseTile.scala
hcook: Thanks for the feedback, I was wondering a bit about whether that bundle needed to be blocked too. There's almost certainly a way to have each Bundle subclass define it's own `blockMe` function (name TBD) that returns a copy of itself but with certain signals tied low. You still have to make one NexusNode per type of bundle bridge, but it's at least very generic code to ensure that the blocking occurs.

PR #2648, opened 2020-09-30
Add scalafix and remove unused imports

richardxia:
**Related issue**: 

**Release Notes**

---

## Changes and motivation

This PR is two parts:

1. https://github.com/chipsalliance/rocket-chip/commit/0a5e2e8ec5e2230311641f00ad2dd0cfc527eebb - adding [Scalafix](https://scalacenter.github.io/scalafix/) to our SBT project/running it in CI, and
2. https://github.com/chipsalliance/rocket-chip/commit/66a60e55a18fae859730b01c04450b2da3630e19 - running Scalafix to automatically delete any unused imports.

My main motivation for adding Scalalint is solely to get rid of unused imports and to prevent future unused imports from being left into the codebase. In my experience with this in other projects in other languages (Python, JavaScript), this is not a controversial thing to add as a lint check and enforce unconditionally. However, I don't have a good sense of whether this kind of thing is normal or uncontroversial in a Scala project.

My main reason for wanting to get rid of the unnecessary imports is so that it becomes easier to reason about the actual internal architecture of the rocket-chip codebase, since the unnecessary imports make it much more difficult to tell which parts of the codebase depend on which other parts. As a vim user who is not using a language server (yet), I understand that my experience here is not representative of a typical Scala programmer, since IDEs normally make code navigation easier.

That said, I hope this is not too much of a burden for developers, since this will mean that unused imports will be enforced by CI and you will be required to delete them before you can pass CI checks.

## Question about rollout

Normally I do not like enabling and enforcing new lint rules on large, preexisting projects, since it often ends up touching a large number of files and can cause merge conflicts with other open PRs. For this PR, however, I think that because it's restricted to just unused imports, all the merge conflicts will be located in a single place at the top of any affected file, rather than potentially spanning the entirety of a file.

I'm open to a much more gradual rollout strategy, however, where we whitelist the existing violations but enforce the rule for any files that are not currently in violation.

## Other lint rules
Scalafix comes with [7 builtin rules](https://scalacenter.github.io/scalafix/docs/rules/overview.html), but I've only enabled four of them here. I found no violations of [`DisableSyntax`](https://scalacenter.github.io/scalafix/docs/rules/DisableSyntax.html), [`NoAutoTupling`](https://scalacenter.github.io/scalafix/docs/rules/NoAutoTupling.html), or [`NoValInForComprehension`](https://scalacenter.github.io/scalafix/docs/rules/NoValInForComprehension.html), so I decided to enable those checks.

[`RemoveUnused`](https://scalacenter.github.io/scalafix/docs/rules/RemoveUnused.html) obviously has violations, but I enabled the rule and fixed the violations in this PR.

The remaining 3 rules do have violations in this codebase, and some of them I think may not work well in the Chisel DSL. I _think_ [`LeakingImplicitClassVal`](https://scalacenter.github.io/scalafix/docs/rules/LeakingImplicitClassVal.html) and [`ProcedureSyntax`](https://scalacenter.github.io/scalafix/docs/rules/ProcedureSyntax.html) are probably worth enabling, but I don't feel that I know Scala and Chisel well enough to have a sense of whether these should unconditionally enforced.

[`ExplicitResultTypes`](https://scalacenter.github.io/scalafix/docs/rules/ExplicitResultTypes.html) gave me the most trouble. Although it offers a lot of customizability, I couldn't really figure out how to configure it such that it doesn't touch almost every file in the codebase. One specific scenario that I saw over and over in our codebase but could not figure out how to whitelist was defining a method or a `val` whose right-hand side is a single expression involving a case class factory method (i.e. `apply()`). For example, it would want every instance of `def foo = Bar()` to be rewritten as `def foo: Bar = Bar()`.

I can understand why it wants to do this; the case class factory is just a convention, and it can't know whether you are just using it as shorthand for `new Bar()` or if it is actually doing something else. The lint rule has a way to configure it to ignore simple definitions that just involve `new MyClass()`, but not for case class factory methods.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -87,7 +87,11 @@ lazy val rocketchip = dependOnChisel(project in file("."))
       mappings in (Compile, packageBin) ++= (mappings in (`rocket-macros`, Compile, packageBin)).value,
       mappings in (Compile, packageSrc) ++= (mappings in (`rocket-macros`, Compile, packageSrc)).value,
       exportJars := true,
-      Test / unmanagedBase := baseDirectory.value / "test_lib"
+      Test / unmanagedBase := baseDirectory.value / "test_lib",
+      // Settings for scalafix
+      semanticdbEnabled := true,
+      semanticdbVersion := scalafixSemanticdb.revision,
+      scalacOptions += "-Ywarn-unused-import"
   )
 
 
--- a/.github/workflows/continuous-integration.yml
+++ b/.github/workflows/continuous-integration.yml
@@ -98,7 +98,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        bucket: [1, 2, 3, 4, 5, 6, 7, 8]
+        bucket: [1, 2, 3, 4, 5, 6, 7, 8, 9]
     steps:
       - uses: actions/checkout@v2
 
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -1,0 +1,6 @@
+rules = [
+  RemoveUnused,
+  DisableSyntax,
+  NoAutoTupling,
+  NoValInForComprehension,
+]
--- a/.travis.yml
+++ b/.travis.yml
@@ -104,3 +104,7 @@ jobs:
       name: Scala tests
       script:
         - ./regression/run-test-bucket 8
+    - <<: *test*
+      name: Scala tests
+      script:
+        - ./regression/run-test-bucket 9
--- a/README.md
+++ b/README.md
@@ -84,6 +84,19 @@ To generate FPGA- or VLSI-synthesizable Verilog (output will be in `vsim/generat
     $ cd vsim
     $ make verilog
 
+To run the Scala tests (`sbt test`) or linter (`sbt scalafix`):
+
+    $ cd regression
+
+    # Scala tests
+    $ make scalatest SUITE=foo
+
+    # Scala linter, automatically modifying files to correct issues
+    $ make scalafix SUITE=foo
+
+    # Scala linter, only printing out issues
+    $ make scalafix-check SUITE=foo
+
 
 ### Keeping Your Repo Up-to-Date
 
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -15,3 +15,5 @@ addSbtPlugin("org.scoverage" % "sbt-scoverage" % "1.5.1")
 addSbtPlugin("com.typesafe" % "sbt-mima-plugin" % "0.6.1")
 
 addSbtPlugin("org.scalameta" % "sbt-mdoc" % "2.2.5" )
+
+addSbtPlugin("ch.epfl.scala" % "sbt-scalafix" % "0.9.21")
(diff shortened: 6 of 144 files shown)


build.sbt
```diff
@@ -87,7 +87,11 @@ lazy val rocketchip = dependOnChisel(project in file("."))
       mappings in (Compile, packageBin) ++= (mappings in (`rocket-macros`, Compile, packageBin)).value,
       mappings in (Compile, packageSrc) ++= (mappings in (`rocket-macros`, Compile, packageSrc)).value,
       exportJars := true,
-      Test / unmanagedBase := baseDirectory.value / "test_lib"
+      Test / unmanagedBase := baseDirectory.value / "test_lib",
+      // Settings for scalafix
+      semanticdbEnabled := true,
+      semanticdbVersion := scalafixSemanticdb.revision,
+      scalacOptions += "-Ywarn-unused-import"
```
richardxia, line 94: @jackkoenig, let me know if this is the best way to add in the configuration for Scalafix. I don't even know what SemanticDB is, but it said I needed to enable it by adding these lines to the project. `-Ywarn-unused-import` is apparently required for the `RemoveUnused` rule to function correctly.

hcook (approved):
Thank you for doing this. By my count this removes 5 or 6 circular dependencies in the imports, undoubtedly left behind when some class was moved between packages without causing breakages.

[Past attempts](https://github.com/chipsalliance/rocket-chip/pull/2097) at using scala-fix ran aground based on some behavior that would convert `val x = f()` into `f()` in situations where `x` was not used in the local scope; but actually the correct thing to do in those situations was either delete the line entirely (with  human knowledge that `f` has no side-effects) or in one case fix the fact that `x` was being calculated and then _not_ put into some metadata case class. I mention this because I am not sure which rule it was that triggered the behavior but it will be one that requires special auditing before being activated in the future.

I would be in favor of enabling both `LeakingImplicitClassVal` and `ProcedureSyntax`.

richardxia:
> Thank you for doing this. By my count this removes 5 or 6 circular dependencies in the imports, undoubtedly left behind when some class was moved between packages without causing breakages.
> 
> [Past attempts](https://github.com/chipsalliance/rocket-chip/pull/2097) at using scala-fix ran aground based on some behavior that would convert `val x = f()` into `f()` in situations where `x` was not used in the local scope; but actually the correct thing to do in those situations was either delete the line entirely (with human knowledge that `f` has no side-effects) or in one case fix the actual bug that `x` was being calculated and then _not_ put into some metadata case class. I mention this because I am not sure which rule it was that triggered the behavior but it will be one that requires special auditing before being activated in the future.
> 
> I would be in favor of enabling both `LeakingImplicitClassVal` and `ProcedureSyntax`.

Thanks! Let me merge this in now before people start changing code that may touch the imports, but I can submit followup PRs for the other lint rules where we can discuss whether we like the results or whether we want to tweak the settings further.

richardxia:
I have been informed that this change will break any SBT projects that use rocket-chip as a subproject if the parent project does not explicitly include `sbt-scalafix` in plugins.sbt, even if the parent project doesn't have scalafix enabled. I'll update this PR's description to indicate that this is apparently a backwards-incompatible change for any project that depends on rocket-chip's SBT project.

PR #2651, opened 2020-10-02
Enable ProcedureSyntax lint rule

richardxia:
**Related issue**: 

**Release Notes**

---

Followup to https://github.com/chipsalliance/rocket-chip/pull/2648. Compared to https://github.com/chipsalliance/rocket-chip/pull/2650, this lint rule requires modifying a much larger number of lines of code in rocket-chip.

This PR enables the [`ProcedureSyntax`](https://scalacenter.github.io/scalafix/docs/rules/ProcedureSyntax.html) lint rule in Scalafix. A deprecated syntax of Scala was to allow for a Java/C-style method declaration that did not require an explicit `=` separating the method name and type signature from the body of the method. This lint rule detects instances of that older syntax and will change them to 1) have the explicit return type of `Unit` and 2) to place an `=` between the return type and the opening curly brace of the method body.

### DIFF
--- a/.scalafix.conf
+++ b/.scalafix.conf
@@ -3,4 +3,5 @@ rules = [
   DisableSyntax,
   NoAutoTupling,
   NoValInForComprehension,
+  ProcedureSyntax,
 ]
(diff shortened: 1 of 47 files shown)


aswaterman:
Why do we want to do this?  The procedure-style syntax is unambiguous.

richardxia:
> Why do we want to do this? The procedure-style syntax is unambiguous.

It will no longer supported in Scala 3: https://dotty.epfl.ch/docs/reference/dropped-features/procedure-syntax.html

aswaterman (approved):
That is certainly a valid reason.

seldridge (approved):
Explicit return types look good to me.

hcook (approved):

PR #2654, opened 2020-10-06
Add mill build system

sequencer:
**Related issue**: 

**Release Notes**
This PR adds mill build system.
`common.sc` is used for using rocketchip as in-project build-from-source dependency.
for example(`mychisel3`, `myhardfloat`, `myconfig` are other in-project dependencies.)
```
import $`file.rocket-chip.common`

object myrocketchip extends rocketchip.common.CommonRocketChip {
  override def millSourcePath = os.pwd / "rocket-chip"

  def chisel3Module: Option[PublishModule] = Some(mychisel3)

  def hardfloatModule: PublishModule = myhardfloat

  def configModule: PublishModule = myconfig
}
```
can create a `myrocketchip` module, while other project can directly depend on it

`build.sc` is used for running rocket-chip as top project:
Using `mill rocketchip.run` to run `RocketChipStage`
Using `mill rocketchip.test` to run in-project tests.

### DIFF
--- a/common.sc
+++ b/common.sc
@@ -1,0 +1,126 @@
+import mill._
+import mill.scalalib._
+import mill.scalalib.publish._
+import coursier.maven.MavenRepository
+
+val defaultVersions = Map(
+  "chisel3" -> "3.4.0",
+  "chisel3-plugin" -> "3.4.0"
+)
+
+def getVersion(dep: String, org: String = "edu.berkeley.cs", cross: Boolean = false) = {
+  val version = sys.env.getOrElse(dep + "Version", defaultVersions(dep))
+  if (cross)
+    ivy"$org:::$dep:$version"
+  else
+    ivy"$org::$dep:$version"
+}
+
+/** The reason to split build.sc to two file is
+  * [[CommonRocketChip]] doesn't need to import `$file.chisel3` and `$file.firrtl`.
+  *
+  * You can extends from [[CommonRocketChip]] to use rocket-chip as build-from-source dependency.
+  * When doing this, you may like to override `chisel3Module`, `hardfloatModule`, `configModule`,
+  * setting them to your favorite commit of those packages.
+  *
+  * If you don't override `chisel3Module`, which will default to be `None`,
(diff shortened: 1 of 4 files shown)


common.sc
```diff
@@ -0,0 +1,126 @@
+import mill.{Agg, _}
+import mill.scalalib._
+import mill.scalalib.publish._
+import coursier.maven.MavenRepository
+
+val defaultVersions = Map(
+  "chisel3" -> "3.4.0-RC3",
+  "chisel3-plugin" -> "3.4.0-RC3"
+)
```
jackkoenig, line 9:
```suggestion
val defaultVersions = Map(
  "chisel3" -> "3.4.0",
  "chisel3-plugin" -> "3.4.0"
)
```

common.sc
```diff
@@ -0,0 +1,126 @@
...
+  private val macroParadise = ivy"org.scalamacros:::paradise:2.1.1"
+
+  private val chisel3Plugin = getVersion("chisel3-plugin")
+
+  override def repositories = super.repositories ++ Seq(
+    MavenRepository("https://oss.sonatype.org/content/repositories/snapshots"),
+    MavenRepository("https://oss.sonatype.org/content/repositories/releases")
+  )
+
+  override def compileIvyDeps = Agg(macroParadise) ++ (if(chisel3Module.isDefined) Agg() else Agg(chisel3Plugin))
```
jackkoenig, line 107: Is this needed since the plugin is declared in the `scalacPluginIvyDeps`?

jackkoenig (approved):
Looks good with a couple of suggestions

common.sc
sequencer: I just checked in:
https://github.com/lihaoyi/mill/blob/cc0d0fc2de175d295e7905f11d709ca69536e616/integration/test/resources/play-json/build.sc#L70-L75, so I think `macroParadise` should exist in both cases.
But I think `chisel3Plugin` only needed in `scalacPluginIvyDeps`.

PR #2659, opened 2020-10-09
Misc. fixes to make Rocket tolerant to X-propagation

aswaterman:
grumble grumble

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/util/package.scala
+++ b/src/main/scala/util/package.scala
@@ -185,6 +185,9 @@ package object util {
     def inRange(base: UInt, bounds: UInt) = x >= base && x < bounds
 
     def ## (y: Option[UInt]): UInt = y.map(x ## _).getOrElse(x)
+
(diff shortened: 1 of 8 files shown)


src/main/scala/util/package.scala
```diff
@@ -185,6 +185,9 @@ package object util {
     def inRange(base: UInt, bounds: UInt) = x >= base && x < bounds
 
     def ## (y: Option[UInt]): UInt = y.map(x ## _).getOrElse(x)
+
+    // Like >=, but prevents x-prop for ('x >= 0)
+    def >== (y: UInt): Bool = x >= y || y === 0.U
```
hcook, line 190: how about `>=x̄` (jk or am i)

hcook (approved):

src/main/scala/util/package.scala
aswaterman: ha

PR #2666, opened 2020-10-13
[prci] Add a pair of diplomatic reset synchronizers

davidbiancolin:

**Release Notes**
* [prci] Add a pair of diplomatic reset synchronizers

Some other thoughts: 
- Here I don't bother exposing any of the underlying parameters of ResetCatchAndSync; if that's desired i can add requested ones. 
- ResetWrangler could be made more diplomatic in spirit by removing its direct instantiation of ResetCatchAndSync.

### DIFF
--- a/src/main/scala/prci/ResetSynchronizer.scala
+++ b/src/main/scala/prci/ResetSynchronizer.scala
@@ -1,0 +1,43 @@
+// See LICENSE for license details.
+package freechips.rocketchip.prci
+
+import freechips.rocketchip.config.{Parameters}
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.util.{ResetCatchAndSync}
+
+/**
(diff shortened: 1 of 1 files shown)


colinschmidt:
I feel like the `sync` parameters for `AsyncQueue` and `ResetCatchAndSync` should be coupled by default in my opinion, and having that default settable by some parameter would be all I cared about in terms of params.
This is not a very straightforward change or one thats related to this PR necessarily but it's my 2 cents.

hcook (approved):

hcook:
@colinschmidt I agree that there is some potential value to coupling them, but I wonder what the scope of such coupling is... is it really global? Does it depend how relatively asynchronous the two domains in question are? (I confess I don't personally know much about the criteria that go into successfully choosing this parameter.)

PR #2673, opened 2020-10-21
IDPool: synthesis timing enhancement argument options

ingallsj:

**Release Notes**
add IDPool arguments `lateValid` and `revocableSelect` to shift the deep logic cones from before the `valid/select` registers to after the `bitmap` register


mhtwn (approved):
Looks good to me

PR #2674, opened 2020-10-21
add Select combinators

albertchen-sifive:
This is a sketch of a [Select](https://github.com/freechipsproject/chisel3/blob/master/src/main/scala/chisel3/aop/Select.scala) library for diplomatic nodes and `LazyModules`. The motivation behind this is to enable fine-grained selection of node bundles for things like monitoring, without having to override the `NodeImp.monitor` method or change a `LazyModule`'s `Parameters` or modify any design code.

**Related issue**: 

**Release Notes**

### DIFF
--- a/docs/src/diplomacy/select_tutorial.md
+++ b/docs/src/diplomacy/select_tutorial.md
@@ -120,7 +120,7 @@ Select.filterCollectDeep(top) {
 `LazyModule`s have a `getNodes` method that return all the nodes instantiated
 within that module. This can be combined with the `Select.collectInwardEdges`
 to select `LazyModule`s based on their connectivity.
-`Select.collectInwardEdges` takes a `BaseNode` and a parital function. It
+`Select.collectInwardEdges` takes a `BaseNode` and a partial function. It
 applies the function to all the `InwardEdge`s of the node. There is also a
 `Select.collectOutwardEdges` that does the same for outward edges. The example
 below uses `Select.filterCollectDeep`, `LazyModule.getNodes`, and
@@ -133,6 +133,7 @@ Select.filterCollectDeep (top) {
   case _ => true
 } {
   case a: A =>
+    // Function.unlift used to convert `InwardEdge => Option[String]` to `PartialFunction[InwardEdge, String]
     a.getNodes.flatMap(Select.collectInwardEdges(_)(Function.unlift { edge =>
       edge.node.lazyModule match {
         case l: Leaf => Some(l.pathName)
--- a/docs/src/diplomacy/select_tutorial.md
+++ b/docs/src/diplomacy/select_tutorial.md
@@ -1,0 +1,143 @@
+# Select Library
+Chisel provides a [Select library](https://github.com/freechipsproject/chisel3/blob/master/src/main/scala/chisel3/aop/Select.scala)
+that provides a set of methods for finding specific sets of hardware nodes in a
+Chisel design after it has been elaborated. However, Diplomacy abstracts away a
+lot of the underlying Chisel hardware, so the low-level Chisel `Select`
+combinators can be fragile and difficult to use in designs using Diplomacy. To
+help with this, Rocket Chip provides its own `Select` library that operates on
+`LazyModule`s and `Node`s instead of `Module`s and `Wire`s.
+
+We will use the following `LazyModule`s in the examples below.
+```scala mdoc
+import chisel3.Bool
+import freechips.rocketchip.aop.Select
+import freechips.rocketchip.config.Parameters
+import freechips.rocketchip.diplomacy.{
+  BundleBridgeSink,
+  BundleBridgeSource,
+  LazyModule,
+  LazyModuleImp,
+  SimpleLazyModule
+}
+
+class Top(implicit p: Parameters) extends LazyModule {
+  val a = LazyModule(new A)
+  val foo = LazyModule(new Foo)
+
+  val aInput = BundleBridgeSource[Bool](() => Bool())
+  a.input := aInput
+
+  val aOutput = a.output.makeSink
+
+  val fooInput = BundleBridgeSource[Bool](() => Bool())
+  foo.input := fooInput
+
+  val fooOutput = foo.output.makeSink
+
+  lazy val module = new LazyModuleImp(this) {
+    aInput.makeIO
+    fooOutput.makeIO
+    fooInput.bundle := aOutput.bundle
+  }
+}
+
+class Foo(implicit p: Parameters) extends SimpleLazyModule {
+  val bar = LazyModule(new A)
+
+  val input = bar.input
+  val output = bar.output
+}
+
+class A(implicit p: Parameters) extends LazyModule {
+  val b = LazyModule(new Leaf)
+  val c = LazyModule(new Leaf)
+
+  val input = b.input
+  val output = c.output
+
+  val bOutput = b.output.makeSink
+  val cInput = BundleBridgeSource[Bool](() => Bool())
+  c.input := cInput
+  lazy val module = new LazyModuleImp(this) {
+    cInput.bundle := bOutput.bundle
+  }
+}
+
+class Leaf(implicit p: Parameters) extends LazyModule {
+  val input = BundleBridgeSink[Bool]()
+  val output = BundleBridgeSource[Bool](() => Bool())
+
+  lazy val module = new LazyModuleImp(this) {
+    output.bundle := input.bundle
+  }
+}
+
+val top = LazyModule(new Top()(Parameters.empty))
+```
+
+```scala mdoc:invisible
+// Select methods can only be used after instantiation
+chisel3.stage.ChiselStage.elaborate(top.module)
+```
+
+The instance `top` has the following hierarchy:
+```
+        top:Top
+        /      \
+      a:A      foo:Foo
+     /  \        \
+b:Leaf  c:Leaf   bar:A
+                  /  \
+              b:Leaf  c:Leaf
+```
+
+## Examples
+`Select.collectDeep` takes a `LazyModule` and a partial function. It
+recursively applies the partial to the module and all of its children. The
+following example uses `Select.collectDeep` to collect all `Leaf` modules in
+`top`.
+```scala mdoc
+Select.collectDeep(top) {
+  case l: Leaf => l.pathName
+}
+```
+
+`Select.filterCollectDeep` takes a `LazyModule`, a filter function, and a
+partial function. It recursively applies the partial to the module and all of
+its children if the filter function returns true. When the filter function
+returns `false`, the recursion stops.  The following example uses
+`Select.filterCollectDeep` to collect all `Leaf` modules in `top` that do not
+belong to the `foo:Foo` subhierarchy.
+```scala mdoc
+Select.filterCollectDeep(top) {
+  case _: Foo => false
+  case _ => true
+} {
+  case l: Leaf => l.pathName
+}
+```
+
+`LazyModule`s have a `getNodes` method that return all the nodes instantiated
+within that module. This can be combined with the `Select.collectInwardEdges`
+to select `LazyModule`s based on their connectivity.
+`Select.collectInwardEdges` takes a `BaseNode` and a parital function. It
+applies the function to all the `InwardEdge`s of the node. There is also a
+`Select.collectOutwardEdges` that does the same for outward edges. The example
+below uses `Select.filterCollectDeep`, `LazyModule.getNodes`, and
+`Select.collectInwardEdges` to find all instances of `Leaf` modules in `top`
+that are connected to an `A` module and do not belong to the `foo:Foo`
+subhierarchy.
+```scala mdoc
+Select.filterCollectDeep (top) {
+  case _: Foo => false
+  case _ => true
+} {
+  case a: A =>
+    a.getNodes.flatMap(Select.collectInwardEdges(_)(Function.unlift { edge =>
+      edge.node.lazyModule match {
+        case l: Leaf => Some(l.pathName)
+        case _ => None
+      }
+    }))
+}
+```
--- a/src/main/scala/aop/Select.scala
+++ b/src/main/scala/aop/Select.scala
@@ -18,24 +18,6 @@ import freechips.rocketchip.diplomacy.{
   *
   * These can be used for e.g. finding specific TLBundles in a design and
   * placing monitors or annotating metadata.
-  *
-  * For example, the following block will print all instances of MyModule that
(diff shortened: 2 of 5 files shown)


src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -5,14 +5,104 @@ package freechips.rocketchip.diplomacy
...
+import chisel3.experimental.{BaseModule, ChiselAnnotation}
+import chisel3.aop.{Select => CSelect}
 import firrtl.passes.InlineAnnotation
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.util.CompileOptions.NotStrictInferReset
 
 import scala.collection.immutable.{ListMap, SortedMap}
 import scala.util.matching._
 
+object Select {
```
mwachs5, line 17: we laboriously added scaladoc to this file... please continue this for this object!

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -5,14 +5,104 @@ package freechips.rocketchip.diplomacy
...
+  type AnyMixedNode = MixedNode[_, _, _, _ <: Data, _, _, _, _ <: Data]
+
+  case class InwardEdge[Bundle <: Data, EdgeInParams](
+    params: Parameters,
+    bundle: Bundle,
+    edge: EdgeInParams,
+    node: OutwardNode[_, _, Bundle],
+  )
```
jackkoenig, line 25: I see a lot of wildcard type parameters here. Is it possible someone may someday want to care about the first two type parameters of `node`? If so, we should probably go ahead and add them because adding them later will be very painful for anyone using this type.

That being said I don't know much about best practices and maintaining backwards compatibility in using type parameters.

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -5,14 +5,104 @@ package freechips.rocketchip.diplomacy
...
+  }
+
+  def collectInwardEdges[T](node: BaseNode)(collect: PartialFunction[InwardEdge[_ <: Data, _], T]): Iterable[T] = {
+    node match {
+      case node: AnyMixedNode =>
+        require(node.instantiated,
+          s"""Select helpers cannot be used with node "${node.name}" that has not been instantiated yet.""")
+        node.bundlesSafeNow = true
+        val result = getInwardEdges(node).collect(collect)
+        node.bundlesSafeNow = false
```
hcook, line 59: Flipping `bundlesSafeNow` on and off every time you want to access a bundle seems like...contravening the intentionality of the flag. Either the bundles are safe to instantiate/manipulate or they aren't, seem really weird to have this code toggling it. I suspect there's duplication of functionality with your `instantiated` that should just be cleaned up/unified

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -5,14 +5,104 @@ package freechips.rocketchip.diplomacy
...
+import chisel3.aop.{Select => CSelect}
 import firrtl.passes.InlineAnnotation
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.util.CompileOptions.NotStrictInferReset
 
 import scala.collection.immutable.{ListMap, SortedMap}
 import scala.util.matching._
 
+object Select {
+  type AnyMixedNode = MixedNode[_, _, _, _ <: Data, _, _, _, _ <: Data]
```
hcook, line 18: I would just add this to the diplomacy package object

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -1261,6 +1262,7 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
     * and return all the dangles of this node.
     */
   protected[diplomacy] def instantiate(): Seq[Dangle] = {
+    instantiated = true
     bundlesSafeNow = true
```
hcook, line 1266: As I mentioned above, it seems like these are exactly the same thing with the caveat that `bundlesSafeNow` is turned back off again in `finishInstantiate` below. But... it seems that code as originally written had the design assumption that the only time in Scala execution that one would want to manipulate such a bundle is during the execution of its `LazyModuleImp`. Your use case is literally disproving that claim, and I can imagine similar issues happening with e.g. `BoringUtils` or any other spooky-action-at-a-distance libraries. So I think we should retire `bundlesSafeNow` in favor of your `instantiated`, which should preserve the guard against accessing bundles too early while allowing them to be accessed "too late".

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -1235,7 +1235,8 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       data   = bundleIn(i))
   }
 
-  private var bundlesSafeNow = false
+  private[diplomacy] var bundlesSafeNow = false
```
hcook, line 1238: Having to soften this guard is a symptom of something being wrong with the existing API, IMO. E.g. if we just had `instantiated` would you `Select` object even have to be a part of this library? (Not saying it shouldn't be anyways, just that it wouldn't have to be)

src/main/scala/diplomacy/LazyModule.scala
hcook: I would put this in its own file, while also doing as @mwachs5 suggests 😎

src/main/scala/diplomacy/LazyModule.scala
hcook: In this case, those parameters are kind of implicitly captured by the `EdgeInParams` type member, which inevitably will have members of the "missing" `DI` and `UI` types. So I guess I don't see much value in offering those through this API. I do wonder whether since these classes are part of the public `collect` interface, should they become part of top-level diplomacy API instead of embedded in this `object Select`?

hcook:
> without having to override the NodeImp.monitor method or change a LazyModule's Parameters or modify any design code.

This claim is appealing and interesting. Ideally we would actually retire the baked-in monitor injection infrastructure and use the new methodology, rather than having two ways of doing the same thing.

I wonder about applying the technique in practice. Specifically, can you demonstrate how one would replace the extant examples of `DisableMonitors` in the code base with `Select`, while still achieving the same outcome? (i.e. monitors attached to certain edges but not others). Honestly we should also probably re-evaluate which `DisableMonitors` are actually necessary after Chisel 3.4, but for the sake of this exercise, take it as a given that the existing distribution of monitors needs to be preserved.

What I am getting at is, while having to "modify any design code" to provide the `DisableMonitors` directive is a conflation of interests, it is actually quite expedient in terms of being like "turn it off for this subgraph" and I am curious what the equivalent `Select`/`collect` code would look like in practice.

src/main/scala/diplomacy/Nodes.scala
hcook: Might be good to factor that change out into a separate PR though.

src/main/scala/diplomacy/LazyModule.scala
azidar: I think the reasoning behind the `bundlesSafeNow` implicitly assumed any time you accessed the bundle, it was because you wanted to assign to it. However, the Aspect model contradicts that assumption because you access bundles to annotate it, but not assign to it. I'm not arguing for any change in particular, just adding context.

hcook: I'm open to refining things into `instantiated`(=> implies `bundlesSafeToAccessNow`) and rename the existing one to `bundlesSafeToAssignNow`, but, in the case where you are deploying an aspect that is intended to force some signals to have certain values for verification purposes, wouldn't that count as "assigning" to them?

src/main/scala/diplomacy/Nodes.scala
albertchen-sifive: yeah `Select` could be a completely separate project, but it would require a public accessor for `instantiated`.

hcook: Hmm, I agree as written it would since you put a `require` on `instantiated` directly, but thinking about it more, is that require in your user code even necessary? Put another way, if we retire `bundlesSafeNow` and replace it with `instantiated` isn't there already some internal requirement on `bundlesSafeNow` that trips if you access the bundle? I am guessing you hit that yourself, since you went in and toggled it off. So if that internal requirement gets changes to be on `instantiated`, do you select functions even need their own `require` anymore?

src/main/scala/diplomacy/LazyModule.scala
```diff
@@ -5,14 +5,104 @@ package freechips.rocketchip.diplomacy
...
+        node.bundlesSafeNow = false
+        result
+      case node: BaseNode =>
+        collect.lift(node)
+    }
+    val fn = collectOutputs(_: BaseNode)(checkedCollect)
+    lmod.nodes.flatMap(fn)
+  }
+
+  def collectDeepLazyMod[T](mod: RawModule)(collect: PartialFunction[LazyModule, T]): Iterable[T] = {
```
hcook, line 97: Should this be `chisel3.experimental.BaseModule`? I confess I don't know the difference myself but I've seen that used in @azidar's code

src/main/scala/diplomacy/Nodes.scala
albertchen-sifive: yeah you're right

src/main/scala/diplomacy/LazyModule.scala
jackkoenig: `RawModule` is super type of all user modules, `BaseModule` is super type of that and `BlackBoxes`

src/main/scala/diplomacy/LazyModule.scala
mwachs5: i thought that just reading a bundle needs to happen in a "safe" context. not assigning to it. "safe" (from my reading of the code) implied that it had a real hardware meaning at that point, for both assigning and reading.

hcook: Sorry @mwachs5 I don't quite follow what you are saying. If "just reading...needs...safe" but "safe...implies...both assigning and reading" why do we need two flags?

hcook: To rephrase, I'm no longer seeing the utility in this library being a nanny about keeping the user from making cross-module references through either reading from or assigning to this bundle. Seems like there are plenty of reasons to access the bundle and things that one might do with a reference to it outside the scope of its `LazyModuleImp`. Sure some of them are bad but I think that without knowing what the user is going to do with the reference we aren't really in a strong position to forbid it here.

hcook: In contrast keeping the user from accessing the bundle before diplomatic negotiation has resolved seems very valuable still.

docs/src/diplomacy/select_tutorial.md
```diff
@@ -0,0 +1,143 @@
...
+  case _ => true
+} {
+  case l: Leaf => l.pathName
+}
+```
+
+`LazyModule`s have a `getNodes` method that return all the nodes instantiated
+within that module. This can be combined with the `Select.collectInwardEdges`
+to select `LazyModule`s based on their connectivity.
+`Select.collectInwardEdges` takes a `BaseNode` and a parital function. It
```
azidar, line 123:
```suggestion
`Select.collectInwardEdges` takes a `BaseNode` and a partial function. It
```

docs/src/diplomacy/select_tutorial.md
```diff
@@ -0,0 +1,143 @@
...
+`Select.collectInwardEdges` to find all instances of `Leaf` modules in `top`
+that are connected to an `A` module and do not belong to the `foo:Foo`
+subhierarchy.
+```scala mdoc
+Select.filterCollectDeep (top) {
+  case _: Foo => false
+  case _ => true
+} {
+  case a: A =>
+    a.getNodes.flatMap(Select.collectInwardEdges(_)(Function.unlift { edge =>
```
azidar, line 136: Perhaps add a clarifying comment about why you used `Function.unlift`?

src/main/scala/aop/Select.scala
```diff
@@ -0,0 +1,122 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.aop
+
+import chisel3.{Data, RawModule}
+import chisel3.experimental.BaseModule
```
azidar, line 6: I think this is an unused import.

azidar (approved):
LGTM, except for a few minor changes.

albertchen-sifive:
> What I am getting at is, while having to "modify any design code" to provide the `DisableMonitors` directive is a conflation of interests, it is actually quite expedient in terms of being like "turn it off for this subgraph" and I am curious what the equivalent `Select`/`collect` code would look like in practice.

I added some mdoc examples https://github.com/chipsalliance/rocket-chip/blob/select-helpers/docs/src/diplomacy/select_tutorial.md. do you think the examples cover this use case or should there be more examples and/or select methods?

hcook:
@albertchen-sifive Nice! I love these examples, they are very cool.

I don't think they quite hit the (admittedly-high) bar I am setting, which requires targeting specific pieces of a diplomatic graph at an even finer granularity. See [this function](https://github.com/chipsalliance/rocket-chip/blob/16cf15ca5d02d2d8cd5026805a8a9de97dec71be/src/main/scala/subsystem/HasTiles.scala#L270) for an example. We only disable monitoring for the connections coming into the tile that treat tile devices as slaves; we don't deactivate monitoring for the connections going out of the tile where the tile is the master device; these pieces of the graph are both put inside the same module hierarchy. This different treatment is because the slaves might have unique names and address ranges, checking of which by a Monitor breaks tile deduplication. I will freely admit that this is a bit of an esoteric use case that we might be able to avoid entirely with better test code extraction. And maybe a solution like a (later-inlined) LazyModule wrapper scope that can serve as the target of your Select would be an appropriate solution? @azidar @jackkoenig Any thoughts?

Another example I would be interested in seeing is the equivalent of [this `Config` alteration](https://github.com/chipsalliance/rocket-chip/blob/d6bd3c61993637c3f10544c59e861fae8af29f39/src/main/scala/subsystem/Configs.scala#L364), allowing the monitor injection to be toggled off through the `Config`-layer interface. What would such a thing look like if we were deciding whether or not to deploy an application of one of your `Selects` to a rocket-chip design? Would it be applied in the `TestHarness` class? The `Generator` App?

You don't have to answer either of these questions now in order to merge in this PR, but I'd consider them the next phase of development in terms of deploying these capabilities in the wild.

albertchen-sifive:
@hcook thanks for the examples. I'll leave those for another PR.

PR #2675, opened 2020-10-22
TileLink Monitor: C-Channel ProbeAckData/ReleaseData from Trunk permission param to any Trunk/Branch/None

ingallsj:
**Related issue**: 

**Impact**: API modification

**Release Notes**
change the allowed TileLink coherence permission `param` for `ProbeAckData/ReleaseData` from `TtoB/TtoN/BtoN` to `TtoT/TtoB/TtoN`

### DIFF
--- a/src/main/scala/tilelink/Monitor.scala
+++ b/src/main/scala/tilelink/Monitor.scala
@@ -8,7 +8,7 @@ import chisel3.internal.sourceinfo.SourceLine
 import chisel3.experimental.chiselName
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
-import freechips.rocketchip.util.{PlusArg, UIntIsOneOf}
+import freechips.rocketchip.util.PlusArg
 import freechips.rocketchip.formal._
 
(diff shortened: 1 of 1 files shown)


hcook:
Technically the 1.8.1 spec allows `(TtoB, TtoN, BtoN, TtoT, BtoB, NtoN)` regardless of whether data is attached to the Probe or Release, so while I think the current monitor implementation is too picky, I think switching it to a different kind of pickiness for those particular messages probably shouldn't happen without a spec change.

ingallsj:
> Technically the 1.8.1 spec allows `(TtoB, TtoN, BtoN, TtoT, BtoB, NtoN)` regardless of whether data is attached to the Probe or Release, so while I think the current monitor implementation is too picky, I think switching it to a different kind of pickiness for those particular messages probably shouldn't happen without a spec change.
> 
> I recommend making it `TLPermissions.isReport(bundle.param) || TLPermissions.isShrink(bundle.param)` in both cases, and the data-less versions.

Will do.  That reduces to `TLPermissions.isReport(bundle.param)` for all of Probe/Release(Data) messages.

hcook (approved):

PR #2676, opened 2020-10-22
Improvements To AMBA AXI4 Adapters

hcook:
Miscellaneous improvements to several AXI4 AMBA adapters:
- Scala doc
- Clarify some variable names
- More comments
- Unify TLToAXI4 metadata code paths into a single path through `TLtoAXI4IdMap`
- Make any value of `TLToAXI4.stripBits` other than 0 illegal and stop using it internally.
- Remove usage of un-consumed `Annotated.idMapping` and delete associated application and annotation class.

**Impact**: no API changes

### DIFF
--- a/src/main/scala/amba/axi4/Deinterleaver.scala
+++ b/src/main/scala/amba/axi4/Deinterleaver.scala
@@ -13,7 +13,7 @@ import freechips.rocketchip.util.leftOR
   *
   * Deinterleaving guarantees that once the first beat of a read response
   * has been accepted by the recipient, all further presented read responses will
-  * be from the same burst transactions, until the burst is complete.
+  * be from the same burst transaction, until the burst is complete.
   *
   * @param maxReadBytes is the maximum supported read burst size that this adapter
   *   has been provisioned to support.
@@ -24,12 +24,12 @@ class AXI4Deinterleaver(maxReadBytes: Int, buffer: BufferParams = BufferParams.d
   require (maxReadBytes >= 1, s"AXI4Deinterleaver: maxReadBytes must be at least 1, not $maxReadBytes")
   require (isPow2(maxReadBytes), s"AXI4Deinterleaver: maxReadBytes must be a power of two, not $maxReadBytes")
 
-  private def beats(slave: AXI4SlavePortParameters): Int =
+  private def maxBeats(slave: AXI4SlavePortParameters): Int =
     (maxReadBytes+slave.beatBytes-1) / slave.beatBytes
 
   // Nothing to do if R channel only uses a single beat
   private def nothingToDeinterleave(slave: AXI4SlavePortParameters): Boolean =
-    beats(slave) <= 1
+    maxBeats(slave) <= 1
 
   val node = new AXI4AdapterNode(
     masterFn = { mp => mp },
@@ -43,9 +43,9 @@ class AXI4Deinterleaver(maxReadBytes: Int, buffer: BufferParams = BufferParams.d
   lazy val module = new LazyModuleImp(this) {
     (node.in zip node.out) foreach { case ((in, edgeIn), (out, edgeOut)) =>
       val endId = edgeOut.master.endId
-      val beats = beats(edgeOut.slave)
+      val beats = maxBeats(edgeOut.slave)
 
-      // This adapter leaves the control + write paths completely untouched
+      // This adapter passes through the AR/AW control + W/B write data channels
       out.ar :<> in.ar
       out.aw :<> in.aw
       out.w :<> in.w
@@ -55,15 +55,15 @@ class AXI4Deinterleaver(maxReadBytes: Int, buffer: BufferParams = BufferParams.d
       if (nothingToDeinterleave(edgeOut.slave)) {
         in.r.asInstanceOf[ReadyValidIO[AXI4BundleR]] :<> buffer.irrevocable(out.r)
       } else {
-        // We only care to do something for ids that are in use
+        // We only care to deinterleave ids that are actually in use
         val maxFlightPerId = Seq.tabulate(endId) { i =>
           edgeOut.master.masters.find(_.id.contains(i)).flatMap(_.maxFlight).getOrElse(0)
         }
 
         // Queues to buffer R responses
-        val qs = maxFlightPerId.map { mf =>
+        val qs = maxFlightPerId.zipWithIndex.map { case (mf, i) =>
           if (mf > 0) {
-            val q = Module(new Queue(out.r.bits.cloneType, beats))
+            val q = Module(new Queue(out.r.bits.cloneType, entries = beats))
             q.suggestName(s"queue_${i}")
             q.io
           } else {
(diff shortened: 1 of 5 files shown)


src/main/scala/amba/axi4/UserYanker.scala
```diff
@@ -8,6 +8,14 @@ import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.util._
 
+/** This adapter prunes all user bit fields of the echo type from request messages,
+  * storing them in queues and echoing them back when matching response messages are recevied.
+  *
+  * It also optionally rate limits the number of transactions that can be in flight simultaneously
+  * per FIFO domain / AXID.
+  *
+  * @param capMaxFlight is an optional maximum number of transactions that can be in flight per AXID.
```
mwachs5, line 17: AXI Id?

src/main/scala/amba/axi4/IdIndexer.scala
```diff
@@ -10,9 +10,17 @@ import freechips.rocketchip.util._
 case object AXI4ExtraId extends ControlKey[UInt]("extra_id")
 case class AXI4ExtraIdField(width: Int) extends SimpleBundleField(AXI4ExtraId)(UInt(OUTPUT, width = width), UInt(0))
 
+/** This adapter limits the set of FIFO domain ids used by outbound transactions.
+  *
+  * Extra AWID and ARID bits from upstream transactions are stored in a User Bits field called AXI4ExtraId,
+  * which values are expected to be echoed back to this adapter alongside any downstream response messages,
+  * and are then prepended to the RID and BID field to restore the original identifier.
+  *
+  * @param idBits is the desired number of AXID bits to be used
```
mwachs5, line 19: AXI Id?

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -9,38 +9,61 @@ import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.util.leftOR
 
+/** This adapter deinterleaves read responses on the R channel.
```
mwachs5, line 12: 🎉 TIL

src/main/scala/amba/axi4/UserYanker.scala
hcook: hmm "AXID" is their term for `A[W|R]ID` which is the actual signal field name but maybe i will change it to be the latter for added clarity

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -9,38 +9,61 @@ import freechips.rocketchip.config.Parameters
...
 
-      if (beats <= 1) {
-        // Nothing to do if only single-beat R
+      // Only the R channel has the possibility of being changed
+      if (nothingToDeinterleave(edgeOut.slave)) {
         in.r.asInstanceOf[ReadyValidIO[AXI4BundleR]] :<> buffer.irrevocable(out.r)
       } else {
+        // We only care to deinterleave ids that are actually in use
+        val maxFlightPerId = Seq.tabulate(endId) { i =>
+          edgeOut.master.masters.find(_.id.contains(i)).flatMap(_.maxFlight).getOrElse(0)
```
jackkoenig, line 60:
```suggestion
          edgeOut.master.masters.collectFirst { case x if x.id.contains(i) => x.maxFlight }.getOrElse(0)
```

jackkoenig: Longer line but `collect` is cooler than `find` + `flatMap`

src/main/scala/amba/axi4/Deinterleaver.scala
```diff
@@ -9,38 +9,61 @@ import freechips.rocketchip.config.Parameters
...
-        val qs = Seq.tabulate(endId) { i =>
-          val depth = edgeOut.master.masters.find(_.id.contains(i)).flatMap(_.maxFlight).getOrElse(0)
-          if (depth > 0) {
-            val q = Module(new Queue(out.r.bits.cloneType, beats))
+        val qs = maxFlightPerId.map { mf =>
+          if (mf > 0) {
+            val q = Module(new Queue(out.r.bits.cloneType, depth = beats))
             q.suggestName(s"queue_${i}")
             q.io
           } else {
```
jackkoenig, line 69: This and following block looks like it should just use the `Queue` factory method which is just a Wire when the size is 0. Other than the assertion that is.

src/main/scala/amba/axi4/UserYanker.scala
mwachs5: ah, thats fine

src/main/scala/amba/axi4/Deinterleaver.scala
hcook: Hmm. Do you think it is actually better to unconditionally wire the interface through while applying some assertions to it for some ids than to feed the `DontCare` tieoff into the downstream logic?

jackkoenig: I don't know, I think having the assertion is good. You could use the Queue factory method and still check the size to see about adding the assertion but YMMV.

hcook: Yeah I understand how to keep having the assertion, my question is more about QoR in the downstream circuitry. Like
is injecting "tied off" `DontCare` constants actually better or worse than just connecting all those wires to the external IO (but asserting failing if that IO is used in certain ways). The former sounds like there will be more muxes but with one side being a constant... I'm not sure I have time to evaluate this at the moment so am inclined to leave it alone for the moment

hcook:
@jackkoenig @mwachs5 can one of you approve this?

src/main/scala/amba/axi4/Deinterleaver.scala
jackkoenig: Yeah I think it's fine as is.

src/main/scala/tilelink/ToAXI4.scala
```diff
@@ -23,16 +23,32 @@ case class AXI4TLStateField(sourceBits: Int) extends BundleField(AXI4TLState) {
...
+/** TLtoAXI4IdMap serves as a record for the translation performed between id spaces.
+  *
+  * Its member [axi4Masters] is used as the new AXI4MasterParameters in diplomacy.
+  * Its member [mapping] is used as the template for the circuit generated in TLToAXI4Node.module.
+  */
+class TLtoAXI4IdMap(tlPort: TLMasterPortParameters) extends IdMap[TLToAXI4IdMapEntry]
 {
-  private val axiDigits = String.valueOf(axi4.endId-1).length()
-  private val tlDigits = String.valueOf(tl.endSourceId-1).length()
+  val tlMasters = tlPort.masters.sortBy(_.sourceId).sortWith(TLToAXI4.sortByType)
```
jackkoenig, line 33: I was a little confused at first why there are two sorts in a row. It may be worth adding another `TLToAXI4.sortBy<something>` method to provide the desired behavior. Perhaps `sortByType` should break ties with `sourceId`? Just a thought.

jackkoenig (approved):
LGTM

PR #2680, opened 2020-10-23
loosen Node bundle accessor requirements

albertchen-sifive:
**Related issue**:  https://github.com/chipsalliance/rocket-chip/pull/2674
This PR replaces the `bundlesSafeNow` guard with a looser `instantiated` guard that is set to true when `instantiate` is called and is never reset to false afterwards. This allows the `in`/`out` bundles to be accessed by things like Aspects after the `LazyModuleImp` has been instantiated.

protected function `finishInstantiate` is removed from `BaseNode` and `LazyModule` as it is no longer necessary.

**Release Notes**

### DIFF
--- a/src/main/scala/diplomacy/Nodes.scala
+++ b/src/main/scala/diplomacy/Nodes.scala
@@ -1235,14 +1235,14 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       data   = bundleIn(i))
   }
 
-  private var bundlesSafeNow = false
+  private[diplomacy] var instantiated = false
 
   /** Gather Bundle and edge parameters of outward ports.
     *
     * Accessors to the result of negotiation to be used within [[LazyModuleImp]] Code.
     */
   def out: Seq[(BO, EO)] = {
-    require(bundlesSafeNow, s"$name.out should only be called from the context of ${scope.get.name}'s LazyModuleImp, but the current scope is ${LazyModule.scope}.")
+    require(instantiated, s"$name.out should only be called after its parent LazyModule.module has been instantiated")
     bundleOut zip edgesOut
   }
 
@@ -1251,7 +1251,7 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
     * Accessors to the result of negotiation to be used within [[LazyModuleImp]] Code.
     */
   def in: Seq[(BI, EI)] = {
-    require(bundlesSafeNow, s"$name.in should only be called from the context of ${scope.get.name}'s LazyModuleImp, but the current scope is ${LazyModule.scope}.")
+    require(instantiated, s"$name.in should only be called after its parent LazyModule.module has been instantiated")
     bundleIn zip edgesIn
   }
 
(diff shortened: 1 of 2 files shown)


src/main/scala/diplomacy/Nodes.scala
```diff
@@ -1273,7 +1273,6 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
     * wires as the [[LazyModuleImp]] has been completely evaluated to create a [[Module]].
     */
   protected[diplomacy] def finishInstantiate(): Unit = {
```
hcook, line 1275: I think we should propagate a removal of the `finishInstantiate` functions here and in `LazyModule` since they have incorrect scala doc and are now useless. If something else ends up needing the functionality we can always restore it. The `finishInstantiate` in `LazyModule` does apply an annotation, but I think there is no reason that application couldn't live in `instantiate` instead of `finishInstantiate`, and the rest can be deleted.

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -1235,14 +1235,14 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
-  private var bundlesSafeNow = false
+  private[diplomacy] var instantiated = false
 
   /** Gather Bundle and edge parameters of outward ports.
     *
     * Accessors to the result of negotiation to be used within [[LazyModuleImp]] Code.
     */
   def out: Seq[(BO, EO)] = {
-    require(bundlesSafeNow, s"$name.out should only be called from the context of ${scope.get.name}'s LazyModuleImp, but the current scope is ${LazyModule.scope}.")
+    require(instantiated, s"$name.out should only be called after its parent LazyModule.module has been instantiated")
```
hcook, line 1245: Feel like text this isn't quite correct, since you can still use the thing during the execution of the `LazyModuleImp`, although I guess you would never trip this requirement at that time. Maybe `s"$name.out should not be called until after instantiation of its parent LazyModule.module has begun"`?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -1235,14 +1235,14 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
       data   = bundleIn(i))
   }
 
-  private var bundlesSafeNow = false
+  private[diplomacy] var instantiated = false
 
   /** Gather Bundle and edge parameters of outward ports.
     *
     * Accessors to the result of negotiation to be used within [[LazyModuleImp]] Code.
```
hcook, line 1242: Maybe update this to `should only be used within [[LazyModuleImp]] code or after its instantiation has completed.`?

src/main/scala/diplomacy/Nodes.scala
```diff
@@ -1251,7 +1251,7 @@ sealed abstract class MixedNode[DI, UI, EI, BI <: Data, DO, UO, EO, BO <: Data](
...
     */
   def in: Seq[(BI, EI)] = {
-    require(bundlesSafeNow, s"$name.in should only be called from the context of ${scope.get.name}'s LazyModuleImp, but the current scope is ${LazyModule.scope}.")
+    require(instantiated, s"$name.in should only be called after its parent LazyModule.module has been instantiated")
```
hcook, line 1254: Same as above

hcook (approved):
Thank you!

PR #2684, opened 2020-10-26
BuiltInDevices: Optional instantiation of TL adapters

rmac-sifive:
**Related issue**: 

**Impact**: API modification

**Release Notes**

Makes optional the instantiation of previously mandatory adapters for these devices, and also adds a CacheCork option for the Zero device

### DIFF
--- a/src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
+++ b/src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
@@ -6,9 +6,18 @@ import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.tilelink._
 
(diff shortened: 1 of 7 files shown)


hcook (approved):
I realllly want to refactor this such that each `TLBusWrapper` has a `Seq[BuiltInDeviceParams]`, each of which has an `attach` method that contains the logic currently smooshed together into `BuiltInDevices.attach` and then that method just becomes /is replaced by a loop over calling those methods. But I guess we should think more about upstreaming `HasDevices` before attempting that?

rmac-sifive:
Meaning we'd get rid of special-casing of the zero/error devices and they would be instantiated by `CanHaveDevices`, which is mixed into `TLBusWrapper`? Mixed into `TLBusWrapper` or not, that does sound cleaner to me than the current instantiation API

PR #2689, opened 2020-10-29
[prci] Fix ResetSynchronizers to use AdapterNodes instead of IdentityNodes

jerryz123:
Identity nodes perform an initial `out <> in` assignment, which causes the outwards resets to receive assignments to both `AsyncReset` and `Bool` types.

**Type of change**: bug fix

**Impact**:  API modification


hcook (approved):

davidbiancolin:
@jackkoenig I'm curious why reset inference can't play nice with last connect semantics. Is that discussed in the FIRRTL PR?

davidbiancolin:
Merged on Jerry's behalf.

PR #2693, opened 2020-10-30
CSR mcountinhibit

ingallsj:
**Related issue**: 

**Release Notes**
add CSR `mcountinhibit` as was defined in RISC-V Privileged ISA version 1.11

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -419,11 +419,14 @@ class CSRFile(
   val reg_vxsat = usingVector.option(Reg(Bool()))
   val reg_vxrm = usingVector.option(Reg(UInt(io.vector.get.vxrm.getWidth.W)))
 
-  val reg_instret = WideCounter(64, io.retire)
-  val reg_cycle = if (enableCommitLog) reg_instret else withClock(io.ungated_clock) { WideCounter(64, !io.csr_stall) }
+  val reg_mcountinhibit = RegInit(0.U((2 + nPerfCounters).W))
+  val reg_instret = WideCounter(64, io.retire, inhibit = reg_mcountinhibit(2))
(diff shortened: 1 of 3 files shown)


ingallsj:
Questions / discussion points:

- I instantiated `reg_mcountinhibit` as a `RegInit` with reset to zero, to maintain compatibility with software that might not initialize this CSR.  Could we save some fanout on the reset by instantiating `reg_mcountinhibit` as a plain `Reg`?

- I instantiated `reg_mcountinhibit` unconditionally.  Should I make it conditional on a parameter, such as 

- I changed `reg_cycle` to be inhibited by `reg_mcountinhibit(0)` for both options of parameter `if (enableCommitLog)`.  Should I have kept that as `if (enableCommitLog) reg_instret`?

src/main/scala/rocket/CSR.scala
```diff
@@ -419,11 +419,14 @@ class CSRFile(
   val reg_vxsat = usingVector.option(Reg(Bool()))
   val reg_vxrm = usingVector.option(Reg(UInt(io.vector.get.vxrm.getWidth.W)))
 
-  val reg_instret = WideCounter(64, io.retire)
-  val reg_cycle = if (enableCommitLog) reg_instret else withClock(io.ungated_clock) { WideCounter(64, !io.csr_stall) }
+  val reg_mcountinhibit = RegInit(0.U((3 + nPerfCounters).W))
```
aswaterman, line 422: Minor nit, use `CSRs.firstHPM` instead of the constant 3 so it's clear what the 3 represents.

aswaterman:
My sense is that in this case we should cope with the slightly higher reset fanout to avoid the PITA for software.

I think instantiating it unconditionally is OK, since it's just 1 flop per relatively expensive counter.

I'm not certain what the right thing to do for enableCommitLog is, but my sense is that what you did is best.  Certainly the HW cost is not a concern, since it's not meant to be used in production.

aswaterman (approved):
LGTM minor the one nit.

PR #2702, opened 2020-11-03
Cutpoint annotations

debs-sifive:
**Related issue**: N/A

**Release Notes**

Adds the `Cutpoint` object, which can be used to mark signals as cutpoints. These signals will be written to a `cutpoints.<label>.txt` file.

### DIFF
--- a/src/main/scala/util/Cutpoint.scala
+++ b/src/main/scala/util/Cutpoint.scala
@@ -1,0 +1,49 @@
+// See LICENSE.Berkeley for license details.
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import chisel3._
+import firrtl.annotations.{Annotation, ReferenceTarget, SingleTargetAnnotation}
+import firrtl.{CircuitState, LowForm, Transform}
+import chisel3.experimental.{ChiselAnnotation, annotate}
+import firrtl.stage.RunFirrtlTransformAnnotation
+
+/** Marks a signal with an annotation so it can be parsed by post-processors as a cutpoint */
+object Cutpoint {
+  def cutpoint (signal: Bits): Unit = {
+    // create annotation for a cutpoint
+    annotate(new ChiselAnnotation {
+      override def toFirrtl: Annotation = CutpointAnnotation(signal.toTarget)
+    })
+    // run transform that prints out cutpoint annotation information
+    annotate(new ChiselAnnotation {
+      override def toFirrtl: Annotation = RunFirrtlTransformAnnotation(new PrintCutpointsTransform)
+    })
+  }
+}
+
+/** Annotation containing information about a cutpointed wire */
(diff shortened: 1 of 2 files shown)


src/main/scala/util/Cutpoint.scala
```diff
@@ -0,0 +1,93 @@
...
+import chisel3.experimental.{ChiselAnnotation, annotate}
+import firrtl.annotations.TargetToken.{Instance, Ref}
+import firrtl.stage.RunFirrtlTransformAnnotation
+import firrtl.options.CustomFileEmission
+
+/** Marks a signal with a [[CutpointAnnotation]]. Optionally, a label can be
+  * passed in.
+  *
+  * The Verilog paths to each cutpoint will be emitted in a metadata file,
+  * called cutpoints.<label>.txt.
```
mwachs5, line 17: should there also be an output file which is a list of all the cutpoints.<label>.txt files...?

debs-sifive: do you think that's something that should be added in right now?

I'm open to adding it in later if it's needed, but I think it's unnecessary for our current purposes.

debs-sifive:
I'm thinking that maybe I should generalize this to not be cutpoint-specific. Make it `MarkBits` or something...

```
MarkBits.mark(signal, <required file path>, <optional extra info>)
```

mwachs5:
would "optional extrea info" just be a string?

debs-sifive:
Yeah, I was thinking a string or an array of strings. This generalization would also allow for marking constraints, like:

```
MarkBits(signal, "constraints.txt", "!= 5'b0001")
```

Which would dump out something like
`constraints.txt`:
```
dut.submodule.submodule.signal : != 5’b00001
```

which can be transformed via some post-processing constraints-specific script into `constraint dut.submodule.submodule.signal != 5’b00001`

**Alternatively:** we could do this
```
//           Bits,   filename,          optional prefix, optional suffix
MarkBitPath(signal, "constraints.txt", "constraint", "!= 5'b0001")
```
And then we could directly dump `constraint dut.submodule.submodule.signal != 5’b00001`

But I'm not sure if that's getting too specific with our use case/overcomplicating things...

mwachs5:
i think there are a few parts to this... the user API ("i want a cutpoint") and how it's actually implemented. I think if you want to make a lower level API that says "add random data to arbitrary wire and put it in a file" (does that not already exist??) and use that with the user-facing "cutpoint" API that makes sense. I don't think users should be hard-coding "cutpoint.txt" as the way of indicating they want a cutpoint.

debs-sifive:
> make a lower level API that says "add random data to arbitrary wire and put it in a file" (does that not already exist??) and use that with the user-facing "cutpoint" API that makes sense

I like this idea. I think the "add random data to arbitrary wire and put it in a file" part doesn't quite exist yet (hence this PR), but that is probably what I should generalize this PR to include. I will discuss with @azidar about how to structure it.

debs-sifive:
Closing this PR to do the API thing Megan suggested

PR #2706, opened 2020-11-06
Documentation to TLBroadcast

sequencer:
**Related issue**: None

**Type of change**: Documentation

### DIFF
--- a/src/main/scala/tilelink/Broadcast.scala
+++ b/src/main/scala/tilelink/Broadcast.scala
@@ -17,6 +17,12 @@ case class TLBroadcastControlParams(
   address:   AddressSet,
   beatBytes: Int)
 
+/**
+  * @param lineBytes cacheline size
+  * @param numTrackers maximum transaction size to record on each port.
+  * @param bufferless if this is true. ready signal in A channel will directly decoupled from manger to client.
+  * @param control Add a control node for filter notification.
+  */
 case class TLBroadcastParams(
   lineBytes:     Int,
   numTrackers:   Int = 4,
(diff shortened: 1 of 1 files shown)


src/main/scala/tilelink/Broadcast.scala
```diff
@@ -28,6 +34,94 @@ case class TLBroadcastParams(
   require (numTrackers > 0)
 }
 
+/** [[TLBroadcast]] is the last level coherence manger for TileLink. Simplest topology can be this:
+  * {{{
+  *   A           B
+  *   │           │
+  *   └─────┬─────┘
```
hcook, line 41: This is also an Xbar, otherwise A cannot probe B.

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -74,72 +173,143 @@ class TLBroadcast(params: TLBroadcastParams)(implicit p: Parameters) extends Laz
...
+  /** Control Node for filter.
+    * [[ProbeFilter.useRegFields]] can be defined to add register to control.
+    */
   val controlNode = params.control.map(x => TLRegisterNode(
     address   = Seq(x.address),
     beatBytes = x.beatBytes,
     device    = new SimpleDevice("cache-controller", Seq("sifive,broadcast0"))))
 
   lazy val module = new LazyModuleImp(this) {
+    /* Number of inward node will decide the number bank. */
```
hcook, line 187: This statement is true only for the generic topology that we use by default. You could design a network using `_ :*= bh :*= _` instead of `_ :=* bh :=* _` where the  outward side decides how many banks there are.

src/main/scala/tilelink/Broadcast.scala
hcook: I think you need to be really very clear in this diagram about whether A and B are different banks or different masters.

sequencer: Sorry missed xbar in this diagram, I'll fix it tomorrow.

src/main/scala/tilelink/Broadcast.scala
sequencer: Got it. I'll make it more clear.

hcook: Put another way, this node is just an adapter node. The number of banks is the number of edges going through it, and the number of inward edges must equal the number of outwards edges. Either direction could be made to pick how many edges there are.

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -74,72 +173,143 @@ class TLBroadcast(params: TLBroadcastParams)(implicit p: Parameters) extends Laz
...
-      // For the probe walker, we need to identify all the caches
-      val caches = clients.filter(_.supports.probe).map(_.sourceId)
-      val cache_targets = caches.map(c => c.start.U)
-
-      // Create the probe filter
-      val flatAddresses = AddressSet.unify(edgeOut.manager.managers.flatMap(_.address))
-      val addressMask = AddressDecoder(flatAddresses.map(Seq(_)), flatAddresses.map(_.mask).reduce(_|_))
+      /** This identify all the caches for the probe walker.  */
+      val caches: Seq[IdRange] = clients.filter(_.supports.probe).map(_.sourceId)
+      /** Use start bit in IdRange to locate which client is target. */
```
hcook, line 201: I think it is the lowest id value in the IdRange, which isn't the same as a bit index within a UInt

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -17,6 +17,12 @@ case class TLBroadcastControlParams(
   address:   AddressSet,
   beatBytes: Int)
 
+/**
+  * @param lineBytes cacheline size
+  * @param numTrackers maximum transaction size to record on each port.
```
hcook, line 22: i would say "the number of outstanding transactions per bank". "maximum transaction size" sounds to me like "maximum number of bytes per transaction"

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -28,6 +34,94 @@ case class TLBroadcastParams(
...
+  *          if A is Tip, B acquire this cacheline:
+  *          - (t0) B -> BH: AcquireBlock
+  *          - (t1) BH -> A: ProbeBlock
+  *          - (t1) A -> BH: ProbeAckData/ProbeAck
+  *          -   (t2) if ProbeAckData: BH -> C: PutFullData
+  *          -   (t2) if ProbeAckData: C -> BH: AccessAck
+  *          - (t3) BH -> C: Get
+  *          - (t3) C -> BH: AccessAckData
+  *          - (t0) BH -> B: GrantData
+  *          - (t0) B -> BH: GrantAck
```
hcook, line 67: Maybe use "M1" and "M2"  instead of "A" and "B"? Since A and B are also TL priority channel names I think it is a little confusing.

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -38,33 +132,38 @@ class TLBroadcast(params: TLBroadcastParams)(implicit p: Parameters) extends Laz
...
         name     = "TLBroadcast",
+        /* Add 2 bits encoded in [[TLBroadcastConstants]] to record transaction type for each sourceId in clients. */
         sourceId = IdRange(0, 1 << log2Ceil(cp.endSourceId*4)))))
     },
     managerFn = { mp =>
       mp.v1copy(
         endSinkId  = params.numTrackers,
         managers   = mp.managers.map { m =>
-          // We are the last level manager
+          /** This is the last level coherence manager, so it cannot be probed by mangers. */
```
hcook, line 142: "outward managers"

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -74,72 +173,143 @@ class TLBroadcast(params: TLBroadcastParams)(implicit p: Parameters) extends Laz
...
+        * 1. tracker need d reply
+        * 2. tracker source is same with the client source.
+        *
+        * This is used for:
+        * 1. encode tracker id into sink of D, make client reply in E to this tracker.
+        * 2. filter will need this D.
+        */
       val d_trackerOH = VecInit(trackers.map { t => t.need_d && t.source === d_normal.bits.source }).asUInt holdUnless d_first
 
+      /** if d is valid, and not drop. */
```
hcook, line 261: [Good comments don't repeat the code or explain it. They clarify its intent.](https://en.wikipedia.org/wiki/Comment_(computer_programming)#Code_description)  The next line of code literally says `!out.d.valid || !d_drop` so I don't think this comment adds any clarity.

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -38,33 +132,38 @@ class TLBroadcast(params: TLBroadcastParams)(implicit p: Parameters) extends Laz
...
         managers   = mp.managers.map { m =>
-          // We are the last level manager
+          /** This is the last level coherence manager, so it cannot be probed by mangers. */
           require (!m.supportsAcquireB)
-          // We only manage addresses which are uncached
+          /* We only manage addresses which are uncached. */
           if (m.regionType == RegionType.UNCACHED) {
-            // The device had better support line transfers
+            /* The device had better support line transfers.
+             * @todo why?
```
hcook, line 147: Because this device is going to turn `ReleaseData`s of `lineSize` into `PutFull`s of `lineSize` and `AcquireBlocks` of `lineSize` into `Gets` of `lineSize`. So the outwards managers must visibly support such operations. This is basically checking the legality of the C->UH conversion.

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -38,33 +132,38 @@ class TLBroadcast(params: TLBroadcastParams)(implicit p: Parameters) extends Laz
...
+              /* truncate supported accesses to lineBytes (we only ever probe for one line). */
               supportsPutFull    = TransferSizes(m.supportsPutFull   .min, min(m.supportsPutFull   .max, params.lineBytes)),
               supportsPutPartial = TransferSizes(m.supportsPutPartial.min, min(m.supportsPutPartial.max, params.lineBytes)),
               supportsGet        = TransferSizes(m.supportsGet       .min, min(m.supportsGet       .max, params.lineBytes)),
               supportsHint       = TransferSizes(m.supportsHint      .min, min(m.supportsHint      .max, params.lineBytes)),
               supportsArithmetic = TransferSizes(m.supportsArithmetic.min, min(m.supportsArithmetic.max, params.lineBytes)),
               supportsLogical    = TransferSizes(m.supportsLogical   .min, min(m.supportsLogical   .max, params.lineBytes)),
-              fifoId             = None // trackers do not respond in FIFO order!
+              /* trackers do not respond in FIFO order! */
+              fifoId             = None
```
hcook, line 166: I am not sure it is necessarily better to change these single line comments to be separate line with `/* */`? They won't show up as scala-doc anyways, right?

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -74,72 +173,143 @@ class TLBroadcast(params: TLBroadcastParams)(implicit p: Parameters) extends Laz
...
+       * client will reply it in the sink.
+       * this decode sink from E channel, then notify correspond tracker.
+       */
       (trackers zip UIntToOH(in.e.bits.sink).asBools) foreach { case (tracker, select) =>
         tracker.e_last := select && in.e.fire()
       }
 
-      // Depending on the high source bits, we might transform D
+      /* Client D Channel:
+       * Forward or transform manger D channel based on the payload of the source.
```
hcook, line 233: manger -> manager

src/main/scala/tilelink/Broadcast.scala
mwachs5: nit: can you use the `//` style comments for those that are not explicitly scaladoc for something? That is the style adopted in the code base

src/main/scala/tilelink/Broadcast.scala
hcook: you should find/replace this change, i see more

src/main/scala/tilelink/Broadcast.scala
mwachs5: i would specifically ask that we NOT make this change

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -149,12 +319,7 @@ class TLBroadcast(params: TLBroadcastParams)(implicit p: Parameters) extends Laz
       filter.io.update.bits.gaveT := d_what === TRANSFORM_T
       filter.io.update.bits.cacheOH := Mux1H(d_trackerOH, trackers.map(_.cacheOH))
 
-      // Incoming C can be:
-      // ProbeAck     => decrement tracker, drop 
-      // ProbeAckData => decrement tracker, send out A as PutFull(DROP)
-      // ReleaseData  =>                    send out A as PutFull(TRANSFORM)
-      // Release      => send out D as ReleaseAck
```
hcook, line 156: Did this information get lost? Or is it replicated elsewhere now?

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -391,45 +559,81 @@ object TLBroadcast
 }
 
 class ProbeTrackInfo(val caches: Int) extends Bundle {
+  /** how many clients left need to be probed. */
   val count   = UInt(log2Ceil(caches+1).W)
+  /** probe which client now. */
   val cacheOH = UInt(caches.W)
 }
 
+/** Tracker to track trans */
```
hcook, line 568: I would say something here about how some people might call this a Miss Status Handling Register (MSHR)

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -391,45 +559,81 @@ object TLBroadcast
...
 class TLBroadcastTracker(id: Int, lineBytes: Int, caches: Int, bufferless: Boolean, edgeIn: TLEdgeIn, edgeOut: TLEdgeOut) extends Module
 {
   val io = IO(new Bundle {
-    val in_a_first = Input(Bool())
-    val in_a  = Flipped(Decoupled(new TLBundleA(edgeIn.bundle)))
+    /** first packet come from a. */
+    val in_a_first: Bool = Input(Bool())
+    /** incoming A channel from clients. */
+    val in_a: DecoupledIO[TLBundleA] = Flipped(Decoupled(new TLBundleA(edgeIn.bundle)))
+    /** output A channel to mangers. */
```
hcook, line 576: I'd say "inwards" and "outwards" here to match what we wrote for the diplomacy node docs

src/main/scala/tilelink/Broadcast.scala
```diff
@@ -446,10 +651,13 @@ class TLBroadcastTracker(id: Int, lineBytes: Int, caches: Int, bufferless: Boole
     user   :<= io.in_a.bits.user
     echo   :<= io.in_a.bits.echo
     address := io.in_a.bits.address
+    /* we need to fetch from lower level at least. */
     count   := 1.U
   }
-
+  /* cache [[cacheOH]] by [[io.clearOH]] */
```
hcook, line 657: Again I think this repeats the code without clarifying its intent. (I think its intent is to clear some of the pending bits identifying which caches still need to be probed).

src/main/scala/tilelink/Broadcast.scala
sequencer: In has been moved to Top of this file under
`Behavior of [[TLBroadcast]]:`
Actually original comments was wrong or not accurate, since `ProbeAckData` won't decrement tracker(MSHR)
Real behavior should be:

1. ClientA -Acquire-> BH
1. BH -Probe-> ClientB 
1. ClientB -ProbeAckData-> BH
1. BH -Put-> Manger
1. Manger -AccessAckData-> BH

decrement by Acquire should happen in 5. rather than 3.

src/main/scala/tilelink/Broadcast.scala
sequencer: Yes, I finally understand it after documenting to MSHR of InclusiveCache, I found they have a similar behavior.

src/main/scala/tilelink/Broadcast.scala
sequencer: @mwachs5 Sure I’ll change it back:)
I’ll change all my modifications with `/** xx */` for scala doc and `//` for comments.

src/main/scala/tilelink/Broadcast.scala
sequencer: changed to the number of outstanding transactions per bank, aka MSHR sizes.

PR #2711, opened 2020-11-07
NMI implementation

ernie-sifive:
**Related issue**: 

**Release Notes**

- New NMI inputs covering both resumable and unresumable NMI
- NMI wired through and integrated into the CSR module's interrupt handling infrastructure
- New CSRs for NMI created.
- New mnret instruction created.
- Object model items now reflects presence of NMI
- Bus Error Unit local interrupt is now wired to cause RNMI and populate mncause with a unique exception code.

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -133,7 +133,7 @@ object CSR
   }
 
   val ADDRSZ = 12
-  def busErrorIntCause = 128  //TODO: Remove once other clients no longer need this
+  def busErrorIntCause = 128
   def debugIntCause = 14 // keep in sync with MIP.debug
   def debugTriggerCause = {
     val res = debugIntCause
@@ -345,13 +345,14 @@ class CSRFile(
     sup.debug := false
     sup.zero2 := false
     sup.lip foreach { _ := true }
+    val supported_high_interrupts = if (io.interrupts.buserror.nonEmpty && !usingNMI) UInt(BigInt(1) << CSR.busErrorIntCause) else 0.U
 
     val del = Wire(init=sup)
     del.msip := false
     del.mtip := false
     del.meip := false
 
-    (sup.asUInt, del.asUInt)
+    (sup.asUInt | supported_high_interrupts, del.asUInt)
   }
   val delegable_exceptions = UInt(Seq(
     Causes.misaligned_fetch,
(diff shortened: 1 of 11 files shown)


src/main/scala/subsystem/HasTiles.scala
```diff
@@ -124,6 +124,18 @@ trait HasTileInterruptSources
   }
 }
 
+trait HasTileNMISources extends InstantiatesTiles { this: BaseSubsystem =>
```
mwachs5, line 127:
```suggestion
/** A trait for providing a Non-Maskable Interrupt (NMI) input bundle to each tile
  *   
  */
trait HasTileNMISources extends InstantiatesTiles { this: BaseSubsystem =>
```

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -320,6 +334,12 @@ trait CanAttachTile {
     }
   }
 
+  /** Connect NMI inputs to the tile. No clock crossing needed. */
```
mwachs5, line 337: why is no crossing needed? "Synchronization must be done prior to this connection"? or...

src/main/scala/subsystem/HasTiles.scala
ernie-sifive: done

src/main/scala/subsystem/HasTiles.scala
ernie-sifive: done

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -219,6 +233,7 @@ trait HasTileNotificationSinks { this: LazyModule =>
 trait DefaultTileContextType
   extends Attachable
   with HasTileInterruptSources
+  with HasTileNMISources
```
hcook, line 236: I'm not sure that `HasTileNMISources`/`connectNMI` needed to be separate from `HasTileInterruptSources`/`connectInterrupts` respectively, which I say because a) they are notionally interrupts even if they are unmaskable b) `val nmi` is part of the `TileInterrupts` bundle, not it's own bundle. But I also don't have a strong reason to say there is a problem with doing it this way either.

hcook (approved):

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -122,12 +123,7 @@ trait HasTileInterruptSources
...
-}
-
-/** Source of Non-maskable Interrupt (NMI) input bundle to each tile.
-  */
-trait HasTileNMISources extends InstantiatesTiles { this: BaseSubsystem =>
-  /** NMI input bundle */
+  /** Source of Non-maskable Interrupt (NMI) input bundle to each tile. */
   val tileNMINode = BundleBridgeEphemeralNode[NMI]()
   val tileNMIIONodes: Seq[BundleBridgeSource[NMI]] = {
     Seq.fill(tiles.size) {
```
hcook, line 129: I think we need to make some kind of `NexusSourceNode` that accepts any number of left-hand-side connections so that you don't need this reference to `tiles.size` here. The `miepNode` above is playing a similar trick. Anyways, fine as-is and I will look at handling all three of these similarly in a future PR.

src/main/scala/rocket/CSR.scala
```diff
@@ -719,7 +760,16 @@ class CSRFile(
...
+    }.elsewhen (trapToNmiInt) {
+      when (reg_rnmie || reg_unmie && causeIsUnmiInt) {
+        reg_rnmie := false.B
+        reg_unmie := !causeIsUnmiInt
+        reg_mnepc := epc
+        reg_mncause := (BigInt(1) << (xLen-1)).U | Mux(causeIsUnmiInt, 1.U, Mux(causeIsRnmiBEU, 3.U, 2.U))
+        reg_mnstatus.mpp := trimPrivilege(reg_mstatus.prv)
+        new_prv := PRV.M
+      }
+    }.elsewhen (delegate && nmie) {
```
solomatnikov, line 772: Why is `nmie` here?

src/main/scala/rocket/CSR.scala
```diff
@@ -395,6 +398,16 @@ class CSRFile(
...
 
+  val reset_mnstatus = Wire(init=new MStatus().fromBits(0))
+  reset_mnstatus.mpp := PRV.M
+  val reg_mnscratch = Reg(Bits(width = xLen))
+  val reg_mnepc = Reg(UInt(width = vaddrBitsExtended))
+  val reg_mncause = RegInit(0.U(xLen.W))
+  val reg_mnstatus = Reg(init=reset_mnstatus)
+  val reg_rnmie = RegInit(true.B)
+  val reg_unmie = RegInit(true.B)
+  val nmie = WireInit(reg_rnmie && reg_unmie)
```
solomatnikov, line 409: I don't think you need `WireInit` here

src/main/scala/rocket/CSR.scala
ernie-sifive: The draft RNMI spec says that exceptions that occur while an NMI handler is active are handled in M-mode and affect mepc, mcause, etc.  So they are never delegated.  The next case of this when/else is taken in that case.

src/main/scala/rocket/CSR.scala
ernie-sifive: Ok

aswaterman (approved):
Thanks, @ernie-sifive - I reviewed only the RocketCore/CSRFile changes, but they LGTM.

PR #2718, opened 2020-11-09
PLIC Priority RegFieldDesc fix off-by-one

mwachs5:
**Related issue**: None

**Impact**:  API modification

**Release Notes**

PLIC Priority registers are now explicitly named with a suffix. This prevents tooling which assumes that the "first" register in a group should have the 0th index ending up with Priority[2] ending up in a register called priority_1, and so on.

### DIFF
--- a/src/main/scala/devices/tilelink/Plic.scala
+++ b/src/main/scala/devices/tilelink/Plic.scala
@@ -187,12 +187,17 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
       harts(hart)   := ShiftRegister(Reg(next = fanin.io.max) > threshold(hart), params.intStages)
     }
 
+    // Priority registers are 32-bit aligned so treat each as its own group.
+    // Otherwise, the off-by-one nature of the priority registers gets confusing.
+    require(PLICConsts.priorityBytes == 4,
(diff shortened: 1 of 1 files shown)


src/main/scala/devices/tilelink/Plic.scala
```diff
@@ -187,12 +187,17 @@ class TLPLIC(params: PLICParams, beatBytes: Int)(implicit p: Parameters) extends
...
+      s"PLIC Priority register descriptions assume 32-bits per priority, not ${PLICConsts.priorityBytes}")
+
     def priorityRegDesc(i: Int) =
       RegFieldDesc(
         name      = s"priority_$i",
         desc      = s"Acting priority of interrupt source $i",
-        group     = Some("priority"),
-        groupDesc = Some("Acting priorities of each interrupt source."),
+        group     = Some(s"priority_${i}"),
+        groupDesc = Some("Acting priority of interrupt source $i")
```
aswaterman, line 200: Need a `s` before the opening `"` here?

aswaterman (approved):
lgtm, modulo that one comment

src/main/scala/devices/tilelink/Plic.scala
mwachs5, line 200:
```suggestion
        groupDesc = Some(s"Acting priority of interrupt source $i")
```

PR #2725, opened 2020-11-10
Create TargetDirKey to expose Stage's --target-dir to Config

timsnyder-siv:
`TargetDirKey` is a `Field[String]` that gets populated with the `TargetDirectoryAnnotation`
during `PreElaboration`.

Enables other `Config` `Field`s to refer to the Stage's TargetDir.  Specifically, the current [chipyard.config.WithBootROM](https://github.com/ucb-bar/chipyard/blob/0685812c342c31d735472ad13573ce8f21ed3687/generators/chipyard/src/main/scala/ConfigFragments.scala#L36-L38) depends on finding a file relative to the current-working directory of the `Stage`.  Adding `TargetDirKey` will allow `chipyard.config.WithBootROM` to look for contents relative to the TargetDir.  This will be helpful as we develop distributed elaboraton flows for Firesim in firesim#651

**Related issue**: 

**Release Notes**

Create a `TargetDirKey` that `PreElaboration` populates with value of `TargetDirAnnotation`

### DIFF
--- a/src/main/scala/stage/phases/PreElaboration.scala
+++ b/src/main/scala/stage/phases/PreElaboration.scala
@@ -14,10 +14,6 @@ import freechips.rocketchip.util.HasRocketChipStageUtils
 
 case object TargetDirKey extends Field[String](".")
 
-case class WithTargetDir(dir: String) extends Config((site, here, up) => {
-  case TargetDirKey => dir
-})
-
 /** Constructs a generator function that returns a top module with given config parameters */
 class PreElaboration extends Phase with PreservesAll[Phase] with HasRocketChipStageUtils {
 
@@ -30,7 +26,9 @@ class PreElaboration extends Phase with PreservesAll[Phase] with HasRocketChipSt
     val rOpts = view[RocketChipOptions](annotations)
     val topMod = rOpts.topModule.get
 
-    val config = getConfig(rOpts.configNames.get) ++ WithTargetDir(stageOpts.targetDir)
+    val config = getConfig(rOpts.configNames.get).alterPartial {
+	case TargetDirKey => stageOpts.targetDir
+    }
 
     val gen = () =>
       topMod
--- a/src/main/scala/stage/phases/PreElaboration.scala
+++ b/src/main/scala/stage/phases/PreElaboration.scala
@@ -7,7 +7,7 @@ import chisel3.stage.ChiselGeneratorAnnotation
 import firrtl.AnnotationSeq
 import firrtl.options.Viewer.view
 import firrtl.options.{Dependency, Phase, PreservesAll, StageOptions}
-import freechips.rocketchip.config.{Config, Field, Parameters}
+import freechips.rocketchip.config.{Field, Parameters}
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.stage.RocketChipOptions
 import freechips.rocketchip.util.HasRocketChipStageUtils
@@ -27,7 +27,7 @@ class PreElaboration extends Phase with PreservesAll[Phase] with HasRocketChipSt
     val topMod = rOpts.topModule.get
 
     val config = getConfig(rOpts.configNames.get).alterPartial {
-	case TargetDirKey => stageOpts.targetDir
+      case TargetDirKey => stageOpts.targetDir
     }
 
     val gen = () =>
--- a/src/main/scala/stage/phases/PreElaboration.scala
+++ b/src/main/scala/stage/phases/PreElaboration.scala
@@ -6,11 +6,17 @@ import chisel3.RawModule
 import chisel3.stage.ChiselGeneratorAnnotation
 import firrtl.AnnotationSeq
 import firrtl.options.Viewer.view
-import firrtl.options.{Dependency, Phase, PreservesAll}
-import freechips.rocketchip.config.Parameters
+import firrtl.options.{Dependency, Phase, PreservesAll, StageOptions}
+import freechips.rocketchip.config.{Config, Field, Parameters}
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.stage.RocketChipOptions
 import freechips.rocketchip.util.HasRocketChipStageUtils
+
+case object TargetDirKey extends Field[String](".")
+
+case class WithTargetDir(dir: String) extends Config((site, here, up) => {
+  case TargetDirKey => dir
+})
 
 /** Constructs a generator function that returns a top module with given config parameters */
 class PreElaboration extends Phase with PreservesAll[Phase] with HasRocketChipStageUtils {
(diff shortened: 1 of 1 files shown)


sequencer:
Just take a glance to this PR, It seems that adding `TargetDirAnnotation` can alter `TargetDirKey`, while adding `TargetDirKey` won't change real `targetDir`, since no `TargetDirAnnotation` is added by `PreElaboration`.

src/main/scala/stage/phases/PreElaboration.scala
```diff
@@ -6,12 +6,18 @@ import chisel3.RawModule
...
+import freechips.rocketchip.config.{Config, Field, Parameters}
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.stage.RocketChipOptions
 import freechips.rocketchip.util.HasRocketChipStageUtils
 
+case object TargetDirKey extends Field[String](".")
+
+case class WithTargetDir(dir: String) extends Config((site, here, up) => {
+  case TargetDirKey => dir
+})
```
hcook, line 19: I think including `WithTargetDir` as a public class is a bit confusing since the intention isn't to allow users to select a new target dir by directly manipulating this Field, but rather only to record what was done via the command line, right? To prevent confusion on this point I suggest  dropping this class in favor of a more direct alteration below.

src/main/scala/stage/phases/PreElaboration.scala
```diff
@@ -20,10 +26,11 @@ class PreElaboration extends Phase with PreservesAll[Phase] with HasRocketChipSt
 
   override def transform(annotations: AnnotationSeq): AnnotationSeq = {
 
+    val stageOpts = view[StageOptions](annotations)
     val rOpts = view[RocketChipOptions](annotations)
     val topMod = rOpts.topModule.get
 
-    val config = getConfig(rOpts.configNames.get)
+    val config = getConfig(rOpts.configNames.get) ++ WithTargetDir(stageOpts.targetDir)
```
hcook, line 33:
```
val config = getConfig(rOpts.configNames.get).alterPartial {
  case TargetDirKey => stageOpts.targetDir
}
```

timsnyder-siv: I agree that this is much better.  I'll make the change.

src/main/scala/stage/phases/PreElaboration.scala
sequencer: Or we can add `TargetDirAnnotation(config(TargetDirKey))` at `PreElaboration`? then modification to config will change `TargetDirAnnotation`.

hcook:
My understanding is that `TargetDirKey` is meant to be a record of the  `targetDir` selected via the command line; the user is not meant to manipulate `TargetDirKey` themselves in order to change the `targetDir`. I made a suggestion that might help clarify this point, assuming my understanding is correct.

I confess I don't understand how or whether `TargetDirAnnotation` itself is added by `PreElaboration` as part of which project.

src/main/scala/stage/phases/PreElaboration.scala
hcook: I agree that seems it could be better to make them able to change together, but if they conflict with each other, which one wins? Config or command line?

I also don't understand where `TargetDirAnnotation` is being added right now. Is it problematic to apply/change it during `PreElaboration` based on the `Config` value?

hcook: @debs-sifive @jackkoenig

timsnyder-siv:
`TargetDirAnnotation` is ensured to be on the design by [Checks phase](https://github.com/timsnyder-siv/rocket-chip/blob/7a1ebd872bcc5cb76a56bde04db21db3765def39/src/main/scala/stage/phases/Checks.scala#L38) and it is the direct prereq of PreElaboration.

@hcook , your understanding of what I'm trying to do is exactly correct.  Just trying to get the CLI value into Config so that it can be used, not changed.

src/main/scala/stage/phases/PreElaboration.scala
sequencer: > I agree that seems it could be better to make them able to change together, but if they conflict with each other, which one wins? Config or command line?

Yes, that's a problem, I also encounter this at chiseltest refactoring, if conflict, I choose to throw this Error to user, but I don't think this is a good way.

debs-sifive: > I also don't understand where TargetDirAnnotation is being added right now. Is it problematic to apply/change it during PreElaboration based on the Config value?

Not sure if I understand your question correctly, but I believe it's assigned with the `-td` flag in the make flow here: https://github.com/sifive/federation/blob/master/mk/common/firrtl.mk#L33

hcook (approved):
I'm going to approve this because I think the current functionality of recording the command-line set value is immediately useful. But if @sequencer would like to work on some way of combining a Config-based value with a CLI-based value I do think that could be a good follow-up PR.

src/main/scala/stage/phases/PreElaboration.scala
hcook: Right, my question was more: which phase in charge of turning a `-td` flag into a Scala string, and does it happen before or after `rocketchip.phases.PreElaboration`. Below @timsifive  said
> TargetDirAnnotation is ensured to be on the design by Checks phase and it is the direct prereq of PreElaboration.
and I think what @sequencer is proposing is that rather than being a pre-req, it becomes something that happens / gets applied during `PreElaboration` instead, so that `Config` `Field` values can affect it. But I don't understand the potential consequences of making that change.

debs-sifive: > which phase is in charge of turning a -td flag into a Scala string
I don't think there's any one phase that does this. Phases that need the target dir take in the `TargetDirAnnotation` and then convert it/use it as needed.

The conversion from flag to annotation is in FIRRTL: https://github.com/freechipsproject/firrtl/blob/bbcee06d406c3755c10d41a71d69a69eb7d6f321/src/main/scala/firrtl/options/StageAnnotations.scala#L84

@sequencer's proposal seems safe in the sense that nothing before `PreElaboration` (besides `Checks`) uses the target dir for anything.

davidbiancolin:
FWIW, i think (in the short term) the `Key` should be made package private (per Henry's initial suggestion) and the command line option should always win. 

Indeed no phase is responsible for adding that annotation, the command line options are translated to annotations before any phase runs. See [firrtl.options.Shell](https://github.com/freechipsproject/firrtl/blob/bbcee06d406c3755c10d41a71d69a69eb7d6f321/src/main/scala/firrtl/options/Shell.scala#L49).
The definition of a CLI argument extends `HasShellOptions` and defines how it should be translated to an annotation. [Here's where that happens for TargetDir.](https://github.com/freechipsproject/firrtl/blob/ecb96e83324ea17cf38b7b90753d745d3c7f51bd/src/main/scala/firrtl/options/StageAnnotations.scala#L84).

sequencer:
＞ But if @sequencer would like to work on some way of combining a Config-based value with a CLI-based value I do think that could be a good follow-up PR.
I’ll give a new PR for this.

PR #2727, opened 2020-11-11
add ElaborationArtefactAnnotation

albertchen-sifive:
Stopgap solution to issue of `ElaborationArtefacts` outputting stale instance paths. This adds `ElaborationArtefactAnnotation` which contains the filename and list of `Token`s that contain the contents of the file. `Token` subclasses are `StringToken`, `ModulePathToken`, and `ReferencePathToken`. `ModulePathToken` and `ReferencePathToken` contain `Target`s that are updated by firrtl. `ElaborationArtefactsTransform` serializes the `ElaborationArtefactAnnotation`s

`ElaborationArtefactAnnotation`s can be annotated with the `ElaborationArtefactAnnotation.annotate` and `Token.apply` methods e.g.
```
ElaborationArtefactAnnotation.annotate("foo.json",
  Seq(Token(s"""{ foo: $foo, bar: $bar, mod: """"), Token(module), Token("""", ref: """"), Token(module.reset), Token("""" }"""))
```

there is a `TokensInterpolator` implicit to make constructing `Token`s easier.
```
ElaborationArtefactAnnotation.annotate("foo.json",
  tokens"""{ foo: $foo, bar: $bar, mod: $module, ref: ${module.reset} }""")
```

`Token.apply` requires a `Tokenizer` typeclass instance. instances are defined for `String`, `Data`, `BaseModule`, `SyncReadMem`

**Release Notes**

### DIFF
--- a/src/main/scala/util/ElaborationArtefactAnnotation.scala
+++ b/src/main/scala/util/ElaborationArtefactAnnotation.scala
@@ -95,7 +95,7 @@ object Token {
     * converted to Tokens, and all other values will be turned into
     * [[StringToken]]s using their .toString method.
     */
-  implicit class TokensInterpolator(val sc: StringContext) extends AnyVal {
+  implicit class TokensInterpolator(private val sc: StringContext) extends AnyVal {
     def tokens(args: Any*): Seq[Token] = {
       val strings = sc.parts.map(StringContext.treatEscapes).iterator
       val expressions = args.iterator
--- a/src/main/scala/util/ElaborationArtefactAnnotation.scala
+++ b/src/main/scala/util/ElaborationArtefactAnnotation.scala
@@ -4,13 +4,12 @@ package freechips.rocketchip.util
 
 import firrtl.RenameMap
 import firrtl.annotations.{Annotation, HasSerializationHints}
-import firrtl.annotations.{IsModule, ModuleTarget, ReferenceTarget, Target}
+import firrtl.annotations.{IsModule, ReferenceTarget}
 
 import chisel3.{Data, SyncReadMem}
 import chisel3.experimental.{BaseModule, ChiselAnnotation}
 
 import scala.collection.mutable
-import java.io.File
 
 /** Like [[ElaborationArtefact]] but in annotation form.
   *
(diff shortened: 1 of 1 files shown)


hcook (approved):

jerryhethatday:
so, what is the purpose of ElaborationArtefacts in general?  I have no idea what it is for? Thanks!

albertchen-sifive:
ElaborationArtefacts is a way to emit metadata to the generator build directory during elaboration. it's used to emit graphml and dts along with a bunch of other metadata files https://github.com/chipsalliance/rocket-chip/blob/147bdcc4a26c74e5d7a47e3d667d456699d6d11f/src/main/scala/subsystem/BaseSubsystem.scala#L32-L35

the `ElaborationArtefacts.add` method takes two arguments, first one is the file extension and the second is the contents of that file as a String. so `ElaborationArtefacts.add("foo", "bar")` would write a file called `$configName.foo` with contents `bar`, where `$configName` would be either the config name or the name specified by the `--name` cmdline argument if it is present.

this PR adds `ElaborationArtefactsAnnotation` which works similarly to `ElaborationArtefacts` except that it is passed as an annotation through the firrtl compiler and takes a list of `Token`s which may embed firrtl `Target`s. this allows for emitting metadata that relies on having correct instance paths and signal names.

jerryhethatday:
> ElaborationArtefacts is a way to emit metadata to the generator build directory during elaboration. it's used to emit graphml and dts along with a bunch of other metadata files
> 
> https://github.com/chipsalliance/rocket-chip/blob/147bdcc4a26c74e5d7a47e3d667d456699d6d11f/src/main/scala/subsystem/BaseSubsystem.scala#L32-L35
> 
> the `ElaborationArtefacts.add` method takes two arguments, first one is the file extension and the second is the contents of that file as a String. so `ElaborationArtefacts.add("foo", "bar")` would write a file called `$configName.foo` with contents `bar`, where `$configName` would be either the config name or the name specified by the `--name` cmdline argument if it is present.
> 
> this PR adds `ElaborationArtefactsAnnotation` which works similarly to `ElaborationArtefacts` except that it is passed as an annotation through the firrtl compiler and takes a list of `Token`s which may embed firrtl `Target`s. this allows for emitting metadata that relies on having correct instance paths and signal names.

Sorry for the delayed reply, Thank you for your detailed reply, I got it!

PR #2729, opened 2020-11-12
allow renames to multiple targets for MemoryPathToken

albertchen-sifive:
`ElaborationArtefactAnnotaion` will fail to rename if it contains a `MemoryPathToken` pointing at a seq mem of an aggregate type. e.g. for `mem: SyncReadMem[Vec[UInt]]`, firrtl will rename `mem` to `mem_0`, `mem_1`, `mem_2`, etc.

This updates `MemoryPathToken` to handle this case by renaming the token to multiple `MemoryPathToken`s separated by spaces. so the resulting output will look like `"mem_0 mem_1 mem_2 ...`

**Impact**: API modification

**Release Notes**


hcook (approved):

timsnyder-siv (approved):
I confirmed that these changes resolve the issue I encountered in my usecase.

hcook:
This fixed our use case! Thanks!

PR #2743, opened 2020-11-21
PTW.scala: enabling configurable L2 TLB (sets/ways)

ncppd:

**Release Notes**

With this PR I contribute my configurable L2 TLB implementation that can output any organization from direct-mapped to fully-associative -with the 8/16-way L2 TLB organizations being the most preferable-. The L2 TLB supports the already implemented True/Pseudo LRU replacement policies. In a larger set-associative TLB -more than 128 entries- a Random replacement policy should perhaps be preferred, in order to avoid resource/ciritical path overhead. 
Note that the one-entry TLB flush `sfence rs1, zero` stalls the TLB for one cycle in order to fetch the indexed set and complete the flush. Another (costly) approach would be to flush the indexed set in order to avoid the one-cycle delay, but this would lead to (more) costly page table walks.

### DIFF
--- a/src/main/scala/rocket/PTW.scala
+++ b/src/main/scala/rocket/PTW.scala
@@ -79,9 +79,9 @@ class PTE(implicit p: Parameters) extends CoreBundle()(p) {
   def sx(dummy: Int = 0) = leaf() && x
 }
 
-class L2TLBEntry(implicit p: Parameters) extends CoreBundle()(p)
+class L2TLBEntry(nSets: Int)(implicit p: Parameters) extends CoreBundle()(p)
     with HasCoreParameters {
-  val idxBits = log2Ceil(coreParams.nL2TLBEntries)
+  val idxBits = log2Ceil(nSets)
   val tagBits = vpnBits - idxBits
   val tag = UInt(width = tagBits)
   val ppn = UInt(width = ppnBits)
@@ -92,7 +92,7 @@ class L2TLBEntry(implicit p: Parameters) extends CoreBundle()(p)
   val w = Bool()
   val r = Bool()
 
-  override def cloneType = new L2TLBEntry().asInstanceOf[this.type]
+  override def cloneType = new L2TLBEntry(nSets).asInstanceOf[this.type]
 }
 
 @chiselName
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/PTW.scala
```diff
@@ -108,14 +108,16 @@ class PTW(n: Int)(implicit edge: TLEdgeOut, p: Parameters) extends CoreModule()(
   val s_ready :: s_req :: s_wait1 :: s_dummy1 :: s_wait2 :: s_wait3 :: s_dummy2 :: s_fragment_superpage :: Nil = Enum(UInt(), 8)
   val state = Reg(init=s_ready)
   val l2_refill_wire = Wire(Bool())
+  val l2_flush_entry_wire = Wire(Bool())
```
ingallsj, line 111: > Note that the one-entry TLB flush sfence rs1, zero stalls the TLB for one cycle in order to fetch the indexed set and complete the flush. Another (costly) approach would be to flush the indexed set in order to avoid the one-cycle delay, but this would lead to (more) costly page table walks.

In the existing direct-mapped L2TLB, we are doing the other (costly) approach: flushing the indexed set without comparing the tag.
I agree that yours is a worthwhile optimization to make, but I don't want to do that in this same pull request, so I will back out those changes.

src/main/scala/rocket/PTW.scala
```diff
@@ -199,66 +201,100 @@ class PTW(n: Int)(implicit edge: TLEdgeOut, p: Parameters) extends CoreModule()(
...
 
   val l2_refill = RegNext(false.B)
+  val l2_flush_entry = RegNext(false.B)
   l2_refill_wire := l2_refill
+  l2_flush_entry_wire := l2_flush_entry
   io.dpath.perf.l2miss := false
   io.dpath.perf.l2hit := false
   val (l2_hit, l2_error, l2_pte, l2_tlb_ram) = if (coreParams.nL2TLBEntries == 0) (false.B, false.B, Wire(new PTE), None) else {
-    val code = new ParityCode
+    val nL2TLBSets = 32
```
ingallsj, line 210: rather than hard-coding `nL2TLBSets = 32`, I will add a parameter `coreParams.nL2TLBWays` and calculate `val nL2TLBSets = coreParams.nL2TLBEntries/coreParams.nL2TLBWays`

src/main/scala/rocket/PTW.scala
```diff
@@ -199,66 +201,100 @@ class PTW(n: Int)(implicit edge: TLEdgeOut, p: Parameters) extends CoreModule()(
...
     io.dpath.perf.l2hit := s2_hit
+    when (s2_valid && s2_v_lookup(hit_way) && s2_error(hit_way)) {
+      valid_array foreach { case x => x := 0.U }
+    }
+    when (s2_hit) { l2_plru.access(r_idx, hit_way) }
+
     val s2_pte = Wire(new PTE)
-    s2_pte := s2_entry
-    s2_pte.g := s2_g
+    s2_pte := s2_set(hit_way)
```
ingallsj, line 286: this would be faster timing as `Mux1H(hit_vec, s2_set)`

ingallsj (changes requested):
Thank you for the contribution!
I'll continue working on this, and make a new PR with fixes for my feedback.

ingallsj:
I have submitted code changes for my review suggestions in https://github.com/chipsalliance/rocket-chip/pull/2748.
@ncppd can you please review those, and see if they work in your environment?

ingallsj:
merged in https://github.com/chipsalliance/rocket-chip/pull/2748

PR #2771, opened 2021-01-02
ReadyValidCancelRRArbiter: round-robin select wrong rotate

ingallsj:

**Impact**: This should have had no impact so far, because I am unaware of anything instantiating this with `rr = true`.

**Release Notes**
fix `ReadyValidCancelRRArbiter` output value when Round-Robin parameter `rr = true`


ingallsj:
I have tested this in a larger project and confirmed that it is now working as it should.

jsmithsf (approved):

Issue #2765, opened 2020-12-21
AXI4ToTL doesn't handle AXI4 ordering model


tymcauley:
**Type of issue**: bug report | feature request

**Impact**: API modification

**Development Phase**: proposal

**Other information**
See conversation from Chipyard Google Group: https://groups.google.com/g/chipyard/c/dGCYZZB12is

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

**What is the current behavior?**

#### Summary

The `AXI4ToTL` widget currently depends on only being able to send one TileLink transaction at a time. Effectively, it must be paired with a `TLFIFOFixer`, or it can generate illegal TileLink transactions (multiple outstanding transactions with the same source ID), and also return illegally-ordered AXI4 responses. This is a performance issue (only allowing one in-flight transaction severely limits AXI4 bandwidth, especially when paired with a high-latency TileLink memory system), and then turns into a logic bug when trying to fix the performance issue.

#### Expected behavior

Right now, we depend on pairing an `AXI4ToTL` widget with an `AXI4UserYanker` to define the `maxFlight` parameter in `AXI4ToTL`. This defines the maximum supported in-flight transactions for each AXI ID.

The "Operation Ordering" section of the TileLink spec says:

> Within a TileLink network, there may be multiple outstanding operations inflight at any given time. These operations may be completed in any order.

Section 5.1 of the AXI4 spec says:

> All transactions with a given AXI ID value must remain ordered, but there is no restriction on the ordering of transactions with different ID values.

These semantics are different between TileLink and AXI4, in particular AXI4 has special ordering restrictions within a single AXI ID. When `maxFlight` is set to 1, there are no ordering issues, since each AXI ID can only have a single transaction in flight at once, and there are no ordering constraints between transactions with different AXI IDs. In this case, the AXI4 ordering model matches TileLink's ordering model. However, once `maxFlight` exceeds 1, the AXI network depends on `AXI4ToTL` to maintain ordering of transactions within the same AXI ID, which is not the current behavior.

#### Current behavior (read requests)

Looking at the example of read requests, this is the logic `AXI4ToTL` uses to generate multiple TileLink requests for AXI requests on a single AXI ID:

```scala
val r_count = RegInit(Vec.fill(numIds) { UInt(0, width = txnCountBits) })
val r_id = if (maxFlight == 1) {
  Cat(in.ar.bits.id, UInt(0, width=1))
} else {
[...]

PR #2773, opened 2021-01-14
Update AXI4ToTL with read response reordering

tymcauley:
Adds the `numTlTxns` parameter to `AXI4ToTL`, which controls the number of read/write TileLink transactions that can be in flight at once. Setting `numTlTxns` to 1 minimizes added hardware, and retains the performance of the previous version of this module when coupled with `TLFIFOFixer`. Setting `numTlTxns` above 1 (it must be a power of 2) adds SRAMs to reorder read responses that return out-of-order.

With this change, `AXI4ToTL` no longer needs to be paired with `TLFIFOFixer`; `AXI4ToTL` can maintain legal read response ordering within each supported AXI ID.

All of these features work for multi-beat bursts.

Open questions/notes:
- Should the generated TileLink master request FIFO if any of the upstream AXI masters are requesting FIFO? And therefore, should we just force `numTlTxns` to 1 in that case?
- `maxBeats` (and therefore the number of SRAMs in `ReservableListBuffer`) could be minimized if we added a `maxBurstLen` parameter to `AXI4MasterParameters`. That could be added in a subsequent PR if that seems like a good idea.
- I've written a pretty ad-hoc testbench to verify these features, and didn't include it in this PR. I'm open to suggestions as to how to test this block, and how to integrate that testing into the normal RocketChip regression flow.

**Related issue**: #2765

**Type of change**: bug report | feature request

**Release Notes**

### DIFF
--- a/src/main/scala/amba/axi4/ToTL.scala
+++ b/src/main/scala/amba/axi4/ToTL.scala
@@ -2,218 +2,666 @@
 
 package freechips.rocketchip.amba.axi4
 
-import Chisel._
+import chisel3._
+import chisel3.util._
 import freechips.rocketchip.amba._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.tilelink._
 import freechips.rocketchip.util._
 
-case class AXI4ToTLIdMapEntry(tlId: IdRange, axi4Id: IdRange, name: String)
-  extends IdMapEntry
-{
-  val from = axi4Id
-  val to = tlId
-  val isCache = false
-  val requestFifo = false
-  val maxTransactionsInFlight = Some(tlId.size)
+class ReorderData(val dataWidth: Int, val respWidth: Int, val userFields: Seq[BundleFieldBase]) extends Bundle {
+  val data = UInt(dataWidth.W)
+  val resp = UInt(respWidth.W)
+  val last = Bool()
+  val user = BundleMap(userFields)
 }
 
-case class AXI4ToTLNode(wcorrupt: Boolean)(implicit valName: ValName) extends MixedAdapterNode(AXI4Imp, TLImp)(
-  dFn = { case mp =>
-    mp.masters.foreach { m => require (m.maxFlight.isDefined, "AXI4 must include a transaction maximum per ID to convert to TL") }
-    val maxFlight = mp.masters.map(_.maxFlight.get).max
-    TLMasterPortParameters.v1(
-      clients = mp.masters.filter(_.maxFlight != Some(0)).flatMap { m =>
-        for (id <- m.id.start until m.id.end)
-          yield TLMasterParameters.v1(
-            name        = s"${m.name} ID#${id}",
-            sourceId    = IdRange(id * maxFlight*2, (id+1) * maxFlight*2), // R+W ids are distinct
-            nodePath    = m.nodePath,
-            requestFifo = true)
-      },
-      echoFields    = mp.echoFields,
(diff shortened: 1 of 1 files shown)


RaphaelKlink:
hey,

I tried to add your axi4totl fix t my project to check if the performance of the cva6 core improves. Unfortunately, i got an error. You bumped the TLMasterPortParameters and TLMasterParameters from v1 to v2, but failed to add the new versions of these classes. Could you extend your merge request with these?

michael-etzkorn:
Hopefully this gets reviewed. I'm honestly tempted to experiment with it anyway. In my current design, AXI4ToTL is causing some issues in simulation as far as writing the correct data to the peripheral, but I have no TLFIFOFixer. Unfortunately, it seems that adding the TLFIFOFixer to my slave port causes the simulation to crash.

abejgonzalez:
I'm surprised this hasn't been looked at / reviewed.

Throwing this out there, maybe for adoption it might make sense to initially fork the `AXI4ToTL` code into a new adapter (`AXI4ToTLExperimental`) so that the old "tested" adapter is unmodified for now. 

As for testing, it might make sense to do some unit testing with the built-in unit test framework like https://github.com/chipsalliance/rocket-chip/blob/25e2c63567689ebe1fc5e60fdfe3375a8dba071c/src/main/scala/unittest/Configs.scala#L3 (for extra examples you can look at https://github.com/ucb-bar/testchipip/blob/master/src/main/scala/Unittests.scala (and how to run it in CY: https://github.com/ucb-bar/chipyard/blob/3b303ba37d3a9f197ca1cb1b4865649f7b586cb8/variables.mk#L91-L101).

@jerryz123 Any thoughts on this?

jerryz123:
I support adding this as AXI4ToTLExperimental

jerryz123:
Have we considered adding this directly to FireSim sources, since that seems to be the only place where this becomes performance-critical?

abejgonzalez:
> Have we considered adding this directly to FireSim sources, since that seems to be the only place where this becomes performance-critical?

Originally this was used for external IP that was AXI4-based (i.e. NVDLA or something like CVA6). So I think it makes sense to be committed here. @tymcauley I assume you used this to test NVDLA in the past?

tymcauley:
> > Have we considered adding this directly to FireSim sources, since that seems to be the only place where this becomes performance-critical?
> 
> Originally this was used for external IP that was AXI4-based (i.e. NVDLA or something like CVA6). So I think it makes sense to be committed here. @tymcauley I assume you used this to test NVDLA in the past?

Exactly right, I used this to improve NVDLA performance in the past. Adding this as an experimental block (or putting it in a new `freechips.rocketchip.amba.axi4.experimental` package, whatever seems best) sounds like a great idea.

terpstra:
So I only just now looked at this. I am not going to approve this.

The AXI4 requirement isn't just that reads return their responses data in order. It's also a requirement that effects of those reads happen in order AND that the sampling occurs in order.

For example, suppose there were two AXI4 reads with ARID=3 for addresses A + B. If another core were to update (in order) A and then B concurrently, it would be illegal to respond with the new value of B and the old value of A. However, this change would make this illegal re-ordering possible.

If your goal is to boost the performance of AXI4 accesses to main memory, there are other techniques that can be employed, such as translating the requests into unordered cache block acquisitions which then commit/release back to the system in order.

terpstra:
An even easier way to boost performance, which doesn't involve changing the bridge at all, is to only ever issue a single request to a given ARID. Then the operations will be able to legally execute out-of-order.

PR #2780, opened 2021-01-31
Removed .sbtopts which contains '$PWD' by default, and remove the '\'…

tiehexue:
… to ensure the command is executed. Should run './scripts/swap-sbt-build' manually at first.

**Related issue**: .sbtopts is incorrected to set to -Dsbt.workspace=$PWD, which causes loading the project in IDEA failed 

**Release Notes**

### DIFF
--- a/.sbtopts
+++ b/.sbtopts
@@ -1,2 +1,0 @@
--Dsbt.sourcemode=true
--Dsbt.workspace=$PWD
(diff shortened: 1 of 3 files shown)


jackkoenig:
The purpose of `.sbtopts` is to make this repo work well when using SBT in the command-line which is the intended / default use mode. There are instructions in the README for how to use an IDE (https://github.com/chipsalliance/rocket-chip#-building-rocket-chip-with-an-ide). Obviously, I'd prefer a solution that works out-of-the-box for both SBT on the command-line and IDEs, but in the case where it's a tradeoff, command-line is the default.

PR #2802, opened 2021-03-30
div width fix

sequencer:

**Impact**: API modification

I believe this was a typo.


aswaterman:
Not a typo.  The multiplier hardware also uses the `remainder` register and expects it to have `2*mulw+2` bits.

sequencer:
Got it. Thanks for explaination!

PR #2805, opened 2021-04-01
TLToAXI4 should not block AXI4-b writeacks indefinitely

jerryz123:
https://github.com/chipsalliance/rocket-chip/blob/025184e1a0fdd8338eb0f66ad9ab33587b619854/src/main/scala/tilelink/ToAXI4.scala#L205-L209

The current arbitration policy in the TLToAXI4 converter always prioritizes AXI4 r-channel responses. This policy leads to starvation for clients waiting for a AccessAck, when the AXI4 r-channel is saturated. For example, a system where the L2 cache evicts a line, then saturates AXI4 with reads, may stall indefinitely waiting for a WriteAck.

This PR implements an alternate policy that stalls for b-channel acks if the b-channel ack has been blocked for 7 consecutive cycles. This is just one possible approach to avoid the indefinite stall. 

**Type of issue**: bug report 

### DIFF
--- a/src/main/scala/tilelink/ToAXI4.scala
+++ b/src/main/scala/tilelink/ToAXI4.scala
@@ -205,10 +205,16 @@ class TLToAXI4(val combinational: Boolean = true, val adapterName: Option[String
       // R and B => D arbitration
       val r_holds_d = RegInit(Bool(false))
       when (out.r.fire()) { r_holds_d := !out.r.bits.last }
-      // Give R higher priority than B
-      val r_wins = out.r.valid || r_holds_d
+      // Give R higher priority than B, unless B has been delayed for 8 cycles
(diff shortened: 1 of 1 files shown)


jerryz123:
@hcook @terpstra

hcook:
Ideally this would be a configurable policy, but I agree this approach will prevent starvation and so is better than the status quo. What led you to choose the number 7?

hcook (approved):

terpstra (changes requested):
I believe @andrew-gouldey-sifive has a more complete solution which made this configurable.

Should this PR be closed?

terpstra (approved):
I agree a configurable policy would be better, but this is better than nothing.

PR #2811, opened 2021-04-11
convert TestHarness to chisel3, work around compatibility mode naming issue.

sequencer:
This is the work around for TestHarness io cannot be named correctly chipsalliance/chisel3#1862.
But I think this should be safe to merge.

**Related issue**: chipsalliance/chisel3#1862

### DIFF
--- a/src/main/scala/amba/axi4/Bundles.scala
+++ b/src/main/scala/amba/axi4/Bundles.scala
@@ -3,6 +3,7 @@
 package freechips.rocketchip.amba.axi4
 
 import Chisel._
+import chisel3.chiselTypeOf
 import chisel3.util.Irrevocable
 import freechips.rocketchip.util._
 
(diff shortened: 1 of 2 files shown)


src/main/scala/amba/axi4/Bundles.scala
```diff
@@ -82,11 +83,16 @@ class AXI4Bundle(params: AXI4BundleParameters) extends AXI4BundleBase(params)
...
         b.valid  := Bool(false)
+        b.bits := 0.U.asTypeOf(chiselTypeOf(b.bits))
       case OUTPUT =>
         ar.valid := Bool(false)
+        ar.bits  := 0.U.asTypeOf(chiselTypeOf(ar.bits))
         aw.valid := Bool(false)
+        aw.bits  := 0.U.asTypeOf(chiselTypeOf(aw.bits))
         w.valid  := Bool(false)
+        w.bits   := 0.U.asTypeOf(chiselTypeOf(w.bits))
         r.ready  := Bool(false)
```
jackkoenig, line 96: @mwachs5 do you think these `.bits` are better tied to zero or should they be `DontCare`?

mwachs5: i think they should be `DontCare`

sequencer: sure

hcook (approved):

PR #2832, opened 2021-05-03
DCache does not block Acquire/ProbeAck by outstanding Release

ingallsj:
TileLink includes these message ordering rules:

> Release
> Once the Release is issued, the master should not issue a ProbeAck, ProbeAckData, Acquire, Get, Put, ArithmeticData, LogicalData, or further Release on the block until it receives a ReleaseAck from the slave acknowledging completion of the writeback.

**Related issue**: The Rocket/Bullet DCache relaxed this in https://github.com/chipsalliance/rocket-chip/pull/2696 to allow a ProbeAck or Acquire to overtake a Release, but not overtake a ReleaseData.

**Impact**: functional change

### DIFF
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -574,7 +574,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   tl_out_a.valid := !io.cpu.s2_kill &&
     (s2_valid_uncached_pending ||
       (s2_valid_cached_miss &&
-       !release_ack_wait &&
+       !(release_ack_wait && (tl_out_a.bits.address ^ release_ack_addr)(((pgIdxBits + pgLevelBits) min paddrBits) - 1, idxLSB) === 0) &&
        (cacheParams.acquireBeforeRelease && !release_ack_wait && release_queue_empty || !s2_victim_dirty)))
   tl_out_a.bits := Mux(!s2_uncached, acquire(s2_vaddr, s2_req.addr, s2_grow_param),
     Mux(!s2_write, get,
--- a/src/main/scala/rocket/DCache.scala
+++ b/src/main/scala/rocket/DCache.scala
@@ -735,7 +735,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
 
   // Handle an incoming TileLink Probe message
   val block_probe_for_core_progress = blockProbeAfterGrantCount > 0 || lrscValid
-  val block_probe_for_pending_release_ack = release_ack_wait && (tl_out.b.bits.address ^ release_ack_addr)(idxMSB, idxLSB) === 0
+  val block_probe_for_pending_release_ack = release_ack_wait && (tl_out.b.bits.address ^ release_ack_addr)(((pgIdxBits + pgLevelBits) min paddrBits) - 1, idxLSB) === 0
   val block_probe_for_ordering = releaseInFlight || block_probe_for_pending_release_ack || grantInProgress
   metaArb.io.in(6).valid := tl_out.b.valid && (!block_probe_for_core_progress || lrscBackingOff)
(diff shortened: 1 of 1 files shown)


aswaterman:
Paging all this back in, I think this PR will have a significant impact on our memory-latency benchmarks.
- Removing `release_ack_dirty` from the `tl.a.valid` will prevent issuing a cache miss while the drop of an unrelated clean line is pending, which is unnecessary serialization.
- The reason I added `release_ack_dirty` to the block_probe logic is that it fixed a perf problem with back-probes.  The issue is that we don't always perform noisy drops: if there's already a pending release that hasn't been acked, we instead do a silent drop, because otherwise we'd have to stall.  So back-probes do happen.  And when they do, the address comparison always matches, because of course the address being probed has the same index as the most recently dropped line.

Unless I'm missing something, we should just undo the change to `tl.a.valid`.

As for the second change, I see a few options:
- Include more bits in the address comparison to eliminate most false positives (say, an additional `pgLevelBits`).
- Change the TL spec as discussed on the ticket.

In the interest of forward progress, I'll suggest backing out the `tl.a.valid` change, and changing the comparison logic to use bits `((pgIdxBits + pgLevelBits) min paddrBits) - 1` .. `idxLSB`.  That should work OK in practice without coupling this PR to the TL discussion.

Of course, I'm open to alternate suggestions, and if it appears the TL spec will actually change, it certainly would be nice to make this whole point moot.

ingallsj:
> - Removing `release_ack_dirty` from the `tl.a.valid` will prevent issuing a cache miss while the drop of an unrelated clean line is pending, which is unnecessary serialization.
> Unless I'm missing something, we should just undo the change to `tl.a.valid`.
> In the interest of forward progress, I'll suggest backing out the `tl.a.valid` change,

Is this scenario handled elsewhere?  Issuing a cache miss (AcquireBlock) while the drop (Release) of the same clean line.  If yes, then I agree, I should not have put in the `tl.a.valid` change.  If not, then the race starts looking similar for Acquire-Release as ProbeAck-Release.

> The reason I added `release_ack_dirty` to the block_probe logic is that it fixed a perf problem with back-probes. The issue is that we don't always perform noisy drops: if there's already a pending release that hasn't been acked, we instead do a silent drop, because otherwise we'd have to stall. So back-probes do happen. And when they do, the address [index] comparison always matches, because of course the address being probed has the same index as the most recently dropped line.
> As for the second change, I see a few options:
> Include more bits in the address comparison to eliminate most false positives (say, an additional `pgLevelBits`).
> In the interest of forward progress, I'll suggest backing out the tl.a.valid change, and changing the comparison logic to use bits `((pgIdxBits + pgLevelBits) min paddrBits) - 1` .. `idxLSB`. That should work OK in practice without coupling this PR to the TL discussion.

Will do now.

aswaterman:
@ingallsj Hmm, I think you might be right about the `tl.a.valid` case.  Ugh.

Perhaps we stick with your change but add an address check--I think comparing the same set of bits between `release_ack_addr` and `s2_req.addr` would suffice.  Agreed?

aswaterman (approved):

src/main/scala/rocket/DCache.scala
```diff
@@ -575,7 +574,7 @@ class DCacheModule(outer: DCache) extends HellaCacheModule(outer) {
   tl_out_a.valid := !io.cpu.s2_kill &&
     (s2_valid_uncached_pending ||
       (s2_valid_cached_miss &&
-       !(release_ack_wait && release_ack_dirty) &&
+       !(release_ack_wait && (tl_out_a.bits.address ^ release_ack_addr)(((pgIdxBits + pgLevelBits) min paddrBits) - 1, idxLSB) === 0) &&
```
aswaterman, line 577: Actually, I think we should follow my original suggestion here and use `s2_req.addr` instead of `tl_out_a.bits.address`.  This clause is already qualified by by `!s2_uncached` (via `s2_valid_cached_miss`), so we know for sure `tl_out_a.bits.address` will be `s2_req.addr`, and we can remove a few levels of logic in the process.

Issue #2843, opened 2021-06-11
Add Protobuf output from Chisel as a CLI option


sam-shahrestani:
**Type of issue**: feature request

**Impact**:  API addition (no impact on existing code)

**Development Phase**: proposal

**Other information**
 
A working minimal example of how to do this reasonably: https://github.com/sam-shahrestani/rocket-chip/commit/65f3232db338fd9a35e5049440986d91352febbf

I'm not sure if this meets code guidelines for this repo, so I haven't PR'd, but we'll be using this as a solution internally for now.

**What is the use case for changing the behavior?**
We've had some issues in the past with memory running out on a medium sized design, partly due to the JVM being a bit dodgy with memory management. Using the Protobuf format gave us a decent speed/memory advantage, and we implemented it using  ` Driver.dumpProto` in an much older checkout. Pulling all the new changes completely overwhelmed this (I'm not sure `GeneratorUtils` even exist anymore?), but it could be a lot nicer with the Cli mix-in style.

The way I see it, this could be very useful for others looking to keep memory usage down somewhat.

PR #2845, opened 2021-06-14
Add '--dump-protobuf' and DumpProtoBufAnnotation

sam-shahrestani:
**Related issue**: Add Protobuf output from Chisel as a CLI option #2843

**Release Notes**

Added command line argument "--dump-protobuf" to generate Chisel output in the ProtoBuf format.

### DIFF
--- a/src/main/scala/stage/RocketChipAnnotations.scala
+++ b/src/main/scala/stage/RocketChipAnnotations.scala
@@ -50,3 +50,19 @@ private[stage] object OutputBaseNameAnnotation extends HasShellOptions {
     )
   )
 }
+
+/** Optionally use ProtoBuf as the output format for Chisel */
+case object DumpProtoBufAnnotation 
+  extends NoTargetAnnotation
+  with RocketChipOption
+  with HasShellOptions {
+
+  override val options = Seq(
+    new ShellOption[Unit](
+      longOption = "dump-protobuf",
+      toAnnotationSeq = _ => Seq(DumpProtoBufAnnotation),
+      helpText = "Change the Chisel output format to ProtoBuf and name the generated FIRRTL output with a .pb extension",
+      shortOption = Some("dpb")
+    )
+  )
+}
--- a/src/main/scala/stage/RocketChipAnnotations.scala
+++ b/src/main/scala/stage/RocketChipAnnotations.scala
@@ -62,7 +62,7 @@ case object DumpProtoBufAnnotation
       longOption = "dump-protobuf",
       toAnnotationSeq = _ => Seq(DumpProtoBufAnnotation),
       helpText = "Change the Chisel output format to ProtoBuf and name the generated FIRRTL output with a .pb extension",
-      shortOption = Some("dpb")
+      shortOption = Some("pb")
     )
   )
 }
--- a/src/main/scala/stage/RocketChipOptions.scala
+++ b/src/main/scala/stage/RocketChipOptions.scala
@@ -41,6 +41,5 @@ class RocketChipOptions private[stage] (
       if (!topPackage.isEmpty && !configClass.isEmpty) Some(s"${topPackage.get}.${configClass.get}") else None
   }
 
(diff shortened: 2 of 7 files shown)


src/main/scala/stage/RocketChipOptions.scala
```diff
@@ -37,5 +40,7 @@ class RocketChipOptions private[stage] (
     case _ =>
       if (!topPackage.isEmpty && !configClass.isEmpty) Some(s"${topPackage.get}.${configClass.get}") else None
   }
+
+  lazy val chiselFileExtension: Option[String] = if (dumpProtobuf) Some(".pb") else Some(".fir")
```
jackkoenig, line 44:
```suggestion
  lazy val chiselFileExtension: String = if (dumpProtobuf) ".pb" else ".fir"
```
I don't think there's any reason to make this an `Option` since it will always have a file extension

src/main/scala/stage/RocketChipAnnotations.scala
```diff
@@ -50,3 +50,19 @@ private[stage] object OutputBaseNameAnnotation extends HasShellOptions {
...
+  extends NoTargetAnnotation
+  with RocketChipOption
+  with HasShellOptions {
+
+  override val options = Seq(
+    new ShellOption[Unit](
+      longOption = "dump-protobuf",
+      toAnnotationSeq = _ => Seq(DumpProtoBufAnnotation),
+      helpText = "Change the Chisel output format to ProtoBuf and name the generated FIRRTL output with a .pb extension",
+      shortOption = Some("dpb")
```
jackkoenig, line 65:
```suggestion
      shortOption = Some("pb")
```
It doesn't really matter, but I think just `pb` is more clear.

jackkoenig (commented):
Sorry for my delay in reviewing, this looks good to me, just a couple of minor nits.

src/main/scala/stage/phases/TransformAnnotations.scala
```diff
@@ -17,6 +17,9 @@ class TransformAnnotations extends Phase with PreservesAll[Phase] with HasRocket
 
   override def transform(annotations: AnnotationSeq): AnnotationSeq = {
     /** Construct output file annotation for emission */
-    new ChiselOutputFileAnnotation(view[RocketChipOptions](annotations).longName.get) +: annotations
+    new ChiselOutputFileAnnotation(
+      view[RocketChipOptions](annotations).longName.get +
+      view[RocketChipOptions](annotations).chiselFileExtension.getOrElse("")
```
sam-shahrestani, line 22:
```suggestion
      view[RocketChipOptions](annotations).chiselFileExtension
```

sam-shahrestani: Need to remove this too if we don't have it as an Option.

jackkoenig: Yep, missed that.

sequencer:
Although @jackkoenig approved this. I'm personally down to this since after migrating to MFC, which doesn't support the `protobuf` emitting. We may need reverting this in the future?

jackkoenig:
> Although @jackkoenig approved this. I'm personally down to this since after migrating to MFC, which doesn't support the `protobuf` emitting. We may need reverting this in the future?

I agree with moving away from protobuf. Historically, the ANTLR parser had terrible performance pathologies for large designs. The worst of these performance issues [were fixed in FIRRTL v1.5.0](https://github.com/chipsalliance/firrtl/pull/2343) (although I'll note protobuf parsing remains better).

sequencer:
Thanks @jackkoenig, I'll close this PR for now. @sam-shahrestani fell free to reopen if any other thoughts about this. 
FIRRTL 1.5 provides a good performance on serialization and deserialization.
If `protobuf` lands to MFC(MLIR FIRRTL Compiler, aka CIRCT) in the future, we can revisit this.

PR #2844, opened 2021-06-11
Bump Chisel and FIRRTL along 3.4.x/1.4.x branches

seldridge:
Bump chisel3 to the HEAD of the 3.4.x branch and FIRRTL to the head of the 1.4.x branch.

**Related issue**: 

**Type of change**: other enhancement

**Impact**: no functional change

**Development Phase**: proposal

**Release Notes**

PR #2846, opened 2021-06-16
switch chisel3 dependency to ivy and bump to chisel 3.4.3

sequencer:
**Related issue**: suppress #2844

**Release Notes**
Use ivy chisel3.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,12 +4,6 @@
 [submodule "torture"]
 	path = torture
 	url = https://github.com/ucb-bar/riscv-torture.git
-[submodule "chisel3"]
-	path = chisel3
-	url = https://github.com/freechipsproject/chisel3.git
-[submodule "firrtl"]
-	path = firrtl
-	url = https://github.com/freechipsproject/firrtl.git
 [submodule "api-config-chipsalliance"]
 	path = api-config-chipsalliance
 	url = https://github.com/chipsalliance/api-config-chipsalliance.git
--- a/build.sbt
+++ b/build.sbt
@@ -6,7 +6,7 @@ import scala.sys.process._
 enablePlugins(PackPlugin)
 
 // This needs to stay in sync with the chisel3 and firrtl git submodules
-val chiselVersion = "3.4.1"
+val chiselVersion = "3.4.3"
 
 lazy val commonSettings = Seq(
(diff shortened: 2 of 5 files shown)


jackkoenig:
I definitely think we should switch to using the published dependencies, but I'm not sure if we should delete the ability to use the submodules. Projects like Chipyard sometimes use a different commit of Chisel and FIRRTL and this obviously precludes that. The whole purpose of the `sourceDependency` stuff in the `build.sbt` is that it enables us to use the published artifacts while allowing users of rocket-chip to override and use submodules if they want.

That being said, if Chipyard folks are down to ditch the submodules, then I'm in favor. @alonamid @a0u @jerryz123 @ abejgonzalez, any thoughts?

sequencer:
BTW, If it is possible, I'm willing to experimentally help Chipyard guys to migrate to mill, which provides a better build-from-source infra then sbt.

jackkoenig:
@sequencer Since we merged https://github.com/chipsalliance/rocket-chip/pull/2847, can you rebase? I'm still of two minds about just deleting the submodules, but let's go for it. We can add them back if it's a problem for Chipyard or others.

sequencer:
@jackkoenig Done :)

jackkoenig (approved):

jackkoenig:
I think there's a question if we should turn on `autoclonetype2` (which will break downstream users) or just wait for it to be on by default in Chisel 3.5

itviewer:
Should we remove the `.sbtopts` file and clean up the content in https://github.com/chipsalliance/rocket-chip/blob/master/README.md#-building-rocket-chip-with-an-ide and https://github.com/chipsalliance/rocket-chip/blob/master/CONTRIBUTING.md#-bumping-chisel-and-firrtl

jackkoenig:
> Should we remove the .sbtopts file and clean up the content in https://github.com/chipsalliance/rocket-chip/blob/master/README.md#-building-rocket-chip-with-an-ide and https://github.com/chipsalliance/rocket-chip/blob/master/CONTRIBUTING.md#-bumping-chisel-and-firrtl

Yes, and any sbt plugins that aren't used by rocket-chip

sequencer:
I’ll PR later this week.

PR #2847, opened 2021-06-18
remove wake build script.

sequencer:


jackkoenig (approved):

davidmlw:
What is the consideration of removing build.wake? It is a cool feature. @sequencer @jackkoenig

sequencer:
There is a big maintain burden, and leak of users, so we choose to remove it.
In the future, a better way is splitting RC and switching to ivy-based compiling strategy.

davidmlw:
I see. Wish to see ivy- and mill-based flow. I would like to offer help on wake based scripts, when the RC is splitted.

PR #2851, opened 2021-06-29
Clean up unnecessary sbt-launch.jar

itviewer:
Type of change: other enhancement
Impact: no functional change
Development Phase: implementation

I think this file may be introduced due to some historical reasons. It has not been updated for a long time and I doubt whether this file is still used. 
This launcher file is based on sbt 1.1.1, but the default sbt version of [project/build.properties](https://github.com/chipsalliance/rocket-chip/blob/master/project/build.properties) is 1.3.13.


sequencer (approved):

itviewer:
https://github.com/chipsalliance/rocket-chip/actions/runs/1118824846

> First-time contributors need a maintainer to approve running workflows

@sequencer Thanks

itviewer:
@sequencer 
It seems that sbt cache does not work for rocket-chip? (It can be observed through a successful action log, for example https://github.com/chipsalliance/rocket-chip/runs/3246794390?check_suite_focus=true)



Every Test job uses sbt-launch.jar in this repository to download sbt dependencies?



It seems that if we do not cache sbt as in [chisel3](https://github.com/chipsalliance/chisel3/blob/75c00606df6a2ba48d5354f32d5dbe327552ad45/.github/workflows/test.yml#L81), we cannot remove sbt-launch.jar.

sequencer:
maybe you need to add setup-scala to GitHub action

itviewer:
Let's keep it as it is ……

PR #2914, opened 2021-12-08
Fix GenericParameterizedBundle deprecation note

michael-etzkorn:

Maybe more info can be provided in this deprecation note about the Compiler Plugin and auto-cloning, but for now, just fixing the grammar. 

**Release Notes**


jerryz123 (approved):

michael-etzkorn:
I'd like to re-introduce versioning before merging this so we can have consistent deprecation notes.

sequencer:
Should we directly removing this API? if we decide to maintain the deprecation flow, we may a more complex flow.

Issue #2920, opened 2021-12-21
TraceGenConfig Incompatibility


jrarnold123:
**Type of issue**: bug report

**Impact**: unknown

**Development Phase**: request

**Other information**

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

Running make CONFIG=freechips.rocketchip.system.TraceGenConfig fails, after TraceGenConfig and associated components are added to the main Config.scala file.  Specifying a BusTopology does not resolve the issue.

**What is the current behavior?**

PreElaboration failure resulting from invalid network graph.

**Please tell us about your environment:**

- version: git commit 330c2b808e19fec66f11fbae33c939b196f0bc42
- OS: Ubuntu 18.04.6 LTS
- PATH includes the scripts directory

**What is the use case for changing the behavior?**
Testing GroundTest traces on the memory system

michael-etzkorn:
So I don't know what your custom `TraceGenConfig` looks like or your `TestHarness`, but assuming you're using the groundtest TestHarness, and you've set up to use `GroundTestBaseConfig`, I was able to get past Diplomacy errors by commenting out this line in `GroundTestSubsystem`

`IntSinkNode(IntSinkPortSimple()) :=* ibus.toPLIC`

I'm not sure why it was included if `GroudnTestSubsystem` sets PLIC's Key to `None`? But yeah looks like `HeterogeneousTileExampleConfig` for the groundtest TestHarness doesn't work out of the box with `WithTraceGen` because of a failed diplomatic graph for the interrupt node and PRCI which is fixed with setting `SubsystemExternalResetVectorKey => true`. It also doesn't work with the ExampleRocketSystem's TestHarness.

PR #2923, opened 2021-12-28
Removed TraceGen in HeterogeneousTileExampleConfig

michael-etzkorn:
**Related issue**: 

**Impact**: API modification

**Release Notes**

The `HeterogeneousTileExampleConfig` will not work with `ExampleRocketSystem` or `GroundTestSubsystem`. Therefore, it should be put firmly into one camp to be congruent with the examples provided. 

This pull request also makes parentheses stylistically more consistent in `src/main/scala/system/Configs.scala`


michael-etzkorn:
Related issue: 
- https://github.com/chipsalliance/rocket-chip/issues/2920

Some relation to the style guide discussion we had today in the working group since I moved some parentheses to a new line if the config was more than one line.

PR #2532, opened 2020-06-24
Adding V2 Constructors in Paramaters and Emits Parameters

brrmorre:
**Related issue**: 

**Type of change**: bug report | feature request | other enhancement

**Impact**: no functional change | API addition (no impact on existing code) | API modification

**Development Phase**: proposal |  implementation

**Release Notes**

PR #2925, opened 2022-01-05
Doc fixes and add comments to axi4 bundles

jiegec:
**Related issue**: None

**Release Notes**


sequencer:
IMHO, porting like this makes RC more friendly to newbies, most of them never reads Chisel2 codes before. And Chisel also leaks of compatibility layer docs. 
I’d like to see this happen if there is no objection from @aswaterman

aswaterman:
I’m ok with it if and only if some representatives designs are proven equivalent.

jerryz123:
Splitting this up into separate PRs to add documentation/upgrade to chisel3 is the way to go.

michael-etzkorn (approved):
Now that the controversial chisel3 changes are moved to here https://github.com/chipsalliance/rocket-chip/pull/2982 

I think these NFC are worth merging. They're in line with the direction of better documentation #2960

src/main/scala/amba/axi4/AsyncCrossing.scala
```diff
@@ -96,6 +106,8 @@ class AXI4RAMAsyncCrossing(txns: Int)(implicit p: Parameters) extends LazyModule
   val island = LazyModule(new CrossingWrapper(AsynchronousCrossing()))
   val ram  = island { LazyModule(new AXI4RAM(AddressSet(0x0, 0x3ff))) }
 
+  // topology: AXI4RAM <-< AXI4CrossingSink <-< AXI4CrossingSource <-<
+  //   TLToAXI4 <-< TLRAMModel <-< TLFuzzer
```
sequencer, line 110: Maybe this comments can be moved scaladoc of this Module.

src/main/scala/amba/axi4/AsyncCrossing.scala
```diff
@@ -86,7 +96,7 @@ class AXI4AsyncCrossing(params: AsyncQueueParams = AsyncQueueParams())(implicit
   }
 }
 
-/** Synthesizeable unit tests */
+/** Synthesizable unit tests */
```
sequencer, line 99: actually this is not a Scala doc.
```suggestion
// Synthesizable unit tests
```

src/main/scala/amba/axi4/AsyncCrossing.scala
```diff
@@ -29,6 +34,11 @@ class AXI4AsyncCrossingSource(sync: Option[Int])(implicit p: Parameters) extends
   }
 }
 
+/**
+  * Sink(Manager) side for AXI4 crossing clock domain
```
sequencer, line 38: there is no Client-Manager concept in AXI. Just Master/Slave(according to spec of AXI) is enough.

src/main/scala/amba/axi4/AsyncCrossing.scala
```diff
@@ -9,6 +9,11 @@ import freechips.rocketchip.tilelink._
 import freechips.rocketchip.subsystem.CrossingWrapper
 import freechips.rocketchip.util._
 
+/**
+  * Source(Client) side for AXI4 crossing clock domain
```
sequencer, line 13: there is no Client-Manager concept in AXI. Just Master/Slave(according to spec of AXI) is enough.

src/main/scala/amba/ahb/Protocol.scala
```diff
@@ -37,8 +37,8 @@ object AHBParameters
   def RESP_SPLIT = UInt(3, width=2)
 
   def PROT_DATA        = UInt(1, width = protBits)
-  def PROT_PRIVILEDGED = UInt(2, width = protBits)
+  def PROT_PRIVILEGED  = UInt(2, width = protBits)
   def PROT_BUFFERABLE  = UInt(4, width = protBits)
   def PROT_CACHEABLE   = UInt(8, width = protBits)
-  def PROT_DEFAULT = PROT_DATA | PROT_PRIVILEDGED
+  def PROT_DEFAULT = PROT_DATA | PROT_PRIVILEGED
```
sequencer, line 43: Thanks for this typo fix:
However for the happiness of RocketChip downstream, would you mind adding a function alias to not break the rocket-chip downstream?:
```
@deprecated(xxx)
def PROT_PRIVILEDGED = PROT_PRIVILEGED
```

src/main/scala/amba/apb/Protocol.scala
```diff
@@ -9,8 +9,8 @@ object APBParameters
   // These are all fixed by the AHB standard:
   val protBits  = 3
 
-  def PROT_PRIVILEDGED = UInt(1, width = protBits)
+  def PROT_PRIVILEGED  = UInt(1, width = protBits)
   def PROT_NONSECURE   = UInt(2, width = protBits)
   def PROT_INSTRUCTION = UInt(4, width = protBits)
-  def PROT_DEFAULT = PROT_PRIVILEDGED
+  def PROT_DEFAULT = PROT_PRIVILEGED
```
sequencer, line 15: Thanks for this typo fix:
However for the happiness of RocketChip downstream, would you mind adding a function alias to not break the rocket-chip downstream?:
```
@deprecated(xxx)
def PROT_PRIVILEDGED = PROT_PRIVILEGED
```

sequencer (changes requested):

sequencer:
Sorry, @michael-etzkorn I just found the author has rebased the PR, but there I do have some nitpick on this, since it needs a deprecation notice and type/function alias for the typo fix.

src/main/scala/amba/axi4/AsyncCrossing.scala
michael-etzkorn: I'm in favor of picking up client/manager terminology within RC for AXI as well. The other terms are offensive.

sequencer: See: https://developer.arm.com/documentation/ihi0022/e/AMBA-AXI3-and-AXI4-Protocol-Specification

michael-etzkorn: I am aware of the AXI spec's terminology. I disagree with it and think the new terminology is equally clear and non-controversial. The spec has no normative requirement on terminology use. Source / Client and Sink / Manager are valid and understandable regardless of protocol context.

sequencer: Basically the manager-client terminology was from https://ieeexplore.ieee.org/document/7851474, it is designed for coherency system.
> Manager-Client Pairing (MCP) as a unifying methodology for designing multi-tier coherence protocols by formally defining and limiting the interactions between levels within a coherence hierarchy to enable composition.

We can use master-slave terminology for TileLink see PR #2532. But for the manager-client terminology, it can only be applied to a coherency system, while AXI4 is not, even for ACE or CHI they are not designed based on the idea of MCP...

src/main/scala/amba/ahb/Protocol.scala
sequencer: I remember in the last dev meeting, we decided not maintain API stability so aggressively.

src/main/scala/amba/apb/Protocol.scala
sequencer: I remember in the last dev meeting, we decided not maintain API stability so aggressively.

sequencer (approved):

PR #2926, opened 2022-01-07
DCache prioritizes misaligned above guest/page fault

ingallsj:
**Related issue**: replaces https://github.com/chipsalliance/rocket-chip/pull/2921

**Impact**: functional change

**Release Notes**
PTW fault prioritization: if a PTW _doesn't_ finish with a valid PTE, then we cannot check if the memory address has side effects to take a misaligned exception.

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -634,14 +634,14 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
 
   val (wb_xcpt, wb_cause) = checkExceptions(List(
     (wb_reg_xcpt,  wb_reg_cause),
-    (wb_reg_valid && wb_ctrl.mem && io.dmem.s2_xcpt.ma.st, UInt(Causes.misaligned_store)),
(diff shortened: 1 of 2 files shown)


aswaterman (approved):

sammy17:
Hi, could you please explain what are the side-effects mentioned in this release notes? 
Can't we check whether the address is misaligned or not when we only know the address? 
Thank you in advance!

ingallsj:
> Hi, could you please explain what are the side-effects mentioned in this release notes?

Memory-Mapped I/O devices can have side-effects when accessed.  Some examples:
- setting or clearing a flag in another memory-mapped register location.
- sending or clearing an interrupt.
- flushing a cache block.

Crucially: accesses to memory-mapped I/O devices _with side-effects_ **cannot** be emulated by splitting them into one-byte-at-a-time accesses.

> Can't we check whether the address is misaligned or not when we only know the address?

Yes, we can, but Rocket distinguishes between an emulate-able misaligned access, which takes a misaligned fault for machine mode to emulate it, or a non-emulate-able misaligned access to a side-effect memory address, which takes an access fault to not be emulated.

PR #2836, opened 2021-05-12
Chisel decoder

sequencer:
This PR use new chisel API to replace original RocketChip Decoder.
Chisel Decoder depends on the explicit width, so I adds explicit width for most usages. This will be a breaking change to RC.
**Related issue**: 

**Type of change**: new feature

**Impact**: no functional change

**Development Phase**: implementation

**Release Notes**
Switch to Chisel Decoder

PR #2932, opened 2022-01-18
[DO NOT MERGE] NBDCache Broken with Chisel 3.5.0 Bump

abejgonzalez:
**Related issue**: 

**Impact**: API modification

**Release Notes**
@jackkoenig Here is the branch that you can use to debug the NBDCache issue. You can see the error when running `make` under `emulator/`.


jackkoenig:
This one is actually really difficult. The basic cause is around the "autowrap val io" in Chisel.Modules not being null-safe for when the autowrapping occurs before the val is initialized. In this case, https://github.com/chipsalliance/rocket-chip/blob/7bd6b1a3f3294caad5c4b2b610a237ad7cd315bf/src/main/scala/rocket/HellaCache.scala#L311, there is a function call that creates a `Wire` the line before `val io` (which means at that point, `val io` is `null`). This could be worked around by calling that function after val io, but that would require adding another `gen` parameter argument which might be a pain. I'm trying to see what the bugfix looks like in chisel3.

jackkoenig:
@sequencer we may not be able to bump rocket-chip to 3.5.0 anyway if this is too painful to work around. If you can get the fix for https://github.com/chipsalliance/rocket-chip/pull/2836 into chisel3 we may just cut a 3.5.1 soon, maybe next week.

sequencer:
OK! I will work on it this today.

abejgonzalez:
This should be solved with bumping to 3.5.1 so I'll close this for now.

PR #2933, opened 2022-01-20
Fix Fuzzer over-aggressively reusing source ids

jerryz123:
From the following excerpts of the tilelink spec:

> Each inflight request identifier of a channel in a particular link must be unique

>  An identifier is said to be inflight if an outstanding request using it has not yet received a response.

> A burst is said to be in progress after the first beat has been accepted and until the last beat has been accepted

If I interpret "received a response" to mean "the response burst is no longer in progress", then re-using source IDs before the response burst has completed is illegal, and thus the TLFuzzer is nonconformant.

**Impact**: functional change

**Release Notes**


jerryz123:
It seems like the identifier is considered non in-flight as soon as the first beat is received, so the Fuzzer is conformant then,

PR #2945, opened 2022-03-14
Fix false-positive FP data hazards

a0u:
**Related issue**: 

**Impact**:

**Release Notes**

Avoid false RAW/WAW hazards for integer instructions that access an `x` register whose numeric specifier coincides with the `f` register destination of an earlier floating-point instruction.

Example testcase:
```asm
fmv.w.x f10, zero
addi x10, x10, 1
```

It appears this phenomenon is not uncommon in compiled code since gcc's register allocator frequently tends to use `a0-a7` (`x10-x17`) and `fa0-fa7` (`f10-f17`) as temporaries.


a0u:
Another solution is to set `wen`/`ren1`/`ren2`/`ren3` in FPUCtrlSigs to `N` instead of `X` for non-floating-point instructions.

https://github.com/chipsalliance/rocket-chip/blob/b36060ebe7bd821ea4b4d1dd2aa9e6d38e7d01b4/src/main/scala/tile/FPU.scala#L59

aswaterman:
The original proposal is probably less HW than changing the decoder.

aswaterman (approved):

PR #2906, opened 2021-11-17
Zk: Scalar Cryptography Extension

phthinh:
**Related issue**: 

**Type of change**: other enhancement

**Impact**: no functional change 

**Development Phase**:  implementation

**Release Notes**

This pull request is to add the hardware implementation to support the cryptographic extensions **Zk** on the Rocket-chip. This supports the Scalar and Entropy Source Instructions specified in the v1.0.0-rc5 version. The implementation can work with 32-bit and 64-bit Architectures. The plug-in supports of the Zk extension can be enabled by using the WithZBK, WithZKN, WithZKS, and WithZKR configuration options.

Notes that: There is no random number/bit generator implemented for the pollentropy (aka. reading random seed value) instruction. But the implementation supports the related CSR registers (i.e., the seed and mseccfg CSRs). So, reading the seed CSR will return an invalid value (i.e., 0xFFFFDEAD). These registers should be modified to support an actual random generator.

The other instructions of the Zk extensions are fully implemented and tested on an FPGA platform against their function specified in the v1.0.0-rc5 version.

PR #2950, opened 2022-03-19
Zk(Zbk, Zkn, Zks)/Zb: Scalar Cryptography/Bitmanip Extension

ZenithalHourlyRate:
**Related PR**: #2906

**Development Phase**: implementation (needs cleanup)

**Note**
This has borrowed a lot of code (mainly decode logic) from #2906. The decode logic has been rebased on the master due to the merging of H extension.

Some of the logics are also inspired by #2906, thanks!

While #2906 has different implementations for rv32 and rv64, this PR merges them together.

As for Zkr, as #2906 is just a stub, I think a Chisel stub is fairly easy to implement. In the meantime, the CSR logics should be PRed by the owner of #2906 instead of me. Maybe users can just cherry-pick that commit and implement their own Zkr?

There are some notes on Zkt (like ROM optimization), but it is not verified. As this is a single-cycle implementation, Zkt ought to be satisfied (not verified though).

A lot of helpers are temply embedded here, and should be cleaned up. barrel.rightRotate is borrowed from [here](https://github.com/sequencer/arithmetic/tree/master/arithmetic/src/utils/shifter)

This has passed riscv-arch-test on [rv64i_m/K](https://github.com/riscv-non-isa/riscv-arch-test/tree/master/riscv-test-suite/rv64i_m/K) and [rv32i_m/K](https://github.com/riscv-non-isa/riscv-arch-test/tree/master/riscv-test-suite/rv32i_m/K) when verilated to an emulator

Cc @sequencer @phthinh @ben-marshall

Also thanks @liweiwei90 for his help on riscv-arch-test

**TODO**

 1. Clean up of the code on mergeing common datapaths and bad handling of some logics (like multiple levels of Mux of io.out)
 2. Clean up of some helpers (e.g. move them into zk/utils.scala) / get helpers upstreamed to Chisel
 3. Add CI tests on Zk for rocket-chip (How?)
 4. Test against FPGA

**Discussion**

1. Should we merge zbk into ALU as they share some logics (like `~io.rs2` and logical operation gates for andn/orn/xnor)
2. Should we embed SBox as ROM here or just write real-time computing logics, as #2906 and [Xiangshan](https://github.com/OpenXiangShan/XiangShan/blob/master/src/main/scala/xiangshan/backend/fu/Bku.scala) have done

**Release Notes**

This pull request is to add the hardware implementation to support the Zbk, Zkn, Zks of the cryptographic extensions Zk on the Rocket-chip. This supports the Scalar Instructions specified in the v1.0.1 version. The implementation can work with 32-bit and 64-bit Architectures. The plug-in supports can be enabled by using the WithZBK, WithZKN, WithZKS configuration options.


src/main/scala/zk/zbk.scala
```diff
@@ -0,0 +1,158 @@
...
+
+  // according to FN_xxx above
+  io.rd := VecInit(Seq(
+    shout, shout, shout,
+    andn, orn, xnor,
+    pack, packh,
+    brev8, rev8,
+    zip, unzip,
+    clmul, clmul,
+    xperm8, xperm4))(io.zbk_fn)
```
sequencer, line 157: bad design :(

src/main/scala/zk/zkn.scala
```diff
@@ -0,0 +1,419 @@
...
+    val out = Output(UInt(8.W))
+  })
+
+  // TODO some note on optimization: constant time look up!
+  io.out := VecInit(table.map(_.U(8.W)).toSeq)(io.in)
+}
+
+class AESSBox extends Module {
+  val io = IO(new Bundle {
+    val zkn_fn = Input(UInt(ZKN.FN_Len.W))
```
sequencer, line 186: 1-bit control signal.

src/main/scala/zk/zkn.scala
```diff
@@ -0,0 +1,419 @@
...
+}
+
+class SBox(table: Seq[Int]) extends Module {
+  val io = IO(new Bundle {
+    val in = Input(UInt(8.W))
+    val out = Output(UInt(8.W))
+  })
+
+  // TODO some note on optimization: constant time look up!
+  io.out := VecInit(table.map(_.U(8.W)).toSeq)(io.in)
```
sequencer, line 181: I'm not sure the ROM compiler can provide a good PPA here.
Need physical implementation.

src/main/scala/zk/zkn.scala
```diff
@@ -0,0 +1,419 @@
...
+  })
+
+  def m(x: UInt, y: Int): UInt = {
+    val m = Module(new GFMul(y))
+    m.io.in := x
+    m.io.out
+  }
+
+  val out = if (enc) Cat(m(io.in, 3), io.in, io.in, m(io.in, 2))
+    else Cat(m(io.in, 0xb), m(io.in, 0xd), m(io.in, 9), m(io.in, 0xe))
```
sequencer, line 253: refer to spec here plz...

src/main/scala/zk/zkn.scala
```diff
@@ -0,0 +1,419 @@
...
+  // aes
+  val aes = if (xLen == 32) {
+    val si = asBytes(io.rs2)(io.bs)
+    val so = {
+      val m = Module(new AESSBox)
+      m.io.in := si
+      m.io.zkn_fn := io.zkn_fn
+      m.io.out
+    }
+    val mixed_so = Mux(io.zkn_fn === ZKN.FN_AES_ES || io.zkn_fn === ZKN.FN_AES_DS, Cat(0.U(24.W), so), {
```
sequencer, line 303: OH

src/main/scala/zk/zkn.scala
sequencer: See https://github.com/riscv/riscv-crypto/blob/master/rtl/crypto-fu/riscv_crypto_fu_sboxes.v

src/main/scala/zk/zkn.scala
```diff
@@ -0,0 +1,419 @@
...
+    val sha512sum1_rs1  = (io.rs1 << 23) ^ (io.rs1 >> 14) ^ (io.rs1 >> 18)
+    val sha512sum1_rs2  = (io.rs2 >>  9) ^ (io.rs2 << 18) ^ (io.rs2 << 14)
+    sha512sum1_rs1 ^ sha512sum1_rs2
+  } else {
+    require(xLen == 64)
+    io.rs1.rotateRight(14) ^ io.rs1.rotateRight(18) ^ io.rs1.rotateRight(41)
+  }
+
+  // according to FN_xxx above
+  io.rd := VecInit(Seq(
```
sequencer, line 413: OH

ZenithalHourlyRate:
Progress report:

1. Implemented Zb in ALU
2. Moved part of Zkb into ALU
3. Refactored ALU

Note that this definitely can not compile now, and wont work even with syntax fixed as there is no decode nor control signal now.

src/main/scala/zk/zbk.scala
```diff
@@ -0,0 +1,158 @@
...
+  def FN_PACK  =  6.U(FN_Len.W)
+  def FN_PACKH =  7.U(FN_Len.W)
+  def FN_BREV8 =  8.U(FN_Len.W)
+  def FN_REV8  =  9.U(FN_Len.W)
+  def FN_ZIP   = 10.U(FN_Len.W)
+  def FN_UNZIP = 11.U(FN_Len.W)
+  def FN_CLMUL = 12.U(FN_Len.W)
+  def FN_CLMULH= 13.U(FN_Len.W)
+  def FN_XPERM8= 14.U(FN_Len.W)
+  def FN_XPERM4= 15.U(FN_Len.W)
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Will redesign based on the refactored ALU when implementing the decode logic

src/main/scala/zk/zbk.scala
```diff
@@ -0,0 +1,158 @@
...
+class ZBKInterface(xLen: Int) extends Bundle {
+  val zbk_fn = Input(UInt(ZBK.FN_Len.W))
+  val dw     = Input(Bool())
+  val valid  = Input(Bool())
+  val rs1    = Input(UInt(xLen.W))
+  val rs2    = Input(UInt(xLen.W))
+  val rd     = Output(UInt(xLen.W))
+}
+
+class ZBKImp(xLen: Int) extends Module {
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Merged some of them now. CLMUL/XPERM is anticipated to be implemented in a multi-cycle style, but I would work on this after I had correctly implemented the single cycle implementation (i.e. tested the refactored ALU)

src/main/scala/zk/zbk.scala
```diff
@@ -0,0 +1,158 @@
...
+    else {
+      require(xLen == 64)
+      Mux(io.dw, shout_raw, sext(shout_raw(31,0)))
+    }
+
+  // bool
+  val rs2n = ~io.rs2
+  val andn = io.rs1 & rs2n
+  val orn = io.rs1 | rs2n
+  val xnor = io.rs1 ^ rs2n
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Done

ZenithalHourlyRate:
Review in RC working group suggested putting the refactored ALU into a separate file (maybe `ABLU.scala`?) instead of directly changing it. and use an option (e.g. `withZB` or `withExperimentalALU`) to switch between old and new implementations. Only if the refactored one turns out to be better after extensive test could it replace the old ALU.

ZenithalHourlyRate:
Now this passes the Zk arch-test and the [Zb arch-test](https://github.com/riscv-non-isa/riscv-arch-test/pull/228), it could be used now, but clean-up is needed. More specifically, options should be added so that users can choose whether to use the new ABLU.

Among the new commits there are cherry-picked commits from #2956 (otherwise I cannot write decode logics). As I understand, that PR should be reworked after [the PR on riscv-opcodes](https://github.com/riscv/riscv-opcodes/pull/106) gets merged.

Phantom1003:
Hi, we recently tested this commit using our hardware fuzzer, the arithmetic function is consistent with the spike implementation, nice work!
We detected some mismatches in some corner cases:
- Rocket has a custom instruction extension, the mstatus XS bit is always enabled. Our co-simulate environment was forced to disable the original custom extension and found that the K extension do not enable this bit.
https://github.com/chipsalliance/rocket-chip/blob/045d03c54bce7636401b1b4b17d6a3677356dfe0/src/main/scala/rocket/CSR.scala#L598-L603
- The current design always enables some B extension. When disable zbb extension, spike throws an exception on rev8 while rocket executes it as a normal instruction.
```
spike config: core 1, isa: rv32gc_xdummy MSU vlen:128,elen:64
...
[rocket] 3 0x80000182 (0x698dd493) x 9 0x1a3e7d6a
[spike] core   0: 0x80000182 (0x698dd493) rev8    s1, s11
[spike] core   0: exception trap_illegal_instruction, epc 0x80000182
[spike] core   0:           tval 0x698dd493
[spike] core   0: 0x80000004 (0x34202f73) csrr    t5, mcause
[error] PC SIM 0000000080000004 , DUT 0000000080000182
[error] INSN SIM 34202f73 , DUT 698dd493
[CJ] Commit Failed
```
[kext-0.zip](https://github.com/chipsalliance/rocket-chip/files/8543061/kext-0.zip)
- As described in the spec, aes64ks1i should throw an illegal instruction when rcon is bigger than 0xA. Rocket did not throw an illegal instruction exception as spike did.
```
spike config: core 1, isa: rv64gc_xdummy_zk_zkn_zks MSU vlen:128,elen:64
...
[rocket] 0 0x0080000178 (0x310d1413) x 8 0x1616161716161617
[spike] core   0: 0x0000000080000178 (0x310d1413) aes64ks1i s0, s10, 0x0
[rocket] 0 0x008000017c (0x31ff9493) x 9 0x6330636363306363
[spike] core   0: 0x000000008000017c (0x31ff9493) aes64ks1i s1, t6, 0xf
[spike] core   0: exception trap_illegal_instruction, epc 0x000000008000017c
[spike] core   0:           tval 0x0000000000000000
[spike] core   0: 0x0000000080000004 (0x34202f73) csrr    t5, mcause
[error] PC SIM 0000000080000004 , DUT 000000008000017c
[error] INSN SIM 34202f73 , DUT 31ff9493
[CJ] Commit Failed
```
[kext-1.zip](https://github.com/chipsalliance/rocket-chip/files/8543062/kext-1.zip)

ZenithalHourlyRate:
Thanks for your attention on this PR!

> * The current design always enables some B extension. When disable zbb extension, spike throws an exception on rev8 while rocket executes it as a normal instruction.

Yes that's what "but clean-up is needed" in my [last comment](https://github.com/chipsalliance/rocket-chip/pull/2950#issuecomment-1101660951) meant. You can see that I have manually enabled both Zb/Zk extensions [for this commit](https://github.com/OpenRigil/rocket-chip/blob/18927be0a4f66e929e5dc2400a941b7903158906/src/main/scala/rocket/RocketCore.scala#L187-L193). I will add a new commit for new options soon. Actually, if these hard-encoded "true" are all turned to "false" then instructions like rev8 will cause illegal instruction exception.

> * As described in the spec, aes64ks1i should throw an illegal instruction when rcon is bigger than 0xA. Rocket did not throw an illegal instruction exception as spike did.

Thanks for detecting this! Yes I have noticed this when reading the spec but as I was following the IO interface of #2906, I did not implement this mechanism. I will fix this issue later.

> * found that the K extension do not enable or check this bit.

By the privileged spec (20211203), 

> In systems without additional user extensions **requiring new state**, the XS field is read-only zero. Every additional extension **with state** provides a CSR field that encodes the equivalent of the XS.

I think Zk/Zb does not have states. Even though Zkr added a `seed` CSR, I do not think that is a state. Also this PR has not implemented Zkr. I am not so sure about this, correct me if there is any misunderstanding.

Phantom1003:
Sorry for the confusion. 
Since misa is writable, for the extended instruction set we need to detect whether the corresponding extension is turned on during execution, which means that there are two places to be checked, one is the misa register and the other is XS bit. 
The current implementation seems unable to turn off these extended.

ZenithalHourlyRate:
> Since misa is writable, for the extended instruction set we need to detect whether the corresponding extension is turned on during execution

AFAIK, there is no `misa` bits for Zb/Zk extension, thus the detection and enabling/disabling of Zb/Zk can not be done via `misa` CSR. Currently I do not know how to detect whether the platform supports Zb/Zk in software, please tell me if anyone knows that.

> and the other is XS bit.

XS bits are also not related to Zb/Zk as Zb/Zk has no CPU state that needs to be saved by supervisor; Zb/Zk also does not define a CSR field that "encodes the qeuivalent of the XS states" as the spec said.

ZenithalHourlyRate:
I've tried to synthesis the PR using vendor tools against [nangate45](https://github.com/The-OpenROAD-Project/OpenROAD-flow-scripts/tree/master/flow/platforms/nangate45). According to the synthesis result I've made changes to SBox implementation.

## Area

At commit [18927be](https://github.com/chipsalliance/rocket-chip/pull/2950/commits/18927be0a4f66e929e5dc2400a941b7903158906) with RV64, the area is

|Hierarchical cell|Absolute Total|
|-|-|
|Rocket|80570|
|ablu|4306|
|div|8015|
|zbk|7612|
|zkn|19378|
|zkn/aes_so_m|1899|
|zkn/aes_so_m/dec|950|
|zkn/aes_so_m/enc|923|
|zks|1471|
|zks/so_m|934|
 
Note that for RV64, there are 16 AES SBoxes (8 Enc, 8 Dec) and one SM4 Sbox, each takes about 950 area. This is quite unbearable as zkn takes around 1/4 of the total area. As a comparison, `div` only takes 8000.

Then I added [8240ffd](https://github.com/chipsalliance/rocket-chip/pull/2950/commits/8240ffd06add867661e496e966091876198d9375) which is inspired from [riscv-crypto/rtl](https://github.com/riscv/riscv-crypto/blob/master/rtl/crypto-fu/riscv_crypto_fu_sboxes.v) and the area result is

|Hierarchical cell|Absolute Total|
|-|-|
|Rocket|67377|
|ablu|4309|
|div|8015|
|zbk|7617|
|zkn|6804|
|zkn/aes_so_m|323|
|zks|709|

After this commit `zkn` only takes nearly the same area as `div`, which is acceptable for RV64 IMHO. Note that for RV32 there is only three SBox (AES enc/dec, SM4) and their `AESMid` can be reused (marked in TODO but not done yet), which is also acceptable.

Originally RC working group had suggested reducing the number of Sboxes and reuse them across multiple cycles. I think the current implementation is acceptable for extremely high crypto performance when using RV64 (RV32 does not have this single/multi-cycle trade-off). I can work on reducing the number by two  (half the area), just similar to the `SCIEPipelined`, but I want to know if it is worth the time. I agree that providing a parameter (do one encryption in 1, 2, 4, 8 cycle) is more flexible for end users, but this complicates the design and needs a lot of effort.

As a reference, this is the area report of b36060ebe

|Hierarchical cell|Absolute Total|
|-|-|
|Rocket|49154|
|alu|1721|
|div|8015|

ABLU doubles the area of the original ALU. Note that ABLU does not contain `clmul` and `xperm` (in zbk above) and the result showed that they are quite costly (two operations along are bigger than a whole ABLU).

## Timing

The timing part is a tricky part. In the original Rocket, the critical path actually lies in `div` part. But in this PR the ABLU becomes the critical path. The slack for `div` is at most 2.6 when synthesis to 100MHz but the slack for `ablu` is at 1.25~1.8

I doubt that this bad timing has to do with the `pla` in ABLU. Actually we have done two decoding here: First IDecode from instruction to FN; then FN to ctrl signals. This should all be done in Decode stage. I will experiment the timing result when that `pla` is moved into Decode stage.

However, decoding all the ctrl signals in Decode stage will increase the footprint of registers (e.g. ex_ctrl). Since the original FN is only 4 bits but the ctrl signals for ABLU along is about 40 bits, it is worth arguing whether we should do it.

Another thing to note is that the timing of the decode stage is not so good either.

Surprisingly, ZKN/ZKS/ZBK does not have timing issue, at least not as bad as `div`. It is worth noting that we originally thought clmul/xperm need to be pipelined as it is deep (64 xor) but the timing report does not have any complaint about them. Based on this result I will merge ZBK into ABLU; ZBK was originally separated for a multi-cycle implementation.

## Cleanup

I've added some configs `withABLU`, `withBitManip`, `withBitManipCrypto`, `withCryptoNIST` and `withCryptoSM`, thus @Phantom1003 can disable an extension now. 

Note that the `misa` string now will not contain "B".

There are still further cleanups like merging the decode table, upstreaming some utils and fixing issues from Phantom1003.

ZenithalHourlyRate:
> I think the current implementation is acceptable for extremely high crypto performance when using RV64

I've made a benchmark for RV64 at https://gist.github.com/ZenithalHourlyRate/7b5175734f87acb73d0bbc53391d7140 using this PR against my software AES implementation in OpenSSL. It turns out that the result is considerable but not too high (up to 9.5X for aes-256-ecb). As the assembly snippet has shown, `aes64esm` are often executed consecutively, thus if a multi-cycle version was implemented, we would need to stall the pipeline most of the time. I think the single cycle implementation is acceptable for now (both area and latency). Multi-cycle version shall be done in future PRs if there is indeed this usage.

I'll work on fixing the issue discovered by Phantom1003 and the timing issue (pla decode) and cleaning up, further architectural changes on crypto would not probably be made in this PR.

ZenithalHourlyRate:
With the `Add aes64ks1i exception` commit the issue @Phantom1003 mentioned should be fixed.

I have run a sample program using `spike pk test` and `emulator pk test`, the result is consistent
```
# spike
core   0: 0x0000000000011480 (0x31069713) aes64ks1i a4, a3, 0
core   0: 0x0000000000011484 (0x7e670333) aes64ks2 t1, a4, t1 
core   0: 0x0000000000011488 (0x7e7303b3) aes64ks2 t2, t1, t2 
core   0: 0x000000000001148c (0x02060613) addi    a2, a2, 32 
core   0: 0x0000000000011490 (0x00663023) sd      t1, 0(a2)
core   0: 0x0000000000011494 (0x00763423) sd      t2, 8(a2)
core   0: 0x0000000000011498 (0x31c39713) aes64ks1i a4, t2, 12 
core   0: exception trap_illegal_instruction, epc 0x0000000000011498
core   0:           tval 0x0000000000000000
core   0: 0x0000000080000004 (0x34011173) csrrw   sp, mscratch, sp
core   0: 0x0000000080000008 (0x1a010863) beqz    sp, pc + 432
# emulator
C0:    1753565 [1] pc=[0000000000011480] W[r14=6363636263636362][1] R[r13=0000000000000000] R[r16=0000000000000000] inst=[31069713] aes64ks1i a4, a3, 0
C0:    1753566 [1] pc=[0000000000011484] W[r 6=6363636263636362][1] R[r14=6363636263636362] R[r 6=0000000000000000] inst=[7e670333] aes64ks2 t1, a4, t1 
C0:    1753567 [1] pc=[0000000000011488] W[r 7=6363636263636362][1] R[r 6=6363636263636362] R[r 7=0000000000000000] inst=[7e7303b3] aes64ks2 t2, t1, t2 
C0:    1753568 [1] pc=[000000000001148c] W[r12=0000003ffffff928][1] R[r12=0000003ffffff908] R[r 0=0000000000000000] inst=[02060613] addi    a2, a2, 32 
C0:    1753569 [1] pc=[0000000000011490] W[r 0=0000000000000000][0] R[r12=0000003ffffff928] R[r 6=6363636263636362] inst=[00663023] sd      t1, 0(a2)
C0:    1753570 [1] pc=[0000000000011494] W[r 0=0000000000000000][0] R[r12=0000003ffffff928] R[r 7=6363636263636362] inst=[00763423] sd      t2, 8(a2)
C0:    1753571 [0] pc=[0000000000011498] W[r14=0000000000000000][0] R[r 7=0000000031c39713] R[r28=0000000000000006] inst=[31c39713] aes64ks1i a4, t2, 12 
C0:    1753598 [1] pc=[0000000080000004] W[r 2=0000000080011ec0][1] R[r 2=0000003ffffff8f0] R[r 0=0000000000000000] inst=[34011173] csrrw   sp, mscratch, sp
C0:    1753601 [1] pc=[0000000080000008] W[r 0=0000000000000000][0] R[r 2=0000000080011ec0] R[r 0=0000000000000000] inst=[1a010863] beqz    sp, pc + 432

```

src/main/scala/rocket/RocketCore.scala
```diff
@@ -119,6 +124,7 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
   val gated_clock =
     if (!rocketParams.clockGate) clock
     else ClockGate(clock, clock_en, "rocket_clock_gate")
+  val ALUClass = if (usingABLU) ABLU else ALU
```
sequencer, line 127: use `AbstractAlu` here.

src/main/scala/rocket/CryptoNIST.scala
```diff
@@ -0,0 +1,343 @@
...
+
+  def rightShift[T <: Data](inputs: Vec[T], shift: UInt, layerSize: Int = 1): Vec[T] =
+    apply(inputs, shift, RightShift, layerSize)
+
+  def leftRotate[T <: Data](inputs: Vec[T], shift: UInt, layerSize: Int = 1): Vec[T] =
+    apply(inputs, shift, LeftRotate, layerSize)
+
+  def rightRotate[T <: Data](inputs: Vec[T], shift: UInt, layerSize: Int = 1): Vec[T] =
+    apply(inputs, shift, RightRotate, layerSize)
+}
```
sequencer, line 70: since chipsalliance/chisel3#2518 merged, you should split these lines into a separate file, add note saying: use `chisel3.stdlib.BarrelShifter` after bumping to Chisel 3.6

src/main/scala/rocket/CryptoNIST.scala
```diff
@@ -0,0 +1,343 @@
...
+  def FN_AES_KS1     =  5.U(FN_Len.W)
+  def FN_AES_KS2     =  6.U(FN_Len.W)
+  def FN_SHA256_SIG0 =  7.U(FN_Len.W)
+  def FN_SHA256_SIG1 =  8.U(FN_Len.W)
+  def FN_SHA256_SUM0 =  9.U(FN_Len.W)
+  def FN_SHA256_SUM1 = 10.U(FN_Len.W)
+  def FN_SHA512_SIG0 = 11.U(FN_Len.W)
+  def FN_SHA512_SIG1 = 12.U(FN_Len.W)
+  def FN_SHA512_SUM0 = 13.U(FN_Len.W)
+  def FN_SHA512_SUM1 = 14.U(FN_Len.W)
```
sequencer, line 88: For a better PPA result. I doubt these MicroOp need more careful designing, but for now(functionally work), it LGTM.

src/main/scala/util/SBox.scala
```diff
@@ -0,0 +1,344 @@
...
+    Proc. EUROCRYPT '93, LNCS 765, Springer, pp. 55-64 (1993)
+    DOI: https://doi.org/10.1007/3-540-48285-7_6
+*/
+
+package freechips.rocketchip.util
+
+import chisel3._
+
+// top (inner) linear layer for AES
+object SBoxAESEncIn {
```
sequencer, line 25: I really like the SBox impl here! wondering if is possible to upstream this to `chisel.stdlib`.

src/main/scala/util/SBox.scala
```diff
@@ -0,0 +1,344 @@
+// See LICENSE.SiFive for license details.
```
sequencer, line 1: Actually this file is under the Apache license(not SiFive), although SiFive is Apache as well. 
```suggestion
// SPDX-License-Identifier: Apache-2.0
```

src/main/scala/rocket/CryptoSM.scala
```diff
@@ -0,0 +1,62 @@
+// See LICENSE.SiFive for license details.
```
sequencer, line 1: Actually this file is under the Apache license(not SiFive), although SiFive is Apache as well. 
```suggestion
// SPDX-License-Identifier: Apache-2.0
```

src/main/scala/rocket/CryptoNIST.scala
```diff
@@ -0,0 +1,343 @@
+// See LICENSE.SiFive for license details.
```
sequencer, line 1: Actually this file is under the Apache license(not SiFive), although SiFive is Apache as well. 
```suggestion
// SPDX-License-Identifier: Apache-2.0
```

src/main/scala/rocket/ABLU.scala
```diff
@@ -0,0 +1,280 @@
+// See LICENSE.SiFive for license details.
+// See LICENSE.Berkeley for license details.
```
sequencer, line 2: Actually this file is under the Apache license(not SiFive), although SiFive is Apache as well. 
```suggestion
// SPDX-License-Identifier: Apache-2.0
```

src/main/scala/rocket/RocketCore.scala
```diff
@@ -313,7 +327,8 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
     id_ctrl.rocc && csr.io.decode(0).rocc_illegal ||
     id_ctrl.scie && !(id_scie_decoder.unpipelined || id_scie_decoder.pipelined) ||
     id_csr_en && (csr.io.decode(0).read_illegal || !id_csr_ren && csr.io.decode(0).write_illegal) ||
-    !ibuf.io.inst(0).bits.rvc && (id_system_insn && csr.io.decode(0).system_illegal)
+    !ibuf.io.inst(0).bits.rvc && (id_system_insn && csr.io.decode(0).system_illegal) ||
+    id_ctrl.zkn && id_ctrl.alu_fn(ZKN.FN_Len-1,0) === ZKN.FN_AES_KS1 && id_inst(0)(23,20) > 0xA.U(4.W)
```
sequencer, line 331: `id_ctrl.alu_fn(ZKN.FN_Len-1,0) === ZKN.FN_AES_KS1` needs additional comparing logic, which should be better to wire a `isAESKS1` signal out form the decoder.
you can add a TODO here.

src/main/scala/tile/BaseTile.scala
```diff
@@ -103,9 +103,8 @@ trait HasNonDiplomaticTileParameters {
     val f = if (tileParams.core.fpu.nonEmpty) "f" else ""
     val d = if (tileParams.core.fpu.nonEmpty && tileParams.core.fpu.get.fLen > 32) "d" else ""
     val c = if (tileParams.core.useCompressed) "c" else ""
-    val b = if (tileParams.core.useBitManip) "b" else ""
     val v = if (tileParams.core.useVector) "v" else ""
-    s"rv${p(XLen)}$ie$m$a$f$d$c$b$v"
+    s"rv${p(XLen)}$ie$m$a$f$d$c$v"
```
sequencer, line 107: Why?

ZenithalHourlyRate: There is no "b" extension

src/main/scala/rocket/RocketCore.scala
ZenithalHourlyRate: I'll redesign the uop into control signals just as ABLU had done.

src/main/scala/util/SBox.scala
ZenithalHourlyRate: Will do so

src/main/scala/tile/BaseTile.scala
sequencer: haha! I know.

src/main/scala/rocket/CryptoNIST.scala
ZenithalHourlyRate: > since [chipsalliance/chisel3#2518](https://github.com/chipsalliance/chisel3/pull/2518) merged, you should split these lines into a separate file, add note saying: use `chisel3.stdlib.BarrelShifter` after bumping to Chisel 3.6

Done

src/main/scala/rocket/RocketCore.scala
jerryz123: Does `useABLU` need to be its own parameter? Should `useABLU = useBitManip || useBitManipCrypto || useCryptoNIST || useCryptoSM`

ZenithalHourlyRate: > Should `useABLU = useBitManip || useBitManipCrypto || useCryptoNIST || useCryptoSM`

Yes it is but conditioning on this everywhere is not maintainable.

I think the `useABLU` in `CoreParams` is not so necessary (I thought there might be some one just trying to use the ABLU alone, but yes generally it is not the case) but `usingABLU` in `HasCoreParameters` is useful for internal usage

jerryz123: You don't need to condition on this everywhere. We shouldn't expose `useABLU` as a parameter in `RocketCoreParams` if its actually a derived field.
You can set `val useABLU = ...` in the body of `RocketCoreParams`

ZenithalHourlyRate: > You can set `val useABLU = ...` in the body of `RocketCoreParams`

Fixed

ZenithalHourlyRate:
Merged master, redesigned control signals and did some cleanup.

The `IDecode: use DW_XPR for branch insts` needs special attention as it is a regression introduced by the new decoder. See more in its commit message.

We have many things to test, but the test framework is not ready (#2991), like both RV32 and RV64, enabling both Zk/Zb, enabling only Zk and enabling only Zb. For `aes64ks1i`, the illegal instruction exception codepath should also be covered in CI, it seems that arch-test has not tested that.

I also added some `if` condition to prevent instantiating some datapath when Zb is enabled but Zbk is not.

I mark this PR as ready for review now. Some utils that could be upstreamed to Chisel, if merged, still need a longer time to hit rocket chip so I assume adding them here and removing them in future PRs is appropriate.

src/main/scala/rocket/CryptoNIST.scala
ZenithalHourlyRate: Done in the same way as ABLU.

src/main/scala/rocket/RocketCore.scala
ZenithalHourlyRate: Now it uses one signal from fn: `id_ctrl.alu_fn(15)`.

The comparing logic for `id_inst(0)(23,20) > 0xA.U(4.W)` could not be avoided, though.

src/main/scala/rocket/RocketCore.scala
```diff
@@ -568,7 +607,7 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
     mem_reg_mem_size := ex_reg_mem_size
     mem_reg_hls_or_dv := io.dmem.req.bits.dv
     mem_reg_pc := ex_reg_pc
-    mem_reg_wdata := Mux(ex_scie_unpipelined, ex_scie_unpipelined_wdata, alu.io.out)
+    mem_reg_wdata := Mux(ex_scie_unpipelined, ex_scie_unpipelined_wdata, Mux(ex_ctrl.zbk,ex_zbk_wdata, Mux(ex_ctrl.zkn,ex_zkn_wdata, Mux(ex_ctrl.zks,ex_zks_wdata,alu.io.out))))
     mem_br_taken := alu.io.cmp_out
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Now it is Mux1H.

src/main/scala/zk/zbk.scala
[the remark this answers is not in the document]
ZenithalHourlyRate: Done

src/main/scala/zk/zbk.scala
```diff
@@ -0,0 +1,158 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.zk
+
+import chisel3._
+import chisel3.util._
+
+object ZBK {
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Done

src/main/scala/zk/zkn.scala
```diff
@@ -0,0 +1,419 @@
...
+
+  def rightShift[T <: Data](inputs: Vec[T], shift: UInt, layerSize: Int = 1): Vec[T] =
+    apply(inputs, shift, RightShift, layerSize)
+
+  def leftRotate[T <: Data](inputs: Vec[T], shift: UInt, layerSize: Int = 1): Vec[T] =
+    apply(inputs, shift, LeftRotate, layerSize)
+
+  def rightRotate[T <: Data](inputs: Vec[T], shift: UInt, layerSize: Int = 1): Vec[T] =
+    apply(inputs, shift, RightRotate, layerSize)
+}
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Upstreamed. It is in the `util` folder now and would be removed when rocket-chip has migrated to chisel 3.6.0.

src/main/scala/zk/zkn.scala
ZenithalHourlyRate: Done

src/main/scala/zk/zkn.scala
ZenithalHourlyRate: Done

src/main/scala/zk/zks.scala
```diff
@@ -0,0 +1,96 @@
...
+  // sm3
+  val r1 = io.rs1(31,0)
+  val sm3 = sext(Mux(io.zks_fn === ZKS.FN_SM3P0,
+    r1 ^ r1.rotateLeft(9) ^ r1.rotateLeft(17),
+    r1 ^ r1.rotateLeft(15) ^ r1.rotateLeft(23)))
+
+  // according to FN_xxx above
+  io.rd := VecInit(Seq(
+    sm4, sm4,
+    sm3, sm3))(io.zks_fn)
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Mux1H now

src/main/scala/zk/zbk.scala
```diff
@@ -0,0 +1,158 @@
...
+    }
+  val packh = Cat(0.U((xLen-16).W), io.rs2(7,0), io.rs1(7,0))
+
+  // rev
+  val rs1_bytes = asBytes(io.rs1)
+  val brev8 = VecInit(rs1_bytes.map(Reverse(_)).toSeq).asUInt
+  val rev8 = VecInit(rs1_bytes.reverse.toSeq).asUInt
+
+  // zip
+  val unzip = if (xLen == 32) {
```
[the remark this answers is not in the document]
ZenithalHourlyRate: Do not know if `Option` works with `Mux1H`. In the end we still need `0.U` when we are `getOrElse`

src/main/scala/zk/zbk.scala
```diff
@@ -0,0 +1,158 @@
...
+      rs1_nibbles(x)
+    }
+  ).toSeq).asUInt
+
+  // clmul
+  val clmul_rs1 = Mux(io.zbk_fn === ZBK.FN_CLMUL, io.rs1, Reverse(io.rs1))
+  val clmul_rs2 = Mux(io.zbk_fn === ZBK.FN_CLMUL, io.rs2, Reverse(io.rs2))
+  val clmul_raw = clmul_rs2.asBools.zipWithIndex.map({
+    case (b, i) => Mux(b, clmul_rs1 << i, 0.U)
+  }).reduce(_ ^ _)(xLen-1,0)
```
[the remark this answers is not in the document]
ZenithalHourlyRate: In my config timing turned out to be acceptable. Critical path is not here.

src/main/scala/rocket/ABLU.scala
```diff
@@ -0,0 +1,278 @@
...
+    exth,
+    extb,
+    //
+    rev8,
+    orc_brev8,
+    pack,
+    packh,
+    //
+    zip,
+    unzip))
```
sequencer, line 269: Seems to be a high fanout `out`...
But I don't have so much idea here to provide better circuit performance...

src/main/scala/rocket/ALU.scala
```diff
@@ -26,6 +25,46 @@ object ALU
...
+  def FN_BCLR     = UInt(0)
+  def FN_BEXT     = UInt(0)
+  def FN_BINV     = UInt(0)
+  def FN_BSET     = UInt(0)
+  // Zbk
+  def FN_BREV8    = UInt(0)
+  def FN_PACK     = UInt(0)
+  def FN_PACKH    = UInt(0)
+  def FN_ZIP      = UInt(0)
+  def FN_UNZIP    = UInt(0)
```
sequencer, line 66: why these are all 0?

ZenithalHourlyRate: > why these are all 0?

Because ALU does not use them, and ABLU has different encoding for them

ZenithalHourlyRate:
FYI, this PR is part of my undergraduate thesis (in Chinese) and I have made it public in my blog. Any one interested could find it [here](https://blog.zenithal.me/2022/06/25/%E6%9C%AC%E7%A7%91%E6%AF%95%E4%B8%9A%E8%AE%BA%E6%96%87-%E5%9F%BA%E4%BA%8E-Chisel-%E5%AE%9E%E7%8E%B0-RISC-V-%E6%8C%87%E4%BB%A4%E9%9B%86%E6%A0%87%E9%87%8F%E5%AF%86%E7%A0%81%E5%AD%A6%E6%89%A9%E5%B1%95/) and there are two presentations (in English) on that thesis. I thought that presentation (and the figures in it) could be a documentation.

sequencer (approved):
Good job!

src/main/scala/rocket/RocketCore.scala
```diff
@@ -25,6 +25,11 @@ case class RocketCoreParams(
   useCompressed: Boolean = true,
   useRVE: Boolean = false,
   useSCIE: Boolean = false,
+  useBitManip: Boolean = false,
+  useBitManipCrypto: Boolean = false,
+  useABLU: Boolean = false,
```
jerryz123, line 30: Why would user want `useBitManip && !useABLU`. Can we make `useABLU` derived?

jerryz123: Actually is this still used when there is using ABLU

src/main/scala/tile/Core.scala
```diff
@@ -22,9 +22,12 @@ trait CoreParams {
   val useAtomics: Boolean
   val useAtomicsOnlyForIO: Boolean
   val useCompressed: Boolean
-  val useBitManip: Boolean = false
+  val useBitManip: Boolean
+  val useBitManipCrypto: Boolean
```
jerryz123, line 26: Does anything else in `BaseTile/BaseCore` use `useBitManipCrypto/useCryptoNIST/useCryptoSM`? maybe those should be rocket-core only params

src/main/scala/tile/BaseTile.scala
```diff
@@ -103,10 +103,9 @@ trait HasNonDiplomaticTileParameters {
     val f = if (tileParams.core.fpu.nonEmpty) "f" else ""
     val d = if (tileParams.core.fpu.nonEmpty && tileParams.core.fpu.get.fLen > 32) "d" else ""
     val c = if (tileParams.core.useCompressed) "c" else ""
-    val b = if (tileParams.core.useBitManip) "b" else ""
     val v = if (tileParams.core.useVector) "v" else ""
     val x = tileParams.core.customIsaExt.map(s => s"_$s").getOrElse("")
-    s"rv${p(XLen)}$ie$m$a$f$d$c$b$v$x"
+    s"rv${p(XLen)}$ie$m$a$f$d$c$v$x"
```
jerryz123, line 108: Shouldn't `bitmanip` be reflected somehow in the dts? I don't know what the exact correct isa string composiiton should be, but it should be implemented here.

jerryz123 (commented):
I need some time to review this. Other cores depend on this stuff, and I need to think about making sure migration is not too painful (and does not cause bugs).

sequencer:
> Other cores depend on this stuff, and I need to think about making sure migration is not too painful (and does not cause bugs).

Agree! But notice eventually we should split rocket into a single package that only publish RocketCore as public api. For other cores depending on internal modules inside core will be deprecated.

src/main/scala/rocket/RocketCore.scala
ZenithalHourlyRate: Removed `useABLU`.

src/main/scala/tile/Core.scala
ZenithalHourlyRate: > maybe those should be rocket-core only params

useABLU is a rocket only param, but these params are not. They correspond to ratified isa extensions and could be reused by other cores instead of each core defining their own parameters.

Also, making these params private to rocket params will cause `isaDTS` not able to find corresponding options (unless with type cast).

I do know this will cause pain for other cores to migrate to new CoreParams. One solution is to provide default value and override them in RocketCoreParams.

src/main/scala/tile/BaseTile.scala
ZenithalHourlyRate: Added.

src/main/scala/tile/Core.scala
jerryz123: Ok in a past commit, isaDTS did not read these params. You have fixed this now.

The migration burden is small here, I think this is currently ok.

src/main/scala/rocket/ALU.scala
```diff
@@ -27,6 +26,46 @@ object ALU
   def FN_SLTU = 14.U
   def FN_SGEU = 15.U
 
+  // from Zb
+  // Zba: UW is encoded here becuase it is DW_64
+  def FN_ADDUW    = 0.U
```
jerryz123, line 31: Can you make these throw an exception if they are accessed? If `ALU` is selected instead of `ABLU`, then no code should try to access `ALU.FN_ADDUW` right?

src/main/scala/rocket/ALU.scala
```diff
@@ -8,8 +8,7 @@ import chisel3.util.{BitPat, Fill, Cat, Reverse}
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.tile.CoreModule
 
-object ALU
-{
+trait ALUFN {
```
jerryz123, line 11: Can we make this:
```
class ALUFN
...
class ABLUFN extends ALUFN
```

Get rid of the companion object.

src/main/scala/rocket/ALU.scala
```diff
@@ -47,17 +89,19 @@ object ALU
 
 import ALU._
 
-class ALU(implicit p: Parameters) extends CoreModule()(p) {
+trait HasALUIO extends Module with HasRocketCoreParameters {
   val io = IO(new Bundle {
     val dw = Input(UInt(SZ_DW.W))
-    val fn = Input(UInt(SZ_ALU_FN.W))
+    val fn = Input(UInt(if (usingABLU) ABLU.SZ_ALU_FN.W else SZ_ALU_FN.W))
```
jerryz123, line 95: See comment on `abstract class AbstractALU`

src/main/scala/rocket/ALU.scala
```diff
@@ -47,17 +89,19 @@ object ALU
 
 import ALU._
 
-class ALU(implicit p: Parameters) extends CoreModule()(p) {
+trait HasALUIO extends Module with HasRocketCoreParameters {
```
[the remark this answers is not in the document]
jerryz123: You can do `abstract class AbstractALU(fn: ALUFN)` here

src/main/scala/rocket/ABLU.scala
```diff
@@ -0,0 +1,278 @@
...
+
+  override def FN_MUL    = FN_ADD
+  override def FN_MULH   = FN_SL
+  override def FN_MULHSU = FN_SEQ
+  override def FN_MULHU  = FN_SNE
+}
+
+class ABLU(implicit p: Parameters) extends CoreModule()(p) with HasALUIO {
+  // note that it is reversed
+  val isSub :: isIn2Inv :: isZBS :: isUW :: Nil = io.fn(25,22).asBools
```
jerryz123, line 82: Can these be done in `ABLUFN.isSub`?

jerryz123: IMO all of this stuff which depends on the func bit arrangements should be in `ABLUFN`

src/main/scala/rocket/CSR.scala
```diff
@@ -594,7 +594,6 @@ class CSRFile(
     (if (fLen >= 32) "F" else "") +
     (if (fLen >= 64) "D" else "") +
     (if (usingVector) "V" else "") +
-    (if (usingBitManip) "B" else "") +
```
jerryz123, line 597: Instead of deleting this, can you comment out a line, and add that the current spec does not define what sub-extensions constitute the 'B' misa bit?

src/main/scala/rocket/Multiplier.scala
```diff
@@ -6,10 +6,12 @@ package freechips.rocketchip.rocket
 import chisel3._
 import chisel3.util.{Cat, log2Up, log2Ceil, log2Floor, Log2, Decoupled, Enum, Fill, Valid, Pipe}
 import freechips.rocketchip.util._
-import ALU._
+import freechips.rocketchip.config.Parameters
 
-class MultiplierReq(dataBits: Int, tagBits: Int) extends Bundle {
-  val fn = Bits(SZ_ALU_FN.W)
+class MultiplierReq(dataBits: Int, tagBits: Int)(implicit val p: Parameters) extends Bundle with HasRocketCoreParameters {
+  val alu = if (usingABLU) ABLU else ALU
```
jerryz123, line 12: Maybe you can add a method to `RocketCoreParams`: `rocketCoreParams.alufn`. I want to avoid this `if (usingABLU)` everywhere

jerryz123, line 12: I don't want to make this depend on `HasRocketCoreParameters`.

jerryz123 (changes requested):
I trust the circuit implementations, but there should be some refactors

src/main/scala/rocket/ALU.scala
ZenithalHourlyRate: Used `???`

src/main/scala/rocket/ALU.scala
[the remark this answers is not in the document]
ZenithalHourlyRate: Done.

src/main/scala/rocket/ALU.scala
ZenithalHourlyRate: Done.

src/main/scala/rocket/ABLU.scala
ZenithalHourlyRate: Done. Also done for other modules like CryptoNIST.

src/main/scala/rocket/Multiplier.scala
ZenithalHourlyRate: I also do not want to do so. However, Multipler internal staff depends on the rocket ALU and its specific FN encoding. Unless we refactor IDecode to include the decode table defined in Multipler, we have to depend on the rocket core parameters.

src/main/scala/rocket/CSR.scala
ZenithalHourlyRate: Done.

src/main/scala/rocket/Multiplier.scala
```diff
@@ -6,10 +6,10 @@ package freechips.rocketchip.rocket
 import chisel3._
 import chisel3.util.{Cat, log2Up, log2Ceil, log2Floor, Log2, Decoupled, Enum, Fill, Valid, Pipe}
 import freechips.rocketchip.util._
-import ALU._
+import freechips.rocketchip.config.Parameters
 
-class MultiplierReq(dataBits: Int, tagBits: Int) extends Bundle {
-  val fn = Bits(SZ_ALU_FN.W)
+class MultiplierReq(dataBits: Int, tagBits: Int)(implicit val p: Parameters) extends Bundle with HasRocketCoreParameters {
+  val fn = Bits(alufn.SZ_ALU_FN.W)
```
jerryz123, line 12: Would `class MultiplierReq(dataBits: Int, tagBits: Int, alufn: ALUFN = ALUFN())` work?

ZenithalHourlyRate: Done.

jerryz123: Please provide default for alufn to be the base, this makes migration much easier.

src/main/scala/rocket/ALU.scala
```diff
@@ -45,51 +85,51 @@ object ALU
...
+    val fn = Input(UInt(alufn.SZ_ALU_FN.W))
     val in2 = Input(UInt(xLen.W))
     val in1 = Input(UInt(xLen.W))
     val out = Output(UInt(xLen.W))
     val adder_out = Output(UInt(xLen.W))
     val cmp_out = Output(Bool())
   })
+}
 
+class ALU(fn: ALUFN)(implicit p: Parameters) extends AbstractALU(fn)(p) {
```
jerryz123, line 100: I think this can just be `class ALU(implicit p: Parameters) extends AbstractALU(ALUFN())(p) {

jerryz123:
I will clone this branch myself to check some things.

src/main/scala/rocket/Multiplier.scala
ZenithalHourlyRate: Done.

src/main/scala/rocket/ALU.scala
ZenithalHourlyRate: Done.

src/main/scala/tile/BaseTile.scala
```diff
@@ -103,10 +103,24 @@ trait HasNonDiplomaticTileParameters {
     val f = if (tileParams.core.fpu.nonEmpty) "f" else ""
     val d = if (tileParams.core.fpu.nonEmpty && tileParams.core.fpu.get.fLen > 32) "d" else ""
     val c = if (tileParams.core.useCompressed) "c" else ""
-    val b = if (tileParams.core.useBitManip) "b" else ""
     val v = if (tileParams.core.useVector) "v" else ""
-    val x = tileParams.core.customIsaExt.map(s => s"_$s").getOrElse("")
-    s"rv${p(XLen)}$ie$m$a$f$d$c$b$v$x"
+    val multiLetterExt = Seq(
+      (if (tileParams.core.useBitManip) "Zba" else ""),
```
jerryz123, line 108: Try 
```
val multiLetterExt = (Option.when(tileParams.core.useBitManip)(Seq("Zba", "Zbb", "Zbc")) ++
                      Option.when(tileParams.core.hasBitManipCrypto)(Seq("Zbkb", "Zbkc", "Zbkx")) ++
                      ...).flatten
val multiLetterString = multiLetterExt.mkString("_")
```

src/main/scala/tile/BaseTile.scala
```diff
@@ -103,10 +103,24 @@ trait HasNonDiplomaticTileParameters {
...
+      (if (tileParams.core.useBitManip) "Zbs" else ""),
+      (if (tileParams.core.useCryptoNIST) "Zknd" else ""),
+      (if (tileParams.core.useCryptoNIST) "Zkne" else ""),
+      (if (tileParams.core.useCryptoNIST) "Zknh" else ""),
+      (if (tileParams.core.useCryptoSM) "Zksed" else ""),
+      (if (tileParams.core.useCryptoSM) "Zksh" else ""),
+      tileParams.core.customIsaExt.getOrElse("")
+    )
+    val multiLetterString = multiLetterExt.filter(_ != "").mkString("_")
+    s"rv${p(XLen)}$ie$m$a$f$d$c$v$multiLetterString"
```
jerryz123, line 123: Shouldn't there be a `_` before `$multiLetterString`?

src/main/scala/tile/Core.scala
```diff
@@ -54,6 +57,7 @@ trait CoreParams {
   def customCSRs(implicit p: Parameters): CustomCSRs = new CustomCSRs
 
   def hasSupervisorMode: Boolean = useSupervisor || useVM
+  def hasBitManipCrypto: Boolean = useBitManipCrypto || useCryptoNIST || useCryptoSM
   def instBytes: Int = instBits / 8
```
jerryz123, line 61: Can we add checks for invalid configurations? Is `useBitManipCrypto && !useBitManip` legal?

src/main/scala/tile/BaseTile.scala
ZenithalHourlyRate: Well that is accepted format as underscores could be added between extensions.

> Underscores “_” may be used to separate ISA extensions to improve readability and to provide disambiguation

The only required underscore is between the multi-letter extensions, and in the spec the example is

> Extensions with the “Z” prefix must be separated from other multi-letter extensions by an underscore, e.g., “RV32IMACZicsr Zifencei”.

Also strangely riscv-software-src/riscv-config accepts rv32imaczicsr but does not accept rv32ima_zicsr.

src/main/scala/tile/Core.scala
ZenithalHourlyRate: Yes it is legal. People may do not want zba/zbs extension introduced by useBitManip hence only enabling useBitManipCrypto.

src/main/scala/tile/BaseTile.scala
ZenithalHourlyRate: Fixed.

ZenithalHourlyRate:
Should squash&merge for this PR because the history is too long..

jerryz123:
Yes we should squash+merge. This PR looks fine for me now.

jerryz123:
Once CI is done I will squash+merge

ZenithalHourlyRate:
> Once CI is done I will squash+merge

CI passed.

jerryz123 (approved):

PR #2956, opened 2022-04-02
Update Instructions with riscv-opcodes

ZenithalHourlyRate:
**Related issue**: dependency of #2950, but not specific to #2950 only.

General speaking, keeping synchronized with riscv-opcodes is easier to maintain.

More notes are in commit messages.

**Impact**: API modification

**Release Notes**

* Updated Instructions/CSRs from riscv-opcodes
* Separate rocket specific and currently not ratified instructions/CSRs into CustomInstructions.scala


src/main/scala/rocket/CustomInstructions.scala
```diff
@@ -0,0 +1,32 @@
+// See LICENSE.SiFive for license details.
+// See LICENSE.Berkeley for license details.
+
+package freechips.rocketchip.rocket
+
+import Chisel._
```
sequencer, line 6: use chisel3 plz

ZenithalHourlyRate: Fixed in the new commit.

src/main/scala/rocket/CSR.scala
```diff
@@ -626,18 +627,18 @@ class CSRFile(
...
   read_mnstatus.mie := reg_rnmie
   val nmi_csrs = if (!usingNMI) LinkedHashMap() else LinkedHashMap[Int,Bits](
-    CSRs.mnscratch -> reg_mnscratch,
-    CSRs.mnepc -> readEPC(reg_mnepc).sextTo(xLen),
-    CSRs.mncause -> reg_mncause,
-    CSRs.mnstatus -> read_mnstatus.asUInt)
+    CustomCSRs.mnscratch -> reg_mnscratch,
+    CustomCSRs.mnepc -> readEPC(reg_mnepc).sextTo(xLen),
+    CustomCSRs.mncause -> reg_mncause,
+    CustomCSRs.mnstatus -> read_mnstatus.asUInt)
```
sequencer, line 641: I wonder whether this should use the customCSR API:
https://github.com/chipsalliance/rocket-chip/blob/5ebac8f75926677f026cb8dab5181503300b28c6/src/main/scala/rocket/CSR.scala#L359

ZenithalHourlyRate: Review in RC Workshop suggested not using customCSRs API, which is for "really" custom CSRs. The current custom instructions/CSRs in `Instructions.scala` is just not ratified.

In my opinion, using this API with NMI logics is not trivial.

Review also suggested submodule riscv-opcodes and auto generate `Instructions.scala`.

I think these can be separated into a file named `NotRatifiedInstructions.scala` to distinguish those existing API named CustomCSRs (e.g. in `CustomCSRs.scala`). When instructions get ratified, they can be removed from this file and riscv-opcodes submodule can bump to cover them.

sequencer (approved):
LGTM, I retied locally to generated the same file. In the long term, we should split the auto-generated codes out from RC code base.

Issue #2963, opened 2022-04-14
Documentation for CRC util


michael-etzkorn:
**Type of issue**: other enhancement

**Impact**: enhanced user experience

**Development Phase**: proposal

**Other information**
 

**What is the current behavior?**
User is directed to CRC zoo where implicit 1 notation is used but CRC object uses explicit 1 notation which can cause unexpected behavior if require is not hit.  
**What is the expected behavior?**

Document that this is explicit 1 notation and that crc zoo uses implicit 1 notation. 

**What is the use case for changing the behavior?**
This was only discoverable while reading the code and noticing the require. This'll prevent user misuse and headache. 

https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/util/CRC.scala#L14

Issue #2960, opened 2022-04-08
[WIP] Documentation Plan


sequencer:
**Type of issue**: feature request

**Impact**: no functional change

**Development Phase**: request

**Other information**

michael-etzkorn:
A small list of things to provide quality documentation for: 
- FPGA and ASIC flow resources
- Configurations and Parameters
- RC's microarchitecture. Doc similar to BOOM's. 
    - helps users with situations like #2955 
- Running unit tests
- Running benchmarking 
- RoCC
- Bus Topology and Bus APIs. e.g.:
    - Diplomatic Widgets 
    - Location API
- Device tree
- Version releases - PRs should correlate to version updates and there should be patch notes with each release

rstarc:
What is the status of this issue? I would like to contribute
I see that there are multiple seemingly related branches named `*_doc` containing unmerged commits. Is anyone actively else actively working on this?

PR #2964, opened 2022-04-14
[CRC.scala][NFC] Note CRC's object explicit+1 notation

michael-etzkorn:
#2963 just so users can follow the CRC object's API a bit better. 

Perhaps #2960 should include a scala doc

**Related issue**: 
#2963

**Release Notes**

### DIFF
--- a/src/main/scala/util/CRC.scala
+++ b/src/main/scala/util/CRC.scala
@@ -6,7 +6,8 @@ import Chisel._
 
 object CRC
 {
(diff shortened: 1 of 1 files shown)


sequencer (approved):
Thanks for this PR!
I think the CRC utils can be merged into chisel utilities after full documented and formally tested.
Merge this since no functionally change and it can make CRC documentation more detail.

PR #2965, opened 2022-04-16
Fix decoding of unimplemented CSR addresses

a0u:
**Related issue**: 

**Impact**:

**Release Notes**

For instance, with `nPMPs = 0`, writes to PMP CSRs were observed to modify unrelated CSRs.

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -795,7 +795,7 @@ class CSRFile(
     val addr = Cat(io.status.v, io.rw.addr)
     val pats = for (((k, _), i) <- read_mapping.zipWithIndex)
       yield (BitPat(k.U), (0 until read_mapping.size).map(j => BitPat((i == j).B)))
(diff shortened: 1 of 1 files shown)


aswaterman:
Observed in what context?  If the CSR address is invalid, the Rocket core will mark the instruction illegal earlier in the pipeline, suppressing the write.

a0u:
Agreed that this isn't problematic for Rocket.  The change in the decode logic just manifests a latent bug in another core design of ours that reuses the CSR file.

aswaterman:
To add some more color: the CSRFile interface assumes that commands other than reads have already been checked for their validity.  In most pipelined processors, this tends to happen automagically, since the illegal-instruction detection logic happens earlier in the pipeline.

jerryz123:
The bug in our core was not checking `read_illegal` as the general "illegal csr" signal for instructions that only write CSRs. `write_illegal` by itself is not sufficient.

PR #2968, opened 2022-04-23
bump json-jackson to 4.0.5

sequencer:
this PR make RC be able to compile against FIRRTL 1.6.X after chipsalliance/firrtl#2519 merged.
This should be merged as a part of 1.6.X bumping.
**Release Notes**
bumping jackson to 4.0.5

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -15,7 +15,7 @@ lazy val commonSettings = Seq(
   traceLevel   := 15,
   scalacOptions ++= Seq("-deprecation","-unchecked"),
   libraryDependencies ++= Seq("org.scala-lang" % "scala-reflect" % scalaVersion.value),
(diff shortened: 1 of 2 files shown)


michael-etzkorn:
I believe this will no longer throw an error once the json4s v4.x-dependent version of FIRRTL is being used by RC. Since RC currently uses FIRRTL 1.5 (which is on json4s 3.x), we can't upgrade RC's json4s to 4.x until a chisel release that uses FIRRTL 1.6-SNAPSHOT (which will then be updated to json4s 4.x and require this bump in RC). 

Really, we could probably get away with not including json4s in the build dependency and just rely on whatever FIRRTL is using at the time.

sequencer:
this PR only can be merged after x.6 Chisel/FIRRTL releasing.

sequencer:
fixed in #3243

PR #2969, opened 2022-04-26
Bind Verilog path with EICG_wrapper

Phantom1003:
I found the EICG_wrapper.v file was not copied to the build directory automatically, the original flow works because it uses rc_resource_dir as the include path.

### DIFF
--- a/src/main/scala/util/ClockGate.scala
+++ b/src/main/scala/util/ClockGate.scala
@@ -3,6 +3,7 @@
 package freechips.rocketchip.util
 
 import chisel3._
+import chisel3.util.HasBlackBoxResource
 import freechips.rocketchip.config.{Field, Parameters}
 
 case object ClockGateImpl extends Field[() => ClockGate](() => new EICG_wrapper)
@@ -37,4 +38,6 @@ object ClockGate {
 }
 
 // behavioral model of Integrated Clock Gating cell
-class EICG_wrapper extends ClockGate
+class EICG_wrapper extends ClockGate with HasBlackBoxResource {
+  addResource("/vsrc/EICG_wrapper.v")
+}
--- a/src/main/scala/util/ClockGate.scala
+++ b/src/main/scala/util/ClockGate.scala
@@ -3,18 +3,29 @@
 package freechips.rocketchip.util
 
 import chisel3._
-import chisel3.util.HasBlackBoxResource
+import chisel3.util.{HasBlackBoxResource, HasBlackBoxPath}
 import freechips.rocketchip.config.{Field, Parameters}
 
-case object ClockGateImpl extends Field[() => ClockGate](() => new EICG_wrapper)
+import java.nio.file.{Files, Paths}
 
-abstract class ClockGate extends BlackBox {
+case object ClockGateImpl extends Field[() => ClockGate](() => new EICG_wrapper)
+case object ClockGateModelFile extends Field[Option[String]](None)
+
+abstract class ClockGate extends BlackBox
+  with HasBlackBoxResource with HasBlackBoxPath {
   val io = IO(new Bundle{
     val in = Input(Clock())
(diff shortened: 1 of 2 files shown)


sequencer:
See https://github.com/chipsalliance/rocket-chip/pull/1981#issuecomment-496505302
I personally wanna have this PR. So maybe another trade-off is providing an CDE configuration and set `EICG_wrapper` to empty by default to resolve @aswaterman's concern.

sequencer (changes requested):
See https://github.com/chipsalliance/rocket-chip/pull/1981#issuecomment-496505302
I personally wanna have this PR. So maybe another trade-off is providing an CDE configuration and set EICG_wrapper to empty by default to resolve @aswaterman's concern.

src/main/scala/util/ClockGate.scala
```diff
@@ -23,6 +35,8 @@ object ClockGate {
       name: Option[String] = None)(implicit p: Parameters): Clock = {
     val cg = Module(p(ClockGateImpl)())
     name.foreach(cg.suggestName(_))
+    p(ClockGateModelFile).map(cg.addVerilogResource(_))
```
sequencer, line 38: I'm not so sure if this correct. Have you tried this? using `addPath` and `addResource` out of `BlackBox` Module is a little anti-pattern. IMHO, you can directly use the `CDE` parameter there.

sequencer (changes requested):

src/main/scala/util/ClockGate.scala
Phantom1003: Yes, following is from the annotation file, the `target` field is correctly set to EICG_wrapper:
```
  {
    "class":"firrtl.transforms.BlackBoxPathAnno",
    "target":"TestHarness.EICG_wrapper",
    "path":"/.../rocket-chip/src/main/resources/vsrc/EICG_wrapper.v"
  },
```
```
  {
    "class":"firrtl.transforms.BlackBoxInlineAnno",
    "target":"TestHarness.EICG_wrapper",
    "name":"EICG_wrapper.v",
    "text":"..."
  },
```

src/main/scala/util/ClockGate.scala
```diff
@@ -3,17 +3,29 @@
 package freechips.rocketchip.util
 
 import chisel3._
+import chisel3.util.{HasBlackBoxResource, HasBlackBoxPath}
 import freechips.rocketchip.config.{Field, Parameters}
 
+import java.nio.file.{Files,Paths}
+
 case object ClockGateImpl extends Field[() => ClockGate](() => new EICG_wrapper)
+case object ClockGateModelFile extends Field[Option[String]](None)
```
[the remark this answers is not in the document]
Phantom1003: We need to make it empty here to generate a link error if the user forgets to provide one.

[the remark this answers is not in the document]
sequencer: I think you are right.

aswaterman (commented):
I'll leave this to y'all to decide.

sequencer (approved):

sequencer:
Thanks @aswaterman :)

src/main/scala/util/ClockGate.scala
```diff
@@ -3,17 +3,29 @@
...
+  with HasBlackBoxResource with HasBlackBoxPath {
   val io = IO(new Bundle{
     val in = Input(Clock())
     val test_en = Input(Bool())
     val en = Input(Bool())
     val out = Output(Clock())
   })
+
+  def addVerilogResource(vsrc: String): Unit = {
+    if (Files.exists(Paths.get(vsrc)))
```
jerryz123, line 24: Is the a reason for this to not just error immediately if the specified file does not exist?
The fallback might be unexpected

PR #2931, opened 2022-01-18
Bump to Chisel 3.5

davidbiancolin:
This pulls in all the necessary changes from currently open PRs to complete a bump to Chisel 3.5. For various reasons, some of those earlier PRs cannot pass CI standalone, so it was easiest (for me) to bundle them together. 
These PRs include:
Bumping hardfloat (closes #2930)
Removing clonetype (closes #2889)
Removing the old property and updating cover (closes #2890)

Other notes: 
- I bumped macro paradise to 2.1.1 so that a published version could be found for scala 2.12.15. We already do this in the mill build, it appears. If that is undesired, can someone point me at a place where it is published and we can update the resolvers?
- I tried to update the mill files. @sequencer how should i test them?

**Related issue**: 

**Type of change**: other enhancement

**Impact**: API addition (no impact on existing code)

**Development Phase**:  implementation

**Release Notes**

Bump to Chisel 3.5.

PR #2973, opened 2022-05-03
[AXISBundleBits] re-add clonetype override

michael-etzkorn:
* AXISBundleBits extends off BundleMap which is of type Record and therefore needs cloneType
* resolves a user issue that can be seen at the repo in this post after compiling the current main's rocket-jar:
* https://stackoverflow.com/questions/72043687/implementing-a-diplomatic-axi-stream-interface-in-chisel-bundlemap-clonetype-e

* git repo: https://github.com/jurevreca12/temp_dspblock_example
* git commit: aca81ab9df437e8cea7fa8abd361b6bab7f40326
* note: replace `lib/rocketchip.jar` with a compiled `rocketchip.jar` from main
* bug fixed by using a `rocketchip.jar` generated from this PR's branch

Signed-off-by: Michael Etzkorn <michael-etzkorn@sage.micro.com>

**Related issue**: 
https://stackoverflow.com/questions/72043687/implementing-a-diplomatic-axi-stream-interface-in-chisel-bundlemap-clonetype-e

**Type of change**: bug fix

**Release Notes**

Notes to be added to version release:
- Fixed an issue where AXIS nodes reliance on AXISBundleBits would throw BundleMap's IllegalArgumentException due to removal of cloneType override. 

Though because we've yet to add release notes for updating to 3.5. The version notes can simply read:
- removed cloneType from Bundles for bumping to Chisel 3.5.x (still required by custom Bundles inheriting BundleMap)

### DIFF
--- a/src/main/scala/amba/axis/Bundles.scala
+++ b/src/main/scala/amba/axis/Bundles.scala
@@ -24,6 +24,7 @@ case class AXISDataField(width: Int) extends BundleField(AXISData) {
 }
 
 class AXISBundleBits(val params: AXISBundleParameters) extends BundleMap(AXISBundle.keys(params)) {
+  override def cloneType: this.type = (new AXISBundleBits(params)).asInstanceOf[this.type]
   def last = if (params.hasLast) apply(AXISLast) else true.B
   def id   = if (params.hasId)   apply(AXISId)   else 0.U
   def dest = if (params.hasDest) apply(AXISDest) else 0.U


sequencer (changes requested):
IIRC. this should blame to me when bumping chisel 3.5.
Thanks @michael-etzkorn to figure it out.
currently the only class that extends from `BundleMap` is `AXISBundleBits`
however `BundleMap.apply` takes a sequence of `BundleFieldBase` which might need to be fixed as well(add a cloneType function to base class might resolve this.)

michael-etzkorn:
I could be wrong on this, but elements are cloned within `BundleMap` so I don't believe any further changes are immediately necessary. 

- https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/util/BundleMap.scala#L108

At the very least, this hot fixes AXIS nodes failing to generate and should be merged before any further proposals to BundleMap.

sequencer:
I checked #2931, seems originally the `BundleMap.apply` function doesn't have a `cloneType` at all. merging this for now.

sequencer (approved):

PR #2978, opened 2022-05-06
Support incoherent access to ExtMem through SBus

jerryz123:
I'm not sure if this approach is the best, but the provided functionality is useful.

This PR provides a new `incohBase` field to the `MemoryPortParams`. When set, this field creates an alias for the backing AXI4 mem and attaches that alias to the systembus. This provides a mechanism through which masters can directly access cacheable memory while bypassing a coherence manager.

For example, the default ExtMem region `0x8000_0000 : 0x9000_0000` can now also also be accessed at `0x10_8000_0000 - 0x10_9000_0000`. 

**Type of change**: new feature

**Release Notes**


src/main/scala/tilelink/ResourceRemover.scala
```diff
@@ -0,0 +1,29 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.tilelink
+
+import freechips.rocketchip.config.Parameters
+import freechips.rocketchip.diplomacy._
+
+
+class TLResourceRemover(implicit p: Parameters) extends LazyModule
```
hcook, line 9: I think you could have done this by writing a new `ManagerFn` for `TLFilter`

Issue #2980, opened 2022-05-08
[Bug Report] incorrect xs field check under VS


Phantom1003:
**Type of issue**: bug report
**Impact**: unknown
**Development Phase**: proposal
**Other information**
In the following test case, we first enable mpv and xs bit in mstatus, and then set up xs field in vsstatus. With the xs field, any custom instructions should not raise illegal signal.
From the co-simulation result, spike successfully executes the rocc instruction, while rocket throws an exception.
```
[rocket] 3 0x00080000178 (0x30005073)
[spike]  core   0: 0x0000000080000178 (0x30005073) csrwi   mstatus, 0
[rocket] 3 0x0008000017c (0x000182b7) x 5 0x0000000000018000
[spike]  core   0: 0x000000008000017c (0x000182b7) lui     t0, 0x18
[rocket] 3 0x00080000180 (0x3002a073)
[spike]  core   0: 0x0000000080000180 (0x3002a073) csrs    mstatus, t0
[rocket] 3 0x00080000184 (0x0010031b) x 6 0x0000000000000001
[spike]  core   0: 0x0000000080000184 (0x0010031b) addiw   t1, zero, 1
[rocket] 3 0x00080000188 (0x02731313) x 6 0x0000008000000000
[spike]  core   0: 0x0000000080000188 (0x02731313) slli    t1, t1, 39
[rocket] 3 0x0008000018c (0x30032073)
[spike]  core   0: 0x000000008000018c (0x30032073) csrs    mstatus, t1
...
[rocket] 3 0x000800001a4 (0x20029073)
[spike]  core   0: 0x00000000800001a4 (0x20029073) csrw    vsstatus, t0
[rocket] 3 0x000800001a8 (0x20002573) x10 0x8000000200018000
[spike]  core   0: 0x00000000800001a8 (0x20002573) csrr    a0, vsstatus
...
[rocket] 3 0x000800001bc (0x30200073)
[spike]  core   0: 0x00000000800001bc (0x30200073) mret
[rocket] 1 0x000800001c0 (0x00000613) x12 0x0000000000000000
[spike]  core   0: 0x00000000800001c0 (0x00000613) li      a2, 0
[rocket] 3 0x00080000004 (0x34302f73) x30 0x000000000000700b
[spike]  core   0: 0x00000000800001c4 (0x0000700b) custom0.rd.rs1.rs2 (args unknown)
[error] PC SIM 00000000800001c4, DUT 0000000080000004
[...]

PR #2984, opened 2022-05-12
Disallow writes to VSSTATUS.XS

Phantom1003:
From https://github.com/chipsalliance/rocket-chip/issues/2980#issuecomment-1123218779, make vsstatus.xs read only, and hard-wired to dirty when RoCC is enabled.

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -950,7 +950,6 @@ class CSRFile(
   io.gstatus.sd := io.gstatus.fs.andR || io.gstatus.xs.andR || io.gstatus.vs.andR
   io.gstatus.uxl := (if (usingUser) log2Ceil(xLen) - 4 else 0)
   io.gstatus.sd_rv32 := xLen == 32 && io.gstatus.sd
-  io.gstatus.xs := (if (usingRoCC) UInt(3) else UInt(0))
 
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/CSR.scala
```diff
@@ -950,6 +950,7 @@ class CSRFile(
   io.gstatus.sd := io.gstatus.fs.andR || io.gstatus.xs.andR || io.gstatus.vs.andR
   io.gstatus.uxl := (if (usingUser) log2Ceil(xLen) - 4 else 0)
   io.gstatus.sd_rv32 := xLen == 32 && io.gstatus.sd
+  io.gstatus.xs := (if (usingRoCC) UInt(3) else UInt(0))
```
jerryz123, line 954: Can't we just hardcode `reg_vsstatus.xs` to dirty?

jerryz123 (changes requested):

src/main/scala/rocket/CSR.scala
Phantom1003: Yes, you're right. And make gstatus.xs won't help to fix the bug mentioned in https://github.com/chipsalliance/rocket-chip/pull/2983, since rocc_illegal use reg_vsstatus.xs.

jerryz123 (approved):

Issue #2953, opened 2022-03-25
Refactor Rocket Chip test framework


sequencer:
Most of configurations in RC is untested. Making RC almost impossible to accept RTL changes to new RV extension from community. 
So here I wanna request a new test framework:
1. A Scala `TestFramework` based test framework.
Currently RC uses Makefile, making Scala guys hard to add new test. It's reasonable to migrate most of test to Scala `TestFramework`, `ScalaTest` or `uTest` are both OK.  
2. Get rid of dependency to SW, upstream compiler is almost mature for RISC-V now, user don't need to compile their toolchain at 2022, directly depends to LLVM seems reasonable.
3. Use test case from
https://github.com/riscv-software-src/riscv-tests
https://github.com/riscv-non-isa/riscv-arch-test
4. frameworks for different developing levels
- a spike-cosim library is required for debug, it should support verilator at least.
- JTAG bitbang should be replaced to JTAG VPI to speed up binary loading.

**Type of issue**: feature request

**Impact**: no functional change

**Development Phase**: request

michael-etzkorn:
@sequencer this is maybe something we can discuss April 6th, but I believe for (4), it'd be reasonable to consider (and I'd be willing to help develop) UVM-based or generic SystemVerilog testbenches for testing the Verilog IR.

sequencer:
>  it'd be reasonable to consider (and I'd be willing to help develop) UVM-based or generic SystemVerilog testbenches for testing the Verilog IR.

Thanks @michael-etzkorn!
I think before ventilator(or any other Open-Source high-performance simulator) providing the full UVM support. We should continue stick in the DPI/VPI for the best compatibility.
And also there is a needs to test RC with commercial tools(like VCS+Verdi). Design with GTKWave is painful.
Currently the RC do support a DPI-based simulation, see LazyModules prefix with `Sim`(`SimJTAG`, `SimDTM`) etc...
And it works good with Verilator and VCS, but needs more documentation and standardization.

So basically, from my point of view: the "test framework" should be a refactor to current test framework. I think we needs hands on it, we can discuss on the RCW meeting!

PR #2991, opened 2022-05-19
Add test framework

geekLucian:
**Related issue**: https://github.com/chipsalliance/rocket-chip/issues/2953

**Release Notes**

Add test framework from @sequencer's [playground](https://github.com/sequencer/playground/tree/master/sanitytests/rocketchip).


build.sc
```diff
@@ -35,6 +36,50 @@ object hardfloatRocket extends hardfloat.build.hardfloat {
...
+    //    if (!helper.isMac) {
+    //      val x86Dir = T.ctx.dest
+    //      val riscv64Dir = T.ctx.dest / "riscv64"
+    //      os.proc("make", s"DESTDIR=${x86Dir}", "install").call(spike.compile())
+    //      os.proc("make", s"DESTDIR=${riscv64Dir}", "install").call(compilerrt.compile())
+    //      os.proc("make", s"DESTDIR=${riscv64Dir}", "install").call(musl.compile())
+    //      os.copy.into(pk.compile(), riscv64Dir)
+    //    }
+    T.ctx.dest
+  }
```
sequencer, line 77: Since the test used the fesvr, you may need introduce spike as well.

michael-etzkorn:
Relevant issue:
https://github.com/chipsalliance/rocket-chip/issues/2953 

I've yet to look over the details of this but I'd like to see more `test/scala` at some point for development. 

My two cents on structuring the RTL based testing directory-wise is that it should be mergable https://github.com/chipsalliance/rocket-chip/tree/master/src/main/scala/unittest and ideally be invokable within `regression`

sequencer:
> I don't think I understand why this is in a separate sanitytests directory instead of src/test. The changes also introduce more places where chisel/firrtl version numbers are encoded, which increases maintenance burden slightly.

I agree, I thought @geekLucian just copy-paste from my own repository, which obviously needs more refactor to make it fit into rocket-chip.

> This approach also seems to reimplement some functionality already in chiseltest. What functionality does this approach support over chiseltest?

IIRC, this approach is copy-paste from the rocket-chip original test suite, and has no relation to the chiseltest.
programs are supported to be loaded via DM, since the DMI, or JTAG-VPI depends on DPI, which makes chiseltest impossible to test it.

> Lastly, if we do move forwards with this approach, I think the csrc files should symlink the existing files, and the bootrom should use the default one in RC (Don't we want to test that that bootrom works?).

Agree, `bootrom` should use the original `bootrom` from Rocket-Chip.

build.sc
```diff
@@ -35,6 +36,50 @@ object hardfloatRocket extends hardfloat.build.hardfloat {
   ) else Agg.empty[Dep]
 }
 
+object sanitytests extends ScalaModule with TestModule.Utest {
```
sequencer, line 39: This should live under `rocketchip` project

build.sc
```diff
@@ -35,6 +36,50 @@ object hardfloatRocket extends hardfloat.build.hardfloat {
...
+
+  override def scalacPluginClasspath = T { super.scalacPluginClasspath() }
+
+  object ivys {
+    val chisel3 = ivy"edu.berkeley.cs::chisel3:3.5.3"
+    val firrtl = ivy"edu.berkeley.cs::firrtl_2.13:1.5.3"
+    val scalatest = ivy"org.scalatest::scalatest:3.2.0"
+    val oslib = ivy"com.lihaoyi::os-lib:0.7.8"
+    val utest = ivy"com.lihaoyi::utest:0.7.10"
+  }
```
sequencer, line 54: this should inherit from `rocketchip` object

build.sc
```diff
@@ -35,6 +36,50 @@ object hardfloatRocket extends hardfloat.build.hardfloat {
...
+
+  override def ivyDeps = Agg(
+    ivys.chisel3,
+    ivys.scalatest,
+    ivys.oslib,
+    ivys.utest
+  )
```
sequencer, line 61: same here.

build.sc
```diff
@@ -35,6 +36,50 @@ object hardfloatRocket extends hardfloat.build.hardfloat {
...
+  def libraryResources = T.sources {
+    //    if (!helper.isMac) {
+    //      val x86Dir = T.ctx.dest
+    //      val riscv64Dir = T.ctx.dest / "riscv64"
+    //      os.proc("make", s"DESTDIR=${x86Dir}", "install").call(spike.compile())
+    //      os.proc("make", s"DESTDIR=${riscv64Dir}", "install").call(compilerrt.compile())
+    //      os.proc("make", s"DESTDIR=${riscv64Dir}", "install").call(musl.compile())
+    //      os.copy.into(pk.compile(), riscv64Dir)
+    //    }
+    T.ctx.dest
```
sequencer, line 76: You should introduce dependency of `spike` aka riscv-isa-sim. for `csrc` you need use the original `csrc`.

sanitytests/src/TestConfig.scala
```diff
@@ -0,0 +1,45 @@
...
+            elf,
+            bin
+          ).call()
+          proc(
+            "dd",
+            s"if=$bin",
+            s"of=$img",
+            "bs=128",
+            "count=1"
+          ).call()
```
sequencer, line 41: I think compiling of `bootrom` should live outside `TestFramework`, it should somehow live in mill, which makes it being able to be cached by the build system.

sanitytests/src/VerilatorTest.scala
```diff
@@ -0,0 +1,49 @@
...
+        ).call(outputDirectory)
+        test("Hello World!") {
+          os.proc(
+            s"$emulator",
+            s"${resource("riscv64/pk")}",
+            "hello",
+          ).call(outputDirectory)
+        }
+      }
+    }
```
sequencer, line 47: The build of test cases should also live inside the build system.
and it depends a C compiler, which should be optional as well, IMO, the RISCV toolchain is stable enough, which is possible to use the toolchain from upstream(LLVM or GCC).
You can install it in the GItHub Action like playground did.

sequencer (changes requested):

sequencer:
> I don't think I understand why this is in a separate sanitytests directory instead of `src/test`.

The `src/test` uses `ScalaTest` by default, while I use the `uTest`, I think it is reasonable to stick on `ScalaTest`, since I didn't depends on the special features of `uTest`.

As for the packaging issue. IMHO, a more fine granularity provides flexibility, the architect part can be split out from the `src/test` living in another package, like what chisel did [here](https://github.com/chipsalliance/chisel3/tree/master/integration-tests/src/test/scala).

build.sc
```diff
@@ -14,7 +16,7 @@ object configRocket extends `api-config-chipsalliance`.`build-rules`.mill.build.
   }
 
   override def pomSettings = T {
-    rocketchip.pomSettings()
+    rocketchip.`pomSettings`()
```
sequencer, line 19: why

build.sc
```diff
@@ -47,4 +154,82 @@ object rocketchip extends common.CommonRocketChip {
...
+      ).call()
+      os.proc(
+        "dd",
+        s"if=$bin",
+        s"of=$img",
+        "bs=128",
+        "count=1"
+      ).call()
+      T.ctx.dest
+    }
```
sequencer, line 230: make this as an external function, so that others can add programs as test payloads.

build.sc
```diff
@@ -47,4 +154,82 @@ object rocketchip extends common.CommonRocketChip {
...
+      os.proc("make", s"DESTDIR=${riscv64Dir}", "install").call(musl.compile())
+      os.copy.into(pk.compile(), riscv64Dir)
+      // build hello world
+      val outputDirectory = x86Dir / "VerilatorTest"
+      os.remove.all(outputDirectory)
+      os.makeDir(outputDirectory)
+      os.proc(
+        "clang",
+        "-o", "hello",
+        s"${resource("resources/csrc/hello.c")}",
```
sequencer, line 184: I personal prefer using a `testcase/hello` directory to store these tests.

common.sc
```diff
@@ -4,8 +4,8 @@ import mill.scalalib.publish._
 import coursier.maven.MavenRepository
 
 val defaultVersions = Map(
-  "chisel3" -> "3.5.0",
-  "chisel3-plugin" -> "3.5.0"
+  "chisel3" -> "3.5.3",
+  "chisel3-plugin" -> "3.5.3"
```
sequencer, line 8: You need to create another PR to bump this.

sequencer (changes requested):
You are going to the right direction!

PR #3000, opened 2022-06-23
[TestDriver.v] Set initial `clock = 1'b1;`

Waxpple:
**Related issue**: Potential timing violation in the initial condition.

**Release Notes**

Set `clock = 1'b1;` to avoid the first reset signal only using half of the clock cycle.
If want to avoid the first reset cycle using only half clock period causing timing violation, modify `TestDriver.v` like this:
image

The waveform will be:
image

The original testbench will look like this:
image
The waveform will be
image

### DIFF
--- a/src/main/resources/vsrc/TestDriver.v
+++ b/src/main/resources/vsrc/TestDriver.v
@@ -8,8 +8,8 @@
 `endif
 
 module TestDriver;
-
-  reg clock = 1'b0;
+  // Set `clock = 1'b1;` to avoid the first reset signal only using half of the clock cycle.
+  reg clock = 1'b1;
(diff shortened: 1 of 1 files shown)


jerryz123:
What timing violation are you seeing? Are you assuming sync or async reset?

Waxpple:
> What timing violation are you seeing? Are you assuming sync or async reset?

Sorry for the late reply.

Setup time violation like this:
image
I assume the reset signal is synchronized.
However, It will reset for a couple of cycles, so the actual timing violation is not critical. 
But if the initial condition is set as above, it is not easy to trigger the timing check and makes it easier to debug the timing violation if there is any.

PR #3014, opened 2022-07-27
fix XLen breakage in `WithRV32` after CDE 1.2

OceanS2000:
**Related issue**: #3013

In CDE 1.2, the behavior of `up` query was changed. The documentation states that `up` gives the value prior to current `Parameter` layer; however, before 1.2, if a `site` query is encountered, then the value of the whole `Parameter`, instead of the value prior to current layer, is given. CDE 1.2 changed this to match documentation.

It turns out there is some code rely on old (presumably incorrect) behavior. In `WithRV32` layer, the `up` query to change FPU configuration cannot see `XLen => 32` assignment anymore. This PR introduces makes `WithRV32` a two `Config` stack so that `up` query can see new `XLen`.

In the long term, we probably need another solution. AFAIK, we can

- change documentation of CDE and revert back to old behavior, or
- rearrange current `Config`s in `subsystem` so its use of `site(XLen)` and `site(CacheBlockBits)` more easy to override

### DIFF
--- a/src/main/scala/subsystem/Configs.scala
+++ b/src/main/scala/subsystem/Configs.scala
@@ -254,14 +254,19 @@ class WithIncoherentTiles extends Config((site, here, up) => {
   )
 })
 
-class WithRV32 extends Config((site, here, up) => {
-  case XLen => 32
-  case RocketTilesKey => up(RocketTilesKey, site) map { r =>
-    r.copy(core = r.core.copy(
-      fpu = r.core.fpu.map(_.copy(fLen = 32)),
-      mulDiv = Some(MulDivParams(mulUnroll = 8))))
-  }
-})
+class WithRV32 extends Config(
+  new Config((site, here, up) => {
+    case RocketTilesKey => up(RocketTilesKey) map { r =>
+      r.copy(core = r.core.copy(
+        fpu = r.core.fpu.map(_.copy(fLen = 32)),
+        mulDiv = Some(MulDivParams(mulUnroll = 8))
+      ))
+    }
+  }) ++
+    new Config((site, here, up) => {
+      case XLen => 32
+    })
+)
 
 class WithNonblockingL1(nMSHRs: Int) extends Config((site, here, up) => {
   case RocketTilesKey => up(RocketTilesKey, site) map { r =>


src/main/scala/subsystem/Configs.scala
```diff
@@ -42,7 +42,7 @@ class BaseSubsystemConfig extends Config ((site, here, up) => {
   case DebugModuleKey => Some(DefaultDebugModuleParams(site(XLen)))
   case CLINTKey => Some(CLINTParams())
   case PLICKey => Some(PLICParams())
-  case TilesLocated(InSubsystem) => 
+  case TilesLocated(InSubsystem) =>
```
sequencer, line 45: non-related change

src/main/scala/subsystem/Configs.scala
```diff
@@ -361,7 +366,7 @@ class WithRationalRocketTiles extends Config((site, here, up) => {
 class WithEdgeDataBits(dataBits: Int) extends Config((site, here, up) => {
   case MemoryBusKey => up(MemoryBusKey, site).copy(beatBytes = dataBits/8)
   case ExtIn => up(ExtIn, site).map(_.copy(beatBytes = dataBits/8))
-  
+
```
sequencer, line 369: same here

src/main/scala/subsystem/Configs.scala
```diff
@@ -254,14 +254,19 @@ class WithIncoherentTiles extends Config((site, here, up) => {
...
+    case RocketTilesKey => up(RocketTilesKey) map { r =>
+      r.copy(core = r.core.copy(
+        fpu = r.core.fpu.map(_.copy(fLen = 32)),
+        mulDiv = Some(MulDivParams(mulUnroll = 8))
+      ))
+    }
+  }) ++
+    new Config((site, here, up) => {
+      case XLen => 32
+    })
```
sequencer, line 268: LGTM, since original CDE behavior was exactly a bug, however my daily CI(playground) didn't figure it out.
I believe this is a necessary bugfix in RC, but need more documentation on the CDE side.

sequencer (changes requested):
LGTM, need to rebase away some unnecessary changes.

src/main/scala/subsystem/Configs.scala
OceanS2000: It is quite disturbing to me that `WithRV32` is a very different to other `With` configurations. Other `With` configs are only intended to modify cores (`RocketTiles`) that comes before them, but changing `XLen` is a more fundamental thing as many values like `beatBytes` of various buses depend on it. Probably it should be set earlier ...?

jerryz123: We believe the recent cde commit introduced a bug in CDE. 

The design of this config is correct, and consistent with the expected behavior of cde.

jerryz123 (changes requested):
This PR should be closed. CDE should be fixed, and rocketchip should be bumped to the fixed version.

sequencer:
Sorry for the delay, I think we should fix that in CDE as well.

PR #3045, opened 2022-09-14
feat: convert unittest to chisel3

tianrui-wei:
Convert unittest subdir to use chisel3

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

**Release Notes**

### DIFF
--- a/src/main/scala/unittest/UnitTest.scala
+++ b/src/main/scala/unittest/UnitTest.scala
@@ -37,9 +37,9 @@ abstract class UnitTest(val timeout: Int = 4096) extends Module with UnitTestLeg
 case object UnitTests extends Field[Parameters => Seq[UnitTest]]
 
 class UnitTestSuite(implicit p: Parameters) extends Module {
-  val io = new Bundle {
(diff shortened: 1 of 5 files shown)


src/main/scala/unittest/UnitTest.scala
```diff
@@ -38,14 +38,14 @@ case object UnitTests extends Field[Parameters => Seq[UnitTest]]
 
 class UnitTestSuite(implicit p: Parameters) extends Module {
   val io = new Bundle {
```
sequencer, line 40: add IO wrapper

sequencer (changes requested):

tianrui-wei:
Addressed, thanks for reviewing

src/main/scala/unittest/UnitTest.scala
tianrui-wei: Addressed in the new commit

jerryz123 (approved):

sequencer (approved):

PR #3043, opened 2022-09-14
feat: convert devices to use chisel3

tianrui-wei:
This commit converts all files in the devices subdir to use chisel3 instead of chisel2

Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>

**Type of change**: other enhancement

**Impact**: no functional change

**Development Phase**:  implementation

**Release Notes**

PR #3058, opened 2022-09-15
feat: convert devices to use chisel3

tianrui-wei:
This commit converts all files in the devices subdir to use chisel3 instead of chisel2

Supercedes #3043 

**Release Notes**


tianrui-wei:
I think #3059 is needed as the base for this PR for the CI to pass

PR #3099, opened 2022-10-04
add comments to Xbar.scala

Lucas-Wye:
rebase Xbar docs from commit f406d268de014d87afa951adc5636b460b4196b7 and add more comments to Xbar

### DIFF
--- a/src/main/scala/tilelink/Xbar.scala
+++ b/src/main/scala/tilelink/Xbar.scala
@@ -9,7 +9,7 @@ import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.util._
 
 // Trades off slave port proximity against routing resource cost
-object ForceFanout
+object ForceFanout // actually not unused
 {
   def apply[T](
     a: TriStateValue = TriStateValue.unset,
@@ -31,9 +31,10 @@ private case class ForceFanoutParams(a: Boolean, b: Boolean, c: Boolean, d: Bool
 private case object ForceFanoutKey extends Field(ForceFanoutParams(false, false, false, false, false))
 
 class TLXbar(
-  policy: TLArbiter.Policy = TLArbiter.roundRobin // use option "roundRobin" which can check all the ports
+  policy: TLArbiter.Policy = TLArbiter.roundRobin // use option "roundRobin" which has similar property of MUX
   )(implicit p: Parameters) extends LazyModule
 {
+  // initialize parameters of master and slave
   val node = new TLNexusNode(
     clientFn  = { seq =>
       seq(0).v1copy(
@@ -69,6 +70,7 @@ class TLXbar(
     override def circuitIdentity = outputs.size == 1 && inputs.size == 1
   }
 
+  // implementation
   lazy val module = new LazyModuleImp(this) {
     if ((node.in.size * node.out.size) > (8*32)) {
       println (s"!!! WARNING !!!")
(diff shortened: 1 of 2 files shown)


src/main/scala/tilelink/Xbar.scala
```diff
@@ -67,6 +70,7 @@ class TLXbar(policy: TLArbiter.Policy = TLArbiter.roundRobin)(implicit p: Parame
     override def circuitIdentity = outputs.size == 1 && inputs.size == 1
   }
 
+  // implementation
```
michael-etzkorn, line 73: This is not a good code comment. Code comments should point out something unusual that cannot be gleamed from the code itself.

src/main/scala/tilelink/Xbar.scala
```diff
@@ -186,56 +190,105 @@ object TLXbar
...
-      (edgesOut zip reachableO).map { case (mp, reachable) =>
-        reachable && cp.client.anySupportProbe && mp.manager.managers.exists(_.regionType >= RegionType.TRACKED)
-      }.toVector}.toVector
-    val releaseIO = (edgesIn zip reachableIO).map { case (cp, reachableO) =>
-      (edgesOut zip reachableO).map { case (mp, reachable) =>
-        reachable && cp.client.anySupportProbe && mp.manager.anySupportAcquireB
-      }.toVector}.toVector
-
+    // "client" is master, "manager" is slave.
+    // Not every master need connect to every slave on every channel; determine which connections are necessary. It will generate a 2 dimensional array interconnect matrix. which represents the connectability of a master port to a slave port.
```
michael-etzkorn, line 198: The commenting of every scoped item looks to be complete overkill.

src/main/scala/tilelink/Xbar.scala
```diff
@@ -186,56 +190,105 @@ object TLXbar
...
+      (edgesIn zip reachableIO).map { 
+        case (cp, reachableO) =>
+          (edgesOut zip reachableO).map { 
+            case (mp, reachable) =>
+              // Fill true in the matrix, if exist any master support probe, and exist any slave can probe.            
+              reachable && cp.client.anySupportProbe && mp.manager.anySupportAcquireB
+        }.toVector
+      }.toVector
+
+    // Interconnect matrix of A channel from master to slave.
```
michael-etzkorn, line 247: I think this is self-explanatory and if a comment is necessary, I'd rather have one then one for each val.

michael-etzkorn (commented):
In general some code documentation is nice, but I don't think a lot of the comments added in this PR are necessary. 

This seems to breaking rules for [best practices of code comments](https://stackoverflow.blog/2021/12/23/best-practices-for-writing-code-comments/)

Mainly that comments should not duplicate code and that they should dispel confusion, not cause it.

src/main/scala/tilelink/Xbar.scala
```diff
@@ -30,8 +30,11 @@ object ForceFanout
 private case class ForceFanoutParams(a: Boolean, b: Boolean, c: Boolean, d: Boolean, e: Boolean)
 private case object ForceFanoutKey extends Field(ForceFanoutParams(false, false, false, false, false))
 
-class TLXbar(policy: TLArbiter.Policy = TLArbiter.roundRobin)(implicit p: Parameters) extends LazyModule
```
sequencer, line 33: Need a uarch doc here.

PR #3083, opened 2022-09-29
Port devices/tilelink to Chisel 3

sinofp:
**Related issue**: #3025 

**Type of change**: other enhancement

**Impact**: no functional change

**Development Phase**: implementation

**Release Notes**

PR #3104, opened 2022-10-04
Port BusBypass.scala to Chisel 3

sinofp:
**Related issue**: #3083 

**Release Notes**

### DIFF
--- a/src/main/scala/devices/tilelink/BusBypass.scala
+++ b/src/main/scala/devices/tilelink/BusBypass.scala
@@ -2,7 +2,7 @@
 
 package freechips.rocketchip.devices.tilelink
 
-import Chisel._
+import chisel3._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.tilelink._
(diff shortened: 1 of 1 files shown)


src/main/scala/devices/tilelink/BusBypass.scala
```diff
@@ -92,7 +92,7 @@ class TLBusBypassBar(dFn: TLSlavePortParameters => TLSlavePortParameters)(implic
     out0.d.ready := in.d.ready &&  bypass
     out1.d.ready := in.d.ready && !bypass
     in.d.valid   := Mux(bypass, out0.d.valid, out1.d.valid)
-    def cast(x: TLBundleD) = { val out = Wire(in.d.bits); out <> x; out }
+    def cast(x: TLBundleD) = { val out = WireDefault(in.d.bits); out <> x; out }
```
sequencer, line 95: I don't think this is OK.
`Wire(in.d.bits)` only use the hardware type and generate a `Wire` hardware, but what you are proposing here is giving `out` a non constant value with `in.d.bits`.

sequencer (changes requested):
need fix the `cast` function.

src/main/scala/devices/tilelink/BusBypass.scala
sinofp: Chisel.Wire is chisel3.WireDefault:
https://github.com/chipsalliance/chisel3/blob/fc970ca28e562f2ea3ba160963604ea3deaf3467/src/main/scala/chisel3/compatibility.scala#L72-L80

sequencer: oh fk

sequencer (approved):

sequencer:
> Chisel.Wire is chisel3.WireDefault

Sure........

PR #3108, opened 2022-10-07
Port util/HellaQueue.scala to Chisel3

CircuitCoder:

### DIFF
--- a/src/main/scala/util/HellaQueue.scala
+++ b/src/main/scala/util/HellaQueue.scala
@@ -2,7 +2,8 @@
 
 package freechips.rocketchip.util
 
-import Chisel._
+import chisel3._
+import chisel3.util._
 
 class HellaFlowQueue[T <: Data](val entries: Int)(data: => T) extends Module {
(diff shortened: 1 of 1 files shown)


sequencer (approved):

sequencer (approved):

src/main/scala/util/HellaQueue.scala
```diff
@@ -2,35 +2,36 @@
 
 package freechips.rocketchip.util
 
-import Chisel._
+import chisel3._
+import chisel3.util._
 
 class HellaFlowQueue[T <: Data](val entries: Int)(data: => T) extends Module {
   val io = new QueueIO(data, entries)
```
yqszxx, line 9: Need a `IO()` wrapper here.

The `io.count` is not connected.

CircuitCoder: Pending fix submitted in https://github.com/chipsalliance/rocket-chip/pull/3157

PR #3109, opened 2022-10-07
Port util/MultiWidthFifo to chisel3

CircuitCoder:

### DIFF
--- a/src/main/scala/util/MultiWidthFifo.scala
+++ b/src/main/scala/util/MultiWidthFifo.scala
@@ -2,18 +2,19 @@
 
 package freechips.rocketchip.util
 
-import Chisel._
+import chisel3._
+import chisel3.util._
 import freechips.rocketchip.unittest.UnitTest
 
 class MultiWidthFifo(inW: Int, outW: Int, n: Int) extends Module {
   val io = new Bundle {
-    val in = Decoupled(Bits(width = inW)).flip
-    val out = Decoupled(Bits(width = outW))
-    val count = UInt(OUTPUT, log2Up(n + 1))
+    val in = Flipped(Decoupled(Bits(inW.W)))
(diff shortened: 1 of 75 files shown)


src/main/scala/util/MultiWidthFifo.scala
```diff
@@ -23,81 +24,81 @@ class MultiWidthFifo(inW: Int, outW: Int, n: Int) extends Module {
...
-      io.in.fire() -> (size + UInt(1)),
-      io.out.fire() -> (size - UInt(nBeats))))
+      (io.in.fire && io.out.fire) -> (size - (nBeats - 1).U),
+      io.in.fire -> (size + 1.U),
+      io.out.fire -> (size - nBeats.U)))
 
-    io.count := size >> UInt(log2Up(nBeats))
-    io.out.valid := io.count > UInt(0)
+    io.count := size >> log2Up(nBeats)
+    io.out.valid := io.count > 0.U
```
sequencer, line 77: why no `UInt` here

CircuitCoder: Because we are shifting by a fixed amount. These two should be equivalent in generated FIRRTL?

src/main/scala/util/MultiWidthFifo.scala
```diff
@@ -2,18 +2,19 @@
 
 package freechips.rocketchip.util
 
-import Chisel._
+import chisel3._
+import chisel3.util._
 import freechips.rocketchip.unittest.UnitTest
 
 class MultiWidthFifo(inW: Int, outW: Int, n: Int) extends Module {
   val io = new Bundle {
```
sequencer, line 10: IO

sequencer (changes requested):

PR #3137, opened 2022-10-09
Port BusErrorUnit.scala to chisel3

Lucas-Wye:
Port BusErrorUnit.scala to chisel3

### DIFF
--- a/src/main/scala/tile/BusErrorUnit.scala
+++ b/src/main/scala/tile/BusErrorUnit.scala
@@ -2,10 +2,8 @@
 
 package freechips.rocketchip.tile
 
-import Chisel._
-import Chisel.ImplicitConversions._
-import chisel3.util.Valid
-import chisel3.DontCare
+import chisel3._
+import chisel3.util._
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.rocket._
 import freechips.rocketchip.diplomacy._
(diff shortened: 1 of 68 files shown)


src/main/scala/tile/BusErrorUnit.scala
```diff
@@ -115,19 +113,19 @@ class BusErrorUnit[T <: BusErrors](t: => T, params: BusErrorUnitParams)(implicit
     def numberRegs(x: Seq[Seq[RegField]]) = x.zipWithIndex.map {case (f, i) => (i * regWidth / 8) -> f }
 
     val omRegMap = node.regmap(numberRegs(Seq(
-      reg(cause, "cause", cause_desc),
-      reg(value, "value", value_desc),
+      reg(cause.asUInt, "cause", cause_desc),
+      reg(value.asUInt, "value", value_desc),
```
sequencer, line 117:
```suggestion
      reg(cause, "cause", cause_desc),
      reg(value, "value", value_desc),
```

sequencer (approved):

src/main/scala/tile/BusErrorUnit.scala
Lucas-Wye: If we remove ".asUInt", there will be a type mismatch

PR #3163, opened 2022-11-03
Bump to scala 2.13.10 / chisel 3.5.5

jerryz123:
**Related issue**: 

**Type of change** other enhancement

**Impact**: API modification

**Release Notes**

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -5,12 +5,12 @@ import scala.sys.process._
 
 enablePlugins(PackPlugin)
 
-val chiselVersion = "3.5.4"
+val chiselVersion = "3.5.5"
 
 lazy val commonSettings = Seq(
   organization := "edu.berkeley.cs",
   version      := "1.2-SNAPSHOT",
-  scalaVersion := "2.13.8",
+  scalaVersion := "2.13.10",
   parallelExecution in Global := false,
   traceLevel   := 15,
   scalacOptions ++= Seq("-deprecation","-unchecked"),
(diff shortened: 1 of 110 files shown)


src/main/scala/util/HeterogeneousBag.scala
```diff
@@ -10,6 +10,7 @@ final case class HeterogeneousBag[T <: Data](elts: Seq[T]) extends Record with c
   def apply(x: Int) = elts(x)
   def length = elts.length
 
+  override def className: String = stringPrefix
```
davidbiancolin, line 13:
```suggestion
  override def className: String = super.className
```

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -355,21 +355,25 @@ trait CanAttachTile {
...
-        tileClockGroup
+        domain.clockNode := tileClockGroup
       }
     })
 
     domain {
       domain.tile_reset_domain.clockNode := crossingParams.resetCrossingType.injectClockNode := domain.clockNode
-    } := clockSource
+    }
+    //domain.clockNode := clockSource
```
davidbiancolin, line 376: Are you just waiting to test before removing this?

davidbiancolin (approved):
These look sane to me.

build.sbt
```diff
@@ -10,14 +10,13 @@ val chiselVersion = "3.5.4"
 lazy val commonSettings = Seq(
   organization := "edu.berkeley.cs",
   version      := "1.2-SNAPSHOT",
-  scalaVersion := "2.12.15",
+  scalaVersion := "2.13.8",
```
jackkoenig, line 13: Any reason not to just go for 2.13.10?

jerryz123: No reason. I can bump to that

sequencer:
I personally suggest directly remove the bucket 9, and readd in the future, the ScalaFix in 213 is more strict then 212, and we should fix that in another PR.

PR #3164, opened 2022-11-04
[fix] missing PTE Cache tag MSB caused by width mismatch

SihaoLiu:
**Related issue**: 

**Type of change**: bug fix

**Impact**: Using same explicit bit-width for page cache reference/assignment to avoid incorrect MSB /reference assignment

**Release Notes**

Bug fix: Missing MSB assignment of page cache tag result in hfence.gvma has no ability to invalidate corresponding page cache entry which tag with v bit set. Padding the pte_cache_addr to `tag width - 1` fixs this problem

### DIFF
--- a/src/main/scala/rocket/PTW.scala
+++ b/src/main/scala/rocket/PTW.scala
@@ -372,8 +372,10 @@ class PTW(n: Int)(implicit edge: TLEdgeOut, p: Parameters) extends CoreModule()(
       if (s2) do_both_stages && !stage2 && !stage2_final
       else can_hit
     val tag =
-      if (s2) Cat(true.B, stage2_pte_cache_addr.padTo(tags.head.getWidth - 1))
-      else Cat(r_req.vstage1, pte_cache_addr.padTo(tags.head.getWidth - 1))
+      if (s2) Cat(true.B, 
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/PTW.scala
```diff
@@ -372,8 +372,10 @@ class PTW(n: Int)(implicit edge: TLEdgeOut, p: Parameters) extends CoreModule()(
       if (s2) do_both_stages && !stage2 && !stage2_final
       else can_hit
     val tag =
-      if (s2) Cat(true.B, stage2_pte_cache_addr)
-      else Cat(r_req.vstage1, pte_cache_addr)
+      if (s2) Cat(true.B, 
+        {if (tags.head.getWidth > stage2_pte_cache_addr.getWidth) stage2_pte_cache_addr.padTo(tags.head.getWidth - 1) else stage2_pte_cache_addr})
+      else Cat(r_req.vstage1, 
+        {if (tags.head.getWidth > pte_cache_addr.getWidth) pte_cache_addr.padTo(tags.head.getWidth - 1) else pte_cache_addr})
```
ingallsj, line 378: Would this be clearer?
```suggestion
      if (s2) Cat(true.B, stage2_pte_cache_addr.padTo(vaddrBits))
      else Cat(r_req.vstage1, pte_cache_addr.padTo(if (usingHypervisor) vaddrBits else paddrBits))
```

ingallsj (changes requested):

ingallsj (approved):

PR #3166, opened 2022-11-06
Tieoff atomics bundle in DCache.

fzi-hielscher:
On systems without atomics a dummy bundle is created. After switching to chisel3 the explicit tieoff is required to prevent an unconnected wire error during verilog generation.

**Type of change**: bug fix


jerryz123 (approved):

sequencer:
I hit this last week as well.

PR #3169, opened 2022-11-15
replace address decoder of XBar to Espresso decoder, also refactor Parameters.scala to chisel3

Lucas-Wye:


src/main/scala/diplomacy/EspressoAddressDecoder.scala
```diff
@@ -0,0 +1,169 @@
...
+import freechips.rocketchip.diplomacy.{AddressDecoder, AddressSet}
+
+// import chisel3.stage.ChiselGeneratorAnnotation
+// import circt.stage.{CIRCTTarget, CIRCTTargetAnnotation, ChiselStage, FirtoolOption}
+// import firrtl.options.TargetDirAnnotation
+// import firrtl.{AnnotationSeq, ChirrtlEmitter, EmitAllModulesAnnotation}
+
+object EspressoAddressDecoder
+{
+	type Port = Seq[AddressSet]
```
jerryz123, line 18: Use 2-space indents, not tabs

src/main/scala/tilelink/Xbar.scala
```diff
@@ -30,8 +31,11 @@ object ForceFanout
 private case class ForceFanoutParams(a: Boolean, b: Boolean, c: Boolean, d: Boolean, e: Boolean)
 private case object ForceFanoutKey extends Field(ForceFanoutParams(false, false, false, false, false))
 
-class TLXbar(policy: TLArbiter.Policy = TLArbiter.roundRobin)(implicit p: Parameters) extends LazyModule
+class TLXbar(
+  policy: TLArbiter.Policy = TLArbiter.roundRobin // use option "roundRobin" which has similar property of MUX
+  )(implicit p: Parameters) extends LazyModule
```
jerryz123, line 36: What does this comment mean? Also why does arbiter choice affect decoder

src/main/scala/tilelink/Xbar.scala
```diff
@@ -186,56 +191,105 @@ object TLXbar
...
-
+    // "client" is master, "manager" is slave.
+    // Not every master need connect to every slave on every channel; determine which connections are necessary. It will generate a 2 dimensional array interconnect matrix. which represents the connectability of a master port to a slave port.
+    val reachableIO = edgesIn.map {
+      // Each parameters from master.
+      cp => edgesOut.map { 
+        // Each parameters from managers (sinks).
+        mp => 
+          cp
+          // Master port parameters.
```
jerryz123, line 206: These comments should not be inline in the final commit.

src/main/scala/diplomacy/EspressoAddressDecoder.scala
Lucas-Wye: Thank you, I have fixed it.

src/main/scala/tilelink/Xbar.scala
Lucas-Wye: yes, these are some code from another branch, I just removed them, thanks

src/main/scala/tilelink/Xbar.scala
Lucas-Wye: has been removed

src/main/scala/diplomacy/EspressoAddressDecoder.scala
```diff
@@ -0,0 +1,47 @@
...
+
+object EspressoAddressDecoder
+{
+  type Port = Seq[AddressSet]
+  type Ports = Seq[Port]
+
+  def apply(addr: UInt, ports: Ports, mask: Seq[Boolean]): Seq[Bool] = {
+    if (ports.flatten.isEmpty) {
+      return ports.map(_ => true.B)
+    }
```
sequencer, line 18: I think it needs more comments.

src/main/scala/diplomacy/EspressoAddressDecoder.scala
```diff
@@ -0,0 +1,47 @@
...
+    // Verify the user did not give us an impossible problem
+    ports.combinations(2).foreach { case Seq(x, y) =>
+      x.foreach { a =>
+        y.foreach { b =>
+          require(!a.overlaps(b), s"Ports cannot overlap: $a $b")
+        }
+      }
+    }
+    val maxBits = log2Ceil(1 + ports.flatMap(_.map(_.base)).max)
```
sequencer, line 27: same here...

src/main/scala/diplomacy/EspressoAddressDecoder.scala
```diff
@@ -0,0 +1,47 @@
...
+          port.map { range =>
+            (AddressSetToBitPat(range, maxBits), new BitPat(value = 1 << i, mask = -1, maxBits))
+          }
+        case _ => Seq()
+      }
+    }
+    if (truthMap.isEmpty) {
+      return ports.map(_ => true.B)
+    }
+    val dontCarePat: BitPat = new BitPat(value = 0x0, mask = 0x0, maxBits)
```
sequencer, line 40: IIRC
```suggestion
    val dontCarePat: BitPat = BitPat.dontCare(maxBits)
```

src/main/scala/diplomacy/EspressoAddressDecoder.scala
```diff
@@ -0,0 +1,47 @@
...
+    decoder(addr, TruthTable(truthMap, dontCarePat)).asBools.take(ports.length)
```
sequencer, line 41: `take` here scared me, in which case, the decoder width won't match `ports` width

src/main/scala/diplomacy/EspressoAddressDecoder.scala
```diff
@@ -0,0 +1,47 @@
...
+  }
+
+  def AddressSetToBitPat(addrs: AddressSet, XLen: Int): BitPat = {
+    new BitPat(addrs.base, ~addrs.mask, XLen)
+  }
```
sequencer, line 46: nitpick: add public api to `AddressSet`: `toBitPat(xLen: Int)` might be better, if not, make this function private.

sequencer (changes requested):

src/main/scala/diplomacy/EspressoAddressDecoder.scala
```diff
@@ -0,0 +1,47 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.diplomacy
+
+import chisel3._
+import chisel3.util.BitPat
+import chisel3.util.log2Ceil
+import chisel3.util.experimental.decode._
+
+object EspressoAddressDecoder
```
sequencer, line 10: I think it should be the name `EspressoAddressDecoder`: if user not installing `espresso`, this is gone be `QMCDecoder` from chisel

sequencer: I think you can move logic in `EspressoAddressDecoder` to `AddressDecoder`, since after this PR is merged, original `AddressDecoder` is going to be dead code.

sequencer:
FYI:
Here are some examples to use chisel decoder as `AddressDecoder`:
Example from @jerryz123 
https://github.com/ucb-bar/constellation/blob/216eee92be30a8d24a3e3369b6f23c6f43b0d3a3/src/main/scala/router/RouteComputer.scala#L29
Example from @sequencer 
https://github.com/chipsalliance/chisel3/blob/2b48fd15a7711dcd44334fbbc538667a102a581a/src/test/scala/chiselTests/util/BitSetSpec.scala#L90-L116

sequencer:
`decoder.bitset` provide a relatively clear API for `AddressSet`. Basically an `AddressSet` is a `BitSet`(but with unknown width IIRC), so if you can use `decoder.bitset` API, I bet it will be much clear.

src/main/scala/diplomacy/EspressoAddressDecoder.scala
Lucas-Wye: Because the decoder width is determined by maxBits, which is the maximum number of bit in all input address. As a result, the decoder width is usually larger than port width.

sequencer:
This is blocked by chipsalliance/chisel3#2883, we should merge this into `develop` branch

sequencer:
wait for chipsalliance/tilelink

PR #3171, opened 2022-11-18
[WIP] Remove rowBits/cacheBlockBytes params from cache | add tile-level masterPortBeatBytes

jerryz123:
We want to move towards decoupling tile-local port widths from the global interconnect widths.
This adds `masterPortBeatBytes` to `BaseTile`. At tile instantiation and connection, a `WidthWidget` is inserted such that the internal bus width of the tile is `masterPortBeatBytes`. `rowBits` in L1Cache params are now effectively references to this, so explicitly setting `rowBits = p(SystemBusKey).beatBits` is no longer necessary.

This PR also removes many local instances of `blockBytes` params. It seems exceedingly unlikely that someone will need to build a system with varying cache block size, so we should just rely on the global variable.

**Related issue**: 

**Impact**:  API modification

**Release Notes**

### DIFF
--- a/src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
+++ b/src/main/scala/devices/tilelink/CanHaveBuiltInDevices.scala
@@ -5,6 +5,7 @@ package freechips.rocketchip.devices.tilelink
 import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.tilelink._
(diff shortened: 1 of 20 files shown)


sequencer:
Generally this increase the CDE dependencies in the rocket core. Since I'm trying to splitting rocket tile out from rocket-chip SoC generator, I think we should try to resolve the CDE parameter at `LazyModule` layer. Thinking how to make a tradeoff...

PR #3172, opened 2022-11-19
Fix "dump-start" in FSDB-mode

hngenc:
When `USE_FSDB=1`, `+dump-start` doesn't do anything. It seems VCS just begins dumping signals as soon `fsdbDumpfile`/`fsdbDumpVars` is called, rather than waiting for `fsdbDumpon`.

My simple workaround is to just call  `fsdbDumpOff` explicitly right away. The waveform dumping can resume when `fsdbDumpOn` is called.

**Related issue**: 

**Impact**: no functional change (a bug fix)

**Release Notes**


jerryz123:
Does this affect behavior when `dump-start` is not specified?

PR #3179, opened 2022-11-28
rvc: allow decode C.MV to ADDI with useAddiForMv

poemonsense:
**Related issue**: 

**Release Notes**

This commit changes how compressed move instructions are decoded. It adds an useAddiForMv argument to RVCDecoder to allow the user to specify how C.MV is decoded.

From RISC-V spec, C.MV expands to a different instruction than the canonical MV pseudoinstruction, which instead uses ADDI. Implementations that handle MV specially, e.g. using register-renaming hardware, may find it more convenient to expand C.MV to MV instead of ADD, at slight additional hardware cost.

We add a configuration option for decoding C.MV to ADDI.

### DIFF
--- a/src/main/scala/rocket/RVC.scala
+++ b/src/main/scala/rocket/RVC.scala
@@ -16,7 +16,7 @@ class ExpandedInstruction extends Bundle {
   val rs3 = UInt(5.W)
 }
 
-class RVCDecoder(x: UInt, xLen: Int) {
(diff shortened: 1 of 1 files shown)


sequencer:
LGTM.

sequencer (approved):

jerryz123 (approved):

PR #3182, opened 2022-12-07
Add riscv-tests/riscv-arch-tests framework based on mill

ZenithalHourlyRate:
**Related issue**: #2953

**Release Notes**

This PR enhances the original mill build system to support automatic tests. This PR features

* emulator generation with different config
* test building
* flexible combination of tests via trait
* `shell.nix` as the replacement of the original `rocket-tools`, which could be used in CI and is deterministic, see [workflow.yml in playground](https://github.com/chipsalliance/playground/blob/5994344ba065f160011531ad7dda6b85615fee66/.github/workflows/bump.yml#L51)

This has been inspired by #2991, https://github.com/chipsalliance/playground/pull/46 and [sequencer/rocket CI](https://github.com/sequencer/rocket/blob/master/build.sc)

One thing worth mentioning is that with Zk/Zb test in arch-test, #2950 can be automatically tested.

### DIFF
--- a/arch-test/emulator/emulator_isa_rv32i.yaml
+++ b/arch-test/emulator/emulator_isa_rv32i.yaml
@@ -1,0 +1,6 @@
+hart_ids: [0]
+hart0:
+  ISA: RV32I
+  physical_addr_sz: 32
+  User_Spec_Version: '2.3'
+  supported_xlen: [32]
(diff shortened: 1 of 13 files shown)


ZenithalHourlyRate:
Discussion in RC meeting concluded that this PR should be separated into several parts. I'll take this in multiple steps.

1. Add mill script for emulator generation
2. Add mill script for riscv-tests
3. Add mill script for riscv-arch-test

All these steps should be linux distribution agnostic. That is, it should work with either NixOS (for CI) or any other distribution that uses rocket-tools, which we have recently bumped, so that users can also smoothly run tests in their environment once they exports the `RISCV` env var.

Then I will add a CI with nix related files in another PR.

PR #3185, opened 2022-12-20
Add interface for setting custom ISA strings in DTS

jerryz123:
Previously misa always reported `rv64gcx`. Now misa will have the 'X' field set only if customIsaExt is defined, and the DTS string will contain custom extensions. Rocket is now `rv64gc_Xrocket` in the DTS. Could this cause any issues in  DTS-parsing code?


jerryz123:
We should add an option to support building a plain rv64gc

sequencer (approved):

PR #3186, opened 2022-12-21
chore: bump maven release for sbt and mill

tianrui-wei:
**Related issue**: 

**Impact**: API modification

**Release Notes**

Bumps the Maven release number according to actual release

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -9,7 +9,7 @@ val chiselVersion = "3.5.5"
 
 lazy val commonSettings = Seq(
   organization := "edu.berkeley.cs",
-  version      := "1.2-SNAPSHOT",
+  version      := "1.6.0",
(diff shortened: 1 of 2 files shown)


sequencer:
LGMT, although RC doesn't release to maven(for now)

sequencer (approved):

michael-etzkorn:
Also, this would be more of a 1.6-SNAPSHOT, since 1.6 is https://github.com/chipsalliance/rocket-chip/commit/44b0b8249279d25bd75ea693b725d9ff1b96e2ab, but we can worry about that when we start version branches back up.

tianrui-wei:
I agree. In retrospect that's a good idea. We should discuss automating the release and publishing in the next rocketchip meeting.

PR #3189, opened 2022-12-28
mill: add support of emulator generation

ZenithalHourlyRate:
**Related issue**: follow up of #3182

**Type of change**: other enhancement

**Impact**: API addition (no impact on existing code)

**Development Phase**: implementation

**Release Notes**

Generate emulator from mill

PR #3191, opened 2022-12-28
mill: add support of running riscv-arch-test

ZenithalHourlyRate:
**Related issue**: follow up of #3182, and this PR depends on #3189

**Release Notes**

Supports [riscv-arch-test](https://github.com/riscv-non-isa/riscv-arch-test)

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -325,3 +325,124 @@ class RunableTest(top: String, config: String, suiteName: String, exclude: Strin
     }
   }
 }
+
+object `runnable-arch-test` extends mill.Cross[ArchTest](
+  ("freechips.rocketchip.system.TestHarness", "freechips.rocketchip.system.DefaultConfig", "64", "RV64IMAFDCZicsr_Zifencei"),
+  ("freechips.rocketchip.system.TestHarness", "freechips.rocketchip.system.DefaultRV32Config", "32", "RV32IMAFCZicsr_Zifencei"),
+)
+class ArchTest(top: String, config: String, xlen: String, isa: String) extends Module {
+  def ispecString = T {
+    // format: off
+    s"""hart_ids: [0]
+       |hart0:
+       |  ISA: ${isa}
+       |  physical_addr_sz: 32
+       |  User_Spec_Version: '2.3'
+       |  supported_xlen: [${xlen}]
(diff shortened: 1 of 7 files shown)


ZenithalHourlyRate:
Rebased and ready for review now.

This is not added in the new CI PR (#3192) because it takes several hours to finish, even on a powerful machine. Before we have custom runners or do some optimization on the way to run arch-test, we have to manually run them.

That's because there are hundreds of tests, and some test has a big binary for the emulator to load via the debug port, usually taking half an hour to one hour. For example, 
`rv32i_m/F/src/fnmadd_b15-01.S/my.elf` is of size 4.8M, and it have not finished running after 1h45min in my machine.

```
# number of arch-test
# DefaultConfig
$ find . -name "my.elf" | grep dut | wc -l
252
# DefaultRV32Config
$ find . -name "my.elf" | grep dut | wc -l
236
```

One way to handle the big binary loading problem is to use markrom and readmemh to load the binary by the emulator, then a bootloader is needed to move the binary into the SimAXIRAM for actually running.

ZenithalHourlyRate:
It also turned out the rocket did not pass the `ebreak` arch-test with the latest spike as reference.

This is related to spec change https://github.com/riscv/riscv-isa-manual/commit/51522f98b5cae989c7d161e00b20a08483a2301e, where [spike changed behavior](https://github.com/riscv-software-src/riscv-isa-sim/commit/be5af59bcaca4dc90a0d276e1eccc4f4896b0373). One thing worth mentioning is that after reverting that spike commit, the test will pass.

The current behavior still complies with the spec, however. There are people discussing this issue in https://github.com/riscv-non-isa/riscv-arch-test/issues/222

There are two ways to handle this:
1. disable this test
2. change rocket behavior

sequencer:
I don’t think we should change rocket core behavior if it is still comply with the latest riscv-spec. 
Basically, if using arch-test for CI in the future, we should mask this test off until arch-test fix this bug.

ZenithalHourlyRate:
Could be merged now. Tested in https://github.com/chipsalliance/rocket-chip/pull/3201#issuecomment-1374416108

sequencer (approved):
LGTM

PR #3193, opened 2022-12-30
Populate tval with pc when ebreak

ZenithalHourlyRate:
Ref to https://github.com/riscv/riscv-isa-manual/commit/51522f98b5cae989c7d161e00b20a08483a2301e

However this change does not only affect `ebreak` so I need to check further.

**Related issue**: https://github.com/chipsalliance/rocket-chip/pull/3191#issuecomment-1367788856

**Impact**: API modification

**Release Notes**

Populate tval with pc when ebreak

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -966,6 +966,7 @@ class CSRFile(
   assert(!reg_singleStepped || io.retire === 0.U)
 
   val epc = formEPC(io.pc)
+  val tval = Mux(insn_break, epc, io.tval)
 
   when (exception) {
     when (trapToDebug) {
@@ -993,7 +994,7 @@ class CSRFile(
       reg_vsstatus.spp := reg_mstatus.prv
       reg_vsepc := epc
       reg_vscause := Mux(cause(xLen-1), Cat(cause(xLen-1, 2), 1.U(2.W)), cause)
-      reg_vstval := io.tval
+      reg_vstval := tval
       reg_vsstatus.spie := reg_vsstatus.sie
       reg_vsstatus.sie := false.B
       new_prv := PRV.S.U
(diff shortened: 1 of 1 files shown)


sequencer:
We should have a mature tests for this.

sequencer:
We should have a mature tests before merging this.

ZenithalHourlyRate:
> However this change does not only affect `ebreak` so I need to check further.

Checked. The original fix also affected `ecall`.

> We should have a mature tests before merging this.

This version passes the privileged arch-test. Details are in the commit message.

This can be merged after arch-test is merged (I think enabling IMAC_Zicsr_Zifencei in CI will finish running in an acceptable time)

ZenithalHourlyRate:
This has been tested by https://github.com/chipsalliance/rocket-chip/pull/3191#issuecomment-1374416205

sequencer:
LGTM based on my own understanding to RTL and spec. 
It seems passed all tests, however I’m still not so confident with the change to CSR. Ask @jerryz123 for review.
if possible I think we need a cosim framework and opensbi+kernel as payload.

sequencer (approved):

jerryz123:
I wonder if it would be cleaner to set `io.tval` properly inside the core, instead of overriding the value written inside the CSR.

ZenithalHourlyRate:
> I wonder if it would be cleaner to set `io.tval` properly inside the core, instead of overriding the value written inside the CSR.

https://github.com/chipsalliance/rocket-chip/blob/a5fa4e4781bf2882f2de26d7ba6dc77561a3934b/src/main/scala/rocket/CSR.scala#L956

We already have similar behavior that exception is overrided inside the CSR.

> which can be carried by the existing logic into `csr.io.tval`.

To use the existing logic, we need to assert `tval_valid` thus asserting `wb_xcpt`. The appropriate place to raise the exception is `id_xcpt` but the exception handling logic will be changed accordingly. Another way is to assert `tval_valid` with another condition (e.g. insn == ebreak from decode); However, that is already done in CSR, or we need to carry the extra decoded signal from id staged to wb stage.

jerryz123 (approved):
LGTM

PR #3194, opened 2022-12-31
Use c++17 for emulator compile

jerryz123:
Recent fesvrs require c++17


sequencer (approved):
I'm tuning emulator recently...
The tricky thing is I found VCS on centos 7(which VCS lives on) doesn't support C++17 since the system CC is too old. Thus requiring user providing CC in EDA servers.
We can merge this, after switching to new CI infra, we should also have a VCS CI.

PR #3195, opened 2022-12-31
Don't check probeacks in TLMonitor.legalizeCDSource

jerryz123:
Bug fix. Per the spec: 

> Because Channel C responses to Channel B requests are routed to a single slave and uniquely
identified to an ongoing operation based on c address, we can further relax the uniqueness
restriction on Channel B requests, requiring only that the combination of b source and b address
be uniquely inflight. This relaxation is necessary in order to simultaneously probe multiple masters
on the same address, while also probing the same master on multiple addresses. Channel C 
responses may use any c source associated with the sender; this signal is ignored for those
message types.


sequencer (approved):
I remember I encountered this bug two years before.....................................

PR #3203, opened 2023-01-07
[WIP] mill: enable ma_data

ZenithalHourlyRate:
**Related issue**: #3202

ma_data test is disabled for both `rv{32,64}ui-{p,v}`. However, at least rv{32,64}ui-p-ma_data will pass according to https://github.com/riscv-software-src/riscv-tests/issues/419. This PR tries to enable all of them. If the v part still fails, this PR would change to enable ma_data for rv{32,64}-p only

**Release Notes**


ZenithalHourlyRate:
Timeout for rv64ui-p-ma_data

jerryz123:
IMO we should not enable the ma_data tests. They are designed to test hardware support for misaligned load/store, not the emulated support.
That the tests might pass on some Rocket configs seems like a quirk to me. We should not prioritize making those tests run.

Issue #3207, opened 2023-01-09
Implementation of Zicntr


ZenithalHourlyRate:
**Type of issue**: feature request

**Impact**: API modification

**Development Phase**: proposal

**Other information**
 

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

**What is the current behavior?**

For the zicntr extension, we have implemented the `cycle` and `instret` CSR (see https://github.com/chipsalliance/rocket-chip/commit/3ff921de046367dbd114150bfca471b5eb0c1550). However, to fully support zicntr, we still need to implement a `time` CSR.

This can not be simply done by connecting `mtime` to `time` as `mtime` is MMIO and currenly it is provided by clint.

**What is the expected behavior?**

There are two ways to provide the `time` CSR

* Hardwire `time` to `reg_cycle`. The spec said simple platform can do this; also, we have already used this fact by `csr.io.time`.
* Hardwire the RTC counter in [HasRTCModuleImp](https://github.com/chipsalliance/rocket-chip/blob/d706071fdc39233277a74d5332f566aa8358010e/src/main/scala/subsystem/RTC.scala#L24) to the CSR. However, not all possible config has the RTC module and CLINT.

The latter can be compliant with the spec on

> The real-time clocks of all harts must be synchronized to within one tick of the real-time clock.

Another final way is to provide `time` CSR from RTC if there is RTC; and fallback to `reg_cycle` if there is none. That will need to pass the `CLINTKey` to the `CSRFile`.

jerryz123:
I think just doing the simple solution is the best. Adding another clock crossing between the RTC and the tile would be additional complexity.

ZenithalHourlyRate:
Closed by #3208

ZenithalHourlyRate:
https://github.com/chipsalliance/rocket-chip/pull/3208#issuecomment-1379697646

PR #3208, opened 2023-01-09
Implement time[h] CSR using cycle[h] CSR

ZenithalHourlyRate:
**Related issue**: #3207

**Impact**: API modification

**Release Notes**

Implemented time[h] CSR; Support Zicntr extension.

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -686,12 +686,14 @@ class CSRFile(
       read_mapping += CSRs.mcounteren -> read_mcounteren
     }
     read_mapping += CSRs.cycle -> reg_cycle
+    read_mapping += CSRs.time -> reg_cycle
     read_mapping += CSRs.instret -> reg_instret
 
     if (xLen == 32) {
       read_mapping += CSRs.mcycleh -> (reg_cycle >> 32)
       read_mapping += CSRs.minstreth -> (reg_instret >> 32)
       read_mapping += CSRs.cycleh -> (reg_cycle >> 32)
+      read_mapping += CSRs.timeh -> (reg_cycle >> 32)
       read_mapping += CSRs.instreth -> (reg_instret >> 32)
     }
   }


jerryz123 (approved):

sorear:
I think that this will violate the spec requirements that `The period should be constant within
a small error bound.` and `The real-time clocks of all harts must be synchronized to within one tick of the real-time clock.`, since `reg_cycle` is affected by WFI and uses a per-core rather than global clock.

ZenithalHourlyRate:
> I think that this will violate the spec requirements that `The period should be constant within a small error bound.` and `The real-time clocks of all harts must be synchronized to within one tick of the real-time clock.`, since `reg_cycle` is affected by WFI and uses a per-core rather than global clock.

I agree with this concern.

I also found the following requirement in the privileged spec

> The time CSR is a read-only shadow of the memory-mapped mtime register.

It seems that we must connect the timer in CLINT to the CSRFile in some way.

One way is to connect the `int_rtc_tick` to CLINT and CSRFile of each core; and instantiate counters for all of them; 

Another way is to instantiate the `time` counter in the top module instead of CLINT and connect it across boundaries of modules all the way down to CSRFile and CLINT.

Both way may lead to high-fanout (especially when there are multiple cores) and have timing concern.

> Implementations can convert reads of the time and timeh CSRs into loads to the memorymapped mtime register

Another solution is a comment in the privileged spec.

However, we have no infra to support such behavior.

One way is to convert rdtime into ld/lw in decode stage, but we have to handle the privilege issue (e.g. reading mtime in user mode with virtual address)

Another way is to issue ld/lw in CSRFile and stall the pipeline but we have no support for that now. This infra is actually needed. E.g. when implementing Zkr (#2906), the `seedcsr` is needed, and in actual systems seeds are usually produced by some external devices. 

The conclusion is that there is no easy implementation and I'll leave the issue open.

I'll revert this commmit.

PR #3210, opened 2023-01-11
remove SBT

sequencer:
**Related issue**: 

**Release Notes**


regression/Makefile
```diff
@@ -335,19 +335,3 @@ emulator-jtag-dtm-tests-64 : $(EMULATOR_JTAG_DTM_64_TEST_STAMPS)
...
-# Target to run Scalafix 'sbt scalafix'
-.PHONY: scalafix
-scalafix: stamps/other-submodules.stamp $(all_srcs)
-	(cd $(base_dir) && $(SBT) $(SBT_OPTS) scalafix)
-
-# Target to run Scalafix 'sbt scalafix --check', which will not actually modify
-# Scala files.
-.PHONY: scalafix-check
-scalafix-check: stamps/other-submodules.stamp $(all_srcs)
-	(cd $(base_dir) && $(SBT) $(SBT_OPTS) 'scalafix --check')
```
sequencer, line 353: Remove these for now, I'll add them in later commits.

jerryz123:
Why is this needed?

sequencer:
This is basically the build system issue:
1. We are currently using SBT+Make for build system, which is certainly a bad practice, to avoid that, SiFive use [wake](https://github.com/sifive/wake) for their build tool, kudo @terpstra for his amazing job. I like that, but no one in opensource world is using it. I think mill is a replacement to it, but lakes some filesystem utilities, but it's enough for most of small projects.
2. SBT is hard to maintain and edit for new users, see [Lihaoyi's blog](http://www.lihaoyi.com/post/SowhatswrongwithSBT.html), and [Martin Odersky's post](https://docs.google.com/document/d/1QdtRJGxlKTiXcAxsjXWLtWVzeZyUzGVDh8oOYOiuhwg/edit). 
3. In the future rocketchip development, we wanna have a fine granularity of packages, rather than a fat jar that including all bus ip, core, and everything, we need a neat way to split packages, basically maintaining two build system(especially SBT) is really hard when we are going to split repos. I can easily do in the [playground](https://github.com/chipsalliance/playground) project, but SBT is always blocking me to achieve so...

Thus I'm trying to provide a single build system in mill, managing all the build cache artifacts in the `out` folder. Here are the pros in the future:
1. We can cache `out` folder in the build system, we will never need to rerun tests if PR doesn't change generated RTL. 
2. We can start to decouple rocketchip into smaller projects for releasing.
3. Single build system makes depending on rocketchip from source much easier.
4. We can totally switch to mill as our RTL build system, from Scala compiling, RTL elaboration, emulator building, and tests running, they can be totally achieved in mill, see example here:
https://github.com/chipsalliance/rocket-chip/blob/master/build.sc
 
If other projects are using rocket-chip from source, we can provide a `common.sc` for them to include to build with mill, or using `mill rocketchip.assembly` to provide the fat jar and depend on it.

jerryz123:
Can we maintain both build systems for the time being? 

Chipyard is unlikely to ditch SBT, and we need to be able to pull in rocket-chip as a dependency. My proposal for this migration is:
- Switch all user-facing Makefiles to use mill. All make commands should behave identically as before.
- Switch legacy regression Makefile to use mill
- Maintain existing build.sbt, project/ directories
- Add CI to check that `sbt compile` works only.

sequencer:
> Chipyard is unlikely to ditch SBT, and we need to be able to pull in rocket-chip as a dependency. 

Will depending on `jar` be able to work with Chipyard?

> Switch all user-facing Makefiles to use mill. All make commands should behave identically as before.

Agree, I think Makefiles should be eventually deprecated as well, at least define some user targets.

> Switch legacy regression Makefile to use mill

Yes, we to do this.

> Maintain existing `build.sbt`, `project/` directories
Add CI to check that sbt compile works only.

I think we should remove sbt eventually, otherwise I cannot start to spliting project(don't wanna play with a funky build systems), we can have a deprecation flow to wait for chipyard finish the work around:
1. migrate to `mill` 
2. depend on a `jar`
3. migrate rocketchip SBT scripts to chipyard

For 1 or 2, I can send PRs to chipyard for speed up the migration.

jerryz123:
I'm even hesitant to switch dependencies to a published jar in chipyard. Chipyard's build system needs to guarantee the following:
 - The ability to quickly make source-code changes in rocket-chip
 - Automatic-recompilation from source-code changes
 - Support for maintaining many copies of chipyard with different source-code changes.

If such behavior can be provided through a jar dependency, then we can consider it. But we'd have to see the changes.

I'd rather not deprecate the makefiles due to how accessible they are to new users. Just make the targets wrap mill commands.

Is it possible to split parts of rocket-chip out without dropping SBT? For example, switching to released version of hardfloat/api-config in the sbt and mill flows instead of submodule'd sources. I don't think maintaining a build.sbt for such a system is that funky.

sequencer:
> I'd rather not deprecate the makefiles due to how accessible they are to new users. Just make the targets wrap mill commands.

Sure, we can define phony target for user to generate verilog and emulator.

> Is it possible to split parts of rocket-chip out without dropping SBT? For example, switching to released version of hardfloat/api-config in the sbt and mill flows instead of submodule'd sources. I don't think maintaining a build.sbt for such a system is that funky.

> Is it possible to split parts of rocket-chip out without dropping SBT?

Yes, that's possible, but I think maintaining it is a duplicated work, I wonder if it is possible to move the `build.sbt` to Chipyard?

> For example, switching to released version of hardfloat/api-config in the sbt and mill flows instead of submodule'd sources.

Like what I did in [build.sc](https://github.com/chipsalliance/rocket-chip/blob/70ed8b63bd3780b2c6091a9e9fbfd12ab9145200/common.sc#L70-L72), I'd like to let user select in-source building or depending on Ivy. This is a very useful feature that developers build everything from source, and users can pull dependencies from maven.

jerryz123:
> Yes, that's possible, but I think maintaining it is a duplicated work, I wonder if it is possible to move the build.sbt to Chipyard?

Chipyard does not need the build.sbt here to work. But I would still like to keep it here for some time for now. Chisel maintains both mill and sbt flows as well.

We can consider switching Chipyard to mill, but we need to see a concrete examples of using rocketchip as a mill dependency where the neccessary functionality is achieved.

sequencer:
> Chipyard does not need the build.sbt here to work. But I would still like to keep it here for some time for now. Chisel maintains both mill and sbt flows as well.

Emmm, I'm targeting removing SBT from Chisel as well, see the [dev-meeting doc](https://docs.google.com/document/d/1BLP2DYt59DqI-FgFCcjw8Ddl4K-WU0nHmQu0sZ_wAGo/edit#heading=h.oxer358qsbxh)

> We can consider switching Chipyard to mill, but we need to see a concrete examples of using rocketchip as a mill dependency where the neccessary functionality is achieved.

OK, I'll create an draft PR for Chipyard for SBT replacement.

jerryz123:
> Emmm, I'm targeting removing SBT from Chisel as well, see the [dev-meeting doc](https://docs.google.com/document/d/1BLP2DYt59DqI-FgFCcjw8Ddl4K-WU0nHmQu0sZ_wAGo/edit#heading=h.oxer358qsbxh)

Not until post chisel3.6. I think we can drop SBT eventually, but I'd like some transition period to make it easier to migrate everything.

abejgonzalez:
(Chipyard dev. here)

A few points to discuss / things of note (in addition to what Jerry mentioned about using source dependencies v.s. fat jars):
- Another concern is moving FireSim to Mill (https://github.com/firesim/firesim). Since Chipyard and FireSim are intertwined the FireSim folks need to be in the loop and possibly FireSim needs to be migrated (I'll bring it up in the next FireSim meeting).
- In reference to the comment below: Can you describe a bit more about a user can use the source v.s. maven dep? How would you enable this? How do you keep the source dependency matching the same commit as the maven dep (and any checks for this)? I think this feature has been wanted in Chipyard for a while.

> Like what I did in [build.sc](https://github.com/chipsalliance/rocket-chip/blob/70ed8b63bd3780b2c6091a9e9fbfd12ab9145200/common.sc#L70-L72), I'd like to let user select in-source building or depending on Ivy. This is a very useful feature that developers build everything from source, and users can pull dependencies from maven.

sequencer:
> Another concern is moving FireSim to Mill (https://github.com/firesim/firesim). Since Chipyard and FireSim are intertwined the FireSim folks need to be in the loop and possibly FireSim needs to be migrated (I'll bring it up in the next FireSim meeting).

actually they are not intertwined in a fine package granurity see the dependencies which maintained in [playground](https://github.com/chipsalliance/playground/blob/master/build.sc#L226)

> Can you describe a bit more about a user can use the source v.s. maven dep? How would you enable this? 

enable this by adding a common package, let user define using `Dep` or`Module`, see [playground](https://github.com/chipsalliance/playground/blob/a349461f600ee46f1e20934dcff999e41a224be5/build.sc#L68) here for chisel srouce depending on firrtl and [rocket-chip](https://github.com/chipsalliance/rocket-chip/blob/06a2a037ca56072492ee5ddfaf9dc46f0365bc40/common.sc#L64) depend on published chisel. 

How do you keep the source dependency matching the same commit as the maven dep (and any checks for this)? 

> I don’t have any example here but that’s simple, there are different ways: 
ask submodule to follow the released branch, git will handle it for you. 
override compile to depend on the warning target. Use git described to warn user if not one the correct branch.

sequencer:
cherry-picked in #3243

PR #3224, opened 2023-01-14
optimize: reduce time complexity of ALU by eliminating Mux

samadpls:
**Related issue**: 

**Impact**:   API modification

Optimize: Reduce time complexity of ALU by eliminating Mux
- Replaced the Mux used to invert 'in2' when the function is subtraction with bit-wise operations.
- This eliminates the need for the Mux and reduces the number of operations required, leading to a reduction in time complexity.

**Release Notes**

This optimization improves the performance of the ALU module by reducing the time complexity of the function selection

### DIFF
--- a/src/main/scala/rocket/ALU.scala
+++ b/src/main/scala/rocket/ALU.scala
@@ -59,9 +59,9 @@ class ALU(implicit p: Parameters) extends CoreModule()(p) {
   })
 
   // ADD, SUB
-  val in2_inv = Mux(isSub(io.fn), ~io.in2, io.in2)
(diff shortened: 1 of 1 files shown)


jerryz123:
Do you have numbers describing the effect this on the ALU?

jerryz123 (changes requested):
I prefer to not change this micro-architecture. #2950 will make it easier for rocket to support custom ALU implementations.

sequencer:
This change seem introduce another adder by adding `-`, it eliminated Mux for better performance, but the trade off is additional adder in ALU. Since Rocket is a small core that I'd like to keep the original uarch for balance.

Issue #3225, opened 2023-01-14
Failed to compile the emulator - g++ version for `v1.6` release ?


pcotret:
**Type of issue**: bug report 

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
```bash
git clone https://github.com/chipsalliance/rocket-chip
cd rocket-chip
git checkout v1.6
git submodule update --init
#(compile rocket-tools)
cd emulator
make
```

**What is the current behavior?**
 I'm working with the `v1.6` release. I was trying to compile the emulator (setup given below).
 - I had to change a flag in the emulator Makefile: https://github.com/chipsalliance/rocket-chip/pull/3194 (not merged yet when `v1.6` was released).

Error log while compiling the emulator:
```
g++  -I.  -MMD -I/home/test/Documents/rocket-chip/emulator/verilator/install/share/verilator/include -I/home/test/Documents/rocket-chip/emulator/verilator/install/share/verilator/include/vltstd -DVM_COVERAGE=0 -DVM_SC=0 -DVM_TRACE=0 -faligned-new -fcf-protection=none -Wno-bool-operation -Wno-sign-compare -Wno-uninitialized -Wno-unused-but-set-variable -Wno-unused-parameter -Wno-unused-variable -Wno-shadow     -O1 -std=c++17 -I/opt/riscv_rocket/include -DVERILATOR -DTEST_HARNESS=VTestHarness -include /home/test/Documents/rocket-chip/src/main/resources/csrc/verilator.h -include /home/test/Documents/rocket-chip/emulator/generated-src/freechips.rocketchip.system.DefaultConfig.plusArgs -I/home/test/Documents/rocket-chip/emulator/generated-src -include /home/test/Documents/rocket-chip/emulator/generated-src/freechips.rocketchip.system.DefaultConfig/VTestHarness.h  -DVL_THREADED -std=gnu++14  -c -o SimDTM.o /home/test/Documents/rocket-chip/src/main/resources/csrc/SimDTM.cc
In file included from /opt/riscv_rocket/include/fesvr/memif.h:10,
                 from /opt/riscv_rocket/include/fesvr/htif.h:6,
                 from /opt/riscv_rocket/include/fesvr/dtm.h:4,
                 from /home/test/Documents/rocket-chip/src/main/resources/csrc/SimDTM.cc:3:
[...]

PR #3226, opened 2023-01-16
Bump to latest verilator and rocket-tools

ZenithalHourlyRate:
* Bump verilator to 5.004; However, as spike recently needs c++17, and verilator selects the c++ version based on Linux distribution, we have to manually set it by CFG_CXXFLAGS_STD_NEWEST

* Bump riscv-tools to latest; This introduce the c++17 spike.

**Related issue**: #3225, #3194

Personally I also encounter https://github.com/verilator/verilator/pull/2505 as I'm using new bison and old verilator

This is also related to #3210 because we already has a mill flow and this Makefile flow is proposed to be removed. However, users are still using the Makefile flow so we need to maintain it for a while.

I think we do not need to compile verilator by us for now as any reasonably new verilator (e.g. from Linux distribution) should be ok (unlike years ago) 


jerryz123:
Why bump so high? Is 4.226 ok?

ZenithalHourlyRate:
> Why bump so high? Is 4.226 ok?

Bumped to 4.226. But I do not know why the sbt CI does not run.

jerryz123 (approved):

Timmmm:
Thanks, this worked for me. Didn't even need to do a clean build!

Issue #3228, opened 2023-01-16
Upgrade scripts to Python 3


Timmmm:
**Type of issue**: other enhancement
**Impact**: no functional change
**Development Phase**: proposal

There are a few Python scripts in this repo that use *Python 2*. This isn't installed by default in most repos anymore and I think it's even being removed from some. I don't really want to have my system unless some scripts accidentally use it - a lot of Python code (including these scripts!) just runs `python` which is sometimes Python 2 and sometimes Python 3, so you don't know which one it wants.

Fortunately the scripts seem to be quite small so it should be easy to update them.

See also #2629 - I also ran into this error and it's quite annoying because it doesn't say where the error originates from. Even worse, Ubuntu suggests installing `python-is-python3` which I'm guessing based on the name will just link `python` to `python3` and things will break even more.

Timmmm:
In fact even if you `apt-get install python2` it won't give you a `python` command. So at the very least you should change all the `#!/usr/bin/env python` lines to `python2`.

ZenithalHourlyRate:
Actually vlsi_mem_gen and vlsi_rom_gen are in python3 style and should suffice your use. I'll update the shebang to explicitly use python3.

As for trace script, I do not have test cases for them so I can not cover them.

jerryz123:
@ZenithalHourlyRate it would substantially improve test coverage to add the tracegen scripts to CI

sequencer:
FYI those script has been removed by 
https://github.com/chipsalliance/firrtl/pull/2202
and MFC has corresponding features, I think those scripts should be removed rather than bumping to py3

ZenithalHourlyRate:
> those script has been removed

At least vlsi_rom_gen is not implemented by firrtl.

I think we can bump to python3 so that users of old versions could cherry-pick that commit. Then we add the flow from firrtl and remove mem gen scripts.

The mill flow currently does not emit behav sram. It just outputs big reg arrays. That will be fixed later.

PR #3229, opened 2023-01-17
Explicitly use python3 for mem gen scripts

ZenithalHourlyRate:
Related to https://github.com/chipsalliance/rocket-chip/issues/3228

Ubuntu 22.04 explicitly asks python script to use python3 or python2 in their shebang instead of an ambiguous python.

As for the change in `iterate_by_n`, see https://stackoverflow.com/questions/16465313/how-yield-catches-stopiteration-exception where it said

> PEP 479 has made it an error to allow a StopIteration to bubble up uncaught from a generator function. If that happens, Python will turn it into a RuntimeError exception instead. This means that code like the examples in older versions of itertools that used a StopIteration to break out of a generator function needs to be modified.

**Related issue**: #3228

### DIFF
--- a/scripts/vlsi_rom_gen
+++ b/scripts/vlsi_rom_gen
@@ -1,10 +1,6 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 
 # See LICENSE.SiFive for license details.
-
-from __future__ import division
-from __future__ import print_function
-from __future__ import unicode_literals
 
 import doctest
 import sys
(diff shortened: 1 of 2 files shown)


scripts/vlsi_rom_gen
```diff
@@ -81,22 +76,16 @@ def iterate_by_n(it, n):
...
     """
-    it = iter(it)
-    while True:
+    if len(li) % n != 0:
+        raise ValueError(
+            'Iterable length not evenly divisible by {}'.format(n)
+        )
+    for i in range(len(li) // n):
         batch = ()
-        for i in range(n):
```
jerryz123, line 87: I don't have time to review this thoroughly. Is changing this code around necessary?

ZenithalHourlyRate: Yes. The generator behavior changed since python3.7. You could refer to the stackoverflow link for more details.

jerryz123 (approved):

PR #3232, opened 2023-01-18
Add more extensions to dts

ZenithalHourlyRate:
For new version of the unprivileged spec, zicsr/zifencei have been separated out of I (base extension). Rocket always supports them. So should other cores that have the support of the old base extension.

Zihpm (also zicntr), however, is an internal uarch for rocket so I wonder if this isaDTS will affect other cores or not.

**Related issue**: https://github.com/chipsalliance/rocket-chip/pull/3202#discussion_r1064312736

**Impact**: API addition

**Release Notes**

Add H/Zicsr/Zifencei/Zihpm/Zfh to dts


sorear:
Might be less misleading to say that Rocket relies on software to provide Zicntr to S and U modes (riscv-software-src/riscv-pk@66776bbc in 2016, and in opensbi since the beginning)?

jerryz123 (approved):

PR #3233, opened 2023-01-18
RegDescsJson should use actual name of devices instead of "device"

jerryz123:
**Related issue**: 

**Release Notes**


sequencer (approved):
Haha I was frustrated by this a long time ago. However I used to try to solve it by OMIR but failed to achieved that. Binding might be get rid after MFC OMIR becoming stable enough, but it’s ok to get this in at this time.

PR #3234, opened 2023-01-20
Support passing wdata through TracedInstruction

jerryz123:
Rocket can't pass through wdata due to OOO write back, but other cores may support this.

**Related issue**: 

**Impact**: API modification

**Release Notes**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,5 @@ docs-target/
 /test_lib/
 /testbuild/
 rocketchip.jar
+*#
+.#*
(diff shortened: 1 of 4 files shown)


sequencer:
LGTM. 
not related this PR:
For now Trace API is used for two purpose, for [trace spec](https://github.com/riscv-non-isa/riscv-trace-spec), for debug in simulation. 

I think we can still keep Trace for RC 1.X to let other cores depending on this. However, in 2.0, I propose to private the current trace api in Rocket, switching to XMR for simulation debugging. And implement the trace IP for SoC. But we will continue maintain 1.X for a long time, let other cores depending on it.

sequencer (approved):

PR #3238, opened 2023-01-31
Support Tile Clusters (recursive subsystems)

jerryz123:
This work *substantially* refactors the existing Tile/TilePRCIDomain/HasTiles/InstantiatesTiles code into a system that support recursive clusters of tiles.

This PR defines an `Element` as an abstract node in the hierarchy tree. `Element`s can be `BaseTile`s or `Cluster`s, where `BaseTile`s are leaf nodes in the hierarchy, clusters are non-root nodes, and the `BaseSubsystem` is the root node.
`BaseTile` and `Cluster` will support deduplication.

Notably `Cluster`s support recursive instantiation, a `Cluster` can contain `Tile`s or internal `Cluster`s.  The `Cluster` bus topology should be capable of both supporting hierarchical coherence, and the `BaseSubsystem`s topology handle managing all coherence, depending on configuration variables.

This PR still requires much more time, and very thorough review. Some of the interrupt handling is not satisfying, and some decisions over clock/reset crossing ought to be addressed.

**Impact**:  API modification

**Release Notes**


sequencer:
will this PR block the 3.6 bump?

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -2,15 +2,16 @@
 
 package freechips.rocketchip.subsystem
 
+import chisel3.Flipped
+import chisel3.experimental.IO
```
sequencer, line 6: No need in 3.6

jerryz123:
No. I don't think this PR should block anything, we should delay features while more critical things are going in.
(I certainly want to see more thorough testing before this goes in).

sequencer:
the semantic of hartid and tileid is different in MT cores(although there are no opensource MT core for now). This PR seems to remove the hartid, is this a desidered refactor?

jerryz123:
> the semantic of hartid and tileid is different in MT cores(although there are no opensource MT core for now). This PR seems to remove the hartid, is this a desidered refactor?

tileId is static, and should only be referenced by generator (its the old "staticIdForMetadataUseOnly"). 
hartId should be set by IO to allow dedup.

Future work can allow a tile to support multiple hartid.

sequencer:
Great!

jerryz123:
Replaced by #3490

PR #2841, opened 2021-06-03
Add hypervisor extension

aswaterman:
This PR implements v0.6.2 of the RISC-V Hypervisor Extension.

The implementation was inspired by José Martins' and colleagues' work
described in [1].  Much of the microarchitecture and essentially all of
the code is new, but their implementation served as our baseline.
We thank them for trailblazing hypervisor support in rocket-chip.

Note that this PR only includes the mechanisms to virtualize the hart
itself.  Virtualized interrupt controllers, IOMMUs, etc. are future work.
Lots of future work.

Note also that some features are (legally) not implemented.  Currently,
misa.H is not writable, something we may or may not choose to fix.
The mtinst and htinst CSRs are hardwired to 0, placing
additional onus on hypervisor software to use the HLVX instruction.

[1] "A First Look at RISC-V Virtualization from an Embedded Systems Perspective", https://arxiv.org/abs/2103.14951

**Related issue**: 

**Type of change**: feature request

**Impact**: API addition (no impact on existing code)

**Development Phase**: implementation

PR #3243, opened 2023-02-06
add dev branch and switch to chisel master

sequencer:
Since Chisel 3.6 is introducing more and more breaking changes(#3242, #3210, #3200, #3169, #2968), while Chisel 5(the next version of Chisel 3.6) will introduce more, and not back porting them. For the developers who wanna use new feature to develop RocketChip is a burden to wait for new release of Chisel release(almost half to one year for a major version). And I'm also tired of maintaining a bunch of patches which will be eventually merged but blocked by Chisel release.  I'm proposing to switch rocket-chip development to use chisel master, AKA the SNAPSHOT version of Chisel.

For the future rocket-chip releasing: I add the new `dev` branch and the future stable releasing of RC will based that. the backportable PR from `dev` to `master` will be merged. `master` will also be able to bump to new minor version of Chisel, since they are all backward compatible. After new major release of Chisel, `dev` will become the new `master` version, and `dev` will switch to next `master` version of Chisel   

**Related issue**: #3242, #3210, #3200, #3169, #2968

**Type of change**: feature request

**Impact**: no functional change

**Development Phase**: proposal

**Release Notes**

PR #3246, opened 2023-02-10
Remove redundant TLBExceptions V bit

ZenithalHourlyRate:
This was introduced by "(185cac83) Add hypervisor extension (#2841)"

This is a dead code, as no circuit is producing and consuming this bit.

This was discovered when migrating Core.scala to chisel3, where strict checking was applied for IO and firrtl found this is not connected.

In the context of hypervisor extension, V bit, or _virtualization mode_, indicates whether the hart is currently executing in a guest.

For TLBReq, the V bit is needed as it affects the PTW thus TLB behavior on whether to do _Two-Stage Address Translation_.

However, the bit is not needed for TLBException. The exceptions (pf, gf, ae, ma) have no V=1/V=0 variants. Also, there is no point to add V bit for gf (guest page fault).

The io.resp.gf added below in the original patch also does not connect io.resp.gf.v. If this V bit should be added for this specific exception, it should be connected.

I assume this was added accidentally, as the original "extends CoreBundle()(p)" modification seems irrelevant.

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/TLB.scala
+++ b/src/main/scala/rocket/TLB.scala
@@ -59,11 +59,10 @@ class TLBReq(lgMaxSize: Int)(implicit p: Parameters) extends CoreBundle()(p) {
 
 }
 
-class TLBExceptions(implicit p: Parameters) extends CoreBundle()(p) {
+class TLBExceptions extends Bundle {
   val ld = Bool()
   val st = Bool()
   val inst = Bool()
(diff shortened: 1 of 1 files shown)


ZenithalHourlyRate:
Do not merge yet, blocked by #3243

sequencer (approved):

PR #3248, opened 2023-02-15
PTW: only makeFragmentedSuperpagePPN from superpage count level

ingallsj:
**Related issue**: 

**Impact**: functional fix

**Release Notes**
The result of the `makeFragmentedSuperpagePPN` function was being used by a leaf-level 4KB page translation result, resulting in an incorrect VA-to-PA translation result.
The fix is to only use `makeFragmentedSuperpagePPN` if `count < (pgLevels-1)`.


sequencer:
sorry our self-hosted CI encountered some issues, I'm fixing.

aswaterman (approved):

PR #3252, opened 2023-02-16
mill: Use MFC

ZenithalHourlyRate:
**Related issue**: 

**Release Notes**

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -89,51 +89,63 @@ object emulator extends mill.Cross[Emulator](
   ("freechips.rocketchip.system.TestHarness", "freechips.rocketchip.system.CloneTileConfig"),
   ("freechips.rocketchip.system.TestHarness", "freechips.rocketchip.system.HypervisorConfig"),
 )
-class Emulator(top: String, config: String) extends ScalaModule {
-  override def moduleDeps = Seq(rocketchip)
+class Emulator(top: String, config: String) extends Module {
 
-  override def scalaVersion: T[String] = T {
-    "2.13.10"
-  }
-
-  override def repositories = super.repositories ++ Seq(
-    MavenRepository("https://oss.sonatype.org/content/repositories/snapshots"),
-    MavenRepository("https://oss.sonatype.org/content/repositories/releases")
-  )
-
-  def spikeRoot = T { envByNameOrRiscv("SPIKE_ROOT") }
-
-  def generator = T {
-    // class path for `moduleDeps` is only a directory, not a jar, which breaks the cache.
-    // so we need to manually add the class files of `moduleDeps` here.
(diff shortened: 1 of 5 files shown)


build.sc
```diff
@@ -118,18 +118,39 @@ class Emulator(top: String, config: String) extends ScalaModule {
...
-    PathRef(output)
+  def chiselAnno = T {
+    os.walk(generator().path).collectFirst { case p if p.last.endsWith("anno.json") => p }.map(PathRef(_)).get
+  }
+
+  def chirrtl = T {
+    os.walk(generator().path).collectFirst { case p if p.last.endsWith("fir") => p }.map(PathRef(_)).get
+  }
+
+  def firtool = T {
```
sequencer, line 129: rename this target to `compile()`

sequencer (changes requested):

build.sc
```diff
@@ -88,51 +88,78 @@ object emulator extends mill.Cross[Emulator](
...
 
-  override def scalaVersion: T[String] = T {
-    "2.13.10"
-  }
-
-  override def repositories = super.repositories ++ Seq(
-    MavenRepository("https://oss.sonatype.org/content/repositories/snapshots"),
-    MavenRepository("https://oss.sonatype.org/content/repositories/releases")
-  )
+  object generator extends ScalaModule {
```
sequencer, line 93: This doesn't need to be a `ScalaModule`

build.sc
```diff
@@ -88,51 +88,78 @@ object emulator extends mill.Cross[Emulator](
...
-        "-C", config,
-      ),
+    override def repositories = super.repositories ++ Seq(
+      MavenRepository("https://oss.sonatype.org/content/repositories/snapshots"),
+      MavenRepository("https://oss.sonatype.org/content/repositories/releases")
     )
-    PathRef(T.dest)
+
+
+    def generate = T {
```
sequencer, line 106: `elaborate`

sequencer (changes requested):

build.sc
ZenithalHourlyRate: The `Generator` below uses `moduleDeps`, `upstreamCompileOutput()` and `runClasspath()`, which are not provided by `Module`.

sequencer: It actually requires `rocketchip.upstreamCompileOutput()`, this Module is not a `ScalaModule` since it doesn't contain any Scala source codes.

sequencer: Actually I do agree split the elaborate specific codes into a standalone package.

sequencer: But this is not ready for now.
I think `generator` should only depend on the `rocketchip.jar` and `firtool`

sequencer (approved):

PR #3253, opened 2023-02-16
Add HypervisorConfig and cover it in CI

ZenithalHourlyRate:
**Related issue**: 

Making hypervisor extension a config makes it more accessible for users, and the extension should be covered in CI. 

**Release Notes**

Add HypervisorConfig


sequencer (approved):
Xvisor should be used for CI workloads in the future.

Issue #3255, opened 2023-02-16
Spurious data dependency in aes64ks1i instruction


tzwaenn:
**Type of issue**: bug report

**Impact**: performance impact

**Other information**
 
During formal verification with the BitManipCryptoConfig, we see the instruction in question being needlessly replayed. Consider the code sequence

        ld x19,256(x2)
        addi x8,x2,260
        aes64ks1i x11,x25,0x3

where the initial load has a cache miss. This is causing the aes64ks1i to replay without any actual dependency on the load. Apparently, the core thinks that aes64ks1i is using the rs2 field (encoding is actually 00 11000 1 rnum[3:0] rs1 001 rd 0010011), meaning a spurious dependency on x19 in this case with rnum=3.
I assume this is also causing needless stalls in decode.

This has been identified by running the Siemens Processor Verification App on the generated RTL.

ZenithalHourlyRate:
#3256 may fix the issue.

ZenithalHourlyRate:
Should be resolved now.

PR #3256, opened 2023-02-16
IDecode: Fix aes64ks1i imm decode

ZenithalHourlyRate:
It is not rs2, it is imm

**Related issue**: #3255

**Type of change**: bug fix

**Impact**: Functionality will not change, as the original rs2 from regfile is not used. Performance may inprove as #3255 indicates.

**Release Notes**

Fix aes64ks1i decoding


sequencer:
Why RVArchTest didn't capture this bug........

ZenithalHourlyRate:
> Why RVArchTest didn't capture this bug........

The functionality is correct, so it passed the test; decoding is internal state and could not be captured by arch-test.

sequencer (approved):

PR #3271, opened 2023-02-24
Rocket cosim framework 

midnighter95:
This PR provides a Rocket-Spike cosim simulation framework. A standalone rocket core is extracted with its memory port exposed to the top-level and is compiled to c++ model by Verilator . The framework drives rocket model to run testcase while spike runs the same test at the same time. During the test, the rocket result  will be compared to the spike for each instruction.

## Framework Outline
- control the overall test flow.
- stimulate Spike with testcase, record the result.
- stimulate rocket model.
- deal with all the memory access from the core (acts as the memory of rocket core)
- peek rocket result and compare it with Spike.

## Features
-  standalone Rocket core with decouple bus interfaces expose to users.
-  use spike as reference model
-  capture architecture events, for each cycles, comparing event with architecture events generated from spike
- fast binary load and start-up

## Uarch Coverpoint
- [x] Scalar register file write
- [ ] Float register file write
- [ ] CSR write

## Completed tests:

- [x] rv64mi-p except csr and breakpoint test
- [x] rv64mzicbo
- [x] rv64ssvnapot
- [x] rv64si
- [x] rv64ua
- [x] rv64uc
- [x] rv64ud
- [x] rv64uf
- [x] rv64ui
- [x] rv64um
- [x] rv64uzfh
- [ ] rv64um

## Test Flow

nix develop -c make riscvtests

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -22,6 +22,11 @@
           verilator cmake ninja
           python3 python3Packages.bootstrapped-pip
           pkgsCross.riscv64-embedded.buildPackages.gcc
+
+          libargs glog fmt zlib
+          gnused coreutils gnugrep which
+          parallel protobuf antlr4 numactl
+
           circt
 
           spike riscvTests
--- a/src/main/scala/tile/Interrupts.scala
+++ b/src/main/scala/tile/Interrupts.scala
@@ -1,10 +1,10 @@
 // See LICENSE.SiFive for license details.
 
-package tile
+package freechips.rocketchip.tile
 
 import Chisel._
 
-import freechips.rocketchip.config.Parameters
+import org.chipsalliance.cde.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.interrupts._
 import freechips.rocketchip.util._
@@ -31,6 +31,29 @@ trait SinksExternalInterrupts { this: BaseTile =>
   val intInwardNode = intXbar.intnode :=* IntIdentityNode()(ValName("int_local"))
   protected val intSinkNode = IntSinkNode(IntSinkPortSimple())
   intSinkNode := intXbar.intnode
+
+  def cpuDevice: Device
+  val intcDevice = new DeviceSnippet {
+    override def parent = Some(cpuDevice)
+    def describe(): Description = {
+      Description("interrupt-controller", Map(
+        "compatible"           -> "riscv,cpu-intc".asProperty,
+        "interrupt-controller" -> Nil,
+        "#interrupt-cells"     -> 1.asProperty))
+    }
+  }
+
+  ResourceBinding {
+    intSinkNode.edges.in.flatMap(_.source.sources).map { case s =>
+      for (i <- s.range.start until s.range.end) {
+       csrIntMap.lift(i).foreach { j =>
+          s.resources.foreach { r =>
+            r.bind(intcDevice, ResourceInt(j))
+          }
+        }
+      }
+    }
+  }
 
   // TODO: the order of the following two functions must match, and
   //         also match the order which things are connected to the
--- a/src/main/scala/tile/Interrupts.scala
+++ b/src/main/scala/tile/Interrupts.scala
@@ -1,10 +1,10 @@
 // See LICENSE.SiFive for license details.
 
-package freechips.rocketchip.tile
+package tile
 
 import Chisel._
 
-import org.chipsalliance.cde.config.Parameters
+import freechips.rocketchip.config.Parameters
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.interrupts._
 import freechips.rocketchip.util._
@@ -31,29 +31,6 @@ trait SinksExternalInterrupts { this: BaseTile =>
   val intInwardNode = intXbar.intnode :=* IntIdentityNode()(ValName("int_local"))
   protected val intSinkNode = IntSinkNode(IntSinkPortSimple())
   intSinkNode := intXbar.intnode
-
-  def cpuDevice: Device
-  val intcDevice = new DeviceSnippet {
-    override def parent = Some(cpuDevice)
-    def describe(): Description = {
-      Description("interrupt-controller", Map(
-        "compatible"           -> "riscv,cpu-intc".asProperty,
-        "interrupt-controller" -> Nil,
-        "#interrupt-cells"     -> 1.asProperty))
-    }
-  }
-
-  ResourceBinding {
-    intSinkNode.edges.in.flatMap(_.source.sources).map { case s =>
-      for (i <- s.range.start until s.range.end) {
-       csrIntMap.lift(i).foreach { j =>
-          s.resources.foreach { r =>
-            r.bind(intcDevice, ResourceInt(j))
-          }
-        }
-      }
-    }
-  }
 
   // TODO: the order of the following two functions must match, and
   //         also match the order which things are connected to the
--- a/src/main/scala/tile/RocketTile.scala
+++ b/src/main/scala/tile/RocketTile.scala
@@ -1,10 +1,10 @@
 // See LICENSE.SiFive for license details.
 // See LICENSE.Berkeley for license details.
 
-package freechips.rocketchip.tile
+package tile
 
 import Chisel._
-import org.chipsalliance.cde.config._
+import freechips.rocketchip.config._
 import freechips.rocketchip.devices.tilelink._
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.interrupts._
@@ -36,11 +36,11 @@ case class RocketTileParams(
 }
 
 class RocketTile private(
-      val rocketParams: RocketTileParams,
-      crossing: ClockCrossingType,
-      lookup: LookupByHartIdImpl,
-      q: Parameters)
-    extends BaseTile(rocketParams, crossing, lookup, q)
+                          val rocketParams: RocketTileParams,
+                          crossing: ClockCrossingType,
+                          lookup: LookupByHartIdImpl,
+                          q: Parameters)
+  extends BaseTile(rocketParams, crossing, lookup, q)
     with SinksExternalInterrupts
     with SourcesExternalNotifications
     with HasLazyRoCC  // implies CanHaveSharedFPU with CanHavePTW with HasHellaCache
@@ -87,20 +87,7 @@ class RocketTile private(
   val itimProperty = frontend.icache.itimProperty.toSeq.flatMap(p => Map("sifive,itim" -> p))
 
   val beuProperty = bus_error_unit.map(d => Map(
-          "sifive,buserror" -> d.device.asProperty)).getOrElse(Nil)
-
-  val cpuDevice: SimpleDevice = new SimpleDevice("cpu", Seq("sifive,rocket0", "riscv")) {
-    override def parent = Some(ResourceAnchors.cpus)
-    override def describe(resources: ResourceBindings): Description = {
-      val Description(name, mapping) = super.describe(resources)
-      Description(name, mapping ++ cpuProperties ++ nextLevelCacheProperty
-                  ++ tileProperties ++ dtimProperty ++ itimProperty ++ beuProperty)
-    }
-  }
-
-  ResourceBinding {
-    Resource(cpuDevice, "reg").bind(ResourceAddress(staticIdForMetadataUseOnly))
-  }
+    "sifive,buserror" -> d.device.asProperty)).getOrElse(Nil)
 
   override lazy val module = new RocketTileModuleImp(this)
 
@@ -120,9 +107,9 @@ class RocketTile private(
 }
 
 class RocketTileModuleImp(outer: RocketTile) extends BaseTileModuleImp(outer)
-    with HasFpuOpt
-    with HasLazyRoCCModule
-    with HasICacheFrontendModule {
+  with HasFpuOpt
+  with HasLazyRoCCModule
+  with HasICacheFrontendModule {
   Annotated.params(this, outer.rocketParams)
 
   val core = Module(new Rocket(outer)(outer.p))
@@ -133,9 +120,9 @@ class RocketTileModuleImp(outer: RocketTile) extends BaseTileModuleImp(outer)
   // Report when the tile has ceased to retire instructions; for now the only cause is clock gating
   outer.reportCease(outer.rocketParams.core.clockGate.option(
     !outer.dcache.module.io.cpu.clock_enabled &&
-    !outer.frontend.module.io.cpu.clock_enabled &&
-    !ptw.io.dpath.clock_enabled &&
-    core.io.cease))
+      !outer.frontend.module.io.cpu.clock_enabled &&
+      !ptw.io.dpath.clock_enabled &&
+      core.io.cease))
 
   outer.reportWFI(Some(core.io.wfi))
 
@@ -182,8 +169,8 @@ class RocketTileModuleImp(outer: RocketTile) extends BaseTileModuleImp(outer)
   require(h == c, s"port list size was $h, core expected $c")
   require(h == o, s"port list size was $h, outer counted $o")
   // TODO figure out how to move the below into their respective mix-ins
-  dcacheArb.io.requestor <> dcachePorts.toSeq
-  ptw.io.requestor <> ptwPorts.toSeq
+  dcacheArb.io.requestor <> dcachePorts
+  ptw.io.requestor <> ptwPorts
 }
 
 trait HasFpuOpt { this: RocketTileModuleImp =>
--- a/src/main/scala/tile/RocketTile.scala
+++ b/src/main/scala/tile/RocketTile.scala
@@ -1,10 +1,10 @@
 // See LICENSE.SiFive for license details.
 // See LICENSE.Berkeley for license details.
 
-package tile
+package freechips.rocketchip.tile
 
 import Chisel._
-import freechips.rocketchip.config._
+import org.chipsalliance.cde.config._
 import freechips.rocketchip.devices.tilelink._
 import freechips.rocketchip.diplomacy._
 import freechips.rocketchip.interrupts._
@@ -36,11 +36,11 @@ case class RocketTileParams(
 }
 
 class RocketTile private(
-                          val rocketParams: RocketTileParams,
-                          crossing: ClockCrossingType,
-                          lookup: LookupByHartIdImpl,
-                          q: Parameters)
-  extends BaseTile(rocketParams, crossing, lookup, q)
+      val rocketParams: RocketTileParams,
+      crossing: ClockCrossingType,
+      lookup: LookupByHartIdImpl,
+      q: Parameters)
+    extends BaseTile(rocketParams, crossing, lookup, q)
     with SinksExternalInterrupts
     with SourcesExternalNotifications
     with HasLazyRoCC  // implies CanHaveSharedFPU with CanHavePTW with HasHellaCache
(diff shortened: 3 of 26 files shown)


tests/cosim/elaborate/src/Main.scala
```diff
@@ -0,0 +1,94 @@
...
+
+}
+
+object Generator extends StageMain(new RocketChipStage)
+
+
+object RocketTileParamsKey extends Field[RocketTileParams]
+
+class cosimConfig extends Config((site, here, up) => {
+  case MonitorsEnabled => false
```
jerryz123, line 70: Cosim should work with standard rocket configs

tests/cosim/emulator/src/glog_exception_safe.h
```diff
@@ -0,0 +1,35 @@
+#pragma once
+
+#include <glog/logging.h>
```
jerryz123, line 3: Is glog necessary? It seems overkill for this

tests/cosim/emulator/src/spike_event.cc
```diff
@@ -0,0 +1,241 @@
...
+  target_mem = -1;
+  is_committed = false;// j insn should be committed immediately cause it doesn't have wb stage.
+
+  // extension depending parameter
+  is_compress = false;
+  if (fetch.insn.length() == 2) {
+    is_compress = true;
+  }
+  if (!is_compress) {
+    rs1_bits = xr[fetch.insn.rs1()];
```
jerryz123, line 117: I don't think all of this should be necessary... this reimplements details of execution. Why?

jerryz123 (changes requested):
Have you seen https://github.com/ucb-bar/chipyard/pull/1323?

The model for cosim should be 
 - Support (almost) arbitrary configuration of DUT
 - Match DUT arch spec with spike configuration
 - For every instruction committed by DUT, compare against single-stepped spike
 
 The only challenge with rocket is the OOO write-back, but that can be solved easily by adding a giant ROB model to the core (used only for cosim with traces, not for implementation).

flake.nix
```diff
@@ -22,6 +22,11 @@
           verilator cmake ninja
           python3 python3Packages.bootstrapped-pip
           pkgsCross.riscv64-embedded.buildPackages.gcc
+
+          libargs glog fmt zlib
+          gnused coreutils gnugrep which
+          parallel protobuf antlr4 numactl
```
sequencer, line 28: Only need to add necessary dependencies.

src/main/scala/tile/Interrupts.scala
```diff
@@ -1,10 +1,10 @@
 // See LICENSE.SiFive for license details.
 
-package freechips.rocketchip.tile
+package tile
 
 import Chisel._
 
-import org.chipsalliance.cde.config.Parameters
+import freechips.rocketchip.config.Parameters
```
sequencer, line 7:
```suggestion
import org.chipsalliance.cde.config.Parameters
```

src/main/scala/tile/Interrupts.scala
```diff
@@ -1,10 +1,10 @@
 // See LICENSE.SiFive for license details.
 
-package freechips.rocketchip.tile
+package tile
```
sequencer, line 3: Don't need to modify this file

src/main/scala/tile/Interrupts.scala
```diff
@@ -32,29 +32,6 @@ trait SinksExternalInterrupts { this: BaseTile =>
...
-    intSinkNode.edges.in.flatMap(_.source.sources).map { case s =>
-      for (i <- s.range.start until s.range.end) {
-       csrIntMap.lift(i).foreach { j =>
-          s.resources.foreach { r =>
-            r.bind(intcDevice, ResourceInt(j))
-          }
-        }
-      }
-    }
-  }
```
sequencer, line 56: Please keep these as it is.

src/main/scala/tile/RocketTile.scala
```diff
@@ -1,10 +1,10 @@
 // See LICENSE.SiFive for license details.
 // See LICENSE.Berkeley for license details.
 
-package freechips.rocketchip.tile
+package tile
```
sequencer, line 4: don't touc

src/main/scala/tile/RocketTile.scala
```diff
@@ -1,10 +1,10 @@
...
 
 import Chisel._
-import org.chipsalliance.cde.config._
+import freechips.rocketchip.config._
```
sequencer, line 7: don't touch

src/main/scala/tile/RocketTile.scala
```diff
@@ -36,11 +36,11 @@ case class RocketTileParams(
...
-      val rocketParams: RocketTileParams,
-      crossing: ClockCrossingType,
-      lookup: LookupByHartIdImpl,
-      q: Parameters)
-    extends BaseTile(rocketParams, crossing, lookup, q)
+                          val rocketParams: RocketTileParams,
+                          crossing: ClockCrossingType,
+                          lookup: LookupByHartIdImpl,
+                          q: Parameters)
+  extends BaseTile(rocketParams, crossing, lookup, q)
```
sequencer, line 43: don't touch

src/main/scala/tile/RocketTile.scala
```diff
@@ -87,20 +87,7 @@ class RocketTile private(
...
-    override def describe(resources: ResourceBindings): Description = {
-      val Description(name, mapping) = super.describe(resources)
-      Description(name, mapping ++ cpuProperties ++ nextLevelCacheProperty
-                  ++ tileProperties ++ dtimProperty ++ itimProperty ++ beuProperty)
-    }
-  }
-
-  ResourceBinding {
-    Resource(cpuDevice, "reg").bind(ResourceAddress(staticIdForMetadataUseOnly))
-  }
```
sequencer, line 103: don't touch

tests/cosim/elaborate/src/DUT.scala
```diff
@@ -0,0 +1,97 @@
...
+import freechips.rocketchip.diplomacy.{AddressSet, BundleBridgeSource, InModuleBody, LazyModule, RegionType, SimpleLazyModule, TransferSizes}
+import freechips.rocketchip.interrupts.{IntSinkNode, IntSinkPortSimple, IntSourceNode, IntSourcePortSimple}
+import freechips.rocketchip.subsystem._
+import freechips.rocketchip.tilelink.{TLManagerNode, TLSlaveParameters, TLSlavePortParameters}
+import org.chipsalliance.cde.config.Parameters
+import freechips.rocketchip.tile.{NMI, PriorityMuxHartIdFromSeq, RocketTile}
+
+import org.chipsalliance.cde.config.{Config, Field}
+
+class DUT(p: Parameters) extends Module {
```
sequencer, line 13: You can directly use `BaseSubsystem` with `InstantiatesTiles` to construct a minimize version of LazyModule

tests/cosim/elaborate/src/DUT.scala
```diff
@@ -0,0 +1,97 @@
...
+    val hartidNode = BundleBridgeSource(() => UInt(4.W))
+    rocketTile.hartIdNode := hartidNode
+    InModuleBody {
+      hartidNode.bundle := 0.U
+    }
+    val nmiNode = BundleBridgeSource(Some(() => new NMI(32)))
+    rocketTile.nmiNode := nmiNode
+    val nmi = InModuleBody {
+      nmiNode.makeIO()
+    }
```
sequencer, line 78: These logic is already implemented in the `InstantiatesTiles`, please use it.

tests/cosim/elaborate/src/DUT.scala
```diff
@@ -0,0 +1,97 @@
...
+          val io = IO(Output(chiselTypeOf(ele))).suggestName(name)
+          println(s"output $name")
+          io := ele
+        case ActualDirection.Input =>
+          val io = IO(Input(chiselTypeOf(ele))).suggestName(name)
+          println(s"input $name")
+          ele := io
+      }
+    }
+  }
```
sequencer, line 96: IIRC this hack should be fixed.

tests/cosim/elaborate/src/Main.scala
```diff
@@ -0,0 +1,94 @@
...
+    Dependency[freechips.rocketchip.stage.phases.GenerateArtefacts]
+  )
+
+  private val pm = new PhaseManager(targets)
+
+  override def run(annotations: AnnotationSeq): AnnotationSeq = pm.transform(annotations)
+
+}
+
+object Generator extends StageMain(new RocketChipStage)
```
sequencer, line 64: Don't use RC Stage here.
I draft a minimized `Main` object [here](https://github.com/chipsalliance/rocket-chip/tree/new_main), I will find some time to finish it.

tests/cosim/elaborate/src/Main.scala
```diff
@@ -0,0 +1,94 @@
...
+      nMSHRs = 0,
+      blockBytes = site(CacheBlockBytes))),
+    icache = Some(ICacheParams(
+      rowBits = site(SystemBusKey).beatBits,
+      blockBytes = site(CacheBlockBytes))))
+  case SystemBusKey => SystemBusParams(
+    beatBytes = site(freechips.rocketchip.tile.XLen) / 8,
+    blockBytes = site(CacheBlockBytes))
+  case DebugModuleKey => None
+})
```
sequencer, line 93: You should be able to use RC default configs.

tests/cosim/emulator/src/tl_interface.h
```diff
@@ -0,0 +1,42 @@
...
+      AccessAckData = 1,
+      PutFullData = 0,
+      PutPartialData = 1,
+      AccessAck = 4;
+}
+
+// the following macro helps us to access tl interface by dynamic index
+#define TL_INTERFACE(type, name)                      \
+  [[nodiscard]] inline type &get_tl_##name(VV &top) { \
+    return top.memory_0_##name;                       \
```
sequencer, line 16: I don't this using `.` from verilator is a good idea. it is hardcoded. Please switch to VPI or DPI instead.

sequencer (changes requested):
Basically this framework was the initial version from my [vector]( https://github.com/sequencer/vector) project. really appreciate @midnighter95 for trying to migrate it to RC.

For RC, we need a more carefully designed and general test framework.
1. You should use VPI or DPI to make Chisel be able to hook data, like the https://github.com/sequencer/vector/blob/master/tests/elaborate/src/GrandCentral.scala interface with DPI.
2. Tests should fully compatible with the native RocketChip Configs. Spike should read from the RocketCore parameter to config itself.
3. For the difference between model&impl, there should be a hook to inject to model, since there exist some undefined behavior, which is allowed but behavior differs between model and RTL(this is a little tricky, you can do it in the future PR)
4. I personally don't like the `Monitor` implementation in the RC, I always believe all verification codes should be injected remotely, like systemverilog `bind` + Cross Module Reference, or Aspect from Chisel. I strongly suggest switch to DPI directly and use the DPI + GrandCenteral magic.

Anyway, thanks for this work, I understand this PR is hard to implement, but this should be the key feature for the future RocketCore CI(far more faster than the in-tree version).

sequencer:
> The only challenge with rocket is the OOO write-back, but that can be solved easily by adding a giant ROB model to the core (used only for cosim with traces, not for implementation).

I personally prefer to maintain it inside the emulator. what do you think?

sequencer:
> Support (almost) arbitrary configuration of DUT

I think we only need to support single core. Since multi-core cosim will be bitten by consistency.
And I don't think we should test the topology of SoC, which is another domain of verification.

jerryz123:
> * You should use VPI or DPI to make Chisel be able to hook data, like the https://github.com/sequencer/vector/blob/master/tests/elaborate/src/GrandCentral.scala interface with DPI.

I think a verilog blackbox with DPI is sufficient for this.

> * Tests should fully compatible with the native RocketChip Configs. Spike should read from the RocketCore parameter to config itself.

Agreed.

> * For the difference between model&impl, there should be a hook to inject to model, since there exist some undefined behavior, which is allowed but behavior differs between model and RTL(this is a little tricky, you can do it in the future PR)

Yes... I just add the overrides in a ad-hoc manner in https://github.com/ucb-bar/chipyard/blob/main/generators/chipyard/src/main/resources/csrc/cospike.cc. API for this would be better, but I agree that is harder to come up with.

> * I personally don't like the `Monitor` implementation in the RC, I always believe all verification codes should be injected remotely, like systemverilog `bind` + Cross Module Reference, or Aspect from Chisel. I strongly suggest switch to DPI directly and use the DPI + GrandCenteral magic.

I think injecting DPI within the module hierarchy makes some flows more difficult. For example, cosim with DUT on FPGA (or firesim). I prefer to keep all DPI outside the DUT.

> I personally prefer to maintain it inside the emulator. what do you think?

I guess you mean outside the DUT. Then you have to define a new interface for carrying the OOO WB information. I prefer to keep interfaces from core to outside implementation-agnostic, so that interface would need to be carefully designed as well.
Adding a ROB model seems to be much easier ... I am doing this in another project.

Personally, I also want a ROB model inside Rocket so https://github.com/ucb-bar/chipyard/pull/1323 can work with Rocket.


> I think we only need to support single core. Since multi-core cosim will be bitten by consistency.
And I don't think we should test the topology of SoC, which is another domain of verification.

I agree. I think some uncore things ought-to-be captured in cosim ... for example... bootrom contents, dts contents, memory map.

sequencer:
> I think injecting DPI within the module hierarchy makes some flows more difficult. For example, cosim with DUT on FPGA (or firesim). I prefer to keep all DPI outside the DUT.

Yes, those DPI should only exist in the Testbench and inspect RTL with XMR. Design codes should never have any DPI. 

> I guess you mean outside the DUT. Then you have to define a new interface for carrying the OOO WB information.

Not really if we have XMR and can peek any signal we like.


> I think some uncore things ought-to-be captured in cosim ... for example... bootrom contents, dts contents, memory map.

Yes, for verification on those components, how about use `SpikeTile` to test them? i believe it can speed up a lot.

jerryz123:
> Yes, those DPI should only exist in the Testbench and inspect RTL with XMR. Design codes should never have any DPI.
> 
> Not really if we have XMR and can peek any signal we like.

I'm not sure restricting DPI to harness and using XMR to bring signals out solves every use case. For example, with DUT on FPGA, you need to bring the trace signals out of the core to the top level, and then interface it with the host somehow. 
Having a uniform traced-instruction port also enables some other interesting use-cases, beyond cosim. You can do analysis of the trace, or even reconstruct architectural state at an arbitrary point in the trace. 

I agree for many other use cases, XMR is preferable to punching everything out. In this case, specifically, extending the existing TracedInstruction infrastructure is more useful than XMR.

> Yes, for verification on those components, how about use SpikeTile to test them? i believe it can speed up a lot.

What I mean is that the spike cosim should have the same memory map/bootrom as the DUT. Spike natively generates its own bootrom/dts that will differ from the target.

I want regressions of Linux Boot in RTL sim in Chipyard with SpikeTile. Right now, giant memsets and memcpys in boot really slow the system down, since the L2/buses are simulated on every L1 cache miss.

sequencer:
> I'm not sure restricting DPI to harness and using XMR to bring signals out solves every use case. For example, with DUT on FPGA, you need to bring the trace signals out of the core to the top level, and then interface it with the host somehow.

Yep, agree the case on FPGA is quite useful, I think in that case, we should migrate the current custom trace function to [riscv-trace-spec](https://github.com/riscv-non-isa/riscv-trace-spec), so that we can use the software stack.

> What I mean is that the spike cosim should have the same memory map/bootrom as the DUT. Spike natively generates its own bootrom/dts that will differ from the target.

Yeah, I think that is a bad design, I personally prefer using `processor_t`, and disable I$ and D$. 

> I want regressions of Linux Boot in RTL sim in Chipyard with SpikeTile. Right now, giant memsets and memcpys in boot really slow the system down, since the L2/buses are simulated on every L1 cache miss.

Maybe a custom linking script while maintaining a memory in emulator may help.

jerryz123:
> Yep, agree the case on FPGA is quite useful, I think in that case, we should migrate the current custom trace function to [riscv-trace-spec](https://github.com/riscv-non-isa/riscv-trace-spec), so that we can use the software stack.

It should support both. A full trace for cases where spec-compliance is not necessary, and a spec-compliant trace. In fact, this is already the case. TraceCoreInterface should provide the spec-compliant trace, while TracedInstruction is instruction-level trace.

> Yeah, I think that is a bad design, I personally prefer using processor_t, and disable I$ and D$.

Why do you think it is bad design? Cosim should test I$ and D$ as well... especially since D$ microarchitecture is quite complicated, and implements things like atomics, lr/sc, etc.

That approach is simple, and works ok. It requires no special harness, and does not rely on any details of the DUT microarchitecture. Thus it is easy to maintain, and microarchitecture-agnostic.

sequencer:
I mean the cache inside spike is strange, we should disable that.

jerryz123:
> > Why do you think it is bad design?
> 
> I mean the spike cache behavior model is quite strange, we should disable that.
> 
> 
> 
> To make it more clear: I think we should only use `process_t` to construct a spike core and use SoC RTL directly. 
> 
> 

What is the advantage of this approach over having spike model the entire system.

sequencer:
> What is the advantage of this approach over having spike model the entire system.

Basically I think the main purpose to introduce `SpikeTile` is testing the SoC, especially cache system and periphery.

In this case, we just put `text` segment to C++, without creating IF transaction to RTL. This can avoid the I$ issue you described. And the `SpikeTile` will be eventually used as the Load/Store generator for generating transaction, and L1 Cache emulator for Cache system testing.

jerryz123:
I think two separate concerns are getting mixed together. Integrating SpikeTile can be left to another discussion. I'm more concerned at the moment with the design of spike-cosim. 

I believe spike-cosim should be performed by having spike model the entire SoC, as is done in Chipyard's implementation. What are the disadvantages you see with that approach?

sequencer:
> I believe spike-cosim should be performed by having spike model the entire SoC


This is basically what I'm concerning, I think the cosim framework in this PR should only for `RocketCore`, Eventually we should split Rocket Core out from RC, while still be able to testing it standalone.
SoC level cosim will force Rocket Core continue depending on the RocketChip. My approach is emulating the necessary off-core components in the C++ to remove the dependency to RocketChip.

jerryz123:
Ok, to summarize, in this proposed cosim design:
 - REF uses a *c++* uncore and the *processor_t* as the core
 - DUT uses the same *c++* uncore as REF the RocketCore RTL
 
I feel like this approach introduces a lot of unnecessary complexity, you have to write a interface between spike's bus architecture model, and the RocketCore's tilelink interface, which implies writing a TileLink coherence manager model in C++. On the other hand, using RC+Rocket as the DUT enables a trivial (~200 LoC) implementation of cosim. 

Standalone cosim of Rocket should be done in the split-out rocket repo, if it is truly expected to be independent of rocket-chip.

sequencer:
> I feel like this approach introduces a lot of unnecessary complexity, you have to write a interface between spike's bus architecture model, and the RocketCore's tilelink interface, which implies writing a TileLink coherence manager model in C++. On the other hand, using RC+Rocket as the DUT enables a trivial (~200 LoC) implementation of cosim.

Yes, I agree that's pain. However it's necessary for splitting `RocketCore` out. The test framework will only be used for the `core` test. We will continue keep the current test framework.

This test framework will only work for the rocket core, rather than the entire SoC, and will be eventually split out with the rocket core.

jerryz123:
Ok. In that case I would like to mark this PR DO_NOT_MERGE. 

I do not believe we should merge things that will be split out later.

sequencer:
> Ok. In that case I would like to mark this PR DO_NOT_MERGE. I do not believe we should merge things that will be split out later.

So how do you think we can start to split? creating another repo or branch?

jerryz123:
I think you can create another repo, then just push the latest dev branch into it. Then someone can work on modifying the version of rocket in that repo to be standalone. 

We should not lose commit history.

sequencer:
> I think you can create another repo, then just push the latest dev branch into it.

Sure!

Issue #3257, opened 2023-02-16
Migrate IDecoder to use DecoderTable  


sequencer:
Current decoder is brittle. chipsalliance/chisel3#2924 provides a better solution to decoders. We need to swap to it for a better decoder structure.

**Type of issue**: feature request

**Impact**: no functional change

**Development Phase**: request

PR #3272, opened 2023-02-26
Migrate IDecoder to use DecoderTable

Lucas-Wye:
https://github.com/chipsalliance/rocket-chip/issues/3257


sequencer:
Firstly please add more information in your the PR description, CI seems failed, I will review it after CI passed. 15k+ lines of code is hard to maintain, how about try to maintain a Metadata for simplicity.

jerryz123 (changes requested):
It's unclear to me what this PR is trying to achieve. A change of this magnitude requires significant justification

jerryz123:
Please also provide some context for the issue with the original code. #3257 does not explain enough why the current decoder structure is brittle.

Lucas-Wye:
> Please also provide some context for the issue with the original code. #3257 does not explain enough why the current decoder structure is brittle.

Thanks for your sugeestion. This PR is still WIP. 
The purpose is to make the generated verilog code to be more readable. Since current decoder will generate many structural verilog codes and is also difficult to maintain in Chisel. For example, if one want to add one more control signal, he/she will have to add this control signal to all the instructions truth table, which will be very unconvenient and make a big change in the decoder file.

jerryz123:
I don't buy that adding another column to the tabular signal representation is a huge burden, but I'm open to an alternative if it's neater.

jerryz123:
Should we expect for the the generated verilog from the QMC based decoder to be readable?

sequencer:
@Lucas-Wye is not correct, basically there are three issues:
- the current RocketChip style instruction need to maintain two different thing: 1. Decode Table, 2. Decode Bundle. The sequence of which should be manually maintained, this is why I call it brittle.
- For small designs, e.g. RV32I, we don't need to decode some control signals in this case, however they still in the PLA consuming area.
- For the case that we need to add new signals, e.g. #2950, there are almost 1000k changes in the IDecode file, since it adds new signals `zbk`, `zbn`, `zks`, and it will also add area for small designs.

The `DecodeBundle` is a new feature in Chisel 3.6, which provides an interface to user to create decode table based on the metadata of instructions, and inject control signals simply, we can reduce control signals in small designs and add control signals in large designs rather than having the decode width among different designs.

sequencer:
So basically the `DecodeBundle` API makes design more intuitively:
- Datapath need Control
- Control need Decode Result
- The change/addition to datapath may need new control
- new control doesn't affect original decode but only need new fields.

The new API will maintain a set of control signals separately, based on whether the datapath exist or not, which decides the corresponding control signals is used or not.

jerryz123:
Ok. I think moving to DecodeBundle is a good choice. 

However, I think we need to think harder about the API used to describe instruction control signals, such that the instruction definitions can be expressed concisely, and readably. The new instruction definitions ideally should bear some resemblance to the old ones so it is easier to review. 

The problems you describe seem to be "backend" problems with the decoder generator, rather than "front end" problems with the tabular representation. If possible, I would try to fix this as much in the backend as possible, rather than rewriting all the instruction definitions.

sequencer:
> However, I think we need to think harder about the API used to describe instruction control signals, such that the instruction definitions can be expressed concisely, and readably. The new instruction definitions ideally should bear some resemblance to the old ones so it is easier to review.

Agree, the current implement from @Lucas-Wye is certainly not good(no one wanna maintain 15k+ codes only for decoding).
I think these things should be done:
1. tune code to pass all CI.
2. add metadatas to instructions.(can borrow logic from riscv-spec or directly parse from it)
3. reduce decoder to use metadata, rather than the instruction enumeration.

> The problems you describe seem to be "backend" problems with the decoder generator, rather than "front end" problems with the tabular representation. If possible, I would try to fix this as much in the backend as possible, rather than rewriting all the instruction definitions.

I think these problems can be resolved by removing the column which is all or `Y`/`N`/`DC`, this can be done in Chisel decoder. However I still think this adds non-necessary code to Chisel only for maintaining a **compatible** fix for the bad frontend design, a Chisel dev, I personally don't want to add it. And even if we resolve this in Chisel, the problem of adding new decode field will introduce 1k+ changes is not fixed.

sequencer (commented):
I suggest for each instructions, you can generate a `case class` for it.

PR #3287, opened 2023-03-04
Porting Chisel 2 to Chisel 3: tilelink/

SingularityKChen:
**Related issue**: 

**Release Notes**

As [`Chisel._` is Deprecated in Chisel 3.6](https://github.com/chipsalliance/chisel3/blob/5bc236268801861d325af7100922da3d14cd8b07/ROADMAP.md?plain=1#L33), we are porting Chisel 2 to Chisel 3.

### DIFF
--- a/src/main/scala/tilelink/ToAXI4.scala
+++ b/src/main/scala/tilelink/ToAXI4.scala
@@ -149,8 +149,8 @@ class TLToAXI4(val combinational: Boolean = true, val adapterName: Option[String
       val depth = if (combinational) 1 else 2
       val out_arw = Wire(Decoupled(new AXI4BundleARW(out.params)))
       val out_w = Wire(chiselTypeOf(out.w))
-      out.w :<> Queue.irrevocable(out_w, entries=depth, flow=combinational)
-      val queue_arw = Queue.irrevocable(out_arw, entries=depth, flow=combinational)
+      out.w :<> Queue.irrevocable(out_w, entries=depth, combinational)
+      val queue_arw = Queue.irrevocable(out_arw, entries=depth, combinational)
 
       // Fan out the ARW channel to AR and AW
       out.ar.bits := queue_arw.bits
@@ -162,7 +162,7 @@ class TLToAXI4(val combinational: Boolean = true, val adapterName: Option[String
       val beatBytes = edgeIn.manager.beatBytes
       val maxSize   = log2Ceil(beatBytes).U
       val doneAW    = RegInit(false.B)
-      when (in.a.fire()) { doneAW := !a_last }
+      when (in.a.fire) { doneAW := !a_last }
 
       val arw = out_arw.bits
       arw.wen   := a_isPut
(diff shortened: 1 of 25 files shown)


SingularityKChen:
I'd suggest adding more `x :#= DontCare` in `Edge.scala` for some corner cases.

jerryz123 (approved):

jerryz123:
@mergifyio backport master

poemonsense:
I would like to submit a PR with commit https://github.com/OpenXiangShan/rocket-chip/commit/8f5e48a9a65a580efcc25f796d56e07189f4ee1b to fix the missing DontCares. Do you think this is necessary? Will the PR be accepted? @sequencer

sequencer:
sure! Please submit!

src/main/scala/tilelink/ToAXI4.scala
```diff
@@ -149,8 +149,8 @@ class TLToAXI4(val combinational: Boolean = true, val adapterName: Option[String
       val depth = if (combinational) 1 else 2
       val out_arw = Wire(Decoupled(new AXI4BundleARW(out.params)))
       val out_w = Wire(chiselTypeOf(out.w))
-      out.w :<> Queue.irrevocable(out_w, entries=depth, flow=combinational)
-      val queue_arw = Queue.irrevocable(out_arw, entries=depth, flow=combinational)
+      out.w :<> Queue.irrevocable(out_w, entries=depth, combinational)
```
poemonsense, line 152: Is the deletion of `flow=` intended? @sequencer It seems we must pass the argument explicitly by `flow=` unless this argument would be used as `pipe` instead of `flow`.

In https://github.com/chipsalliance/chisel/blame/main/src/main/scala/chisel3/util/Decoupled.scala#L394-L401
```scala
  @chiselName
  def irrevocable[T <: Data](
    enq:            ReadyValidIO[T],
    entries:        Int = 2,
    pipe:           Boolean = false,
    flow:           Boolean = false,
    useSyncReadMem: Boolean = false,
    flush:          Option[Bool] = None
  ): IrrevocableIO[T] = {
...
}
```

We have faced performance regression after bumping rocket to 0586532e47b679d472bed5dd2c70c246ba1da0c6 and are working on finding the root cause. Currently I'm seeing the Verilog differences because of this Queue. Still waiting for the XiangShan CI to see whether fixing this could address the performance issue. But up to now this line is quite suspective.

poemonsense: After reverting this line, the Verilog seems to be aligned with the same number of Queue* modules generated. The performance is to be tested at https://github.com/OpenXiangShan/XiangShan/pull/2353. Will see the results this afternoon

sequencer: This is certainly a bug. please PR ;p

poemonsense: Submitted

PR #3304, opened 2023-03-14
Change CharCountRoCC Example to use dcacheParams

jerryz123:
RoCC accesses D$, not I$


jerryz123:
@mergifyio backport master

ZenithalHourlyRate (approved):
It seems that these params actually have nothing to do with icache/dcache, instead they should be from tilelink.

It may be convenient to access these params from dcache param though.

Issue #3230, opened 2023-01-18
Documentation generation failed


pcotret:
**Type of issue**: bug report 

**Impact**: unknown

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
```bash
git clone https://github.com/chipsalliance/rocket-chip
cd rocket-chip
git submodule update --init
sbt
```
In the SBT prompt:
```bash
docs/mdoc
```
Got following errors:
```
[warn] Multiple main classes detected.  Run 'show discoveredMainClasses' to see the list
[info] running mdoc.Main --cwd docs
[error] (run-main-0) java.lang.NoClassDefFoundError: scala/tools/nsc/reporters/AbstractReporter
[error] java.lang.NoClassDefFoundError: scala/tools/nsc/reporters/AbstractReporter
[error]         at java.base/java.lang.ClassLoader.defineClass1(Native Method)
[error]         at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1012)
[error]         at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
[error] 	at java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:524)
[error] 	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:427)
[error] 	at java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:421)
[error] 	at java.base/java.security.AccessController.doPrivileged(AccessController.java:712)
[error] 	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:420)
[error] 	at sbt.internal.ManagedClassLoader.findClass(ManagedClassLoader.java:98)
[error] 	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:587)
[error] 	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
[error] 	at mdoc.internal.markdown.MarkdownBuilder$.fromClasspath(MarkdownBuilder.scala:92)
[error] 	at mdoc.internal.cli.Context$.fromOptions(Context.scala:33)
[error] 	at mdoc.internal.cli.Settings.validate(Settings.scala:231)
[error] Caused by: java.lang.ClassNotFoundException: scala.tools.nsc.reporters.AbstractReporter
[error] 	at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:445)
[...]

Issue #3312, opened 2023-03-20
Support building mdoc through Mill on dev


michael-etzkorn:
**Type of issue**: feature request 

**Impact**: no functional change 

**Development Phase**: request

**Other information**
 

**What is the current behavior?**
Since merging https://github.com/chipsalliance/rocket-chip/commit/0b7fef44f82b35bf289949b3f8fdc139641f8e10, the steps to build mdoc are no longer supported and need a Mill build equivalent to check the mdoc against the latest dev branch.

FurryAcetylCoA:
We can use [`yoohaemin/mill-mdoc`](https://github.com/yoohaemin/mill-mdoc/) to generate mdoc with mill. Note: the `readme.md` in this fork is not up to date. com.yoohaemin::mill-mdoc::0.0.3-7-355e3e-SNAPSHOT is the first release compatible with mill 0.11.

However, I'm not sure were to mix the MDocModule trait in. I tried to mix it with `RocketChip` and while it did compile the docs but it failed when emitting SystemVerilog(adder_tutorial.md:315). This issue arises because chisel5 compilation requires the plugin, and 'm not familiar with build.sc structure used by rocketchip. I don't know how to fix it ( should I just use another object extended from `RocketChip` ? )

Also, there are problems with executing ChiselStage.elaborate(top.module) in select_tutorial.md:80 even after adjusting the package path. I believe it has something to do with the internal API change of chisel, which, unfortunately, I'm also not familiar with.

PR #3313, opened 2023-03-20
Fix mdoc build failing (sbt)

michael-etzkorn:
**Related issue**: 
Closes: #3230 

Other related issues: #2960 #3312 

**Notes**
Because dev branch does not support mdoc, this was fixed directly off master. If #3312 is fixed, these changes can be downstreamed into dev, but in the mean time, we should fix mdoc building on master.


michael-etzkorn:
I could remove the build.sbt changes and rebase this to dev? I'm just not sure how to build the Diplomacy docs with mill.

PR #3314, opened 2023-03-20
Hypervisor: encodeVirtualAddress extra MSB to canonicalize VS-disabled

ingallsj:
**Related issue**: 

**Impact**: functional fix

**Release Notes**
The `encodeVirtualAddress` function in the cache pipeline relies on setting the sign-extended-mask extra bit to the opposite value, but the `badGPA` (i.e. `badVA(false)`) function in the ITLB/DTLB ignores the sign-extended-mask check for GPA when stage-1-**disabled** and stage-2-enabled.
So, when a VA (a.k.a. GPA in this mode Sv39x4) has bits[63:41]!=0 and bit[40]==1, the `encodeVirtualAddress` function sets bit [41]==0, which passes the `badGPA` (i.e. `badVA(false)`) function check.


aswaterman (approved):

jerryz123:
@mergifyio backport master

PR #3319, opened 2023-03-23
Use chisel 3.6 rc2 instead of SNAPSHOT

ZenithalHourlyRate:
**Related issue**: 

Relying on chisel snapshot is causing a lot of pain for rocket chip and its dependencies like hardfloat/inclusive-cache/blocks, especially chisel master now removes Chisel and moves Cat into chisel3._. It even breaks the CI for PRs of last week.

In the dev meeting it is decided that we should rely on 3.6 instead of snapshot.

**Release Notes**

### DIFF
--- a/common.sc
+++ b/common.sc
@@ -4,8 +4,8 @@ import mill.scalalib.publish._
 import coursier.maven.MavenRepository
 
 val defaultVersions = Map(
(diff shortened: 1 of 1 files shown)


jerryz123 (approved):
Very happy to see this.

Issue #170, opened 2016-07-18
weird output of C emulator


harriss95:
I did a C simulator and this is the output file for rv64ui-p-add.out

>C0:          0 [0] pc=[529b4473c8] W[r  0=0000000000000000][0] R[r 4=b9c1d835c084f7a4] R[r16=1ffebe133e9aa313] inst=[db025a65] DASM(db025a65)
C0:          1 [0] pc=[529b4473c8] W[r  0=0000000000000000][0] R[r 4=b9c1d835c084f7a4] R[r16=1ffebe133e9aa313] inst=[db025a65] DASM(db025a65)
C0:          2 [0] pc=[529b4473c8] W[r  0=0000000000000000][0] R[r 4=b9c1d835c084f7a4] R[r16=1ffebe133e9aa313] inst=[db025a65] DASM(db025a65)
C0:          3 [0] pc=[529b4473c8] W[r  0=0000000000000000][0] R[r 4=b9c1d835c084f7a4] R[r16=1ffebe133e9aa313] inst=[db025a65] DASM(db025a65)

can anyone help me what might be the reason for this ?

terpstra:
This looks correct to me.

Does it at line 19 or so say something like:

C0:         19 [1] pc=[0000001000] W[r  0=0000000000001004][1] R[r
0=0000000000000000] R[r 0=0000000000000000] inst=[0000006f] j       pc + 0x0

harriss95:
yes . that is exactly what i got at line 19

C0:         19 [1] pc=[0000001000] W[r  0=0000000000001004][1] R[r 0=0000000000000000] R[r 0=0000000000000000] inst=[0000006f] DASM(0000006f)

can i know where u get that 

> j       pc + 0x0

terpstra:
If you pipe the output through spike-dasm it will show you the instructions
disassembled.

PR #3328, opened 2023-04-21
FIX (improve-boot-1): change reset vector address

T-K-233:
**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/devices/tilelink/BootROM.scala
+++ b/src/main/scala/devices/tilelink/BootROM.scala
@@ -17,7 +17,7 @@ import java.nio.file.{Files, Paths}
 case class BootROMParams(
   address: BigInt = 0x10000,
   size: Int = 0x10000,
-  hang: BigInt = 0x10040, // The hang parameter is used as the power-on reset vector
+  hang: BigInt = 0x10000, // The hang parameter is used as the power-on reset vector
   contentFileName: String)
 
 class TLROM(val base: BigInt, val size: Int, contentsDelayed: => Seq[Byte], executable: Boolean = true, beatBytes: Int = 4,


T-K-233:
Note: this depends on [testchipip #170](https://github.com/ucb-bar/testchipip/pull/170)

ZenithalHourlyRate:
rocketchip itself does not depend on testchipip, that dependency is not valid.

If you want to change the bootrom reset vector, you should also change `bootrom/bootrom.S`, otherwise it will malfunction.

Or you can use `BootROMLocated` to override the reset vector in your target framework (maybe chipyard?)

One example I can give you is
https://github.com/OpenRigil/openrigil-rtl/blob/710dc7cec0302b79feda1c1c6e0bebe66d776188/sanitytests/rocketchip/src/OpenRigil.scala#L93-L126

ZenithalHourlyRate:
> I wonder if there's a reason why we set the reset vector to 0x40 offset...? If not, I can also make a change to the bootrom and linker script here then.

As you can see in `bootrom/linker.ld`, 0x10000 is for `_start`, which jumps to DRAM directly, and 0x10040 is for `_hang`, which will wait for an interrupt from the debug module. By default, the rocket chip emulator starts by waiting for a debug request from htif.

> I prefer to not change the bootrom here.

Agreed.

T-K-233:
> > I wonder if there's a reason why we set the reset vector to 0x40 offset...? If not, I can also make a change to the bootrom and linker script here then.
> 
> As you can see in `bootrom/linker.ld`, 0x10000 is for `_start`, which jumps to DRAM directly, and 0x10040 is for `_hang`, which will wait for an interrupt from the debug module. By default, the rocket chip emulator starts by waiting for a debug request from htif.


Is there a reason to arrange the `_start` and `_hang` sections as such in the BootROM? It's quite unintuitive that the program does not start from the base of the BootROM. 

One trouble I (and several others) have run into is when we tried to debug our design, we wanted to override the BootROM with some simple code (say, continuously dumping characters to UART), we were so confused why the program was not executing properly. The doc did not point out that the hart starts at this 0x10040 offset address....

If there's not specific reason, I wish to swap the `_start` and `_hang` sections, and make the reset PC at `0x10000` for everyone's convenience. 

Thanks.

jerryz123:
You can configure the start and hang addresses when you specify a custom bootrom. The default here should not need to change

jerryz123:
Closing, the default bootrom should not change

PR #3331, opened 2023-04-24
config: add WithBootFreq configuration

cyyself:
The rocket-chip will set clock-frequency to 0 in the CPU fields of the device tree. And the value is taken from bootFreqHz in the RocketCoreParams which is zero by default. Some devices will use this property to calculate the clock divider ratio, but a zero value will cause division overflow.

This patch allows setting the value by simply adding WithBootFreq to the configuration, so we will have the correct clock-frequency in the dts.

### DIFF
--- a/src/main/scala/subsystem/Configs.scala
+++ b/src/main/scala/subsystem/Configs.scala
@@ -514,6 +514,14 @@ class WithTimebase(hertz: BigInt) extends Config((site, here, up) => {
   case DTSTimebase => hertz
 })
 
+class WithBootFreq(hertz: BigInt) extends Config((site, here, up) => {
+  case TilesLocated(InSubsystem) => up(TilesLocated(InSubsystem), site) map {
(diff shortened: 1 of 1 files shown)


src/main/scala/subsystem/Configs.scala
```diff
@@ -514,6 +514,14 @@ class WithTimebase(hertz: BigInt) extends Config((site, here, up) => {
   case DTSTimebase => hertz
 })
 
+class WithBootFreq(hertz: BigInt) extends Config((site, here, up) => {
```
jerryz123, line 517: Change to WithRocketTileBootFreq

ZenithalHourlyRate:
Can this be done by taking values from clockNode from diplomacy? `ClockParameters` has the `freqMHz` field.

jerryz123:
I think it would be more useful if this also set the ClockSinkParams

jerryz123:
IIRC the tile ClockSinkParams currently are not given a freq

PR #3332, opened 2023-04-25
all isaDTS strings to lowercase

cyyself:
The Linux Kernel requires letters in the riscv,isa string in the device tree must be all lowercase to simplify parsing. [see here](https://github.com/torvalds/linux/blob/v6.3/Documentation/devicetree/bindings/riscv/cpus.yaml#L82)

However, rocket-chip will produce an ISA string with uppercase letters will cause the kernel to misbehave. Especially when using defconfig Linux with rocket-chip without FPU, the ISA string parser will confuse the current kernel that the CPU has "f" from "Zifencei", and the kernel will panic at `__fstate_restore` due to illegal instruction exceptions.

I tried to patch the Linux Kernel, but a maintainer thinks it will make the device tree have no backward compatibility. In my opinion, the rocket-chip should be fixed, and the parser in the kernel should also be fixed to avoid illegal parsing causing the kernel to misbehave.

In this PR, I changed all the ISA strings to lowercase to meet the kernel's requirements, and the issue was solved.

**Related issue**: See [https://lore.kernel.org/lkml/20230425-flyable-prompter-5b1e4cebf9db@wendy/](https://lore.kernel.org/lkml/20230425-flyable-prompter-5b1e4cebf9db@wendy/)

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -64,7 +64,7 @@ case class RocketCoreParams(
   val instBits: Int = if (useCompressed) 16 else 32
   val lrscCycles: Int = 80 // worst case is 14 mispredicted branches + slop
   val traceHasWdata: Boolean = false // ooo wb, so no wdata in trace
-  override val customIsaExt = Some("Xrocket") // CEASE instruction
+  override val customIsaExt = Some("xrocket") // CEASE instruction
   override def minFLen: Int = fpu.map(_.minFLen).getOrElse(32)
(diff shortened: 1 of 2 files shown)


jerryz123 (approved):

ConchuOD:
@cyyself Thanks for doing this!

ZenithalHourlyRate:
Can this be made against the dev branch? It will be backported to the master branch using mergify.

cyyself:
> Can this be made against the dev branch? It will be backported to the master branch using mergify.

OK. Moved to #3333.

PR #3335, opened 2023-04-27
rename M_HLVX to prevent clashing with M_FLUSH

Sh-Anand:

**Release Notes**

Renamed the constant M_HLVX from `10000` to `11000` to prevent clashing with M_FLUSH, as `isRead` would always return true for memory commands with `M_FLUSH`.

Note: The bug was spotted while working with `boom/src/main/scala/lsu/dcache.scala` of the BOOM. This prevents `M_FLUSH` commands from sending a response as `s2_send_resp` would always evaluate to false.


ZenithalHourlyRate (approved):
After inspecting M_HLVX related spec/code, it seems like a mistake instead of an intended clashing.

The order of the constants is also a hint.

Could you make this change against the dev branch? It will be backported to the master branch using mergify.

PR #3354, opened 2023-05-10
Bump master to chisel 3.6

jerryz123:


ZenithalHourlyRate:
chisel 3.6 breaks mill. See https://github.com/chipsalliance/chisel/pull/3045#issuecomment-1523815303

ZenithalHourlyRate:
> What's the solution for this? The SBT release flow for 3.6.0 on master is fine for now.

I assume backporting that fix to chisel 3.6.x branch and then releasing a 3.6.1 should work. @sequencer for this.

sequencer:
I think depending on maven release rather than compiling from source would help.

ZenithalHourlyRate:
> chisel 3.6 breaks mill.

Oh I made a mistake here. I encountered such problem when working with a project that compiles chisel 3.6 from source. rocket-chip uses 3.6 from maven so things should be fine.

jerryz123:
@ZenithalHourlyRate or @sequencer can one of you update this PR with the fixes for mill?

sequencer:
I can do it after going back to keyboard tonight

src/main/scala/util/HeterogeneousBag.scala
```diff
@@ -2,17 +2,17 @@
...
 
-import Chisel._
-import chisel3.Record
+import chisel3._
+import chisel3.experimental.DataMirror
 import scala.collection.immutable.ListMap
 
 final case class HeterogeneousBag[T <: Data](elts: Seq[T]) extends Record with collection.IndexedSeq[T] {
-  def apply(x: Int) = elts(x)
+  val elements = ListMap(elts.zipWithIndex.map { case (n,i) => (i.toString, DataMirror.internal.chiselTypeClone(n)) }:_*)
```
jackkoenig, line 10: Since rocket-chip is on Scala 2.13 now, I would recommend using `VectorMap` instead of `ListMap`, it's **much** faster with large collection sizes.

src/main/scala/util/HeterogeneousBag.scala
```diff
@@ -2,17 +2,17 @@
...
-import scala.collection.immutable.ListMap
+import chisel3._
+import chisel3.experimental.DataMirror
+import scala.collection.immutable.VectorMap
 
 final case class HeterogeneousBag[T <: Data](elts: Seq[T]) extends Record with collection.IndexedSeq[T] {
-  def apply(x: Int) = elts(x)
+  val elements = VectorMap(elts.zipWithIndex.map { case (n,i) => (i.toString, DataMirror.internal.chiselTypeClone(n)) }:_*)
+
+  def apply(x: Int): T = elements.values.toSeq(x)
```
jackkoenig, line 12:
```suggestion
  def apply(x: Int): T = {
    val key = elements.keys(x)
    elements(key)
  }
```
Now that `elements` is a `VectorMap`, we can make `apply` actually `O(1)` (as is implied by extending `IndexedSeq`), whereas the current implementation is `O(n)`.

sequencer:
I think mill doesn't break it since it already depends on the maven version of chisel

jerryz123:
Please don't merge this yet, we are encountering some issues with 3.6 bump

ZenithalHourlyRate:
CI is happy now. Failure on riscv-arch-test CI is irrelevant.

ZenithalHourlyRate:
> we are encountering some issues with 3.6 bump

I assume we could merge now?

jerryz123:
Please wait, I found many bugs, I want to fix them for 3.5.6 first before bumping to 3.6.0

jerryz123:
We may also need to delay this until we have a suitable replacement for fixed-point in 3.6.0

jackkoenig:
> We may also need to delay this until we have a suitable replacement for fixed-point in 3.6.0

FixedPoint is still supported in 3.6.0, it's just deprecated, you can suppress the deprecation warning. Just note that FixedPoint does not work in MFC so a suitable replacement that works in MFC would be very handy.

There are many interested in a Chisel-only FixedPoint implementation: https://github.com/chipsalliance/chisel/issues/3161

jerryz123:
> > We may also need to delay this until we have a suitable replacement for fixed-point in 3.6.0
> 
> 
> 
> FixedPoint is still supported in 3.6.0, it's just deprecated, you can suppress the deprecation warning. Just note that FixedPoint does not work in MFC so a suitable replacement that works in MFC would be very handy.
> 
> 
> 
> There are many interested in a Chisel-only FixedPoint implementation: https://github.com/chipsalliance/chisel/issues/3161

Oh great, that's good to hear, thanks. I'm losing track of what's getting dropped and what's getting deprecated each release.

ZenithalHourlyRate (approved):

PR #3356, opened 2023-05-10
Support custom bundles in Trace

jerryz123:
An alternative approach to #3355, core implementations can set a custom `traceCustom` field to pull out extra information out of the `TraceBundle`.

@bgottschall does this look OK for you?

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -812,7 +812,7 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
   csr.io.rw.addr := wb_reg_inst(31,20)
   csr.io.rw.cmd := CSR.maskCmd(wb_reg_valid, wb_ctrl.csr)
   csr.io.rw.wdata := wb_reg_wdata
-  io.trace := csr.io.trace
+  io.trace.insns := csr.io.trace
   for (((iobpw, wphit), bp) <- io.bpwatch zip wb_reg_wphit zip csr.io.bp) {
     iobpw.valid(0) := wphit
     iobpw.action := bp.control.action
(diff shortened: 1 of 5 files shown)


jerryz123:
@ZenithalHourlyRate @sequencer the CI machines do not provide logs, and the mill flow on dev fails for me with
```
[5/140] mill.scalalib.ZincWorkerModule.worker 
1 targets failed
mill.scalalib.ZincWorkerModule.worker java.lang.ClassNotFoundException: mill.scalalib.worker.ZincWorkerImpl
    java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:476)
    mill.api.ClassLoader$$anon$1.findClass(ClassLoader.scala:47)
    java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:589)
    java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
    mill.scalalib.ZincWorkerModule.$anonfun$worker$3(ZincWorkerModule.scala:58)
make: *** [/nscratch/jerryz/chipyard-proj/rocket-chip/Makefrag:47: /nscratch/jerryz/c
```

bgottschall:
> @bgottschall does this look OK for you?

Yes, that looks good!

sequencer:
I have two concerns, both these are not problem in the near future, and won't block this PR:
- `Data` cannot be serialized, although this is a common issue in most of Chisel designs, but I think we should eventually get rid of this in the user API of rocket-chip. user can use json to config their IP, rather than CDE.
- Trace has two design purpose: one for riscv-trace-spec, one for firesim-like debug. I personally think the later one is a part of verification and should be implemented with XMR.

jerryz123:
> * `Data` cannot be serialized, although this is a common issue in most of Chisel designs, but I think we should eventually get rid of this in the user API of rocket-chip. user can use json to config their IP, rather than CDE.

`traceCustom` doesn't have to be a user API, it is a source API, to allow other implementations of Tile to provide more information through that port. If a user wants to add more signals to that, then they have to edit source code to drive those signals from within the core anyways.

> * Trace has two design purpose: one for riscv-trace-spec, one for firesim-like debug. I personally think the later one is a part of verification and should be implemented with XMR.

Unfortunately firesim does not support XMR yet

sequencer (approved):

jerryz123:
@mergifyio backport master

PR #3358, opened 2023-05-12
Support RoCC accels which define CSRs

jerryz123:

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -357,11 +357,13 @@ class VType(implicit p: Parameters) extends CoreBundle {
 
 class CSRFile(
   perfEventSets: EventSets = new EventSets(Seq()),
-  customCSRs: Seq[CustomCSR] = Nil)(implicit p: Parameters)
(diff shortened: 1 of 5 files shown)


ZenithalHourlyRate (commented):
It took me some time to understand how each rocc csr is connected to the corresponding rocket csrfile rocc csr, especially when there are multiple RoCCs. The flatten magic is quite tricky.

The semantic of `nRoCCCSRs` in `HasCoreIO` and `LazyRoCCModuleImp` are different, and the very thing that they instantiated the same `RoCCCoreIO` makes me very confusing.

I think a different name should clarify the difference, like `nRoCCsCSRs` and `nRoCCCSRs`? Or if the extra _s_ is not significant to the reader, we can use some name like `nAllRoCCCSRs`? Some other names could also be clarified like `roccCSRIOs` and `roccCSRs` in `HasLazyRoCC`

jerryz123:
I updated the naming to make clarify the confusion between single-accelerator CSRs and aggregated CSRs with nTotalRoCCCSRs

ZenithalHourlyRate (approved):

jerryz123:
@mergifyio backport master

PR #3362, opened 2023-05-15
Zcb: Code Size Reduction Extension

Abdulwadoodd:

**Background**
RISC-V [code size reduction extension](https://github.com/riscv/riscv-code-size-reduction) supersedes the existing standard `C` (compressed) extension. It divides instructions in the `C` extension into three sub-extensions (`Zca`, `Zcf`, and `Zcd`) and exploits the reserved encodings in `C` to add more instructions in three sub-extensions (`Zcb`, `Zcmp` and `Zcmt`). 
RocketCore already supports C-extension i.e. Zca, Zcf, and Zcd. The un-privilege spec classifies compressed instructions into three different parts: Quadrant 0, Quadrant 1, and Quadrant 2. This classification is done based on the Opcode of compressed instruction: bits[1:0].

**Implementation**
Zcb instructions fall into Quadrant 0 and Quadrant 1 having opcodes 2'b00 and 2'b01 respectively. Hence, Implementation is done in two parts:

_Part 1_: (Related commits: [00f7580](https://github.com/chipsalliance/rocket-chip/commit/00f75804620ace4acb351ddbe1d5583100ffe32a), [d7b122e](https://github.com/chipsalliance/rocket-chip/commit/d7b122e074b805cc7738008fb0b36bcb3743545f))
In part 1, the Opcodes of the instructions belonging to Quadrant 0 are added, and the corresponding 16-bit->32-bit decode logic is hooked up with Q0 logic for the existing instructions. The instructions in this part are the compressed version of I-type load-store instructions: `lbu`, `lh`, `lhu`, `sb`, `sh`.

_Miscellaneous changes_:
- Updated `gitignore(s)`
- Updated `clean` target in Makefile
- Add a variable in Makefile to pass CONFIG from the command line while building verilator model. 

**Verification**
A simple assembly program is written including all the `Zcb` instructions to test their functionality. It is available [here](https://github.com/Abdulwadoodd/rocket-chip/tree/Zc_dev/tests). It is a bit crude way of testing in the early stage. I'm planning to develop more mature tests covering the corner cases which will come as part of riscv-arch tests. 

_Part 2_: (Related commits: [cf10d81](https://github.com/chipsalliance/rocket-chip/commit/cf10d814b220c385050d30fdd642244f8bc6cd01), [9a44292](https://github.com/chipsalliance/rocket-chip/commit/9a44292df2536891f16abfe748fdf0cdd3a51324))
In part 2, the Opcodes of the instructions belonging to Quadrant 1 are added, and the corresponding 16-bit->32-bit decode logic is hooked up with Q1 logic for the existing instructions. The instructions in this part are the has dependency on I-type, M-type, and  Bit-manip (`Zba`, `Zbb`) instructions: `c.zext.b`, `c.sext.b`, `c.zext.h`, `c.sext.h`, `c.zext.w`, `c.mul`, `c.not`.

**Release Notes**
This PR adds hardware implementation to support the `Zcb` extension which is one of the variants of the recently ratified `code size reduction` extension. RocketCore has the infrastructure for the uncompressed version of `Zcb` instructions. So, mainly there are changes implemented in RVCDecoder to decode the compressed `Zcb` instruction into its 32-bit equivalent instructions. The implementation works for both 32-bit and 64-bit architectures.


src/main/scala/rocket/RVC.scala
```diff
@@ -60,15 +62,24 @@ class RVCDecoder(x: UInt, xLen: Int, useAddiForMv: Boolean = false) {
...
+    }
     def sd = inst(Cat(ldImm >> 5, rs2p, rs1p, 3.U(3.W), ldImm(4,0), 0x23.U(7.W)), rs2p, rs1p, rs2p)
     def sw = inst(Cat(lwImm >> 5, rs2p, rs1p, 2.U(3.W), lwImm(4,0), 0x23.U(7.W)), rs2p, rs1p, rs2p)
     def fsd = inst(Cat(ldImm >> 5, rs2p, rs1p, 3.U(3.W), ldImm(4,0), 0x27.U(7.W)), rs2p, rs1p, rs2p)
     def fsw = {
       if (xLen == 32) inst(Cat(lwImm >> 5, rs2p, rs1p, 2.U(3.W), lwImm(4,0), 0x27.U(7.W)), rs2p, rs1p, rs2p)
       else sd
     }
-    Seq(addi4spn, fld, lw, flw, unimp, fsd, sw, fsw)
+    Seq(addi4spn, fld, lw, flw, zcb_q0, fsd, sw, fsw)
```
ZenithalHourlyRate, line 82: Zcb should be populated with conditions

For example, `c.zext.b` should be added only when `usingBitManip`, and `c.mul` should be populated only when `usingMulDiv`, `zext.w` should be added only when RV64. This could save a tiny bit of area.

Other insts like `c.lbu` and `c.not` will always be added as `I` or `E` are always present. This default behavior will add area overhead for end users as they may not use it. I think we provide an option like `usingCompressedBitManip` to control this behavior. Asking @jerryrzhao for advice.

src/main/scala/rocket/Instructions.scala
```diff
@@ -76,17 +76,26 @@ object Instructions {
   def C_J                = BitPat("b????????????????101???????????01")
   def C_JALR             = BitPat("b????????????????1001?????0000010")
   def C_JR               = BitPat("b????????????????1000?????0000010")
+  def C_LBU              = BitPat("b????????????????100000????????00")
```
ZenithalHourlyRate, line 79: Actually, these are not used in IDecode. But it is OK to include them.

ZenithalHourlyRate (changes requested):
Generally LGTM. Have not checked the detail of the RTL implementation against the spec. Some other small changes are needed.

First, this should be made against the dev branch, it will then be backported the to master branch. @sequencer should we add a contribution guide in the PR template as there are multiple contributors that have fallen into this pitfall?

As for the misc changes, they should be introduced in another PR.

You should also add `zcb` to the dts. `zca` should also be included. In the future when you have implemented all `Zc*`, a `zce` should be added.

src/main/scala/rocket/RVC.scala
ZenithalHourlyRate: A mistake. Should @jerryz123.

src/main/scala/rocket/Instructions.scala
jerryz123: Shouldn't this file be auto-generated?

src/main/scala/rocket/RVC.scala
jerryz123: Parameters to the RVCDecode object should be added that disable the Zcb/Zcmp/Zcmt extensions if not needed.

src/main/scala/rocket/RVC.scala
```diff
@@ -16,7 +16,7 @@ class ExpandedInstruction extends Bundle {
   val rs3 = UInt(5.W)
 }
 
-class RVCDecoder(x: UInt, xLen: Int, useAddiForMv: Boolean = false) {
+class RVCDecoder(x: UInt, xLen: Int, useAddiForMv: Boolean = false, usingBitManip: Boolean = false, usingMulDiv: Boolean = false, usingCompressedSuiteB: Boolean = false) {
```
jerryz123, line 19: Can you add some comments describing how ISA features map to these flags

ZenithalHourlyRate: Following the current convention, I assume we should use

```
class RVCDecoder(x: UInt, xLen: Int) extends HasCoreParameters {
```
then we can use these `using`s freely.

But as [chipsalliance/rocket](https://github.com/chipsalliance/rocket) is trying to remove `HasCoreParameters` everywhere and put `using`s to the class parameter, I do not know which one is recommended.

Abdulwadoodd: I actually tried exactly this in the first place but got the following error:

Abdulwadoodd:
Hi @jerryz123 @ZenithalHourlyRate
- Comments are added in RVC.scala describing the purpose of added parameters in RVCDecoder.
- zca, zcb, zcd, zcf are updated in DTS.
- After rebase, updates the base branch from `master` to `dev`.

Please let me know if additional changes are required.

src/main/scala/rocket/RVC.scala
```diff
@@ -95,16 +119,50 @@ class RVCDecoder(x: UInt, xLen: Int, useAddiForMv: Boolean = false) {
...
+              else unimp
+            }
+            def zexth = {
+              if(usingBitManip) {
+                if (xLen == 32) inst(Cat(0x80.U, rs1p, 4.U(3.W), rs1p, 0x33.U(7.W)), rs1p, rs1p, rs2p)
+                else zexth64
+              }
+              else unimp
+            }
+            Seq(zextb, sextb, zexth, sexth, zextw, not)(x(4,2))
```
ZenithalHourlyRate, line 159: When `usingCompressedSuiteB` but not `usingBitManip`, this will become Mux between a lot of unimp. Is there better way to do it?

ZenithalHourlyRate (commented):
I'm happy with the non-circuit part now. Need more time to check the circuit part.

src/main/scala/rocket/RVC.scala
```diff
@@ -95,16 +119,50 @@ class RVCDecoder(x: UInt, xLen: Int, useAddiForMv: Boolean = false) {
...
+        val opc = Mux(x(12), Mux(x(6), 0x33.U(7.W), 0x3B.U(7.W)), 0x33.U(7.W))
+        def mul = {
+          if(usingMulDiv && usingCompressedSuiteB) Mux(Cat(x(12), x(6,5)) === 6.U, (1 << 25).U, 0.U)
+          else 0.U
+        }
+        def zca = inst(Cat(rs2p, rs1p, funct, rs1p, opc) | sub | mul, rs1p, rs1p, rs2p)
+        def zcb_q1 = {
+          def unimp = inst(Cat(lwImm >> 5, rs2p, rs1p, 2.U(3.W), lwImm(4,0), 0x3F.U(7.W)), rs2p, rs1p, rs2p)
+          if(usingCompressedSuiteB){
+            def zextb = inst(Cat(0xff.U, rs1p, 7.U(3.W), rs1p, 0x13.U(7.W)), rs1p, rs1p, rs2p)
```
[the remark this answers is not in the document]
Abdulwadoodd: Fixed

src/main/scala/rocket/RVC.scala
```diff
@@ -95,16 +119,50 @@ class RVCDecoder(x: UInt, xLen: Int, useAddiForMv: Boolean = false) {
...
+            def not   = inst(Cat(0xFFF.U, rs1p, 4.U(3.W), rs1p, 0x13.U(7.W)), rs1p, rs1p, rs2p)
+            def zexth64 = inst(Cat(0x80.U, rs1p, 4.U(3.W), rs1p, 0x3B.U(7.W)), rs1p, rs1p, rs2p)
```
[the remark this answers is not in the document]
Abdulwadoodd: I followed the existing convention as a selection between `fs` and `sd` based on the `xlen`.
Anyways, it is updated as you suggested

src/main/scala/rocket/RVC.scala
```diff
@@ -95,16 +119,50 @@ class RVCDecoder(x: UInt, xLen: Int, useAddiForMv: Boolean = false) {
     def beqz = inst(Cat(bImm(12), bImm(10,5), x0, rs1p, 0.U(3.W), bImm(4,1), bImm(11), 0x63.U(7.W)), rs1p, rs1p, x0)
     def bnez = inst(Cat(bImm(12), bImm(10,5), x0, rs1p, 1.U(3.W), bImm(4,1), bImm(11), 0x63.U(7.W)), x0, rs1p, x0)
     def arith = {
-      def srli = Cat(shamt, rs1p, 5.U(3.W), rs1p, 0x13.U(7.W))
-      def srai = srli | (1 << 30).U
-      def andi = Cat(addiImm, rs1p, 7.U(3.W), rs1p, 0x13.U(7.W))
+      def srli = inst(Cat(shamt, rs1p, 5.U(3.W), rs1p, 0x13.U(7.W)), rs1p, rs1p, rs2p)
+      def srai = inst(Cat(0x10.U, shamt, rs1p, 5.U(3.W), rs1p, 0x13.U(7.W)), rs1p, rs1p, rs2p)
```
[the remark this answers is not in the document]
Abdulwadoodd: Fixed

src/main/scala/rocket/RVC.scala
```diff
@@ -95,16 +119,50 @@ class RVCDecoder(x: UInt, xLen: Int, useAddiForMv: Boolean = false) {
...
+              else unimp
+            }
+            Seq(zextb, sextb, zexth, sexth, zextw, not)(x(4,2))
+          }
+          else unimp
+        }
+        Mux(Cat(x(12), x(6,5)) === 7.U, zcb_q1, zca)
       }
-      inst(Seq(srli, srai, andi, rtype)(x(11,10)), rs1p, rs1p, rs2p)
+      Seq(srli, srai, andi, rtype)(x(11,10))
```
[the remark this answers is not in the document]
Abdulwadoodd: > Is there any special reason for it? The x0 in zextw?

Yes, this is the reason.

> the user will have a larger area even the user does not enable usingCompressedSuiteB

I'm a little skeptical about it, and I have following two reasons to back my argument:
1. As per my (vague) understanding of hardware (Verilog) generation using CHISEL, computation is done in software by keeping into account the parameters, and then the Verilog is generated including (only) that hardware part that was allowed by the parameter.
2. The above can be verified by observing the generated Verilog (by looking at the reference at the end of each line of generated Verilog that points to the corresponding scala file and its line number). For example, when I build Rocket with `usingCompressedSuiteB = false`, the generated Verilog simply excludes the corresponding definition of 16->32 instructions. But when I build using `usingCompressedSuiteB = true`, generated Verilog does include the references of the definition of those instructions which are defined under the if clause of this parameter.

So, this should not increase the area if `usingCompressedSuiteB` is disabled, unless I'm missing something.

src/main/scala/rocket/RVC.scala
Abdulwadoodd: Actually, not all of these (`zextb, sextb, zexth, sexth, zextw, not`) become Mux. The problem is that only `sextb, zexth, sexth, zextw` depends upon the `usingBitManip` parameter and `zextb` and `not` are decoded in I-type instructions. So, if we keep the `Seq` statement constant in the end, then EITHER `sextb, zexth, sexth, zextw` needs to be decoded OR if `usingBitManip` is enabled OR each one of these must be replaced by `unimp` if `usingBitManip` is disabled. Hence, comes up these bad if/else statements.

src/main/scala/rocket/RVC.scala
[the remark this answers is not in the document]
ZenithalHourlyRate: Further details should be checked. There are two steps of selection, static selection like if/elses in scala and dynamic selection like `x(11,10)` in the circuit.

The DefaultConfig in dev branch

```verilog
// the final mux
  assign io_out_bits = _GEN_19[_io_out_T_2];
  assign io_out_rd = _GEN_20[_io_out_T_2];
  assign io_out_rs1 = _GEN_21[_io_out_T_2];
  assign io_out_rs2 = _GEN_22[_io_out_T_2];
  assign io_out_rs3 = _GEN_23[_io_out_T_2];
  
// lets check rs1
  wire [31:0][4:0]  _GEN_21 =
  ...
    {io_out_s_12_rs1}, // this coresponds to arith in scala
  ...
  
  assign io_out_s_12_rs1 = {2'h1, io_in[9:7]};
```

DefaultConfig in this branch

```verilog
// the final mux
  assign io_out_bits = _GEN_23[_io_out_T_2];
  assign io_out_rd = _GEN_24[_io_out_T_2];
  assign io_out_rs1 = _GEN_25[_io_out_T_2];
  assign io_out_rs2 = _GEN_26[_io_out_T_2];
  assign io_out_rs3 = _GEN_27[_io_out_T_2];
  
// lets check rs1
  wire [31:0][4:0]  _GEN_25 =
  ...
    {io_out_s_12_rd}, // this coresponds to arith in scala
  ...
  
  wire [4:0]        io_out_s_12_rs1 = _GEN_8[io_in[11:10]]; // extra mux
  
  wire [3:0][4:0]   _GEN_8 =
    {{(&_io_out_s_T_235) ? io_out_s_res_4_rs1 : io_out_s_res_5_rs1},
     {io_out_s_res_3_rs1},
     {io_out_s_res_2_rs1},
     {io_out_s_res_1_rs1}};
     
  assign io_out_s_res_4_rs1 = {2'h1, io_in[9:7]};
  assign io_out_s_res_5_rs1 = {2'h1, io_in[9:7]};
  assign io_out_s_res_3_rs1 = {2'h1, io_in[9:7]};
  assign io_out_s_res_2_rs1 = {2'h1, io_in[9:7]};
  assign io_out_s_res_1_rs1 = {2'h1, io_in[9:7]};
```

Chisel is faithful of generating what you write. It might be argued that the synthesizer will optimize this out, but I think avoiding this from source is a better option.

src/main/scala/rocket/RVC.scala
ZenithalHourlyRate: What I meant is something like this

```scala
          if(usingCompressedSuiteB){
            def sexth = inst(Cat(0x605.U, rs1p, 1.U(3.W), rs1p, 0x13.U(7.W)), rs1p, rs1p, rs2p)
            }
            def zextw = inst(Cat(4.U, x0, rs1p, 0.U(3.W), rs1p, 0x3B.U(7.W)), rs1p, rs1p, x0)
            }
            def zexth = {
                val zexth_common = Cat(0x80.U, rs1p, 4.U(3.W), rs1p)
                if (xLen == 32) inst(Cat(zexth_common, 0x33.U(7.W)), rs1p, rs1p, rs2p)
                else inst(Cat(zexth_common, 0x3B.U(7.W)), rs1p, rs1p, rs2p)
              }
            }
        
            def zextb = inst(Cat(0xFF.U, rs1p, 7.U(3.W), rs1p, 0x13.U(7.W)), rs1p, rs1p, rs2p)
            def not   = inst(Cat(0xFFF.U, rs1p, 4.U(3.W), rs1p, 0x13.U(7.W)), rs1p, rs1p, rs2p)
            def sextb = inst(Cat(0x604.U, rs1p, 1.U(3.W), rs1p, 0x13.U(7.W)), rs1p, rs1p, rs2p)
            if (usingBitManip) {
              Seq(zextb, sextb, zexth, sexth, zextw, not)(x(4,2))
            } else {
              Mux(x(4,2) === 5.U, not, Mux(x(4,2) === 0.U, zextb, unimp))
            }
          }
```

And like the comment in `arith`, try to avoid selecting between a lot of rs1p/rs2p

src/main/scala/rocket/RVC.scala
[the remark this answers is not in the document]
Abdulwadoodd: Thanks for the detailed explanation, I understand now. Pushed fixes in the latest commit.

src/main/scala/rocket/RVC.scala
Abdulwadoodd: Updated the changes as suggested. Thanks

PR #3373, opened 2023-05-19
Add [m/s/h]envcfg CSRs to satisfy spec requirements

jerryz123:
I believe wiring them to 0 is ok.

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -124,6 +124,21 @@ class MIP(implicit p: Parameters) extends CoreBundle()(p)
   val usip = Bool()
 }
 
+class Envcfg extends Bundle {
+  val stce = Bool() // only for menvcfg/henvcfg
+  val pbmte = Bool() // only for menvcfg/henvcfg
+  val zero54 = UInt(54.W)
+  val cbze = Bool()
+  val cbcfe = Bool()
+  val cbie = UInt(2.W)
+  val zero3 = UInt(3.W)
+  val fiom = Bool()
+  def write(wdata: UInt) {
+    val new_envcfg = wdata.asTypeOf(new Envcfg)
+    fiom := new_envcfg.fiom // only FIOM is writable currently
+  }
+}
+
 class PTBR(implicit p: Parameters) extends CoreBundle()(p) {
   def additionalPgLevels = mode.extract(log2Ceil(pgLevels-minPgLevels+1)-1, 0)
   def pgLevelsToMode(i: Int) = (xLen, i) match {
@@ -287,6 +302,7 @@ class CSRFileIO(implicit p: Parameters) extends CoreBundle
   val trace = Output(Vec(retireWidth, new TracedInstruction))
   val mcontext = Output(UInt(coreParams.mcontextWidth.W))
   val scontext = Output(UInt(coreParams.scontextWidth.W))
+  val fiom = Output(Bool())
 
   val vector = usingVector.option(new Bundle {
     val vconfig = Output(new VConfig())
@@ -493,6 +509,10 @@ class CSRFile(
   val reg_rnmie = RegInit(true.B)
   val nmie = reg_rnmie
 
+  val reg_menvcfg = RegInit(0.U.asTypeOf(new Envcfg))
+  val reg_senvcfg = RegInit(0.U.asTypeOf(new Envcfg))
+  val reg_henvcfg = RegInit(0.U.asTypeOf(new Envcfg))
+
   val delegable_counters = ((BigInt(1) << (nPerfCounters + CSR.firstHPM)) - 1).U
   val (reg_mcounteren, read_mcounteren) = {
     val reg = Reg(UInt(32.W))
@@ -589,6 +609,7 @@ class CSRFile(
   io.bp := reg_bp take nBreakpoints
   io.mcontext := reg_mcontext.getOrElse(0.U)
   io.scontext := reg_scontext.getOrElse(0.U)
+  io.fiom := (reg_mstatus.prv < PRV.M.U && reg_menvcfg.fiom) || (reg_mstatus.prv < PRV.S.U && reg_senvcfg.fiom) || (reg_mstatus.v && reg_henvcfg.fiom)
   io.pmp := reg_pmp.map(PMP(_))
 
   val isaMaskString =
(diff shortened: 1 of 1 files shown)


ZenithalHourlyRate:
> I believe wiring them to 0 is ok.

For other bits it is ok as we do not have these extensions now (I'm finding people to implement them).

For FIOM, I believe it is not the case for us? Especially when user enabled H extension they might want to enable FIOM

> If S-mode is not supported, or if satp.MODE is read-only zero (always Bare), the implementation may make FIOM read-only zero

Or the software could set FIOM but maybe we can make it a no-op. Though I am not sure if it is OK and is still checking it.

jerryz123:
The spec seems to indicate that FIOM is writable, while in spike, FIOM is hardwired to 0.

ZenithalHourlyRate:
> Though I am not sure if it is OK and am still checking it.

Seems like rocket does not distinguish between PI and PR (similarly for other bits) so the implication requirement by FIOM (e.g. PI implies PR) is automatically there. We can safely do no-op when FIOM is set to 1.

https://github.com/chipsalliance/rocket-chip/blob/284b3608ccdd084e7fe157364c4f3c7e27001652/src/main/scala/rocket/RocketCore.scala#L349-L350

jerryz123:
Ok. I'm happy to change this such that only FIOM is writable in these registers.
I will also expose a `fiom` output from the CSR file since other core implementations may need to check FIOM for ordering fences.

src/main/scala/rocket/CSR.scala
```diff
@@ -688,6 +705,13 @@ class CSRFile(
 
     if (usingUser) {
       read_mapping += CSRs.mcounteren -> read_mcounteren
+      read_mapping += CSRs.menvcfg -> reg_menvcfg.asUInt
+      read_mapping += CSRs.senvcfg -> reg_senvcfg.asUInt
+      read_mapping += CSRs.henvcfg -> reg_henvcfg.asUInt
```
ZenithalHourlyRate, line 710: I assume henvcfg should be added to the read_mapping only when Hypervisor extension is enabled?

>  Additional CSRs are provided to HS-mode, but
not to VS-mode, to manage two-stage address translation and to control the behavior of a VSmode guest: hstatus, hedeleg, hideleg, hvip, hip, hie, hgeip, hgeie, henvcfg, henvcfgh,
hcounteren, htimedelta, htimedeltah, htval, htinst, and hgatp.

After checking the spec, this sentense only exists for menvcfg

> If U-mode is not supported, then registers menvcfg and menvcfgh do not exist.

Simiarly, senvcfg should be defined in the `usingSupervisor` block as it is possible to have `usingUser` but do not have `usingSupervisor`

src/main/scala/rocket/CSR.scala
```diff
@@ -124,6 +124,17 @@ class MIP(implicit p: Parameters) extends CoreBundle()(p)
   val usip = Bool()
 }
 
+class Envcfg extends Bundle {
+  val stce = Bool()
+  val pbmte = Bool()
+  val zero54 = UInt(54.W)
```
ZenithalHourlyRate, line 130: It is OK with me, but I have to say that senvcfg does not define these fields. Maybe a comment will clarify that?

src/main/scala/rocket/CSR.scala
jerryz123: Thanks for the rigorous review, this made me realize my initial implementation bound these registers to haveBasicCounters

src/main/scala/rocket/CSR.scala
```diff
@@ -1381,6 +1414,13 @@ class CSRFile(
     }
     if (usingUser) {
       when (decoded_addr(CSRs.mcounteren)) { reg_mcounteren := wdata }
+      when (decoded_addr(CSRs.menvcfg))    { reg_menvcfg.write(wdata) }
+    }
+    if (usingSupervisor) {
+      when (decoded_addr(CSRs.senvcfg))    { reg_senvcfg.write(wdata) }
+    }
```
ZenithalHourlyRate, line 1421: https://github.com/chipsalliance/rocket-chip/blob/bd3276d398c50af21812bc054422609491aba3a3/src/main/scala/rocket/CSR.scala#L1309

I believe this is a better place?

src/main/scala/rocket/CSR.scala
```diff
@@ -1381,6 +1414,13 @@ class CSRFile(
     }
     if (usingUser) {
       when (decoded_addr(CSRs.mcounteren)) { reg_mcounteren := wdata }
+      when (decoded_addr(CSRs.menvcfg))    { reg_menvcfg.write(wdata) }
```
ZenithalHourlyRate, line 1417: I spend some time to find the path for `xxxh` CSRs, like `mstatush` and `cycleh`. Turned out there was some special utility

https://github.com/chipsalliance/rocket-chip/blob/bd3276d398c50af21812bc054422609491aba3a3/src/main/scala/rocket/CSR.scala#L1573-L1581


I believe the current implementation does not handle `csrw menvcfgh`

I also do not find the path for writing `mstatush`, do you have any clue?

ZenithalHourlyRate: > I also do not find the path for writing mstatush, do you have any clue?

Seems like this is part of privileged spec 1.12 (envcfg is also added in 1.12). Is rocket compatible with privileged spec 1.11 only and we need to spend some time upgrading?

jerryz123: That snippet is for counters only. We ought to work towards making it compliant with 1.12.

ZenithalHourlyRate: Other parts are fine with me now

> That snippet is for counters only. We ought to work towards making it compliant with 1.12.

I think `menvcfgh` and `henvcfgh` are the only problem left.

jerryz123: Writes do menvcfgh and menvcfgh should cause no effects at the moment, since FIOM is the low bit.

ZenithalHourlyRate: Understood. I assumed that we must have a `decoded_addr(menvcfgh)` somewhere to make it writable. Turned out it is not the case.

ZenithalHourlyRate (approved):

jerryz123:
@mergifyio backport master

PR #3379, opened 2023-05-30
Remove SFC

sequencer:
This PR removes the entire usage to `SFC`, some annotation still remains, but no transforms/stage exist. This will pave the path to Chisel5 support. 

**Impact**: API modification

**Release Notes**
Drop support to legacy SFC APIs


ZenithalHourlyRate (approved):

ZenithalHourlyRate:
Extra note on this PR in case someone bisected:

1. This breaks the `make` workflow for the dev branch
2. The `mill` flow (usage demoed in github actions) currently does not support `repl-seq-mem` so the RAM generated is only for emulator, and Xilinx FPGA BRAM inference will fail. If you need to bring rocket to FPGA, try master branch.

PR #3389, opened 2023-06-16
TLB: must_alloc swapped AMO Logical/Arithmetic

ingallsj:
**Related issue**: 

**Release Notes**
TLB: swap back so that `must_alloc` is decided by for the `amo_logical` term based on `!supportsLogical`, and similarly for `amo_arithmetic` based on `!supportsArithmetic`.

In practice, we don’t think this bug in the generator ever manifested in practice, because these supportsLogical and supportsArithmetic have always been the same in Diplomacy as far as we can recall.


aswaterman (approved):

sequencer:
Oh my...
Seems to be a long time standing performance issue...

ZenithalHourlyRate:
@mergifyio backport master

PR #3391, opened 2023-06-16
Zmmul: Multiply without Divide Extension

nananapo:

**Release Notes**

This PR is to add Zmmul extension on the rocket-chip.
This adds a new Config `WithMulDivWithoutDiv` and a new parameter `useMulDivWithoutDiv` in CoreParams.

When `usingMulDivWithoutDiv` is true, MulDivParams.divUnroll is replaced with 0 to disable div logic of MulDiv. And div/rem instruction is not included in a decode table.


src/main/scala/tile/Core.scala
```diff
@@ -30,6 +30,7 @@ trait CoreParams {
   val useCryptoSM: Boolean
   val useRVE: Boolean
   val useConditionalZero: Boolean
+  val useMulDivWithoutDiv: Boolean
```
ZenithalHourlyRate, line 33: context: We had some prior private communication.

After seeing the code things become more clear to me, It seems that `useMulDivWithoutDiv` can be merged into `MulDiv` by setting `divUnroll = 0` instead of a standalone option?

I favor this simple way. And for the exposed interface to the end user, he/she only needs to use some config like `WithoutDiv` and that config will set `divUnroll = 0`

But this is tricky as `divUnroll == 0` has two purpose for developer now

1. Zmmul, namely turning off div.
2. Alternative divider impl suggested in #1228

src/main/scala/tile/BaseTile.scala
```diff
@@ -118,6 +118,7 @@ trait HasNonDiplomaticTileParameters {
       Option.when(tileParams.core.useBitManip)(Seq("zbs")) ++
       Option.when(tileParams.core.useCryptoNIST)(Seq("zknd", "zkne", "zknh")) ++
       Option.when(tileParams.core.useCryptoSM)(Seq("zksed", "zksh")) ++
+      Option.when(tileParams.core.mulDiv.nonEmpty && tileParams.core.useMulDivWithoutDiv)(Seq("zmmul")) ++
```
ZenithalHourlyRate, line 121: 1. `zmmul` should be placed after `zihpm` instead of appending to the end

2. Some interesting thing: `rv32imac_zmmul` is a valid ISA string.

src/main/scala/subsystem/Configs.scala
```diff
@@ -382,6 +382,16 @@ class WithoutMulDiv extends Config((site, here, up) => {
   }
 })
 
+class WithMulDivWithoutDiv extends Config((site, here, up) => {
+  case TilesLocated(InSubsystem) => up(TilesLocated(InSubsystem), site) map {
+    case tp: RocketTileAttachParams => tp.copy(tileParams = tp.tileParams.copy(
+      core = tp.tileParams.core.copy(
+        mulDiv = Some(MulDivParams(mulUnroll = 8)),
```
ZenithalHourlyRate, line 389: Should not set mulUnroll here.

ZenithalHourlyRate (commented):
Overall good to me! Though the software design for enabling Zmmul needs further discussion, including the config name. `WithMulDivWithoutDiv` seems confusing.

Could you demo with an emulator that a div instruction will cause illegal instruction exception in this config?

Could you add an arch-test CI for zmmul?

src/main/scala/tile/Core.scala
ZenithalHourlyRate: After discussion with Jerry, it is concluded that adding another option inside `MulDivParams` like `divUnroll: Option[Int]` or (`divUnroll: Int` and `divEnable: Boolean = true`) should be used, instead of another boolean in `CoreParams`. The later way is prefered as it wont break compatibility.

ZenithalHourlyRate (changes requested):
Other parts look good to me!

src/main/scala/rocket/Multiplier.scala
```diff
@@ -29,6 +29,7 @@ class MultiplierIO(val dataBits: Int, val tagBits: Int, aluFn: ALUFN = new ALUFN
 case class MulDivParams(
   mulUnroll: Int = 1,
   divUnroll: Int = 1,
+  divEnabled: Boolean = false,
```
ZenithalHourlyRate, line 32: This should default to true.

src/main/scala/tile/BaseTile.scala
```diff
@@ -112,6 +112,7 @@ trait HasNonDiplomaticTileParameters {
       //Some(Seq("zicntr")) ++
       Option.when(tileParams.core.useConditionalZero)(Seq("zicond")) ++
       Some(Seq("zicsr", "zifencei", "zihpm")) ++
+      Option.when(tileParams.core.mulDiv.nonEmpty && !tileParams.core.mulDiv.get.divEnabled)(Seq("zmmul")) ++
```
ZenithalHourlyRate, line 115: Forgot to start a review...

Actually it should be 

```
      Option.when(tileParams.core.mulDiv.nonEmpty)(Seq("zmmul")) ++
```

Check https://github.com/riscv/riscv-isa-manual/issues/869 for more details.

And apply the following diff to the `m` instead

```diff
-val m = if (tileParams.core.mulDiv.nonEmpty) "m" else ""
+val m = if (tileParams.core.mulDiv.nonEmpty  && !tileParams.core.mulDiv.get.divEnabled) "m" else ""
``` 

Also you should update the MISA generation in CSR.scala

nananapo: Hi.
If I understand correctly, there is a diferrence between rv32im_zmmul and rv32im about mul inst when misa.M=0.
If I change ISA of rocket-chip to rv32im_zmmul when mulDiv.nonEmpty, the behaviour of mul inst with misa.M=0 will be changed to not cause illegal instruction exception.
Is it OK?

ZenithalHourlyRate: > If I understand correctly, there is a diferrence between rv32im_zmmul and rv32im about mul inst when misa.M=0.

It does exist in that thread, but from my perception it was not the focus of that thread and it was not finally decided. 

From https://github.com/riscv/riscv-isa-manual/issues/869#issuecomment-1186043486, a future `Sm*` extension should come to solve the problem. As it does not exist for now, I assume this detail is undefined, so it is now implementation-defined.

> If I change ISA of rocket-chip to rv32im_zmmul when mulDiv.nonEmpty, the behaviour of mul inst with misa.M=0 will be changed to not cause illegal instruction exception.

I think we should not change the behavior, as it is undefined, and counter-intuitive.

I think we should officially support only two modes: `m` and `zmmul`. As for `m_zmmul`, it is equivalent to `m` and the misa behavior is the same as `m`.

Now I wont insist that we should output `m_zmmul` for `m`. I am OK if you want to just output `m` or `zmmul` in the ISA string. One final commit/comment and I will merge.

Issue #3402, opened 2023-06-29
PTW incorrectly marks the access fault type for out-of-range non-leaf PTEs


poemonsense:
**Type of issue**: bug report

**Impact**: unknown

**Development Phase**: proposal

**Other information**
 

**If the current behavior is a bug, please provide the steps to reproduce the problem:**
In S mode, the PTW gets a valid, non-leaf PTE with a PPN exceeding the physical address space.

**What is the current behavior?**
Rocket-chip checks the PPNs (`!invalid_paddr` in PTW.scala:316) before traversing to the next level of page table. It thens returns `ae_final` to the TLB. However, since the page is non-leaf, it does not have the `r` bit set. Therefore, further a read request to TLB is responded with a page-fault exception.

**What is the expected behavior?**
According to the RISC-V manual 5.3.2, in step 4, if the PTE is valid and `pte.r = 0` and `pte.x = 0`, the PTE is a pointer to the next level of page table. Since we have not finished the three levels of page table, we continue to step 2 and `a = pte.ppn X PAGESIZE` to calculate the base address. If accessing the PTE violates the PMA, it should raise an access-fault exception.

This is the same behavior as in Spike. As seen in `mmu_t::walk`, if the PTE points to the next level of page table and does not have any bits of `PTE_D | PTE_A | PTE_U | PTE_N | PTE_PBMT` set, then we calculate the next base address with `base = ppn << PGSHIFT;` and continue to the next iteration of the for loop. In the next level, if accessing the PTE violates PMA (like exceeding the physical address space), an access-fault exception is raised.

**Please tell us about your environment:**

Found during co-simulating rocket-chip with riscv-isa-sim. Page-fault is delegated so rocket-chip goes to S mode while the Spike goes to M mode because of access-fault.
```
   mode different at pc = 0x0000011184, right= 0x0000000000000003, wrong = 0x0000000000000001
mstatus different at pc = 0x0000011184, right= 0x8000000a000060a2, wrong = 0x8000000a000060a8
sstatus different at pc = 0x0000011184, right= 0x8000000200006022, wrong = 0x8000000200006020
   mepc different at pc = 0x0000011184, right= 0x0000000000011186, wrong = 0x000000000001116e
   sepc different at pc = 0x0000011184, right= 0x0000000000010a1a, wrong = 0x0000000000011186
  mtval different at pc = 0x0000011184, right= 0x0000003f9bdf749e, wrong = 0x0000000000000000
[...]

PR #3407, opened 2023-07-06
PTW: set ae_ptw for out-of-range non-leaf PTEs

poemonsense:
For PTEs whose physical address is out-of-range, we need to set `ae_ptw` instead of `ae_final` to raise access-fault.

Because non-leaf PTEs will not have R or X bits set, `ae_final` will be overrided by page-fault exceptions.

**Related issue**: 

**Impact**: API modification

**Release Notes**

See https://github.com/chipsalliance/rocket-chip/issues/3402#issuecomment-1614229185 for more information.

### DIFF
--- a/src/main/scala/rocket/PTW.scala
+++ b/src/main/scala/rocket/PTW.scala
@@ -720,6 +720,7 @@ class PTW(n: Int)(implicit edge: TLEdgeOut, p: Parameters) extends CoreModule()(
           resp_valid(r_req_dest) := true.B
         }
 
+        resp_ae_ptw := ae && count < (pgLevels-1).U && pte.table()
         resp_ae_final := ae
         resp_pf := pf && !stage2
         resp_gf := gf || (pf && stage2)


poemonsense:
@aswaterman Would you mind take a look at this? I think the current behavior of rocket-chip should violate the ISA. Thanks

aswaterman:
I'm going punt this to @ingallsj.  I believe it also makes sense in the context of two-stage translation, but I haven't thought it through.

ingallsj:
Looking, and attempting to reproduce the bug and fix internally, too.  Standby.

ingallsj (approved):
We have reproduced this failure and this fix passes.

Still looking at https://github.com/chipsalliance/rocket-chip/pull/3408

ingallsj:
https://github.com/chipsalliance/rocket-chip/pull/3408 is now closed, so this PR is ready to merge.

ZenithalHourlyRate:
@mergifyio backport master

PR #3408, opened 2023-07-06
PTW: check all reserved bits for non-leaf ptes

poemonsense:
Non-leaf PTEs should not have any of D, A, and U bits set, otherwise a page-fault exception is raised.

Previously rocket-chip is missing the condition for D, A, and U bits.

**Related issue**: 

**Impact**: API modification

**Release Notes**

See https://github.com/chipsalliance/rocket-chip/issues/3402#issuecomment-1614229185 for more information on the bug.

### DIFF
--- a/src/main/scala/rocket/PTW.scala
+++ b/src/main/scala/rocket/PTW.scala
@@ -135,8 +135,12 @@ class PTE(implicit p: Parameters) extends CoreBundle()(p) {
   val r = Bool()
   /** valid bit */
   val v = Bool()
+  /** the most significant bits are reserved for all types of PTE */
+  def isReserved: Bool = reserved_for_future =/= 0.U
+  /** a non-leaf PTE has more reserved bits than the leaf PTEs */
+  def isReservedIfNonLeaf: Bool = d || a || u
   /** return true if find a pointer to next level page table */
-  def table(dummy: Int = 0) = v && !r && !w && !x && !d && !a && !u && reserved_for_future === 0.U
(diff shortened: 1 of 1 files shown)


poemonsense:
@aswaterman Would you mind take a quick look at this as well? Rocket-chip is missing the conditions for non-leaf page-fault. It checks the conditions in table but fails to report it as a page-fault. Thanks.

aswaterman (commented):
cc @ingallsj again, but this looks right to me.

I'd suggest the name `isReservedIfNonLeaf`.

ingallsj:
Looking, and attempting to reproduce the bug and fix internally, too.  Standby.

poemonsense:
Changed the name to `isReservedIfNonLeaf` as suggested.

ingallsj (changes requested):
Because the definition of a non-leaf "table" PTE is that the R,W,X permission bits are zeroes "off", the L1TLB (ITLB/DTLB) then goes on to report a page fault, even though the PTW response did not explicitly include the `pf` bit set.

So we think that this change would add redundant logic to what the L1TLB (ITLB/DTLB) already performs, and we don't think there is a bug that this change fixes.

poemonsense:
After double-checking the case, I found you are right and this bug depends on a wrong patch for #3407. Before I submit the final and correct PR for that, I had a bad patch trying to fix #3407. This bug is only feasible in that version. 

Sorry for the confusion and thanks for your work. I'll close this PR.

PR #3411, opened 2023-07-11
Fix: wrong value written in reg_vscause for local interrupts

vithursons:
**Related issue**: 

**Impact**:  API modification

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/CSR.scala
+++ b/src/main/scala/rocket/CSR.scala
@@ -1041,7 +1041,7 @@ class CSRFile(
       reg_mstatus.v := true.B
       reg_vsstatus.spp := reg_mstatus.prv
       reg_vsepc := epc
-      reg_vscause := Mux(cause(xLen-1), interruptMSB.U | (Mux(cause_lsbs > 12, cause_lsbs, cause_lsbs - 1)), cause) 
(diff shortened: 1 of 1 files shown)


src/main/scala/rocket/CSR.scala
```diff
@@ -1041,7 +1041,7 @@ class CSRFile(
       reg_mstatus.v := true.B
       reg_vsstatus.spp := reg_mstatus.prv
       reg_vsepc := epc
-      reg_vscause := Mux(cause(xLen-1), Cat(cause(xLen-1, 2), 1.U(2.W)), cause)
+      reg_vscause := Mux(cause(xLen-1), interruptMSB.U | (Mux(cause_lsbs > 12, cause_lsbs, cause_lsbs - 1)), cause)
```
ZenithalHourlyRate, line 1044:
```suggestion
      reg_vscause := Mux(cause(xLen-1) && !cause_lsbs(cause_lsbs.getWidth-1, 4).orR, Cat(cause(xLen-1, 2), 1.U(2.W)), cause)
```

If I understand the code correctly, only `vssip`, `vstip`, `vstip` and local interrupts are delegable to VS. RoCC are not delegable. I think a separation line of 16 should be enough?

vithursons: @ZenithalHourlyRate Commiting the suggestion, but this should be revised when advanced interrupt architecture is implemented.

ZenithalHourlyRate:
CLA issues...The first commit might be made again using your private or working email that is added to your github account?

PR #3478, opened 2023-08-22
Sync dev with master

jerryz123:
Merge master into dev, now both branches are the same


sequencer:
I force merge it, since mergify doesn't need to sign the CLA

PR #3490, opened 2023-09-19
Support Tile Clusters (tree-like hierarchies) 

jerryz123:
#3238 rebased on latest.

This work *substantially* refactors the existing Tile/TilePRCIDomain/HasTiles/InstantiatesTiles code into a system that support recursive clusters of tiles.

This PR defines an `Element` as an abstract node in the hierarchy tree. `Element`s can be `BaseTile`s or `Cluster`s, where `BaseTile`s are leaf nodes in the hierarchy, clusters are non-root nodes, and the `BaseSubsystem` is the root node.
`BaseTile` and `Cluster` will support deduplication.

Notably `Cluster`s support recursive instantiation, a `Cluster` can contain `Tile`s or internal `Cluster`s.  The `Cluster` bus topology should be capable of both supporting hierarchical coherence, and the `BaseSubsystem`s topology handle managing all coherence, depending on configuration variables.

This PR still requires more time. Some of the interrupt handling is not satisfying, and some decisions over clock/reset crossing ought to be addressed.

**Related issue**: 

**Release Notes**


src/main/scala/prci/ClockGroupDriver.scala
```diff
@@ -1,46 +0,0 @@
-// See LICENSE.SiFive for license details.
```
sequencer, line 1: I don't think `ClockGroupDriver` make sense as well;p
Thanks for removing it.

src/main/scala/subsystem/Cluster.scala
```diff
@@ -0,0 +1,230 @@
...
+{
+  val busContextName = thisClusterParams.baseName
+  lazy val clusterId = thisClusterParams.clusterId
+  lazy val location = InCluster(clusterId)
+
+  lazy val allClockGroupsNode = ClockGroupIdentityNode()
+
+  val csbus = tlBusWrapperLocationMap(SBUS) // like the sbus in the base subsystem
+  val ccbus = tlBusWrapperLocationMap(CBUS) // like the cbus in the base subsystem
+  val cmbus = tlBusWrapperLocationMap.lift(MBUS).getOrElse(csbus)
```
sequencer, line 48: For each cluster, they should have these buses?
or we can let user directly query bus via location API.

src/main/scala/subsystem/Cluster.scala
```diff
@@ -0,0 +1,230 @@
...
+    }
+  }
+}
+
+case class ClusterAttachParams(
+  clusterParams: ClusterParams,
+  crossingParams: ElementCrossingParamsLike
+) extends CanAttachCluster
+
+case class CloneClusterAttachParams(
```
sequencer, line 214: Basically I don't like the current Clone implementation in Diplomacy, which is a little brittle, I think in the future, we can try to remove them and use `https://github.com/llvm/circt/blob/8129d5de502a8cd5d861438e99f823babc5cebb0/docs/Dialects/FIRRTL/FIRRTLAnnotations.md?plain=1#L613` to force circt do it.

src/main/scala/subsystem/Element.scala
```diff
@@ -0,0 +1,83 @@
...
+
+/** An interface for describing the parameterization of how a particular element port is connected to an interconnect */
+trait ElementPortParamsLike {
+  /** The subnetwork location of the interconnect to which this element port should be connected. */
+  def where: TLBusWrapperLocation
+  /** Allows port-specific adapters to be injected into the interconnect side of the attachment point. */
+  def injectNode(context: Attachable)(implicit p: Parameters): TLNode
+}
+
+abstract class BaseElement (val crossing: ClockCrossingType)(implicit p: Parameters)
```
sequencer, line 49: I prefer more doc describing the design of `Element`

src/main/scala/system/Configs.scala
```diff
@@ -37,6 +37,17 @@ class DualCoreConfig extends Config(new WithNBigCores(2) ++ new WithCoherentBusT
...
+class ClusterConfig extends Config(
+  new WithNBigCores(2, location=InCluster(3)) ++
+  new WithNBigCores(2, location=InCluster(1)) ++
+  new WithNBigCores(2, location=InCluster(0)) ++
+  new WithCluster(3, location=InCluster(2)) ++
+  new WithCluster(2) ++
+  new WithCluster(1) ++
+  new WithCluster(0) ++
+  new DefaultConfig
+)
```
sequencer, line 49: Good job!

src/main/scala/subsystem/Cluster.scala
jerryz123: I agree the implementation of CloneLazyModule is not good right now, but I like the API of CloneLazyModule. I'm not sure what you are envisioning, but I thought the backend implementation of CloneLazyModule could change, while these APIs remain

src/main/scala/subsystem/Cluster.scala
jerryz123: I will try to remove what I can here.

src/main/scala/subsystem/Element.scala
jerryz123: Yep, will add. I call this "HierarchicalElement" now to be clearer. I will also create a new sub-package to avoid stuffing everything under `subsystem`.

jerryz123:
@sequencer this is ready.

PR #3498, opened 2023-09-26
The new decoder

sequencer:
Original `IDecode.scala` is really hard to maintain which use the view from each instruction rather than each control field. 
This PR remove original decoder with new [DecodeTable] in Chisel.
It will be a non-small refactor, but it can help us adding new instructions and have a better understand to the uarch.
After this PR is landed, all original IDecode will be moved into an "legacy" folder, downstream user can still use them if they wish, but after Boom and Xiangshan migrated, they will be removed. cc @jerryrzhao, @poemonsense 
sequncer/rvdecoderdb will be upstreamed to chipsalliance as a part of RC after all test clean.

BTW, I found the Zb/Zk UOP was really a bad design...

**Release Notes**
Switch to DecodeTable API.


poemonsense:
We are still on the way to upgrading to Chisel 5.

Thank you for cc me. We will bump rocket first to the version before new decoder table and then bump to the latest with changes on the decoder table.

Btw, we still have some local commits in Openxiangshan/rocket-chip. I'll see whether they should/could be upstreamed to chipsalliance. I agree it would be better to follow upstream rocket.

sequencer:
Spot my bug in chipsalliance/chisel#2924
https://github.com/chipsalliance/chisel/blob/440f01addeadd265fca2518c0a4df00b698e4603/src/main/scala/chisel3/util/experimental/decode/DecoderBundle.scala#L61
`asTypeOf` should never be LHS, otherwise it will create a wire and broken in when connection....
I'll vendor a chisel source for debugging, but not going to be merged.

sequencer:
Assign the Zb/Zk to @cyyself, I'll skip them in this PR by mask them off.

After Chisel merging chipsalliance/chisel#3563 I'll bump Chisel version to 3.6.1 and 5.0.x and get this PR merged.

PR #3504, opened 2023-10-06
Fix the missing DontCare in TLEdgeOut

poemonsense:
**Related issue**:  PR #3287 (not fully fixed before)

**Impact**: API modification

**Release Notes**

### DIFF
--- a/src/main/scala/tilelink/Delayer.scala
+++ b/src/main/scala/tilelink/Delayer.scala
@@ -30,6 +30,8 @@ class TLDelayer(q: Double)(implicit p: Parameters) extends LazyModule
       anoise.size    := LFSRNoiseMaker(anoise.params.sizeBits)
       anoise.source  := LFSRNoiseMaker(anoise.params.sourceBits)
       anoise.address := LFSRNoiseMaker(anoise.params.addressBits)
+      anoise.user    := DontCare
+      anoise.echo    := DontCare
       anoise.mask    := LFSRNoiseMaker(anoise.params.dataBits/8)
       anoise.data    := LFSRNoiseMaker(anoise.params.dataBits)
       anoise.corrupt := LFSRNoiseMaker(1)
@@ -50,6 +52,8 @@ class TLDelayer(q: Double)(implicit p: Parameters) extends LazyModule
       cnoise.size    := LFSRNoiseMaker(cnoise.params.sizeBits)
       cnoise.source  := LFSRNoiseMaker(cnoise.params.sourceBits)
       cnoise.address := LFSRNoiseMaker(cnoise.params.addressBits)
+      cnoise.user    := DontCare
+      cnoise.echo    := DontCare
       cnoise.data    := LFSRNoiseMaker(cnoise.params.dataBits)
       cnoise.corrupt := LFSRNoiseMaker(1)(0)
 
@@ -60,6 +64,8 @@ class TLDelayer(q: Double)(implicit p: Parameters) extends LazyModule
       dnoise.source  := LFSRNoiseMaker(dnoise.params.sourceBits)
       dnoise.sink    := LFSRNoiseMaker(dnoise.params.sinkBits)
       dnoise.denied  := LFSRNoiseMaker(1)(0)
+      dnoise.user    := DontCare
+      dnoise.echo    := DontCare
       dnoise.data    := LFSRNoiseMaker(dnoise.params.dataBits)
       dnoise.corrupt := LFSRNoiseMaker(1)(0)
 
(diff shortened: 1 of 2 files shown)


jerryz123:
Which fields were being dropped here?

poemonsense:
For channel A, it may have user bits.

For channel E, I'm not sure. I can remove the DontCare for channel E if not needed.

I think the underlying issue is that, previously LazyModule has DontCare of its IOs by default. Later this implication changes and breaks some existing codebase. Probably it's due to backward compatibility considerations that @sequencer adds the DontCare afterwards. I'm not sure about this.

So actually I think these dontcares are not must-have. If rocket decides to leave it as is without the DontCare, I'm also pleased to fix the missing connections in our design. This PR is just a choice between backward compatibility and behavior deprecations/removal.

jerryz123:
My preference is for us to be precise about the application of these DontCares. So for A channel, we should explicitly DontCare the user field. And not DontCare the E channel unless necessary.

sequencer:
I prefer provide an API to waive user-bits or some fields

jerryz123:
How is the waive connect applicable here?
Something has to drive the user field in this case. We can't leave it disconnected.

poemonsense:
Sure, I fully understand your considerations and agree it would be safer to dontcare specific issues. Do you have any recommendations on how we should achieve the goal?

sequencer:
@poemonsense can you point me the use case in your codebases? I will find some time to figure out how to deal with this issue.

poemonsense:
https://github.com/OpenXiangShan/XiangShan/blob/master/src/main/scala/xiangshan/cache/dcache/mainpipe/MissQueue.scala#L633

If someone tries to create a Wire with the Edges. It will cause errors if rocket does not DontCare it.

Like this:
```scala
  val acquireBlock = edge.AcquireBlock(  // This does have DontCare in rocket
    fromSource = io.id,
    toAddress = req.addr,
    lgSize = (log2Up(cfg.blockBytes)).U,
    growPermissions = grow_param
  )._2

  val acquirePerm = edge.AcquirePerm(  // This does NOT have DontCare in rocket and causes errors
    fromSource = io.id,
    toAddress = req.addr,
    lgSize = (log2Up(cfg.blockBytes)).U,
    growPermissions = grow_param
  )._2
```

poemonsense:
I think we can explicitly DontCare the following fields (only):

- `user` and `echo` in Channel A
- `user` and `echo` in Channel C

poemonsense:
By looking into the codebase of Edges.scala and Bundles.scala, I think only these bits are necessary for DontCare.

poemonsense:
Currently it looks like rocket does not have any DontCare in TLEdgeIn. If it also requires DontCare, we should also DontCare the `user` and `echo` bits in Channel D.

sequencer:
OK it seems that we forgot to put the `DontCare` here...

I personally agree add `DontCare`, and I'll think about what's a better API for TileLink Edge...

poemonsense:
Should we also DontCare channel_c.data for TLMessages.Release? Currently it is assigned with 0.U, but I think it should be DontCare as well.

poemonsense:
I've force pushed a commit. It DontCare only the user and echo bits in channel A/C. It also changes the `data := 0.U` to `data := DontCare`.

sequencer:
I'll go back to spec to carefully review if those dc is allowed. Maybe this afternoon.

jerryz123:
Does the spec require 0.U? Or is DontCare ok

poemonsense:
> I'll go back to spec to carefully review if those dc is allowed. Maybe this afternoon.

Relax. No worry. It's my fault to mix the modifications for DontCare user/echo and data.

poemonsense:
I think EdgeIn should also be DontCare'ed. @sequencer 

See https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/tilelink/CacheCork.scala#L86

So in our design, channel D may have user/echo bits that break this code.

poemonsense:
Similar issues apply to https://github.com/chipsalliance/rocket-chip/blob/master/src/main/scala/tilelink/Delayer.scala#L27-L67

These new Wires may have user/echo bits but are never assigned. How should we fix this? Also add DontCare to it?

sequencer:
> Relax. No worry. It's my fault to mix the modifications for DontCare user/echo and data.

I checked the spec, this PR LGTM.

sequencer (approved):

poemonsense:
I need to push one more commit to fix more DontCares. Please wait tomorrow

sequencer:
> I need to push one more commit to fix more DontCares. Please wait tomorrow

Sure

sequencer:
> Does the spec require 0.U? Or is DontCare ok

For all `data` field in this PR, spec says: 
```
x_data D 8w Ignored; can be any value.
```

sequencer:
> I think EdgeIn should also be DontCare'ed. @sequencer

Do as you wish, I think this is reasonable.

poemonsense:
Sorry for the delay. Please help review it again. I believe all `data := 0.U` can be replaced by DontCare in their messages. Thank you @sequencer

poemonsense:
Full lists of DontCare data in this PR:

- AcquireBlock
- AcquirePerm
- Release
- ProbeAck
- Get
- Hint
- AccessAck
- HintAck
- Probe
- Grant
- ReleaseAck

I look into the TileLink spec and they should all be marked `Ignored; can be any value`

sequencer (approved):

PR #3507, opened 2023-10-08
Support RocketChip build Chisel from source

sequencer:
This PR support RocketChip build chisel from source, this flow is not for users, only for devs to test new version of Chisel and check some backport feature is working when they are not released to maven.

**Related issue**: 

**Release Notes**

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -4,3 +4,6 @@
 [submodule "dependencies/cde"]
 	path = dependencies/cde
 	url = https://github.com/chipsalliance/cde.git
+[submodule "dependencies/chisel"]
+	path = dependencies/chisel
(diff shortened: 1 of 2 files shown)


jerryz123:
I am generally cautious about relying on unreleased versions of dependencies..... 

IMO default build should always build from released versions, which developer-only flag to build from source.

sequencer:
> I am generally cautious about relying on unreleased versions of dependencies.....

RocketChip won't depend on the unreleased version of Chisel, this is only for debug use.

Default build will only build from the release version of Chisel

sequencer:
@jerryz123 I think this is safe to get in, just provide an additional `source` cross target, no influence to our release flow.

jerryz123 (approved):

PR #3512, opened 2023-10-18
Add TLFixedDelayer

cyyself:

**Release Notes**

Add TLFixedDelayer for memory performance exploration with different memory latency.

### DIFF
--- a/src/main/scala/tilelink/Delayer.scala
+++ b/src/main/scala/tilelink/Delayer.scala
@@ -81,11 +81,54 @@ class TLDelayer(q: Double)(implicit p: Parameters) extends LazyModule
   }
 }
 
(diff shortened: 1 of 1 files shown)


jerryhethatday:
Is this only for simulation？

src/main/scala/tilelink/Delayer.scala
```diff
@@ -81,6 +81,40 @@ class TLDelayer(q: Double)(implicit p: Parameters) extends LazyModule
   }
 }
 
+class TLFixedDelayer(cycle: Int, maxLatency: Int = 4096)(implicit p: Parameters) extends LazyModule
+{
+  val node = TLAdapterNode()
```
jerryz123, line 86: This should be an `IdentityNode`

jerryz123 (changes requested):

cyyself:
As a more general #3520 for all DecoupledIO being merged, this PR will not be needed anymore.

PR #3520, opened 2023-11-03
Add DelayQueue

abejgonzalez:
**Related issue**: https://github.com/chipsalliance/rocket-chip/pull/3512

**Release Notes**
Adds two variations of queues that delay elements by a certain cycle count (either static at compile time or configurable during runtime). Re-uses some of the infrastructure used #3512. Replaces some of the work done in https://github.com/ucb-bar/chipyard/pull/1645

### DIFF
--- a/src/main/scala/util/DelayQueue.scala
+++ b/src/main/scala/util/DelayQueue.scala
@@ -5,6 +5,16 @@ package freechips.rocketchip.util
 import chisel3._
 import chisel3.util._
 
+/** A queue that delays elements by a certain cycle count.
+  *
+  * @param gen              queue element type
+  * @param entries          queue size
+  *
+  * @param enq              enqueue
+  * @param deq              dequeue
+  * @param timer            cycle count timer
+  * @param entries          cycle delay
+  */
 class DelayQueue[T <: Data](gen: T, entries: Int) extends Module {
   val io = IO(new Bundle {
     val enq = Flipped(DecoupledIO(gen))
@@ -35,6 +45,13 @@ class DelayQueue[T <: Data](gen: T, entries: Int) extends Module {
 }
 
 object DelayQueue {
+  /** Helper to connect a delay queue.
+   *
+   * @param source           decoupled queue input
+   * @param timer            cycle count timer
+   * @param delay            cycle delay
+   * @param depth            queue size
+   */
   def apply[T <: Data](source: DecoupledIO[T], timer: UInt, delay: UInt, depth: Int): DecoupledIO[T] = {
     val delayQueue = Module(new DelayQueue(chiselTypeOf(source.bits), depth))
     delayQueue.io.enq <> source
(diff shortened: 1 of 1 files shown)


src/main/scala/util/DelayQueue.scala
```diff
@@ -0,0 +1,40 @@
...
+      source.ready := q.io.enq.ready
+
+      sink.bits := q.io.deq.bits.data
+      sink.valid := q.io.deq.valid && ( (timer - q.io.deq.bits.time) >= delay)
+      q.io.deq.ready := sink.fire
+  }
+}
+
+object FixedDelayQueue {
+  def apply[T <: Data](sink: DecoupledIO[T], source: DecoupledIO[T], delay: Int, maxLatency: Int = 4096): Unit = {
```
jerryz123, line 27: This can become
```
  apply(sink, source, delay.U, delay.U, maxLatency)
```

jerryz123 (changes requested):
Can we change the `apply` function signatures to 
`(in: DecoupledIO[T], ...) => DecoupledIO[T]`. Basically match the `Queue.apply` in chisel?

src/main/scala/util/DelayQueue.scala
abejgonzalez: Sorry I don't follow? What do you mean?

jerryz123: Have a single object `DelayQueue`, with method overloading of `apply`. Note the different type signatures, depending on if you want a fixed delay (pass a Int) or a variable delay (pass a UInt and a max).

For `depth`, you would normally want to set this to be equal to the `maxDelay` right?

abejgonzalez: Ah. I see. I can do that.

jerryz123:
@mergifyio backport master

src/main/scala/util/DelayQueue.scala
abejgonzalez: And yea I missed the `maxDelay` comment. Updated as well.

src/main/scala/util/DelayQueue.scala
```diff
@@ -0,0 +1,45 @@
...
+    sink.bits := q.io.deq.bits.data
+    sink.valid := q.io.deq.valid && ((timer - q.io.deq.bits.time) >= delay_r)
+    q.io.deq.ready := sink.fire
+
+    sink
+  }
+
+  def apply[T <: Data](source: DecoupledIO[T], delay: Int): DecoupledIO[T] = {
+    val timer = RegInit(0.U((1 + log2Ceil(delay.max(1))).W))
+    timer := timer + 1.U
```
jerryz123, line 36: Can we also dedup this timer thing?

src/main/scala/util/DelayQueue.scala
```diff
@@ -0,0 +1,49 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import chisel3._
+import chisel3.util._
+
+object DelayQueue {
+  private def apply[T <: Data](source: DecoupledIO[T], timer: UInt, delay: UInt, depth: Int): DecoupledIO[T] = {
+    val q = Module(new Queue(new Bundle {
```
jerryz123, line 10: Can you wrap everything in this block in another module? This way we can avoid instances of `q` in the module hierarchy.

abejgonzalez: Ok I did so. Note: I kept the timer outside of the module so that power-users could re-use the timer across different `DelayQueues` (like what #3512 does).

src/main/scala/util/DelayQueue.scala
```diff
@@ -0,0 +1,61 @@
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import chisel3._
+import chisel3.util._
+
+class DelayQueue[T <: Data](gen: T, entries: Int) extends Module {
+  val io = new Bundle {
```
jerryz123, line 9: This needs to be `IO(new Bundle {`

src/main/scala/util/DelayQueue.scala
```diff
@@ -0,0 +1,61 @@
...
+
+  val delay_r = RegInit(0.U(io.delay.getWidth.W))
+  when (delay_r =/= io.delay) {
+    delay_r := io.delay
+    assert(q.io.count == 0, "Undefined behavior when delay is changed while queue has elements.")
+  }
+
+  q.io.enq.bits.data := enq.bits
+  q.io.enq.bits.time := io.timer
+  q.io.enq.valid := enq.fire
```
jerryz123, line 29: `enq` -> `io.enq`?

jerryz123 (changes requested):

jerryz123 (approved):

PR #3532, opened 2023-11-16
Remove Scalar Crypto and BitManip

cyyself:
**Related issue**: #3529

**Type of change**: feature request

**Impact**: API modification

**Development Phase**: implementation

**Release Notes**
Remove Scalar Crypto and BitManip as discussed in #3529

PR #3534, opened 2023-11-18
The new decoder

cyyself:
This PR takes over #3498 and removed Zb* and Zk* since #3532 has been merged.

**Related issue**: #3498

**Release Notes**
Switch to DecodeTable API.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -7,3 +7,9 @@
 [submodule "dependencies/chisel"]
 	path = dependencies/chisel
 	url = https://github.com/chipsalliance/chisel.git
+[submodule "dependencies/rvdecoderdb"]
+	path = dependencies/rvdecoderdb
+	url = https://github.com/sequencer/rvdecoderdb.git
+[submodule "dependencies/riscv-opcodes"]
+	path = dependencies/riscv-opcodes
+	url = https://github.com/riscv/riscv-opcodes.git
--- a/src/main/scala/rocket/Decoder.scala
+++ b/src/main/scala/rocket/Decoder.scala
@@ -1,0 +1,709 @@
+// SPDX-License-Identifier: Apache-2.0
+// SPDX-FileCopyrightText: 2023 Jiuyang Liu <liu@jiuyang.me>
+
+package org.chipsalliance.rocketcore.decoder
+
+import chisel3._
+import chisel3.util.BitPat
+import chisel3.util.experimental.decode._
+import org.chipsalliance.rvdecoderdb.{Encoding, Instruction, InstructionSet, Utils}
+
+// behave like ChiselEnum, but for compatibility, use UInt for now.
+// This is going to be upstreamed to Chisel in the future.
+trait UOP {
+  def width: Int
+
+  def dontCare: BitPat = BitPat.dontCare(width)
+
+  def chiselType: TPE = UInt(width.W)
+
+  def encode(lit: Int): BitPat = BitPat(lit.U(width.W))
+
+  def encode(strLit: String): BitPat = BitPat(strLit.U(width.W))
+
+  type TPE = UInt
+}
+
+trait UOPDecodeField[T <: DecodePattern] extends DecodeField[T, UInt] {
+  def uopType: UOP
+
+  def chiselType: UInt = uopType.chiselType
+}
+
+/** DecodePattern for an RISC-V instruction */
+case class RocketDecodePattern(instruction: Instruction) extends DecodePattern {
+  def isRoCC: Boolean = instruction.instructionSet.name == "rv_rocc"
+
+  override def bitPat: BitPat = BitPat("b"+instruction.encoding.toString)
+}
+
+object CustomInstructions {
+  private def rocket(name: String, encoding: Encoding) =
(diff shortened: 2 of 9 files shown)


.gitmodules
```diff
@@ -7,3 +7,9 @@
 [submodule "dependencies/chisel"]
 	path = dependencies/chisel
 	url = https://github.com/chipsalliance/chisel.git
+[submodule "dependencies/rvdecoderdb"]
+	path = dependencies/rvdecoderdb
+	url = https://github.com/sequencer/rvdecoderdb.git
```
sequencer, line 12: After @jerryz123 approves, I will transfer this repo to chipsalliance.

src/main/scala/rocket/Decoder.scala
```diff
@@ -0,0 +1,709 @@
...
+      case i if Seq("jalr").contains(i) => y
+      case _ => n
+      // format: on
+    }
+  }
+
+  object rxs2 extends BoolDecodeField[RocketDecodePattern] {
+    override def name: String = "rxs2"
+
+    override def genTable(op: RocketDecodePattern): BitPat = op.instruction.name match {
```
sequencer, line 201: can you sort it by instruction set? for maintaining purpose

src/main/scala/rocket/RocketCore.scala
```diff
@@ -176,30 +177,89 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
...
-    Seq(new IDecode(aluFn))
-  } flatMap(_.table)
-
-  val ex_ctrl = Reg(new IntCtrlSigs(aluFn))
-  val mem_ctrl = Reg(new IntCtrlSigs(aluFn))
-  val wb_ctrl = Reg(new IntCtrlSigs(aluFn))
+
+  val decoderParameter = org.chipsalliance.rocketcore.decoder.InstructionDecoderParameter(
+    // TODO: configurable
+    (org.chipsalliance.rvdecoderdb.fromFile.instructions(os.pwd / "dependencies" / "riscv-opcodes") ++
```
sequencer, line 183: can you fixup this TODO? with out configuring it, downstream users cannot use it.

sequencer (commented):
some TODO need to be resolved, other LGTM(since I wrote it).
Need review by @jerryz123

src/main/scala/rocket/Decoder.scala
```diff
@@ -199,7 +199,7 @@ class InstructionDecoder(p: InstructionDecoderParameter) {
...
+      case i if Seq("amomaxu.w", "amoand.w", "amoor.w", "amoxor.w", "amoswap.w", "lr.w", "amomax.w", "amoadd.w", "amomin.w", "amominu.w", "sc.w", "lr.d", "amomax.d", "amoswap.d", "amoxor.d", "amoand.d", "amomin.d", "amoor.d", "amoadd.d", "amomaxu.d", "amominu.d", "sc.d", "hsv.w", "hsv.b", "hfence.vvma", "hsv.h", "hfence.gvma", "hsv.d", "or", "srl", "sltu", "sra", "sb", "add", "xor", "beq", "bge", "sw", "blt", "bgeu", "bltu", "bne", "sub", "and", "slt", "sh", "sll", "addw", "sd", "sllw", "sraw", "subw", "srlw", "mulhsu", "rem", "div", "mul", "mulhu", "mulh", "remu", "divu", "remuw", "divw", "divuw", "mulw", "remw", "sfence.vma", "czero.nez", "czero.eqz", "custom0.rs1.rs2", "custom0.rd.rs1.rs2", "custom1.rs1.rs2", "custom1.rd.rs1.rs2", "custom2.rs1.rs2", "custom2.rd.rs1.rs2", "custom3.rs1.rs2", "custom3.rd.rs1.rs2").contains(i) => y
```
sequencer, line 202: How about split them with '\n'

jerryz123 (changes requested):
I do not feel confident in this change, as currently designed. The API change is very drastic.

We can keep the existing `IntCtrlSigs` bundle, and have a generator param control whether to use the new decoder (this can even be the default) or the old way.

sequencer:
> We can keep the existing IntCtrlSigs bundle

I don’t think the reason keeping the original IntCtrl is a good design since for different instruction sets support, some control signal will be different.

> And have a generator param control whether to use the new decoder (this can even be the default) or the old way.

i don’t think the maintain burden is acceptable for me :(

jerryz123:
> I don’t think the reason keeping the original IntCtrl is a good design since for different instruction sets, some control signal will be optional. The current design doesn’t support it.

I don't like the premise that additional extensions require adding signals to IntCtrlSigs. IntCtrlSigs are the control signals for the subset of the control signals for the base integer ISA... the fields in this bundle should never need to change. We should develop interfaces and APIs that allow for extensions to define their own control signals, and set the IntCtrlSignals where necessary.

> i don’t think the maintain burden is acceptable for me :(

Aggressively deprecating APIs is a large maintenance burden. I don't see how retaining support for the original decoder is a large burden, if we do it with the correct interfaces. Once done, the original decoder code should never need to change again.

I will look into integrating this without deprecating the existing decoder.

Issue #3548, opened 2023-12-24
Error in AXI4 Xbar with unused fields


tymcauley:
**Type of issue**: bug report

**Impact**: no functional change

**Development Phase**: proposal

**Other information**
 

When constructing an AXI4 Xbar where some master/slave ports have extra fields (such as `AMBACorrupt`), but others don't, the new `:<>=` connectable operators throw the following errors:

```
[error] generators/rocket-chip/src/main/scala/amba/axi4/Xbar.scala:88:24: unconnected consumer field AXI4Xbar.in[1].w.bits.user.corrupt: Wire[Bool]
[error]       in(i).squeezeAll :<>= io_in(i).squeezeAll
[error]                        ^
[error] generators/rocket-chip/src/main/scala/amba/axi4/Xbar.scala:171:28: dangling producer field AXI4Xbar.out[1].w.bits.user.corrupt: Wire[Bool]
[error]       io_out(i).squeezeAll :<>= out(i).squeezeAll
[error]                            ^
[error] There were 2 error(s) during hardware elaboration.
```

It looks like this was [resolved in the TileLink Xbar](https://github.com/chipsalliance/rocket-chip/blob/74d65ca320cf5f296b39dc5218f6f099e08e2794/src/main/scala/tilelink/Xbar.scala#L160) by adding `.waiveAll` to these operators, and I can confirm that this resolves the hardware elaboration error for me.

**If the current behavior is a bug, please provide the steps to reproduce the problem:**

This change will reproduce the error:

```diff
--- a/src/main/scala/amba/axi4/Xbar.scala
+++ b/src/main/scala/amba/axi4/Xbar.scala
@@ -319,7 +319,7 @@ class AXI4XbarFuzzTest(name: String, txns: Int, nMasters: Int, nSlaves: Int)(imp
   masters.zipWithIndex.foreach { case (m, i) => (xbar
     := AXI4Delayer(0.25)
     := AXI4Deinterleaver(4096)
-    := TLToAXI4()
+    := TLToAXI4(wcorrupt = (if (i == 0) false else true))
     := TLFilter(filter(i))
     := TLRAMModel(s"${name} Master $i")
     := m.node) }
```

Then try to emit FIRRTL for the `AMBAUnitTestConfig`.

**Please tell us about your environment:**

- Version: current HEAD on `master`: 74d65ca
- OS: macOS 14.2.1

**What is the use case for changing the behavior?**

[...]

PR #3549, opened 2023-12-24
Fix AXI4 Xbar elaboration error due to missing fields

tymcauley:
**Related issue**: https://github.com/chipsalliance/rocket-chip/issues/3548

**Release Notes**

### DIFF
--- a/src/main/scala/amba/axi4/Xbar.scala
+++ b/src/main/scala/amba/axi4/Xbar.scala
@@ -85,7 +85,12 @@ class AXI4Xbar(
     // Transform input bundles
     val in = Wire(Vec(io_in.size, new AXI4Bundle(wide_bundle)))
     for (i <- 0 until in.size) {
-      in(i).squeezeAll :<>= io_in(i).squeezeAll
+      in(i).aw.bits.user := DontCare
+      in(i).aw.bits.echo := DontCare
+      in(i).ar.bits.user := DontCare
+      in(i).ar.bits.echo := DontCare
+      in(i).w.bits.user := DontCare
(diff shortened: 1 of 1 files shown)


jerryz123 (commented):
Can you rebase this PR on dev?

tymcauley:
Unfortunately I need to wait a bit to get some feedback from my company on the CLA, so resolving that is going to take some time.

tymcauley:
On a recent `firtool` version (1.58.0), I found that I'm getting this error:

```
generators/rocket-chip/src/main/scala/amba/axi4/Xbar.scala:86:18: error: sink "in_1_w_bits_user_corrupt" not fully initialized in "AXI4Xbar"
```

I worked around it by adding this line right above the `in(i).squeezeAll.waiveAll :<>= io_in(i).squeezeAll.waiveAll` statement:

```scala
in(i).w.bits.user := DontCare
```

Not completely sure what the more principled way to solve this is, but I do see that `tilelink/Xbar.scala` also tying off some the `user` signals. Presumably we want `user` signals to propagate through the crossbar when appropriate.

tymcauley:
Okay, I believe I've taken care of all the fields that need to be tied off in case they're not used by a particular destination.

jerryz123 (approved):

tymcauley:
Checking on the status of this, anything else that I need to do before this can be merged @jerryz123 ?

jerryz123 (approved):
Nope, this just slipped past me. Let's merge this

tymcauley:
> Nope, this just slipped past me. Let's merge this

Thanks so much!

PR #3557, opened 2024-01-16
Update ECC SRAM to improve throughput on full write

Kevin99214:

**Release Notes**
Enhance ECC SRAM write throughput when eccBytes > 1 by 2 cycles per write when utilizing the full bus.


jerryz123 (approved):
Thanks!

PR #3564, opened 2024-01-25
Add customDecoder interface to rocket

jerryz123:
This should enable downstream users to implement custom decoders without very heavy-handed changes to the core implementation (#3534)

**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/rocket/RocketCore.scala
+++ b/src/main/scala/rocket/RocketCore.scala
@@ -50,7 +50,8 @@ case class RocketCoreParams(
   fpu: Option[FPUParams] = Some(FPUParams()),
   debugROB: Boolean = false, // if enabled, uses a C++ debug ROB to generate trace-with-wdata
   haveCease: Boolean = true, // non-standard CEASE instruction
-  haveSimTimeout: Boolean = true // add plusarg for simulation timeout
+  haveSimTimeout: Boolean = true, // add plusarg for simulation timeout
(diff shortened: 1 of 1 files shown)


sequencer (approved):
I'm hesitate to expose this to other users.
At least in T1's flow, we will continue use the fork version of rvdeocderdb and the chisel DecoderTable API.
Since it reduces the control signal in datapath significantly for small designs.(~30 -> ~15) and is more type safe.
But Since T1 use the forked version of RocketCore, I don't against provide this feature in the upstream RocketChip;p

jerryz123:
> I'm hesitate to expose this to other users. At least in T1's flow, we will continue use the fork version of rvdeocderdb and the chisel DecoderTable API. Since it reduces the control signal in datapath significantly for small designs.(~30 -> ~15) and is more type safe. But Since T1 use the forked version of RocketCore, I don't against provide this feature in the upstream RocketChip;p

I see, its cool that the rvdecoderdb has that impact. What is the overall area reduction? 

The reason I prefer the simpler decode tables as the default is to
1) Reduce number of library dependencies.  Since rocket-chip is the "base" for most users, it should only depend on chisel
2) Present a simpler interface to users adding new instructions

While your fork is fine, I think it would be nice to have a shared rocket-vector interface in the upstream for both our projects, even if the vector backends target different design points. The `ifv` branch is almost stable.

PR #3568, opened 2024-02-08
Modularize PMA-checking functionality

jerryz123:
Logically equivalent modularization. Useful for systems which want to implement PMA checks elsewhere in the pipeline.

**Release Notes**


sequencer:
I personally think we can pass explicit parameters and resolve the edge and implicit parameters outside of the PMA module. WDYT?

jerryz123:
Good point. A Seq of TLManagerParameters is better.

PR #3575, opened 2024-02-25
Purge xbus in BaseSubsystem

sequencer:
There should be no xbus in the API, while users must use Location API to query and attach their own devices.

### DIFF
--- a/src/main/scala/devices/tilelink/CLINT.scala
+++ b/src/main/scala/devices/tilelink/CLINT.scala
@@ -105,7 +105,7 @@ trait CanHavePeripheryCLINT { this: BaseSubsystem =>
   val (clintOpt, clintDomainOpt, clintTickOpt) = p(CLINTKey).map { params =>
     val tlbus = locateTLBusWrapper(p(CLINTAttachKey).slaveWhere)
     val clintDomainWrapper = tlbus.generateSynchronousDomain.suggestName("clint_domain")
-    val clint = clintDomainWrapper { LazyModule(new CLINT(params, cbus.beatBytes)) }
+    val clint = clintDomainWrapper { LazyModule(new CLINT(params, tlbus.beatBytes)) }
     clintDomainWrapper { clint.node := tlbus.coupleTo("clint") { TLFragmenter(tlbus) := _ } }
     val clintTick = clintDomainWrapper { InModuleBody {
       val tick = IO(Input(Bool()))
--- a/src/main/scala/subsystem/BaseSubsystem.scala
+++ b/src/main/scala/subsystem/BaseSubsystem.scala
@@ -78,11 +78,12 @@ trait HasConfigurableTLNetworkTopology { this: HasTileLinkLocations =>
 
   // Calling these functions populates tlBusWrapperLocationMap and connects the locations to each other.
   val topology = p(TLNetworkTopologyLocated(location))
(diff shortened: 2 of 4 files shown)


src/main/scala/devices/tilelink/CLINT.scala
```diff
@@ -105,7 +105,7 @@ trait CanHavePeripheryCLINT { this: BaseSubsystem =>
   val (clintOpt, clintDomainOpt, clintTickOpt) = p(CLINTKey).map { params =>
     val tlbus = locateTLBusWrapper(p(CLINTAttachKey).slaveWhere)
     val clintDomainWrapper = tlbus.generateSynchronousDomain.suggestName("clint_domain")
-    val clint = clintDomainWrapper { LazyModule(new CLINT(params, cbus.beatBytes)) }
+    val clint = clintDomainWrapper { LazyModule(new CLINT(params, tlBusWrapperLocationMap.lift(CBUS).getOrElse(tlBusWrapperLocationMap(p(TLManagerViewpointLocated(location)))).beatBytes)) }
```
jerryz123, line 108: I prefer adding a `slaveWhere` parameter to CLINTParams, rather than providing this default

sequencer: Agree. AttachParameter is required for these devices. Do you prefer add it in this PR or my following PR to clean up these usages?

jerryz123: Can you do it to this PR?

sequencer: Sure! Seems only CLINT is required. (Where is PLICAttachParam?)

jerryz123 (changes requested):

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -90,18 +90,11 @@ abstract class BaseSubsystem(val location: HierarchicalLocation = InSubsystem)
 
   val busContextName = "subsystem"
```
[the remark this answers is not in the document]
jerryz123: `viewPointBus`, perhaps. I think the subtle difference is that `viewPointBus` does not HAVE to be the sbus, if the user has changed `TLManagerViewPointLocated`.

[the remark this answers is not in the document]
jerryz123: I don't have a strong opinion, but I haven't looked into the structure deeply recently.

src/main/scala/subsystem/BaseSubsystem.scala
```diff
@@ -78,11 +78,12 @@ trait HasConfigurableTLNetworkTopology { this: HasTileLinkLocations =>
 
   // Calling these functions populates tlBusWrapperLocationMap and connects the locations to each other.
   val topology = p(TLNetworkTopologyLocated(location))
-  topology.map(_.instantiate(this))
+  topology.foreach(_.instantiate(this))
   topology.foreach(_.connect(this))
+  def viewpointBus: TLBusWrapper = tlBusWrapperLocationMap(p(TLManagerViewpointLocated(location)))
```
sequencer, line 83: cc @jerryz123 added here.

jerryz123 (approved):

PR #3582, opened 2024-03-10
First try to purge out ClockGroup API

sequencer:


src/main/scala/tilelink/BusWrapper.scala
```diff
@@ -36,24 +36,13 @@ abstract class TLBusWrapper(params: HasTLBusParams, val busName: String)(implici
     with HasTLBusParams
     with CanHaveBuiltInDevices
 {
-  private val clockGroupAggregator = LazyModule(new ClockGroupAggregator(busName){ override def shouldBeInlined = true }).suggestName(busName + "_clock_groups")
-  private val clockGroup = LazyModule(new ClockGroup(busName){ override def shouldBeInlined = true })
-  val clockGroupNode = clockGroupAggregator.node // other bus clock groups attach here
-  val clockNode = clockGroup.node
+  val clockNode = ClockAdapterNode()
```
sequencer, line 39: Need to assert input to this clockNode to be 1? or just use `FixedClockBroadcast` instead?

PR #3591, opened 2024-03-16
PTW Hypervisor bug fixes: check GPA bits higher than HGATP.Mode

ingallsj:
**Related issue**: 

**Impact**: functional fix

**Release Notes**
1. The PTE Cache is supposed to cache the address being requested from memory.
2. The RISC-V Privileged ISA spec states that, for _every_ Stage-2 walk at _every level_ of Stage-1 walk, we need to check
> For Sv39x4, ... Address bits 63:41 must all be zeros, ...
> For Sv48x4, ... Address bits 63:50 must all be zeros, ...
> For Sv57x4, ... Address bits 63:59 must all be zeros, ...
> or else a guest-page-fault exception occurs.

### DIFF
--- a/src/main/scala/rocket/PTW.scala
+++ b/src/main/scala/rocket/PTW.scala
@@ -541,7 +541,7 @@ class PTW(n: Int)(implicit edge: TLEdgeOut, p: Parameters) extends CoreModule()(
   io.mem.req.bits.data := DontCare
   io.mem.req.bits.mask := DontCare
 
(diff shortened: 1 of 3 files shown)


aswaterman (approved):

sequencer:
we should add CI for Hypervisor in RocketChip. e.g. xvisor, kvm-riscv.
which testcases do you(@ingallsj @aswaterman) think, can be appropriate to add for testing? 
I'll schedule someone to add these workloads.

PR #3598, opened 2024-03-20
Fix s1_data when coreDataBits > xLen

jerryz123:
**Related issue**: 

**Release Notes**


sequencer:
decoding vector or custom load store into `mem_ctrl.fp` may trigger this bug?
I can understand, but for the field `fp` can be misleading. I personal think we need add some inline doc to it...

sequencer (approved):
Feel free to add, we can review and refactor it later.

jerryz123:
This does not affect vector instructions with scalar for scalar fp operands... this only affects FP stores when coreDataBits > xLen. So I think the condition is correct, and clear.

PR #3059, opened 2022-09-15
Port legacy chisel code to chisel3 for amba

tianrui-wei:
Signed-off-by: Tianrui Wei <tianrui@tianruiwei.com>
**Related issue**: 

**Type of change**: bug report | feature request | other enhancement

**Impact**: no functional change | API addition (no impact on existing code) | API modification

**Development Phase**: proposal |  implementation

**Release Notes**

PR #3634, opened 2024-05-22
Update APBtoTL scala to not flip apb address when doing conversion

Kevin99214:

**Release Notes**
APB address is getting bitwise flipped to get TL address which then immediately fails the assertion below that checks they are the same. 
`when (out.a.fire) {
        assert(in.paddr === out.a.bits.address, "Do not expect to have to perform alignment in APB2TL Conversion")
      }`

Reverts line back to before commit #3059


Kevin99214:
Sorry for closing and re-opening, misclicked

jerryz123 (approved):

PR #3652, opened 2024-06-28
Allow non-V implementations of vector units, with Zve/Zvl extensions

jerryz123:
**Related issue**: 

**Release Notes**

### DIFF
--- a/src/main/scala/tile/BaseTile.scala
+++ b/src/main/scala/tile/BaseTile.scala
@@ -104,8 +104,17 @@ trait HasNonDiplomaticTileParameters {
     val f = if (tileParams.core.fpu.nonEmpty) "f" else ""
     val d = if (tileParams.core.fpu.nonEmpty && tileParams.core.fpu.get.fLen > 32) "d" else ""
     val c = if (tileParams.core.useCompressed) "c" else ""
-    val v = if (tileParams.core.useVector) "v" else ""
+    val v = if (tileParams.core.useVector && tileParams.core.vLen >= 128 && tileParams.core.eLen == 64 && tileParams.core.vfLen == 64) "v" else ""
     val h = if (usingHypervisor) "h" else ""
+    val zvl = Option.when(tileParams.core.useVector) { Seq(s"zvl${tileParams.core.vLen}b") }
+    val zve = Option.when(tileParams.core.useVector) {
+      val c = tileParams.core.vfLen match {
+        case 64 => "d"
+        case 32 => "f"
+        case 0 => "x"
+      }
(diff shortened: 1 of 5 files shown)


src/main/scala/tile/BaseTile.scala
```diff
@@ -104,8 +104,17 @@ trait HasNonDiplomaticTileParameters {
     val f = if (tileParams.core.fpu.nonEmpty) "f" else ""
     val d = if (tileParams.core.fpu.nonEmpty && tileParams.core.fpu.get.fLen > 32) "d" else ""
     val c = if (tileParams.core.useCompressed) "c" else ""
-    val v = if (tileParams.core.useVector) "v" else ""
+    val v = if (tileParams.core.useVector && tileParams.core.vLen >= 128 && tileParams.core.eLen == 64 && tileParams.core.vfLen == 64) "v" else ""
     val h = if (usingHypervisor) "h" else ""
+    val zvl = Option.when(tileParams.core.useVector) { Seq(s"zvl${tileParams.core.vLen}b") }
```
sequencer, line 109: `vLen > 1024` is not defined in spec, although it also says:
>  Longer vector length extensions should follow the same pattern.

I still think the spec should explicitly define it.

src/main/scala/tile/BaseTile.scala
```diff
@@ -104,8 +104,17 @@ trait HasNonDiplomaticTileParameters {
     val f = if (tileParams.core.fpu.nonEmpty) "f" else ""
     val d = if (tileParams.core.fpu.nonEmpty && tileParams.core.fpu.get.fLen > 32) "d" else ""
     val c = if (tileParams.core.useCompressed) "c" else ""
-    val v = if (tileParams.core.useVector) "v" else ""
+    val v = if (tileParams.core.useVector && tileParams.core.vLen >= 128 && tileParams.core.eLen == 64 && tileParams.core.vfLen == 64) "v" else ""
```
sequencer, line 107: > The single-letter V extension is intended for use in application processor profiles.

Does spec also indicate privileged spec support, e.g. `usingUser`?

sequencer (approved):
Another my complain is [riscv-opcodes](https://github.com/riscv/riscv-opcodes) should use a fine-grained way to declare instructions(so we can parse it.)

src/main/scala/tile/BaseTile.scala
jerryz123: Take it up with the spec

src/main/scala/tile/BaseTile.scala
jerryz123: No.

PR #3667, opened 2024-08-12
B ext

jerryz123:
A simpler approach than the previous attempt.

**Related issue**: 

**Release Notes**

### DIFF
--- a/overlay.nix
+++ b/overlay.nix
@@ -8,12 +8,12 @@ final: prev: {
   });
   riscvTests = final.pkgsCross.riscv64-embedded.stdenv.mkDerivation rec {
     pname = "riscv-tests";
-    version = "55bbcc8c06637a31cc01970881ba8072838a9121";
+    version = "f2f748ebb9cf8ea049103f85c4cbf7e8a2927b16";
     src = final.fetchgit {
       url = "https://github.com/riscv-software-src/riscv-tests.git";
       rev = "${version}";
       fetchSubmodules = true;
-      sha256 = "sha256-TcIU+WFQxPqAG7lvfKPgHm4CnBpTkosqe+fYOxS+J7I=";
+      sha256 = "sha256-E3RfrP+PFIYy9c/pY04jYPxeGpnfgWwjV8iwL5+s+9w=";
     };
 
     enableParallelBuilding = true;
--- a/src/main/scala/rocket/ALU.scala
+++ b/src/main/scala/rocket/ALU.scala
@@ -31,6 +31,11 @@ object ALU {
   def FN_UNARY = 16.U
   def FN_ROL  = 17.U
   def FN_ROR  = 18.U
+  def FN_BEXT = 19.U
+
+  def FN_ANDN = 24.U
+  def FN_ORN  = 25.U
+  def FN_XNOR = 26.U
 
   def FN_MAX  = 28.U
   def FN_MIN  = 29.U
@@ -56,10 +61,11 @@ object ALU {
   def cmpUnsigned(cmd: UInt) = cmd(1)
   def cmpInverted(cmd: UInt) = cmd(0)
   def cmpEq(cmd: UInt) = !cmd(3)
-  def shiftReverse(cmd: UInt) = !cmd.isOneOf(FN_SR, FN_SRA, FN_ROR)
+  def shiftReverse(cmd: UInt) = !cmd.isOneOf(FN_SR, FN_SRA, FN_ROR, FN_BEXT)
+  def bwInvRs2(cmd: UInt) = cmd.isOneOf(FN_ANDN, FN_ORN, FN_XNOR)
 }
 
-  import ALU._
+import ALU._
 
 
 abstract class AbstractALU(implicit p: Parameters) extends CoreModule()(p) {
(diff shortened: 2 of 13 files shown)


sequencer:
omg

src/main/scala/rocket/ALU.scala
```diff
@@ -4,13 +4,14 @@
...
-import chisel3.util.{BitPat, Fill, Cat, Reverse}
+import chisel3.util.{BitPat, Fill, Cat, Reverse, PriorityEncoderOH, PopCount, MuxLookup}
 import org.chipsalliance.cde.config.Parameters
 import freechips.rocketchip.tile.CoreModule
+import freechips.rocketchip.util._
 
-class ALUFN {
-  val SZ_ALU_FN = 4
-  def FN_X    = BitPat("b????")
+object ALU {
```
sequencer, line 12: So question is here: the uop of ALU goes from 4 width to 5 width, a little increasing for all cores, but anyway in current rocket-chip methodology, we don't need to take it seriously.

src/main/scala/rocket/ALU.scala
```diff
@@ -41,21 +56,22 @@ class ALUFN {
 
   def isMulFN(fn: UInt, cmp: UInt) = fn(1,0) === cmp(1,0)
   def isSub(cmd: UInt) = cmd(3)
-  def isCmp(cmd: UInt) = cmd >= FN_SLT
+  def isCmp(cmd: UInt) = (cmd >= FN_SLT && cmd <= FN_SGEU)
+  def isMaxMin(cmd: UInt) = (cmd >= FN_MAX && cmd <= FN_MINU)
```
sequencer, line 60: I'm sightly nervous on these code, maybe we can believe synthesis code? But I personally think we can decode to a onehot signal at the decode stage(but the uop will increase). This is an affordable trade off.

src/main/scala/rocket/ALU.scala
```diff
@@ -64,50 +80,97 @@ abstract class AbstractALU[T <: ALUFN](val aluFn: T)(implicit p: Parameters) ext
...
+  val in1_bytes = io.in1.asTypeOf(Vec(xLen / 8, UInt(8.W)))
+  val orcb = VecInit(in1_bytes.map(b => Fill(8, b =/= 0.U))).asUInt
+  val rev8 = VecInit(in1_bytes.reverse).asUInt
+  val unary = MuxLookup(io.in2(11,0), count)(Seq(
+    0x287.U -> orcb,
+    (if (xLen == 32) 0x698 else 0x6b8).U -> rev8,
+    0x080.U -> io.in1(15,0),
+    0x604.U -> Fill(xLen-8, io.in1(7)) ## io.in1(7,0),
+    0x605.U -> Fill(xLen-16, io.in1(15)) ## io.in1(15,0)
+  ))
```
sequencer, line 149: RISC-V is elegant? :melting_face: 
For other future note, this is defined https://github.com/riscv/riscv-opcodes/blob/07b21cc5143a15959eda12e30aa40cea0971efe0/rv_zbb#L4-L6
Used for zvbb...

This remains some discussion on the where should unary decoding be? at DEC stage or at EXE stage?
For me I'd like to put it to dec, but the rocket-chip dec is hard to maintain...

src/main/scala/rocket/ALU.scala
jerryz123: The high bit is useless when no B ext is enabled. It should be optimized away at synthesis

src/main/scala/rocket/ALU.scala
jerryz123: This is only a handful of gates on `io.fn`, its just a bitmask ignoring the lower 2 bits of `io.fn`.

src/main/scala/rocket/ALU.scala
jerryz123: This should be done at DEC really, but I don't think it makes a big difference in the grand scheme

overlay.nix
```diff
@@ -8,7 +8,7 @@ final: prev: {
   });
   riscvTests = final.pkgsCross.riscv64-embedded.stdenv.mkDerivation rec {
     pname = "riscv-tests";
-    version = "55bbcc8c06637a31cc01970881ba8072838a9121";
+    version = "f2f748ebb9cf8ea049103f85c4cbf7e8a2927b16";
```
sequencer, line 11: Forgot to update sha256 to `sha256-E3RfrP+PFIYy9c/pY04jYPxeGpnfgWwjV8iwL5+s+9w=`

src/main/scala/rocket/ALU.scala
cyyself: I think using `cmd === FN_MAXMIN` at `isMaxMin` will make it easier to maintain. We can also apply this to `isCmp`.

jerryz123: You mean use `BitPat` to do this?

cyyself: > You mean use `BitPat` to do this?

Yes.

sequencer: In the future, with our new Rocket framework, we'd like to use our new [rvdecoderdb](https://github.com/chipsalliance/rvdecoderdb) infra for this to handle the control signal generation, and they can be used for multiple-stage uop decoding.

sequencer:
I saw CI passed in [here](https://github.com/chipsalliance/rocket-chip/actions/runs/10357436845/job/28669468965?pr=3667), let's get it into upstream!

sequencer (approved):

PR #3673, opened 2024-08-19
Solve naming conflict with Yosys

leviathanch:
Yosys and other synthesis tools already have internal ALU blocks which collide with a module named ALU


jerryz123:
@leviathanch you need to amend that commit to include a email that is associated with your Github account, such that the EasyCLA can work.

jackkoenig:
I think this change is pragmatic, but tools like Yosys really should handle collisions with user-defined modules in ways that don't require you to rename your module.

jackkoenig:
The real bug is the lack of packages/namespaces in Verilog though...

PR #3674, opened 2024-08-19
Litex2

leviathanch:
Adding support for the Crypto features back in and adding the configurations from Litex so that I can generate the platforms for pythondata_cpu_rocket

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -185,6 +185,35 @@ trait Emulator extends Cross.Module2[String, String] {
     }
   }
 
+  object litexgenerate extends Module {
+    def compile = T {
+      os.proc("firtool",
+        generator.chirrtl().path,
+        s"--annotation-file=${generator.chiselAnno().path}",
+        "--disable-annotation-unknown",
+        "-dedup",
+        "-O=debug",
+        "--split-verilog",
+        "--preserve-values=named",
+        "--output-annotation-file=mfc.anno.json",
+        "--lowering-options=disallowLocalVariables",
+        s"-o=${T.dest}"
+      ).call(T.dest)
+      PathRef(T.dest)
+    }
+
+    def rtls = T {
+      os.read(compile().path / "filelist.f").split("\n").map(str =>
+        try {
+          os.Path(str)
+        } catch {
+          case e: IllegalArgumentException if e.getMessage.contains("is not an absolute path") =>
+            compile().path / str.stripPrefix("./")
+        }
+      ).filter(p => p.ext == "v" || p.ext == "sv").map(PathRef(_)).toSeq
+    }
+  }
+
   object mfccompiler extends Module {
     def compile = T {
       os.proc("firtool",
@@ -233,7 +262,7 @@ trait Emulator extends Cross.Module2[String, String] {
         "debug_rob.cc",
         "emulator.cc",
         "remote_bitbang.cc",
-        ).map(c => PathRef(csrcDir().path / c))
+      ).map(c => PathRef(csrcDir().path / c))
     }
 
     def CMakeListsString = T {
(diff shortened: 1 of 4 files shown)


jerryz123:
Do you need the crypto extensions? We restructured the B extension implementation, but crypto is not supported currently.

jerryz123:
Also, please refer to the `dev` branch for the latest implementations

leviathanch:
My changes are on top of master, I've now rebased, changed the email in the commit and did a push -f

jerryz123:
Can you confirm if the crypto B extensions are necessary for your use case? 

They were removed from rocket because of the uncertain implications those implementations had on QoR and the intrusive nature of the original implementation. If we were to re-add support for those extensions, we would pursue a different approach.

Note that we have reimplemented the base B extensions (Zba/Zbb/Zbs), which should be sufficient for the most common use cases.

src/main/scala/subsystem/Configs.scala
```diff
@@ -632,3 +664,29 @@ class WithCloneRocketTiles(n: Int = 1, cloneHart: Int = 0, overrideIdOffset: Opt
   }
 })
 
+class WithMemoryDataBits(dataBits: Int) extends Config((site, here, up) => {
```
jerryz123, line 667: To preserve naming consistency, this should be `WithNBitMemoryBus`

src/main/scala/system/Configs.scala
```diff
@@ -91,3 +94,210 @@ class BaseFPGAConfig extends Config(new BaseConfig ++ new WithCoherentBusTopolog
 class DefaultFPGAConfig extends Config(new WithNSmallCores(1) ++ new BaseFPGAConfig)
 
 class CloneTileConfig extends Config(new WithCloneRocketTiles(7) ++ new WithNBigCores(1) ++ new WithCoherentBusTopology ++ new BaseConfig)
+
+class BaseLitexConfig extends Config(
```
jerryz123, line 98: Can you move the Litex-configs to a new file under `system` (LitexConfigs.scala?).

jerryz123 (changes requested):

src/main/scala/system/Configs.scala
leviathanch: OK, I will move the Litex configurations into a separate file.
The problem with the crypto extension is that several projects which use Litex and in turn Rocket Chip have been developed for using the old crypto implementation.

Maybe we can do something like a legacy flag for what crypto extension to use or so?

Also, BTW, now that I fixed yosys and it parses through the System Verilog code produced by rocket chip, it seems to do something very nasty and declares variables (automatic logic) within always blocks

jerryz123: Can you point me to the downstream projects which rely on crypto extensions? I want to see exactly which ones of the many extensions are used. 
I'm happy to do a bit of work adding back in some of the crypto support, but I'd like to know where to start.

leviathanch: OK, I've now decided to get rid of the legacy crypto B extension and using your proper implementation, so much has changed in Rocket Chip that pythondata_cpu_rocket will need a major rework anyway.

jerryz123: Sorry about the churn. We should have taken more care before merging the original crypto+B implementation.

Ultimately what motivated me to revert the entire crypto support was the intrusive change to decode that the crypto+B implementation mandated. Attempting to implement a flag that switches between a B-enabled ALU and the normal ALU made the system quite a bit messier, and caused headaches for downstream projects which reference the rocket ALU or decode. 

Any future extensions to rocket, including if/when we add back the crypto support, must be done in a way to have minimal QoR impact on base components, and must not potentially introduce bugs into base features or instructions.

The good news is, now that we have the crypto FUs in the commit history, it would be only a small lift to re-integrate them in a less intrusive way (although the lack of readily available tests for many crypto instructions still makes me nervous about doing so).

build.sc
```diff
@@ -294,6 +294,41 @@ object emulator extends Cross[Emulator](
   //
   ("freechips.rocketchip.system.TestHarness", "freechips.rocketchip.system.DefaultRV32Config"),
   ("freechips.rocketchip.system.TestHarness", "freechips.rocketchip.system.DefaultFP16Config"),
+  ("freechips.rocketchip.system.TestHarness", "freechips.rocketchip.system.BitManipCryptoConfig"),
+  ("freechips.rocketchip.system.TestHarness", "freechips.rocketchip.system.BitManipCrypto32Config"),
```
jerryz123, line 298: Should these two be removed?

leviathanch: Yeah, I've removed them now

jerryz123 (approved):
Thanks!

PR #3676, opened 2024-08-19
Prioritize coprocessor FPU reqs over core FPU reqs

jerryz123:
This avoids deadlock conditions due to coprocessor requests typically being from older instructions.

**Related issue**: 

**Release Notes**


sequencer:
But why it will deadlock, I don't understand. It sounds like the deadlock is coming from the structural hazard between scalar and vec/rocc. However they basically has no real dependencies?

jerryz123:
Suppose a older vector is trying to use the FPU, while the core is trying to execute a younger vector instruction, which needs the FPU at the X stage to access the FP operand. If the younger instruction is given priority, it will read the operand, but then potentially stall+replay at the W stage due to vector unit being busy.

Generally, in such a simple pipeline, prioritizing the oldest in-flight operation will avoid deadock conditions.

sequencer (approved):

PR #3704, opened 2025-01-03
fixed TL refill request when TL request is not accepted

HakamAtassi:
**Related issue**: 

**Impact**: API modification

**Release Notes**

There is a peculiarity in the rocket chip I$ that seems to be a bug. Namely, it seems that the I$ is coded such that when a miss takes place, the I$ only requests a refill from the TL bus once, and achieves this by checking if the previous cycle has a miss and the I$ is currently missing. This works okay most of the time but doesnt account for the TL.A channel not accepting the request. 

ex: 
I$_Bug

What I suspect the offending line of code is (line 380 in icache.scala):

image

Waveform after the change: 

image

### DIFF
--- a/src/main/scala/rocket/ICache.scala
+++ b/src/main/scala/rocket/ICache.scala
@@ -377,7 +377,7 @@ class ICacheModule(outer: ICache) extends LazyModuleImp(outer)
   /** [[io]] access L1 I$ miss. */
   val s2_miss = s2_valid && !s2_hit && !io.s2_kill
   /** forward signal to stage 1, permit stage 1 refill. */
-  val s1_can_request_refill = !(s2_miss || refill_valid)
+  val s1_can_request_refill = !((s2_miss && refill_fire) || refill_valid)
   /** real refill signal, stage 2 miss, and was permit to refill in stage 1.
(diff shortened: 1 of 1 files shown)


jerryz123:
I'm not sure this is a needed fix. If the A channel does not accept, it is the responsibility of the frontend to request the address again from the Icache.

jerryz123:
I think you should avoid holding the same address on the ICache until it gets a response, as the ICache is designed to be used in a pipelined design. You should fire the request into the ICache once, check on the next cycle if a response was provided (a hit), and reissue a request if no response was provided.

I realize your request for changing the ICache behavior is reasonable, and I can't immediately spot any potential problem, but I'm concerned about it introducing a subtle bug elsewhere that I don't have the bandwidth to resolve.

If you truly desire the described behavior, you can fork the ICache implementation and simplify it to your needs.

HakamAtassi:
Hey, 

That's reasonable. In my implementation I keep a buffer of the previous request and re-request the prev_pc if it never got a response, which effectively holds the address constant on the I$. 

You make a fair point though. I'll keep this open if I or anyone else is end up thoroughly verifying things. Thanks again.

PR #3706, opened 2025-01-15
ADD: Timing-Accurate Core Instruction Tracing

iansseijelly:
**Related issue**: 

**Release Notes**

Added feature for:
1. rocket core to drive the trace ingress bundle, which is a standard interface for different trace encoder standards. 
2. trace encoder builder in rockettile. This allows the building of trace encoders within the tile, whether it is N-trace, E-trace, or custom. 
3. trace sink builder in rockettile. This allows the building of a sequence of trace sinks. Also supports monitoring the data sinking into the arbiter to cross-check the encoding. 
4. arbiter and controller instantiation in rockettile. This instantiates an arbiter that chooses the different trace sinks, and a trace controller that instantiates the necessary MMIO register routings to control encoders.


src/main/scala/util/TraceCoreIngressGen.scala
```diff
@@ -0,0 +1,52 @@
+// See LICENSE.Berkeley for license details.
```
jerryz123, line 1: Can you make a new package for these? `src/main/scala/trace`, and `freechips.rocketchip.trace`.

src/main/scala/util/TraceCoreInterface.scala
```diff
@@ -28,8 +28,8 @@ object TraceItype extends ChiselEnum {
 class TraceCoreParams (
```
jerryz123, line 28: This should be a `case class`

src/main/scala/util/TraceCoreIngressGen.scala
```diff
@@ -0,0 +1,52 @@
...
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import chisel3._
+
+class TraceCoreIngressGen(val params: TraceCoreParams) extends Module {
```
jerryz123, line 8: Why is this named `IngressGen` instead of just `TraceCoreIngress`

src/main/scala/rocket/RocketCore.scala
```diff
@@ -131,6 +131,8 @@ class CoreInterrupts(val hasBeu: Boolean)(implicit p: Parameters) extends TileIn
 trait HasRocketCoreIO extends HasRocketCoreParameters {
   implicit val p: Parameters
   def nTotalRoCCCSRs: Int
+  def ingress_params = new TraceCoreParams(nGroups = 1, iretireWidth = coreParams.retireWidth,
```
jerryz123, line 134:
```suggestion
  def traceIngressParams = new TraceCoreParams(nGroups = 1, iretireWidth = coreParams.retireWidth, 
```

src/main/scala/subsystem/HasTiles.scala
```diff
@@ -175,6 +175,7 @@ trait CanAttachTile {
     connectOutputNotifications(domain, context)
     connectInputConstants(domain, context)
     connectTrace(domain, context)
+    connectTraceSinkDMA(domain, context)
```
jerryz123, line 178: Add a comment here to explain the difference between `connectTrace` and `connectTraceSinkDMA`

src/main/scala/rocket/RocketCore.scala
```diff
@@ -823,6 +828,25 @@ class Rocket(tile: RocketTile)(implicit p: Parameters) extends CoreModule()(p)
                  wb_reg_wdata))))
   when (rf_wen) { rf.write(rf_waddr, rf_wdata) }
 
+  val ingress_gen = Module(new TraceCoreIngressGen(ingress_params))
```
jerryz123, line 831: `RocketCoreParams` should have a new flag `enableTraceCoreIngress: Boolean`, defaults to false. If set, it builds this block and connects it as below.

jerryz123, line 831:
```suggestion
  val trace_ingress = Module(new TraceCoreIngressGen(ingress_params))
```

src/main/scala/tile/BaseTile.scala
```diff
@@ -299,6 +301,8 @@ abstract class BaseTile private (crossing: ClockCrossingType, q: Parameters)
   val traceCoreSourceNode = BundleBridgeSource(() => new TraceCoreInterface(traceCoreParams))
   /** Node for external consumers to source  a V1.0 instruction trace from the core. */
   val traceCoreNode = traceCoreSourceNode
+  /** Node for tile to drive encoded instruction trace to external consumers. */
+  val traceSinkIdentityNode = TLIdentityNode()
```
jerryz123, line 305: Hmm, in this architecture the trace tilelink network is independent of the TileBus <-> SystemBus connections. Is this intentional?

src/main/scala/tile/RocketTile.scala
```diff
@@ -177,7 +223,7 @@ class RocketTileModuleImp(outer: RocketTile) extends BaseTileModuleImp(outer)
 
   // Pass through various external constants and reports that were bundle-bridged into the tile
   outer.traceSourceNode.bundle <> core.io.trace
-  core.io.traceStall := outer.traceAuxSinkNode.bundle.stall
+  // core.io.traceStall := outer.traceAuxSinkNode.bundle.stall
```
jerryz123, line 226: ?

src/main/scala/tile/RocketTile.scala
```diff
@@ -83,6 +90,23 @@ class RocketTile private(
     beu
   }
 
+  val trace_encoder_controller = rocketParams.ltrace.map { t =>
```
jerryz123, line 93: Add some comments here. The trace_encode_controller selects whether the `trace_sink_print` or the `trace_sink_dma` is used? Add comments indicating the difference between them.

src/main/scala/tile/RocketTile.scala
```diff
@@ -83,6 +90,23 @@ class RocketTile private(
...
+    val trace_encoder_controller = LazyModule(new TraceEncoderController(t.encoderBaseAddr, xBytes))
+    connectTLSlave(trace_encoder_controller.node, xBytes)
+    trace_encoder_controller
+  }
+
+  val trace_sink_print = rocketParams.ltrace.map { t =>
```
jerryz123, line 99: This needs to be done carefully... as the printer is not synthesizable. Is there a use case where you need runtime (via trace_encode_controller) control over selecting whether the printer or the DMA is the consumer?

src/main/scala/tile/RocketTile.scala
```diff
@@ -83,6 +90,23 @@ class RocketTile private(
...
+    LazyModule(new TraceSinkPrint(rocketParams.uniqueName))
+  }
+
+  val trace_sink_dma = rocketParams.ltrace.map { t =>
+    val trace_sink_dma = LazyModule(new TraceSinkDMA(t.sinkDMABaseAddr, xBytes))
+    connectTLSlave(trace_sink_dma.regnode, xBytes)
+    traceSinkIdentityNode := trace_sink_dma.node
```
jerryz123, line 106: This should just master the tileMasterXbar, I think, instead of defining a new TL node which goes to the MBUS.

src/main/scala/util/TraceCoreIngressGen.scala
jerryz123: Match the indentation to 2-space indentation in this file

src/main/scala/util/TraceCoreInterface.scala
```diff
@@ -28,8 +28,8 @@ object TraceItype extends ChiselEnum {
...
   val nGroups: Int = 1,
   val iretireWidth: Int = 1,
-  val xlen: Int = 32,
-  val iaddrWidth: Int = 32
+  val xlen: Int = 64,
```
jerryz123, line 31: Why is it ok to change these defaults?

src/main/scala/util/TraceEncoder.scala
```diff
@@ -0,0 +1,422 @@
...
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.util
+
+import chisel3._
+import chisel3.util._
+import scala.math.min
+
+
+class TraceEncoderParams(
```
jerryz123, line 11: This should be a `case class`

src/main/scala/util/TraceEncoderController.scala
```diff
@@ -0,0 +1,99 @@
...
+import freechips.rocketchip.tilelink.TLRegisterNode
+import freechips.rocketchip.regmapper.{RegField, RegFieldDesc}
+
+class TraceEncoderControlInterface() extends Bundle {
+  val enable = Bool()
+  val target = UInt(TraceSinkTarget.getWidth.W)
+  val hpmcounter_enable = UInt(32.W)
+  val hpmcounter_report_interval = UInt(32.W)
+}
+class TraceEncoderController(addr: BigInt, beatBytes: Int)(implicit p: Parameters) extends LazyModule {
```
jerryz123, line 21: Can this thing just be combined with the Arbiter? Since this thing sets the control signals for the arbiter.

src/main/scala/util/TraceSink.scala
```diff
@@ -0,0 +1,166 @@
...
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.diplomacy._
+import org.chipsalliance.cde.config.{Parameters, Field, Config}
+import freechips.rocketchip.subsystem._
+import freechips.rocketchip.prci._
+import freechips.rocketchip.regmapper.{RegField, RegFieldDesc}
+
+trait HasTraceSinkIO {
+  val io = IO(new Bundle {
+    val in = Flipped(Decoupled(UInt(8.W)))
```
jerryz123, line 15:
```suggestion
    val trace_in = Flipped(Decoupled(UInt(8.W)))
```
?

src/main/scala/util/TraceSink.scala
```diff
@@ -0,0 +1,166 @@
...
+  })
+}
+
+class TraceSinkPrint(bb_name: String)(implicit p: Parameters) extends LazyModule {
```
jerryz123, line 19: What is `bb_name`?

src/main/scala/util/TraceSink.scala
```diff
@@ -0,0 +1,166 @@
...
+    "FILE_NAME" -> StringParam(s"tacit_${name}_trace.out")
+  )
+) with HasBlackBoxResource {
+  val io = IO(new Bundle {
+    val clk = Input(Clock())
+    val reset = Input(Reset())
+    val in_valid = Input(Bool())
+    val in_byte = Input(UInt(8.W))
+  })
+  addResource("/vsrc/BytePrinter.v")
```
jerryz123, line 45: This file seems to be missing

src/main/scala/util/TraceSink.scala
```diff
@@ -0,0 +1,166 @@
+package freechips.rocketchip.util
+
+import chisel3._
```
jerryz123, line 3: The printer and DMA sink should be separated into two files

src/main/scala/tile/BaseTile.scala
iansseijelly: Yes, this is to bypass the coherent memory system and directly sink data into DRAM.

src/main/scala/tile/RocketTile.scala
iansseijelly: Depends on whether the trace encoder is used or not, the value of traceStall is either what is currently is, or `core.io.traceStall := outer.traceAuxSinkNode.bundle.stall || trace_encoder.io.stall`. This is in line 196 or RocketTile.scala.

src/main/scala/tile/RocketTile.scala
iansseijelly: I wonder if it's best to just have a spec or datasheet of the functionality of registers? It has enable bits, sets target, and can also control the tile CSR trace encoder, which will be a separate block that will be upstreamed in the future.

src/main/scala/tile/BaseTile.scala
jerryz123: Why? To avoid contention?

src/main/scala/tile/RocketTile.scala
iansseijelly: I added a guard to the printer black-box (ifdef `VCS). So this synthesizes fine in Vivado. It's just not a valid selection and should be optimized out since there's no logic there.

src/main/scala/tile/RocketTile.scala
iansseijelly: As said, the intention is to bypass the coherent memory and directly go to DRAM.

src/main/scala/tile/RocketTile.scala
jerryz123: Regardless of whether or not a datasheet/specs are added, the code can have more descriptive inline comments.

src/main/scala/tile/BaseTile.scala
iansseijelly: Yes exactly. The trace should have minimal perturbation impact to the rest of the system, including not making these traces mess up the cache memory state. I think in an ideal world we can dynamically claim a certain memory region as non-cacheable, but in TL this declaration is static, so this is an alternative approach to achieve the same effect.

src/main/scala/tile/RocketTile.scala
jerryz123: Even if the logic doesn't break the build, this means it is possible to build a arbiter where one of the sinks just sinks the trace to nowhere. 

Simulation-only things should not be so intertwined with synthesized logic.

I think I would prefer a design where the only trace sink is the DMA sink, and a simulation-only flag (`+ltrace_print`, for example), would print the trace as it moves between the trace encoder and the sink. This also simplifies and removes the need for a programmable trace arbiter.

src/main/scala/tile/BaseTile.scala
jerryz123: Hmm, ok. The concern is that if you go to the mbus, and the memory region allocated to the trace has been cached in the LLC, writes through the mbus to that region will be lost when the LLC writes back its (potentially) dirty data. Similarly, reads to the LLC will hit in the LLC, and not see the dirty data.

src/main/scala/tile/RocketTile.scala
iansseijelly: I think an arbiter still has its value here, because I would also like to add other sink targets like some sort of tethered debug port (like UART), as well as other simulation-only sinks (like a FireSim bridge). Maybe there needs to be extra parameters to choose what sinks are instantiated, so simulation-only sinks are not even instantiated in a non-simulation design.

src/main/scala/tile/BaseTile.scala
iansseijelly: I agree, and I have ran into this issue during development. The current way of avoiding this scenario is:
https://github.com/ucb-bar/Baremetal-IDE/blob/main/examples/pmu-tests/src/ltrace-dma-test.c#L10
I understand a software guarantee is not ideal, but I an not sure if there's a better way to resolve this issue.

src/main/scala/tile/RocketTile.scala
jerryz123: I agree having a interface to plugin arbitrarily many sinks would be nice.

What if we upstream the trace encoder, trace arbiter, and abstract interface of a trace-sink into rocket-chip, but leave the implementations of the trace sinks elsewhere? This makes it easier to experiment with new trace sinks without hacking up rocket-chip

iansseijelly: That sounds good to me 👍

src/main/scala/util/Blockable.scala
```diff
@@ -6,6 +6,7 @@ import chisel3._
 import chisel3.util.DecoupledIO
 import freechips.rocketchip.tile.{TraceBundle}
 import freechips.rocketchip.rocket.{TracedInstruction}
+import freechips.rocketchip.trace.{TraceCoreInterface, TraceItype}
```
jerryz123, line 9: This should not be changed?

iansseijelly: `TraceCoreInterface` used to be in utils, but since we're creating a separate trace package I feel it's cleaner to make it part of the new package, and hence Blockable needs to import these interfaces now.

jerryz123: Ah right

src/main/resources/vsrc/TraceSinkMonitor.v
```diff
@@ -0,0 +1,36 @@
...
+    parameter FILE_NAME = "trace_sink_monitor.txt"
+)
+(
+    input clk,
+    input reset,
+    input in_fire,
+    input[7:0] in_byte
+);
+
+`ifdef VCS
```
jerryz123, line 12: This should work in all RTL simulators. I think `ifndef SYNTHESIS` is what you want

src/main/scala/tile/RocketTile.scala
jerryz123: Ok just delete this line then

src/main/scala/trace/TraceSinkAlways.scala
```diff
@@ -0,0 +1,54 @@
...
+class TraceSinkAlways()(implicit p: Parameters) extends LazyModule {
+  println(s"TraceSinkAlways Lazy Init}")
+  override lazy val module = new TraceSinkAlwaysImpl(this)
+  class TraceSinkAlwaysImpl(outer: TraceSinkAlways) extends LazyModuleImp(outer) {
+    println(s"TraceSinkAlways Impl Init")
+    val io = IO(new Bundle {
+      val trace_in = Flipped(Decoupled(UInt(8.W)))
+    })
+    withClockAndReset(clock, reset) {
+      io.trace_in.ready := true.B
```
jerryz123, line 21: You don't need withClockAndReset here... there is no synchronous logic

src/main/scala/trace/TraceSinkAlways.scala
```diff
@@ -0,0 +1,54 @@
+package freechips.rocketchip.trace
+
+import chisel3._
+import chisel3.util._
+import freechips.rocketchip.diplomacy._
+import freechips.rocketchip.prci._
+import org.chipsalliance.cde.config.{Parameters, Config, Field}
+import freechips.rocketchip.tile._
+import freechips.rocketchip.subsystem._
+case object TraceSinkAlwaysKey extends Field[Option[Int]](None)
```
jerryz123, line 10: What does this Int indicate?

src/main/scala/tile/RocketTile.scala
```diff
@@ -83,6 +88,29 @@ class RocketTile private(
...
+   * controlling enable/disable of trace encoder
+   * and selecting trace sink
+   */
+  val trace_encoder_controller = rocketParams.ltrace.map { t =>
+    val trace_encoder_controller = LazyModule(new TraceEncoderController(t.encoderBaseAddr, xBytes))
+    connectTLSlave(trace_encoder_controller.node, xBytes)
+    trace_encoder_controller
+  }
+
+  val (trace_sinks, trace_sink_ids) = rocketParams.ltrace match {
```
jerryz123, line 102:
```suggestion
  val (trace_sinks, traceSinkIds) = rocketParams.ltrace match {
```

Informal convention is snake_case for HW vars, camelCase for SW vars

src/main/scala/tile/RocketTile.scala
```diff
@@ -83,6 +88,29 @@ class RocketTile private(
...
+    case Some(t) =>
```
jerryz123, line 103: I
```suggestion
    case Some(t) => t.buildSinks.map {(p)}.unzip
```

Does this work?

src/main/scala/tile/RocketTile.scala
```diff
@@ -83,6 +88,29 @@ class RocketTile private(
...
+    val trace_encoder_controller = LazyModule(new TraceEncoderController(t.encoderBaseAddr, xBytes))
+    connectTLSlave(trace_encoder_controller.node, xBytes)
+    trace_encoder_controller
+  }
+
+  val (trace_sinks, trace_sink_ids) = rocketParams.ltrace match {
+    case Some(t) => 
+      val sequence = t.buildSinks.map {_(p)}
+      (sequence.map(_._1), sequence.map(_._2))
+    case None => (Seq.empty, Seq.empty)
```
jerryz123, line 106:
```suggestion
    case None => (Nil, Nil)
```

src/main/scala/trace/TraceEncoder.scala
```diff
@@ -0,0 +1,423 @@
...
+import scala.math.min
+
+import org.chipsalliance.cde.config.Parameters
+
+case class TraceEncoderParams(
+  coreParams: TraceCoreParams,
+  bufferDepth: Int,
+  encoderBaseAddr: BigInt,
+  useArbiterMonitor: Boolean,
+  // a seq of functions that takes a parameter and returns a lazymodule and a target id
```
jerryz123, line 17: Can you explain what the target id is for in a comment? I thnk that is not obvious

src/main/scala/trace/TraceEncoder.scala
```diff
@@ -0,0 +1,423 @@
...
+    val control = Input(new TraceEncoderControlInterface())
+    val in = Input(new TraceCoreInterface(params.coreParams))
+    val stall = Output(Bool())
+    val out = Decoupled(UInt(8.W))
+  })
+
+  val MAX_DELTA_TIME_COMP = 0xCF // 63, 6 bits
+
+  when (io.stall) {
+    printf("TraceEncoder stall detected\n")
```
jerryz123, line 214: This printf should probably not be here

src/main/scala/trace/TraceSinkAlways.scala
jerryz123: I think this can be removed now

src/main/scala/trace/TraceSinkArbiter.scala
```diff
@@ -0,0 +1,40 @@
...
+import chisel3.util._
+import org.chipsalliance.cde.config.Parameters
+import org.chipsalliance.diplomacy.lazymodule._
+import freechips.rocketchip.diplomacy.{AddressSet}
+import freechips.rocketchip.resources.{SimpleDevice}
+import freechips.rocketchip.tile._
+import freechips.rocketchip.regmapper.{RegField, RegFieldDesc}
+
+class TraceSinkArbiter(nSeq : Seq[Int], use_monitor: Boolean = false, monitor_name: String = "unknown")(implicit p: Parameters) extends LazyModule {
+  println(s"TraceSinkArbiter nSeq: $nSeq")
```
jerryz123, line 16: Prints should be removed.

src/main/scala/trace/TraceEncoder.scala
iansseijelly: I think this RTL printf is somewhat important, so in simulation we can kind of know when stalls happen. I would prefer to keep it.

jerryz123: It messes up the commit log, and I think there's a compelling reason to include the trace stuff by default in some designs.

You can add a PlusArg flag - `PlusArg("trace_debug")`, for example, which can generate a HW bool that you can use to gate the printf.

Then at run-time, to enable the TraceEncode prints, you just pass (in CY) `EXTRA_SIM_FLAGS="+trace_debug=1"`

src/main/scala/trace/TraceSinkDMA.scala
```diff
@@ -0,0 +1,172 @@
...
+import chisel3.util._
+import freechips.rocketchip.tilelink._
+import freechips.rocketchip.diplomacy._
+import org.chipsalliance.cde.config.{Parameters, Field, Config}
+import freechips.rocketchip.subsystem._
+import freechips.rocketchip.prci._
+import freechips.rocketchip.regmapper.{RegField, RegFieldDesc}
+import freechips.rocketchip.tile._
+
+case object TraceSinkDMAKey extends Field[Option[Int]](None)
```
jerryz123, line 13: We are not putting this in RC, right?

src/main/scala/trace/TraceSinkArbiter.scala
```diff
@@ -0,0 +1,40 @@
...
+    })
+    val nVec = VecInit(nSeq.map(_.U))
+    io.in.ready := Mux(nVec.contains(io.target), io.out(nVec.indexWhere(_ === io.target)).ready, true.B)
+    io.out.zipWithIndex.foreach { case (o, i) => 
+      o.valid := io.in.valid && (io.target === nVec(i))
+      o.bits := io.in.bits
+    }
+
+    if (use_monitor) {
+      val monitor = Module(new TraceSinkMonitor(s"trace_monitor_$monitor_name.out"))
```
jerryz123, line 31: Should `monitor_name` reflect the hartId of the core that is being traced? In a multicore, each core should generate its own trace file, right?

jerryz123 (changes requested):
High level request, can you add in a comment at the beginning of the files you added pointing out what spec exactly you implemented? Like "This is an implementation of the TraceXXX as defined in the RISC-V Trace-XXX spec"

src/main/scala/tile/RocketTile.scala
```diff
@@ -83,6 +88,32 @@ class RocketTile private(
...
+    case Some(t) => Some(t.buildEncoder(p))
+    case None => None
+  }
+
+  val (trace_sinks, traceSinkIds) = rocketParams.traceParams match {
+    case Some(t) => t.buildSinks.map {_(p)}.unzip
+    case None => (Nil, Nil)
+  }
+
+  val trace_sink_arbiter = rocketParams.traceParams.map { t =>
```
jerryz123, line 112: Can you just merge the arbiter and encoder modules?

src/main/scala/tile/RocketTile.scala
```diff
@@ -83,6 +88,32 @@ class RocketTile private(
...
+   * controlling enable/disable of trace encoder
+   * and selecting trace sink
+   */
+  val trace_encoder_controller = rocketParams.traceParams.map { t =>
+    val trace_encoder_controller = LazyModule(new TraceEncoderController(t.encoderBaseAddr, xBytes))
+    connectTLSlave(trace_encoder_controller.node, xBytes)
+    trace_encoder_controller
+  }
+
+  val trace_encoder = rocketParams.traceParams match {
```
jerryz123, line 102:
```suggestion
  val trace_encoder = rocketParams.traceParams.map(_.buildEncoder(p))
```

src/main/scala/tile/RocketTile.scala
jerryz123: @iansseijelly

src/main/scala/trace/TacitEncoder.scala
```diff
@@ -0,0 +1,408 @@
+// See LICENSE.Berkeley for license details.
+// See LICENSE.SiFive for license details.
+
+package freechips.rocketchip.trace
+
+import chisel3._
+import chisel3.util._
+import scala.math.min
+
+import org.chipsalliance.cde.config.Parameters
```
jerryz123, line 10: This should be removed from RC then? The abstract encoder is in RC, but the custom encoder should be elsewhere?

src/main/scala/trace/TraceSinkMonitor.scala
```diff
@@ -0,0 +1,19 @@
+package freechips.rocketchip.trace
+
+import chisel3._
+import chisel3.util._
+import chisel3.experimental.StringParam
+
+class TraceSinkMonitor(name: String) extends BlackBox(
+  Map(
+    "FILE_NAME" -> StringParam(s"tacit_${name}_trace.out")
```
jerryz123, line 9: Clean `tacit` from the naming

jerryz123 (changes requested):
I'll pass through this again once you remove all the things that will not be upstreamed @iansseijelly

iansseijelly:
Hi, I have removed the non-standard components, and addressed the above items.

src/main/scala/trace/TraceSinkArbiter.scala
```diff
@@ -0,0 +1,39 @@
...
+import chisel3._
+import chisel3.util._
+import org.chipsalliance.cde.config.Parameters
+import org.chipsalliance.diplomacy.lazymodule._
+import freechips.rocketchip.diplomacy.{AddressSet}
+import freechips.rocketchip.resources.{SimpleDevice}
+import freechips.rocketchip.tile._
+import freechips.rocketchip.regmapper.{RegField, RegFieldDesc}
+
+class TraceSinkArbiter(nSeq : Seq[Int], use_monitor: Boolean = false, monitor_name: String = "unknown")(implicit p: Parameters) extends LazyModule {
```
jerryz123, line 15: This should just be a Module, not a LazyModule

jerryz123 (changes requested):
Only one minor request, TraceSinkArbiter can be a Module, not a LazyModule

jerryz123 (approved):

src/main/scala/tile/RocketTile.scala
```diff
@@ -153,6 +177,32 @@ class RocketTileModuleImp(outer: RocketTile) extends BaseTileModuleImp(outer)
...
+    val arb = Module(new TraceSinkArbiter(outer.traceSinkIds, use_monitor = t.useArbiterMonitor, monitor_name = outer.rocketParams.uniqueName))
+    arb
+  }
+
+  if (outer.rocketParams.traceParams.isDefined) {
+    core.io.trace_core_ingress.get <> outer.trace_encoder.get.module.io.in
+    outer.trace_encoder_controller.foreach { lm =>
+      outer.trace_encoder.get.module.io.control <> lm.module.io.control
+    }
+
```
jerryz123, line 190: Don't construct the arbiter outside this `if` block... just do it inside.... then you don't need a `Option[Arbiter]`

PR #3717, opened 2025-02-19
[RTL]Fix BTB.scala: update_target is from btb_update

donaldkuck:
**Related issue**: none

**Type of change**: bug fix

**Release Notes**

Fix BTB.scala: update_target is from btb_update


jerryz123:
I believe taking it from `io.req.bits.addr` is a area optimization.... this is just taking it from the s1_pc register in the frontend, instead of adding the target to the r_btb_update pipeline register inside the BTB.

I'm aware this does make the BTB module more tricky to use in a non-Rocket core... is that what you are doing?

PR #3724, opened 2025-03-09
Update to Chisel 6.7, use new annotation API

### LINKED ISSUE #3722: Chisel 7 Support - ChiselAnnotation going away
With https://github.com/chipsalliance/chisel/pull/4717, Chisel has removed the `ChiselAnnotation` API, which breaks a number of spots throughout rocket-chip:

- https://github.com/chipsalliance/rocket-chip/blob/ac5192c76646569ebd47d247b33cd69d6f01ce0c/src/main/scala/util/ElaborationArtefactAnnotation.scala#L41-L49
- https://github.com/chipsalliance/rocket-chip/blob/ac5192c76646569ebd47d247b33cd69d6f01ce0c/src/main/scala/rocket/DCache.scala#L26-L31
- Many uses within https://github.com/chipsalliance/rocket-chip/blob/ac5192c76646569ebd47d247b33cd69d6f01ce0c/src/main/scala/util/Annotations.scala

Can we replace the custom `DCache` `InlineInstance` trait with either `chisel3.util.experimental.InlineInstance` (in Chisel 6.6.0) or [`InlineInstanceAllowDedup`](https://github.com/chipsalliance/chisel/blob/05c87bca07ebaa611b9e4486a7e041c59ff4c628/src/main/scala/chisel3/util/experimental/Inline.scala#L45-L50) (in Chisel 7)?

Do the other annotations have any consumer, or would they be harmless to remove?

tymcauley:
Resolves #3722.

**Related issue**: #3722

**Release Notes**

Update to Chisel 6.7.0

### DIFF
--- a/build.sc
+++ b/build.sc
@@ -12,7 +12,7 @@ object v {
   val scala = "2.13.12"
   // the first version in this Map is the mainly supported version which will be used to run tests
   val chiselCrossVersions = Map(
-    "5.1.0" -> (ivy"org.chipsalliance::chisel:5.1.0", ivy"org.chipsalliance:::chisel-plugin:5.1.0"),
+    "6.7.0" -> (ivy"org.chipsalliance::chisel:6.7.0", ivy"org.chipsalliance:::chisel-plugin:6.7.0"),
     // build from project from source
     "source" -> (ivy"org.chipsalliance::chisel:99", ivy"org.chipsalliance:::chisel-plugin:99"),
   )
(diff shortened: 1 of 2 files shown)


tymcauley:
Looks like because the `mill` files in this repo are a bit old, and the `chisel` submodule is pretty out-of-date, updating to a newer version of the Scala compiler breaks CI. I don't know how the nix flow works too well, so seems like someone who can work with that should do a bump of the Chisel source dependency at some point, but it's not blocking for this PR.

jerryz123 (approved):

PR #3728, opened 2025-03-17
Resolve no-implicit-clock error with AXI4 monitors

tymcauley:
If a user enables AXI4 Monitors (using `AXI4MonitorBuilder`), they'll get a no-implicit-clock error at this point, which isn't easy to track down. Chipyard worked around a similar problem with TileLink Monitors here: https://github.com/ucb-bar/chipyard/pull/1868

**Related issue**: None

**Release Notes**

Resolve build error when using AXI4 monitors.


jerryz123 (approved):

tymcauley:
Thanks for approving @jerryz123! Can we merge this now?

PR #3754, opened 2025-07-15
Clint rocket pr 0715

zhaohong1988:
update rocket-chip for new clint. Add remapclint to be used only for timer, to co-use with hart id.


jerryz123:
This looks like lots of unrelated changes bundles together.

PR #3785, opened 2026-03-24
Fix AHBToTL elaboration error due to missing user fields

tymcauley:
Without this change, the following error shows up when connecting an AHBSlaveSourceNode (which can model an external AHB master port) through AHBToTL:

    [error] generators/rocket-chip/src/main/scala/amba/ahb/ToTL.scala 121:19: unconnected consumer field AHBToTL.d_user.amba_prot: Reg[AMBAProtBundle]
    [error]           d_user  :<= in.hauser
    [error]                   ^
    [error] There were 1 error(s) during hardware elaboration.

**Related issue**: N/A

**Release Notes**

Fix AHBToTL elaboration error due to missing user fields

### DIFF
--- a/src/main/scala/amba/ahb/Test.scala
+++ b/src/main/scala/amba/ahb/Test.scala
@@ -6,12 +6,12 @@ import chisel3._
 
 import org.chipsalliance.cde.config.Parameters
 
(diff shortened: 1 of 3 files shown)


tymcauley:
@jerryz123 or @sequencer are you able to review this?

PR #3786, opened 2026-04-13
fix rocc write back

andrewchen29:
**Related issue**: 
The RoCC Coprocessor will write back the value unconditionally without having `xd` rocc instructions, which might cause data pollution if the Coprocessor intentionally raise `io.rocc.resp.valid`

**Impact**: functional change

**Release Notes**


andrewchen29:
Just found out that it now will cause the second write be masked out.
Maybe it's just Coprocessor's responsibility to not to do that.

PR #3822, opened 2026-08-15
Resolve implicit-width-truncation warnings

tynan-jdwk:
firtool-1.154.0 and later warn on all implicit width truncations (https://github.com/llvm/circt/pull/10621). These width truncations were flagged in stock Chipyard builds.

The commits in this PR are separated by the type of truncation fix, and the commit messages describe why the fix was made, and why the resulting RTL is equivalent to the existing design.

**Related issue**: N/A

**Release Notes**

Resolve implicit-width-truncation warnings

### DIFF
--- a/src/main/scala/devices/debug/Custom.scala
+++ b/src/main/scala/devices/debug/Custom.scala
@@ -84,7 +84,7 @@ class DebugCustomXbar(
     }
     // mux correct 'ready' and 'data' based on address
     sink.ready := (decoded zip sources).foldLeft(false.B){case (result, (d, i)) => result || (d & i.ready)}
-    sink.data := (decoded zip sources).foldLeft(0.U){ case (result, (d, i)) => result | Mux(i.ready, i.data, 0.U)}
+    sink.data :<= (decoded zip sources).foldLeft(0.U(0.W)){ case (result, (d, i)) => result | Mux(i.ready, i.data, 0.U)}
 
   }
 }
--- a/src/main/scala/devices/debug/Debug.scala
+++ b/src/main/scala/devices/debug/Debug.scala
@@ -24,6 +24,7 @@ import freechips.rocketchip.util.{AsyncBundle, AsyncQueueParams, AsyncResetSynch
 import freechips.rocketchip.util.SeqBoolBitwiseOps
 import freechips.rocketchip.util.SeqToAugmentedSeq
 import freechips.rocketchip.util.BooleanToAugmentedBoolean
+import freechips.rocketchip.util.ConnectableBitsOpExtension
 
 object DsbBusConsts {
   def sbAddrWidth = 12
@@ -905,7 +906,7 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
       when (~io.dmactive) {
         selectedHartReg := 0.U
       }.elsewhen (io.innerCtrl.fire){
-        selectedHartReg :<= io.innerCtrl.bits.hartsel.squeeze
+        selectedHartReg :%= io.innerCtrl.bits.hartsel
       }
     }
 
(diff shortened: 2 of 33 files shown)


src/main/scala/devices/debug/Debug.scala
```diff
@@ -1323,18 +1323,18 @@ class TLDebugModuleInner(device: Device, getNComponents: () => Int, beatBytes: I
         val hartselIndex        = UIntToOH(io.innerCtrl.bits.hartsel)
         when (hartHaltedWrEn) {
           // add those harts halting and remove those in reset
-          haltedBitRegs := (haltedBitRegs | hartHaltedIdIndex) & ~(hartIsInResetSync.asUInt)
+          haltedBitRegs :<= ((haltedBitRegs | hartHaltedIdIndex) & ~(hartIsInResetSync.asUInt)).squeeze
         }.elsewhen (hartResumingWrEn) {
           // remove those harts in reset and those in resume
-          haltedBitRegs := (haltedBitRegs & ~(hartResumingIdIndex)) & ~(hartIsInResetSync.asUInt)
+          haltedBitRegs :<= ((haltedBitRegs & ~(hartResumingIdIndex)) & ~(hartIsInResetSync.asUInt)).squeeze
```
jackkoenig:
```suggestion
          haltedBitRegs :%= ((haltedBitRegs & ~(hartResumingIdIndex)) & ~(hartIsInResetSync.asUInt))
```
I personally would use the new truncating connection operator[[1]](https://www.chisel-lang.org/docs/explanations/connectable#connecting-components-with-different-widths) (only defined on `Bits`) because with complex expressions on the RHS it can be easy to miss the `.squeeze` at the end.

src/main/scala/devices/debug/Custom.scala
```diff
@@ -84,7 +84,7 @@ class DebugCustomXbar(
     }
     // mux correct 'ready' and 'data' based on address
     sink.ready := (decoded zip sources).foldLeft(false.B){case (result, (d, i)) => result || (d & i.ready)}
-    sink.data := (decoded zip sources).foldLeft(0.U){ case (result, (d, i)) => result | Mux(i.ready, i.data, 0.U)}
+    sink.data :<= (decoded zip sources).foldLeft(0.U(sink.data.getWidth.W)){ case (result, (d, i)) => result | Mux(i.ready, i.data, 0.U)}
```
jackkoenig:
```suggestion
    sink.data :<= (decoded zip sources).foldLeft(0.U(0.W)){ case (result, (d, i)) => result | Mux(i.ready, i.data, 0.U(0.W)}
```
I'm assuming this only fixes things when `sink.data` is 0-width (because `0.U` is 1-bit[^1]) so you can probably just do this?

[^1]: I want to change it to 0-bit but until implicit truncation is gone that could silently change behavior

src/main/scala/devices/debug/Debug.scala
jackkoenig: If the issue is using older Chisel, that's fine because you can easily copy [the implementation](https://github.com/chipsalliance/chisel/blob/e25e8ca7c3af7eb6605821402ba83327a34f26b1/core/src/main/scala/chisel3/connectable/Connectable.scala#L429) of `:%=` in to rocket-chip, it uses extension methods anyway

tynan-jdwk: Indeed the reason I didn't use `:%=` was to allow Rocket Chip to continue to work with older Chisel, but copying this implementation in sounds like a great idea.

src/main/scala/devices/debug/Custom.scala
tynan-jdwk: Ya, that does work, I'll go with this.

src/main/scala/devices/debug/Debug.scala
tynan-jdwk: Just did this, thanks for the suggestion!

